PR #560, opened 2013-10-14
Real Time Paparazzi using ChibiOS

flixr:
Replacing #556 

Hottest issues:
1) Fix linker script properly so it can be used with Luftboot (offset 8kb)
2) Fix project building path so object files are generated into var/airframe_name directory
If you use airframe "hexarotor_lia_rt.xml" with "default_chibios.xml" telemetry and compile it, it creates new folder "sw/airborne/.deb" with dependency files ( "filename.o.d"). Normal paparazzi doesn't make these files, so I assume it is something with compiler settings/flags.
3) Fix flash erase/lock during JTAG upload


flixr:
Closing as this one is outdated.
Work is continuing in the rt_chibios branch and tracked in #1105

PR #669, opened 2014-03-13
Stm32f3 discovery

flixr:
Preliminary support for F3 based on #625 of @memeruiz

Still needs quite a bit of work (see comments in previous pull request) and testing...


OpenUAS:
This board for testing OK? Shall I buy it?
 http://nl.farnell.com/stmicroelectronics/stm32f3discovery/eval-stm32f3-cortex-m4-discovery/dp/2215108?ost=STM32F3+DISCOVERY

Will buy this board and can perform some test it on if needed before next release.

flixr:
Ugh, this tree is really old and there is still some work to be done to support F3 using libopencm3...
Not sure if it is worth the effort... using ChibiOS (with it's HAL) might actually be an easier/better option.

OpenUAS:
Discovery F3 Board ordered, ETA 20151120, able to test ChiBeOS v3.x or whatever needed. A lot of external IO devices can be connected and tested.

OpenUAS:
Discovery F3 Board available at my desk. Any details on the work to be done for F3 welcomed.

Looks like LibopenCM3 has examples for F3 see:
https://github.com/libopencm3/libopencm3-examples/tree/master/examples/stm32/f3

Willing to spend time to test PPRZ on this board if of any use for PPRZ project, btw also willing to test CheBeOS v3.x on the board if it helps.

flixr:
This branch would need merging with master (and update to same libopencm3 version), then more fixes and testing...
Maybe also an update to latest libopencm3 (but will need changes for other F1/4 stuff as well).

gautierhattenberger:
Do we keep this open ? BTW, it might be easier in some cases to go through the ChibiOS arch as most of the driver differences are already hidden by ChibiOS HAL.

podhrmic:
FYI http://chibios.org/dokuwiki/doku.php?id=chibios:product:hal:platforms
This can be easily done with ChibiOS (basically just add a proper board file), so maybe better to close it?

PR #681, opened 2014-04-02
Activate the use of the barometer by default on the lisa S.

clement-roblot:
The corresponding documentation is available here : https://wiki.paparazziuav.org/wiki/Lisa/S#Barometer

### DIFF
--- a/sw/airborne/boards/lisa_s_0.1.h
+++ b/sw/airborne/boards/lisa_s_0.1.h
@@ -92,6 +92,11 @@
 
 #define BOARD_HAS_BARO 1
 
+/* by default activate onboard baro */
+#ifndef USE_BARO_BOARD
+#define USE_BARO_BOARD 1
+#endif
+
 /* SPI slave mapping */
 
 /* IMU_MPU_CS */


woeileong1104:
Hi, I tried to activate the barometer on my Lisa/S, followed the documentation on the wiki page, but in vain. I have two Lisa/S with me, both of them not working, so I guess it is not hardware problem. 

I tried to put  "<define name="USE_BAROMETER" value="TRUE"/>" at different locations in the airframe files, even inside the "ins subsystem" like the following, but it yields no result.

<subsystem name="ins">
   <define name="USE_BAROMETER" value="TRUE"/>
</subsystem>

Is there anything else I can do?

kirkscheper:
Hey,
Unfortunately it may be a hardware issue after all, the Lisa/s uses a baro
with very tight overvoltage limits which can be exceeded when you plug in a
fresh (brand new and fully charged battery). We have seen this happen quite
often.

If you send the branch you are using I can try the code on a Lisa/s with a
working baro and check your configuration to make sure it is not a software
issue.

Kirk


On 28 Dec 2016 4:08 a.m., "WL Chan" <notifications@github.com> wrote:

Hi, I tried to activate the barometer on my Lisa/S, followed the
documentation on the wiki page, but in vain. I have two Lisa/S with me,
both of them not working, so I guess it is not hardware problem.

I tried to put "" at different locations in the airframe files, even inside
the "ins subsystem" like the following, but it yields no result.

Is there anything else I can do?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/paparazzi/paparazzi/pull/681#issuecomment-269418434>,
or mute the thread

woeileong1104:
Hi Kirk,

I am using this:
https://github.com/woeileong1104/paparazziNUS

Airframe flowerfly_tailed2. Modified it from the Lisa/S Quadcopter.

Woei Leong

kirkscheper:
Hey Woei Leong,

I tested your airframe on a lisa/s I know has a working baro and it 
works just fine. Unfortunately, that would suggest that the baros on 
your lisas are broken. As I mentioned in my previous mail, we have seen 
this to happen when there is an over-voltage on the baro as it has a 
very low over-voltage tolerance. You can try replacing the baro but it 
is not the easiest part to replace on the board. If you replace the baro 
be careful what type of batteries you use, we have seen that some types 
of batteries are more likely to cause an over-voltage than others 
(seemingly correlated with price but not sure).

Good luck.

Kirk Scheper

woeileong1104:
Thanks Kirk, that's bad news. You have a recommended battery brand? We are using e-flite 150mah 45c.

Issue #794, opened 2014-08-17
make settings suck less....


flixr:
They suck... and that really sucks...
We should try to make them at least suck a bit less... e.g. not only send as float, which sucks...
##

Issue #793, opened 2014-08-17
link does not support full range of unsigned integer values


flixr:
Sending an unsigned integer value (e.g. uint32_t) that does not fit in to the respective signed version anymore (int32_t) is not properly handled in link. It is instead converted to a signed int and hence negative if the value is so large that the MSB is set.

flixr:
Issue known, but worked around in all known cases...
Still possible issues with unsigned 64bit ints, but these are not used right now...
@gautierhattenberger doesn't have the time to work on this to make it clean, so closing this as WONT_FIX

PR #797, opened 2014-08-17
[settings] send as double instead of float

flixr:
A bit of a sucky workaround in absense of proper type handling in settings...
Addresses parts of #794 by enabling you to send/receive settings that don't fit in a float (e.g. high uint32 values)


OpenUAS:
IMHO,implementing something like this it would be best using ocaml-uint

flixr:
@OpenUAS see https://github.com/paparazzi/paparazzi/issues/793

Closing this one...

PR #829, opened 2014-09-09
[modules] baro_board as module

flixr:
Since in the long run we want to move more stuff to modules and eventually also convert subsystems to modules to be more flexible and have less explicit calls in main, I started with changing the baro_board subsystem (which is actually automatically added to fixedwing and rotorcraft firmware at the moment) to a module.

So if you try to use this, you have to add the baro_board.xml module (otherwise it will warn you if your board normally has a baro onboard).

Only the microjet_lisa_m and quadrotor_lisa_mx are updated to have the baro_board module loaded right now...

This is mostly meant to have something to look at for discussion about how we should handle things like automatically loading the correct drivers according to the board that you have.

This module does basically the same as the subsystem makefile before:
a raw makefile section that needs to list all available boards with their baro options.

Maybe we can instead think of another way to have some modules automatically loaded, so we really have only one module for each baro?

### DIFF
--- a/conf/airframes/ENAC/fixed-wing/apogee.xml
+++ b/conf/airframes/ENAC/fixed-wing/apogee.xml
@@ -33,6 +33,7 @@
     <!--define name="AGR_CLIMB"/-->
     <!--define name="LOITER_TRIM"/-->
     <!--define name="PITCH_TRIM"/-->
+    <define name="USE_BARO_BOARD" value="FALSE"/>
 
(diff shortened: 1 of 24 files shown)


podhrmic:
@gautierhattenberger any updates here or should we close it?

gautierhattenberger:
I would close this and wait for a proper dependency system to have one module per baro loaded as a default for boards that have one.

PR #856, opened 2014-10-13
AHRS API/Interface refactor

flixr:
Goals:
- to allow multiple different AHRS algos running at the same time
- no more dependencies on the global imu or ahrs_aligner structs in the ahrs implementation

The idea is that every ahrs has it's own unique names for functions and implementation struct and that you then can e.g. register an ahrs algorithm by passing pointers to to the propagate/update functions.

Question if we also want to support running multiple instances of the same AHRS algorithm in the same process (meaning compiled into same blob)? That would also require to pass the AHRS implementation struct to it's own update functions. But you start to get to a point where you implement C++ features in C, but in a less readable way...
-> only one instance per AHRS implementation for now


flixr:
Updated:
- algorithms now (mostly) have a wrapper for registering the default implementation and the ABI and periodic telemetry stuff, that makes it easier to use the algorithms standalone (e.g. in tests without ABI)
- body_to_imu quaternion is also set via ABI
- each ahrs implementation has an `is_aligned` bool for now...

flixr:
@gautierhattenberger, what do you think about merging this "as is"?
It also contains the function to only update heading from mag in MLKF, which is quite useful.

gautierhattenberger:
I want to test it first, especially the ins_float_invariant. Hopefully I can do that tomorrow, it's at the top of my todo list...

gautierhattenberger:
Tested and working on lpc21, stm32 and ardrone2 with ins_invariant, float_dcm and int_cmpl_euler

PR #861, opened 2014-10-19
Gasoline Flow meter module

agressiva:
Add a flow meter to be used as gasoline or glow meter.
To be used with BIOtech low flow sensor with pulse output


OpenUAS:
An update welcomed. BTW sadly testing with real flowmeter not yet possible for us before 20151101

OpenUAS:
This part could be used for testing right?
http://www.conrad.com/ce/en/product/150392/BT-Bio-Tech-FCH-M-POM-LC-AD-6-MM-Low-Flow-Flowmeter-Non-Corrosive-Liquids-Water-Diesel-Oil

If So then I can test this with a pump, regulator valve and a weigting scale (on the density of the fluid used for testing) on other end to see if it works well

Like to see at least one flowmeter in PPRZ soon.

agressiva:
Hi, for sure you can use this sensor.
i used a similar one on mine plane.
The problem is the minimum flow measured by the sensor.
In mine case, the sensor now worked with 20cc gas motor.

gautierhattenberger:
Should we close this due to feedback timeout ?

podhrmic:
Sounds like a good idea.

PR #999, opened 2014-11-29
[link] improve datalink/telemetry report

gautierhattenberger:
- more reported data about up/down-link
- cleanup report messages
- display link page in GCS even in single link mode

Tested in simulation, will be on real hardware soon.
Test with redundant link would be welcome since I don't have working setup for this.


camlee:
@flixr : UAARG member here. I know this is quite late, but I'll try and test these changes with multiple links soon (thinking 1-2 months). I skimmed your code changes and they look good to me so far. Definitely nice to see improvement in this area such as the link page showing data even when redundant links are not used.

PR #1017, opened 2014-12-07
[luftboot] Added support for luftboot-like bootloader for Lisa MX

podhrmic:
(the bootloader itself is in ext repository, this is only support in pprzi code)
Tested on Lisa MX.


OpenUAS:
@flixr OK, was is tested now?

flixr:
We are still missing a proper and clean implementation of luftboot for F4...
The issue @podhrmic had with the broken pipe should be fixed in libopencm3 now...

PR #1018, opened 2014-12-07
[stm32] use desig_get_unique_id from libopencm3

flixr:
- update libopencm3 so we have desig in F4 lib
- use desig_get_unique_id_as_string from libopencm3 instead of our own function


flixr:
Updated to use the latest libopencm3 version that converts the high nibble of the id to hex char first:
https://github.com/libopencm3/libopencm3/commit/f206fcb57ea4bc93a6d2590d878837857ce5e848
Now `desig_get_unique_id_as_string` returns a string that corresponds directly to the id in hex...

flixr:
Ah, yeah.. and also updated luftboot to use the libopencm3 function...

Issue #927, opened 2014-11-05
items in the georef menu of Flight Plan Editor are not updated after closing or editing plan


alonsoac:
Hi. When closing a flight plan I would expect the corresponding items from the menu would be removed.  When deleting a waypoint the item should be removed. When changing the name of a waypoint the item should be renamed.

PR #1059, opened 2015-01-01
fixes to gcs / flight plan editor

alonsoac:
Fixes for issue #927 

- deleting a wp from Waypoint Edit dialog or from XML editor does not remove the item from the georefs menus
- changing the name of a wp in xml editor or in dialog does not update the menus
- closing a flight plan does not clear the menus
- opening a different plan after closing the first does not work because fit to window keeps trying to fix wp from closed fp

Other changes
- keep map center when zooming with the spin button
- When the georef is set to a wp and the mouse is moved to a different UTM zone it would no longer update the distance. Now it shows an error message. TODO: fix so that it works across utm zones.

NOT Fixed
- creating a new wp by inserting a new xml item or by copying an existing item does not add the wp to the window or the menus
- deleting a wp does not remove it from sectors
- changing the name of a wp does not update the sectors

### DIFF
--- a/sw/ground_segment/cockpit/editFP.ml
+++ b/sw/ground_segment/cockpit/editFP.ml
@@ -63,6 +63,7 @@ let close_fp = fun geomap ->
     | Some (fp, _filename) ->
       let close = fun () ->
         fp#destroy ();
+        geomap#clear_georefs ();
         current_fp := None in
       match GToolbox.question_box ~title:"Closing flight plan" ~buttons:["Close"; "Save&Close"; "Cancel"] "Do you want to save/close ?" with
           2 -> save_fp geomap; close ()
--- a/sw/ground_segment/cockpit/gcs.ml
+++ b/sw/ground_segment/cockpit/gcs.ml
@@ -290,7 +290,7 @@ let button_press = fun (geomap:G.widget) ev ->
     GToolbox.popup_menu ~entries:([`I ("Load background tile", display_gm)]@m)
       ~button:3 ~time:(Int32.of_int 0);
     true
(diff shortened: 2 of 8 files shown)


alonsoac:
I am new to Ocaml. Any comments on coding style are welcome.

flixr:
Thx! Really appreciated.
Since @gautierhattenberger knows this stuff a lot better, I hope he can cross-check this and maybe fix some of the remaining ones.

Regarding style:
Commit style wise it is nicer to have each commit represent a "logical change" (where ideally each step compiles) with a good commit message.
Coding style wise it looks ok to me. Only thing I would probably change is to put the added comment in a separate line before instead of at the end... but that is really minor.

gautierhattenberger:
The zoom part looks fine.
For the georefs list, the first thing is to avoid using the List.iteri, which is very convenient but not supported before ocaml 4.0 (that's why Travis is failing). Since you are now manipulating a lot the data in the list, I would suggest to change it for a Hashtbl.

alonsoac:
Ok please apply this latest commit to remove List.iteri.  I am not sure that changing to a Hashtbl will make any difference on this short list but if you think it is important I can give it a try next week.

gautierhattenberger:
@alonsoac please have a look at https://github.com/paparazzi/paparazzi/commit/0c9c90e7659550727ddd8ccb6c2fe6a92e7ccc57
It is functionally the same but written in a more "functional" way. I kept the list and not the hashtable because it was finally easier to keep the original wp order with the list.

alonsoac:
It seems to work as expected so it is fine with me. Thanks.

PR #1062, opened 2015-01-07
opticflow module

dewagter:
todo

 - [x] investigate the changes of @gautierhattenberger and use the best approach
 - [x] move conf
 - [x] remove submodule


dewagter:
how about this... anyone wants to make the code look better?

sw/airborne/modules/computer_vision/cv/resize.h
```diff
@@ -36,13 +36,13 @@ inline void resize_uyuv(struct img_struct *input, struct img_struct *output, int
       *dest++ = *source++; // U
       *dest++ = *source++; // Y
       // now skip 3 pixels
-      if (pixelskip) { source += (pixelskip + 1) * 2; }
+      source += (pixelskip + 1) * 2;
```
gautierhattenberger: I would like to keep my version, otherwise it doesn't work with downsample=1 (which makes a copy of the image)

sw/airborne/modules/computer_vision/lib/paparazzi.h
```diff
@@ -0,0 +1,32 @@
+
+////////////////////////////////////////////////
+// Paparazzi communication interface
```
gautierhattenberger: is this file really needed ?

sw/airborne/modules/computer_vision/opticflow/opticflow_code.c
```diff
@@ -0,0 +1,310 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+// Own Header
+#include "opticflow_code.h"
+
+// Computer Vision
+#include "opticflow/optic_flow_ardrone.h"
+#include "opticflow/fast9/fastRosten.h"
+#include "../opticflow_module.h"
```
gautierhattenberger: better use global path `modules/computer_vision/opticflow_module.h`

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -0,0 +1,225 @@
...
+    computervision_thread_status = -1;
+    return 0;
+  }
+
+  // Video Grabbing
+  struct img_struct *img_new = video_create_image(&vid);
+
+  // Video Resizing
+#define DOWNSIZE_FACTOR 1
+  struct img_struct small;
```
gautierhattenberger: if you don't plan to use DOWNSIZE_FACTOR higher than 1, you should probably remove completely `small` variable

gautierhattenberger:
Thanks. Just one thing missing: in `sw/ground_segment/tmtc/server_globals.ml:7` and `sw/ground_segment/cockpit/livel.ml:1268` you should add the new mode name (or short name, `MOD`)

sw/airborne/firmwares/rotorcraft/guidance/guidance_module.h
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2008-2009 Antoine Drouin <poinix@gmail.com>
```
flixr: Could you please correctly put your name in here?

sw/airborne/firmwares/rotorcraft/guidance/guidance_module.h
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  If the module implements both V and H mode, take into account that the H is called first and later V
+ */
+
+#ifndef GUIDANCE_MODULE_H_
+#define GUIDANCE_MODULE_H_
+
+#include "generated/modules.h"
+
+/**
```
flixr: This dox block it is not documenting any function or variable (as there is nothing coming after it).
So it would make more sense to add this to the file dox block above, where it will actually end up in the generated docs.

sw/airborne/modules/computer_vision/opticflow/opticflow_code.c
```diff
@@ -0,0 +1,310 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/*
```
flixr: should be `/**` instead of `/*`

sw/airborne/modules/computer_vision/opticflow/hover_stabilization.h
```diff
@@ -0,0 +1,60 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/*
```
flixr: should be `/**` instead of `/*`

sw/airborne/modules/computer_vision/opticflow/hover_stabilization.h
```diff
@@ -0,0 +1,60 @@
...
+ * @file paparazzi/sw/ext/ardrone2_vision/modules/OpticFlow/hover_stabilization.h
```
flixr: should be `@file modules/computer_vision/opticflow/hover_stabilization.h`

flixr:
Doesn't seem to compile here, I'm probably missing something...
Could you please also add an aircraft configuration to conf_tests?

dewagter:
@flixr thanks for the updates/tips

 - [x] dox
 - [x] conf_test.xml
 - [x] more generic controller interface: see sw/airborne/modules/ctrl/ctrl_module_demo

sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -464,6 +469,11 @@ void autopilot_set_mode(uint8_t new_autopilot_mode)
       case AP_MODE_NAV:
         guidance_v_mode_changed(GUIDANCE_V_MODE_NAV);
         break;
+      case AP_MODE_MODULE:
+#ifdef GUIDANCE_V_MODE_MODULE_SETTING
+        guidance_v_mode_changed(GUIDANCE_V_MODE_MODULE);
```
flixr: I guess this should be `GUIDANCE_V_MODE_MODULE_SETTING`

dewagter:
yes indeed

sw/airborne/modules/computer_vision/opticflow/visual_estimator.c
```diff
@@ -0,0 +1,309 @@
...
+// Own Header
+#include "visual_estimator.h"
+
+// Computer Vision
+#include "opticflow/optic_flow_int.h"
+#include "opticflow/fast9/fastRosten.h"
+#include "modules/computer_vision/opticflow_module.h"
+
+// Paparazzi Data
+#include "subsystems/ins/ins_int.h"
```
flixr: Should not be needed...

sw/airborne/modules/computer_vision/opticflow/visual_estimator.c
```diff
@@ -0,0 +1,309 @@
...
+#else
+  OFx_trans = dx_sum;
+  OFy_trans = dy_sum;
+#endif
+
+  // Average Filter
+  OFfilter(&OFx, &OFy, OFx_trans, OFy_trans, flow_count, 1);
+
+  // Velocity Computation
+#ifdef USE_SONAR
```
flixr: USE_x flags should be checked with `#if USE_x` and not `#ifdef USE_x`.
Probably best to subscribe to the AGL ABI message to get the scale.

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -0,0 +1,202 @@
...
+ * Sensors from vertical camera and IMU of Parrot AR.Drone 2.0
+ */
+
+
+// Own header
+#include "opticflow_module.h"
+
+// Paparazzi
+#include "state.h" // for attitude
+#include "boards/ardrone/navdata.h" // for ultrasound Height
```
flixr: doesn't seem to be used...

sw/airborne/modules/computer_vision/opticflow/visual_estimator.h
```diff
@@ -0,0 +1,48 @@
...
+extern unsigned int verbose;
+
+// Variables used by the controller
+extern float Velx, Vely;
+extern int count;
+extern int flow_count;
+extern struct FloatVect3 V_body;
+
+// Called by plugin
+void my_plugin_init(void);
```
flixr: any reason to give this such a non-descriptive name?

flixr:
@dewagter what is the status here?
Is it tested, do you need more feedback?

dewagter:
- applied all recommendations
 - for me it can be merged in master
 - no opportunity to test this month
 - still misses sonar direct measurement (AGL for vision scaling) which
apparently moved. What is the good way to get AGL? should be filled here:
https://github.com/paparazzi/paparazzi/pull/1062/files#diff-f82843cf3c615d01d35cd2b6d673cee0R278
and then it should^{TM} fly as before.

-Christophe

flixr:
Well, after looking at it again, there are still some open questions (that I already posted above)....
E.g.
- as I wrote to get AGL you should subscribe to the ABL ABI message.
- what is up with the rather non-descriptive `my_plugin` names?

dewagter:
Will see if I can ask someone to look at it. If its minor, feel free to
merge-squash-update.

-Christophe

flixr:
I can add an ABI subscriber for AGL, but have no way to test it...

@gautierhattenberger what do you think about the guidance_module and `AP_MODE_MODULE` additions?

sw/airborne/modules/computer_vision/opticflow_module.h
```diff
@@ -0,0 +1,51 @@
...
+ * Sensors from vertical camera and IMU of Parrot AR.Drone 2.0
+ */
+
+#ifndef OPTICFLOW_MODULE_H
+#define OPTICFLOW_MODULE_H
+
+#include "std.h"
+
+// Navigate Based On Vision
+#include "opticflow/hover_stabilization.h"
```
flixr: why is this needed in the header?

sw/airborne/modules/computer_vision/opticflow_module.h
```diff
@@ -0,0 +1,51 @@
...
+
+// Module functions
+extern void opticflow_module_init(void);
+extern void opticflow_module_run(void);
+extern void opticflow_module_start(void);
+extern void opticflow_module_stop(void);
+
+extern void guidance_module_run(bool_t inflight);
+
+// Frame Rate
```
flixr: why expose those?

flixr:
There are also some other potential issues, like calling `my_plugin_run` from the computervision thread.
Access to state interface is not thread safe!

conf/modules/cv_opticflow.xml
```diff
@@ -0,0 +1,71 @@
...
+  <header>
+    <file name="opticflow_module.h"/>
+  </header>
+
+  <init fun="opticflow_module_init()"/>
+
+  <periodic fun="opticflow_module_run()" freq="60" start="opticflow_module_start()" stop="opticflow_module_stop()" autorun="TRUE"/>
+  <makefile target="ap">
+    <define name="ARDRONE_VIDEO_PORT" value="2002" />
+    <define name="USE_ARDRONE_VIDEO" />
```
flixr: What is this supposed to be good for, seems unused to me...

flixr:
Tried to cleanup some more stuff and added a subscriber for AGL (to use for scale in velocity computation)...
Before we merge this, it should really be tested...

Accessing the state interface from a thread is probably not a good idea, you can be lucky and it works, but there is no guarantee that you don't try to read it while it is currently updated...

gautierhattenberger:
AP_MODE_MODULE is not the best, but good enough I guess until we can generate modes

dewagter:
@flixr thank for the updates. 

This multithread design initially used UDP to send data from one program to the next: https://github.com/tudelft/ardrone2_vision/blob/master/modules/ObstacleAvoidSkySegmentation/sky_seg_avoid_gst.c#L82

I don't know at which point in time this got lost...

sw/airborne/modules/computer_vision/opticflow/visual_estimator.c
```diff
@@ -0,0 +1,331 @@
...
+#ifndef OPTICFLOW_AGL_ID
+#define OPTICFLOW_AGL_ID ABI_BROADCAST
+#endif
+abi_event agl_ev;
+static void agl_cb(uint8_t sender_id, const float *distance);
+
+static void agl_cb(uint8_t sender_id __attribute__((unused)), const float *distance)
+{
+  if (distance > 0) {
+    estimator_agl = *distance;
```
gautierhattenberger: In this particular case, the data is read only and the copy is atomic, so it will be thread safe.
But testing the pointer is not correct ;)
```
if (*distance > 0) {
```

flixr: Thanks for catching this!
And yes, this one is thread safe, otherwise I wouldn't have added it ;-)

flixr:
And sending telemetry messages via DOWNLINK_SEND_x or pprz_msg_send_x is also not thread safe...
So I would suggest to move the downlink  of `OF_HOVER` to somewhere else and also compute `V_body` in hover_stabilization which is the only other place that actually uses this.

dewagter:
I would prefer to make it 100% safe again: define 1 struct that goes from
pprz->thread and 1 struct that goes from thread->pprz and have all
navigation (fetching state) and messages in the module.

(this was the setup from the start, but after 3 people worked on it, it
changed along the road...)

Any better idea than UDP transfer?  (this also makes it possible to run a
separate program instead of a thread)

-Christophe

flixr:
At least use a unix domain socket instead of a udp socket...
Ideally we would just send it via ABI and it would handle that in the background...

flixr:
@dewagter see my opticflow_socket branch with a dummy example on how you could use unix domain sockets to pass results from the thread, specifically https://github.com/flixr/paparazzi/commit/dfa33e4824db5e23386e6e5e361c398218d265fd

PR #1063, opened 2015-01-09
bebop actuators left/right reversed

EwoudSmeur:
The motor locations originally came from the bebop datasheet (interpreted from french).
When checking which actuator is where, it turns out that they are left/right reverssed. It was only flying because the actuator matrix was also left/right swapped


flixr:
What about the correct assignment for NPS?

flixr:
Merged with ff964e55ac0e97d44eba7d047b633e3f0a2273b6

Issue #981, opened 2014-11-24
Common API interface for waypoints


flixr:
Currently rotorcraft and fixedwing firmwares have different waypoint representations/coordinate systems.
We have no common API interface to get the coordinates of a waypoint in one representation regardless if fw/rc firmware... needed e.g. for #936 

Also related to #455 

An ideas on what the best approach would be?
##

flixr:
About the naming of the API functions for waypoints:
Currently we have `nav_set_waypoint_foo`, `nav_globalize_local_wp`, etc... in subsystems/navigation/waypoints.h
I think that was a bad initial choice on my part!
How about naming them all consistently (and shorter): `waypoint_set_enu_f`, `waypoint_set_xy_i`, `waypoint_get_lla`, `waypoint_localize`, `waypoints_globalize_all`, etc...?

saaib:
Felix,

Is there a pre-agreed naming convenction structure for other functions/APIs?

Regards,

-Urivan

On 04/20/2015 05:33 AM, Felix Ruess wrote:
>
> About the naming of the API functions for waypoints:
> Currently we have |nav_set_waypoint_foo|, |nav_globalize_local_wp|, 
> etc... in subsystems/navigation/waypoints.h
> I think that was a bad initial choice on my part!
> How about naming them all consistently (and shorter): 
> |waypoint_set_enu_f|, |waypoint_set_xy_i|, |waypoint_get_lla|, 
> |waypoint_localize|, |waypoints_globalize_all|, etc...?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/paparazzi/paparazzi/issues/981#issuecomment-94440770>.
>

-- 
-Urivan Flores-Saaib

flixr:
Not really, mostly for historic reasons... apart from what is in the style guide: http://docs.paparazziuav.org/latest/stylec.html

It is quite painful to change the API later, and this is not in any stable branch yet: so it should be decided before the next stable release.

In general functions and variables should start with a lower case letter.
And prefer under_scored_names, but we also have quite a few camelCase ones...
IMHO the most important thing, is to try to be consistent within the module/subsystem functional API.

Issue #455, opened 2013-06-08
absolute waypoint handling for fixedwings


flixr:
There is currently no _real_ support for absolute waypoints, after initialization all are equal and specified in the navigation coordinate system (UTM for fixedwings, local NED for rotorcrafts).
##

flixr:
Done for rotorcrafts with #1065

PR #1065, opened 2015-01-13
[navigation] implement global waypoints for rotorcraft

flixr:
- each waypoint is a struct with multiple representations and flags (e.g. if absolute/global)
- factor out waypoints in separate files so it can be shared between firmwares later
- if a waypoint is defined with lat/lon instead of x/y in the flight plan, it is marked as an absolute/global waypoint and not moved upon GeoInit

Only for rotorcraft so far...

### DIFF
--- a/conf/firmwares/subsystems/rotorcraft/navigation.makefile
+++ b/conf/firmwares/subsystems/rotorcraft/navigation.makefile
@@ -2,4 +2,5 @@
 
 $(TARGET).CFLAGS += -DUSE_NAVIGATION
 $(TARGET).srcs += $(SRC_FIRMWARE)/navigation.c
+$(TARGET).srcs += subsystems/navigation/waypoints.c
 $(TARGET).srcs += subsystems/navigation/common_flight_plan.c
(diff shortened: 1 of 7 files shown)


flixr:
Basically solves #455 (for rotorcraft) and is a start for #981.
For fixedwings we stilll need to figure out how to best deal with the waypoints in UTM (or even if we keep waypoints in UTM there).

gautierhattenberger:
It looks fine for the rotorcraft part.
When updating local waypoints, it would be nice to also compute their global position (based on the state interface ref point), if possible of course.

flixr:
Sure, I can do that... but I was wondering if it is nicer to also create some "getter" functions that compute it on the fly if needed, or just update all possible representations each time a waypoint is updated...

gautierhattenberger:
I was suggesting this since WP are not moved very often, but getter functions are even better

flixr:
Added computation of LLA on ENU waypoint set for now. But we should probably still add a getter function that computes it on demand if not already available (e.g. if origin is not already set from flight plan at startup).

flixr:
Added nav_get_waypoint_lla function, that will compute the LLA representation if not already available.
@gautierhattenberger do you think this is good to merge as is?
It's only rotorcraft so far, we will need to think about how to handle UTM for fixedwings properly...

OpenUAS:
A fix without UTM support for this for Fixedwing would be already much better. But probbably e wait for ABI to be finished ?

flixr:
This is totally unrelated to ABI.
And for fixedwing firmware it *has* to handle UTM properly, as that is the used navigation coordinate system.

Issue #1056, opened 2014-12-24
It should be possible to select which of multiple a/c are seen in each GCS window


alonsoac:
I would like to be able to monitor 2 or more a/c each one in a separate GCS window. 

One way to do this could be to add a -aircraft_id parameter to gcs so that only messages for that a/c are received.
Another more user friendly way could be in the GCS aircraft menu add a Hide item that would completely hide all information from all sections of the GCS window for that aircraft.

OpenUAS:
There is a GCS parameter -auto_hide_fp  is this of help?

alonsoac:
The problems with having multiple a/c in the window when only one needs to be monitored include:
 - alerts from other aircrafts are mixed in with the monitored a/c
 - wind information is mixed from different sources when only interested in monitored a/c
 - papgets mix data from different a/c
 - it is not possible to hide the strips for other aircrafts or at least have the selected aircraft show on top
- it is possible to click on the wrong a/c and make a flight plan change
- it is not easy to go back to the previous location and zoom after having to change it to see another a/c
There are multiple use cases. Someone might want to auto hide fp and strips but keep seeing alerts for all a/c.
It may be possible to make the GCS more robust so that multiple a/c can be monitored easily from just one window but I figure it would be easier first to select only one and then open more windows to monitor other a/c if needed.

OpenUAS:
Thanks for the good issue description. I also see it as benificial. How to implement best is open for discussion add remarks to this issue plz.

PR #1068, opened 2015-01-15
improve Papgets with multi aircraft support

### LINKED ISSUE #181: papgets should also work with more than one aircraft
Papgets only work if you only have one aircraft and are currently not usable when you have several as you can't specify the aircraft id for them.
E.g. if you use a goto button papget with multiple aircrafts you would end up sending the command to all aircrafts.
So it seems to me like this would need quite a bit of refactoring and the functionality to listen to a specific ac_id.

gautierhattenberger:
Makes papgets working with several aircraft. If no target A/C specified, behavior is the same then before.

This will close #181 and a part of #1056

### DIFF
--- a/sw/ground_segment/cockpit/gcs.ml
+++ b/sw/ground_segment/cockpit/gcs.ml
@@ -693,7 +693,13 @@ let () =
   listen_dropped_papgets geomap;
 
   let save_layout = fun () ->
-    let the_new_layout = replace_widget_children "map2d" (Papgets.dump_store ()) the_layout in
+    (* Ask if ac_id parameters from papgets should be saved *)
+    let save_acid =
+      match GToolbox.question_box ~title:"Save Layout" ~buttons:["Yes"; "no"] ~default:1 "Do you want to save A/C id of Papgets if available\nYes: the saved layout will only work with A/C that have the same id (default)\nno: the saved layout will work with any A/C (but will mix data while using multiple A/C)" with
+      | 2 -> false
+      | _ -> true
+    in
+    let the_new_layout = replace_widget_children "map2d" (Papgets.dump_store save_acid) the_layout in
     let width, height = Gdk.Drawable.get_size window#misc#window in
     let the_new_layout = update_widget_size `HORIZONTAL widgets the_new_layout in
     let new_layout = Xml.Element ("layout", ["width", soi width; "height", soi height], [the_new_layout]) in
--- a/sw/ground_segment/cockpit/gcs.ml
+++ b/sw/ground_segment/cockpit/gcs.ml
@@ -695,9 +695,11 @@ let () =
   let save_layout = fun () ->
     (* Ask if ac_id parameters from papgets should be saved *)
     let save_acid =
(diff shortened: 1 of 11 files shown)


flixr:
And how do you specify the aircraft id?

gautierhattenberger:
It is automatic for drag-and-drop.
If you make a papget from a GCS xml file, you just need to add the property to the list. Something like this should work for A/C number 10:
```
<property name="ac_id" value="10"/>
```
It reminds me that I should probably save this when using the save GCS layout option... coming soon

flixr:
Ok, was wondering because you can neither select the ac_id in the papget edit popup, neither was it saved to the config...

gautierhattenberger:
tested and working with 2 A/C in sim, saving and loading layout

flixr:
Can you also add the ac_id as an input to the papget edit popup?
On a sidenote: I also noticed that the ticks in the gauge papget are not correctly resized when zooming in.

gautierhattenberger:
I was wondering if adding the field ac_id to the edit box was really useful. Changing the ac_id dynamically means unbinding and binding again. I did looked at it yet if it is a pain or not, but probably a little.
But I agree that showing the ac_id (without edition) is a good idea. Would it be enough ?

I'll look at the gauge.

flixr:
I guess at least displaying it (or displaying that any id is used) would be good...

flixr:
The fix for the gauge papget works nicely, thanks!

Regarding the ac_id for papgets, IMHO it is still not really great and I think we should have at least one of the following options
- upon drag-n-drop of a message: possibility to choose whether it should bind to that specific ac_id or any
  - maybe even default to "any"?
- possibility to disable filtering on ac_id in the papget edit dialog

Otherwise it becomes rather annoying to create a layout that has the papgets you want, but works with any _single_ aircraft.

gautierhattenberger:
Here, I did option 3: ask if ac_id should be saved or not when saving layout.
The reasons are:
* it is really a annoying to unbind and bind again
* not all papgets have config dialog box
* drag-and-drop to "any" doesn't make mush sense to me
* this solution is easy and covers your main use case

flixr:
Sounds OK to me :-)

flixr:
Some remaining things:
- saving the layout removes `<!DOCTYPE layout SYSTEM "layout.dtd">`
- it asks about saving ac_id even if there are no papgets
- removing a papget doesn't really work, you can drag it out of the map, but it is still there somewhere and saved to the layout...

gautierhattenberger:
Points 1 and 2 should be okay. For the last one, removing messages papgets works for me (and they are not saved afterwards). For other papgets like buttons, it was not possible to remove them anyway (and I don't plan to do it in this pull request).

flixr:
Ok, deleting a papget via the popup edit dialog works indeed.
Somehow I was under the impression that you can also remove a papget by dragging it out of the map...
That is not the case, but it would be good if this could be prevented (don't accept the drop if it is outside of the map, or really remove it in that case)...

gautierhattenberger:
I don't really to try to find the window size, so it is only deleted on left and top side, which are the places you have to chance get them back. On right and bottom side, you can try to make the window bigger to find your papget again...

flixr:
Would have been nice to somehow get the map window size where you can place papgets.
But I guess this is good enough for now and there are more important problems to work on...

gautierhattenberger:
Found it ;)

flixr:
nice :-)
Works and looks good to merge to me...

PR #1069, opened 2015-01-16
[build] use same float_abi for linking as for compiling

flixr:
Make sure we link with same float abi (softfp, vfp) as we compile.
This *should* fix errors like
```
/usr/bin/ld: error: ap.elf uses VFP register arguments, foo.o does not
/usr/bin/ld: failed to merge target specific data of file foo.o
```
Not tested...


gautierhattenberger:
looks good

PR #1073, opened 2015-01-22
Python ivy msg interface

### LINKED ISSUE #1072: ac_id is converted to int in messages_tool.py
At around line 69, the messages_tool converts the ac_Id to an int in the messagesinterface class. This generates a ValueError exception, which is then passed. the messages tool is a python port of the messages window and usually only displays telemetry info, which is why it's probably the correct behavior.
In a different script in python I wanted to reuse the messagesinterface class (the class connecting python to ivy), but also forward messages of any type (ground, datalink, etc). 
We should probably include an option to the initializer to indicate which type of messages should be passed (a list of the types?), which is then matched against the ac_id to see if it should call the callback, then pass it as a string?

flixr:
Rewrote IvyMessagesInterface to also pass the msg_class to the callback and tried to clean it up a little.

Should close #1072


flixr:
For messages of  msg_class `telemetry` you always have the AC_ID on the ivy bus before msg name and then the msg contents/values.

This currently tries to extract the ac_id for messages of the `ground` class as well, but since these are actually part of the message it probably doesn't make that much sense...
Here the ac_id does not specify the sender as in the telemetry case, but e.g. the target it should be sent to.

So I would probably rather set the ac_id to 0 for ground messages and return the msg values as is (which in most but not all cases include the ac_id).
@gtoonstra would that be ok for you?

gtoonstra:
That works for now. The messages_tool afaik is only used for hte messages window and in our case as a proxy between ivy and redis. So it's only really required for telemetry class messages. We should probably introduce a callback per msg class, so that telemetry class gets pushed to a different callback having the ac_id, whereas datalink and ground classes go to a different one that makes more sense for that class of messages?

PR #1074, opened 2015-01-23
[module] rewrite nav_bungee_takeoff

gautierhattenberger:
shorter, with less trigo and more vectors
tested in simulation, to be done in real flight

### DIFF
--- a/sw/airborne/modules/nav/nav_bungee_takeoff.c
+++ b/sw/airborne/modules/nav/nav_bungee_takeoff.c
@@ -14,9 +14,8 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with paparazzi; see the file COPYING.  If not, write to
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/nav/nav_bungee_takeoff.c
```diff
@@ -22,198 +22,183 @@
...
-    <define name="Speed" value="15" unit="m/s"/>
-    <define name="Distance" value="10" unit="m"/>
-    <define name="MinSpeed" value="5" unit="m/s"/>
-    </section>
-    @endverbatim
-*/
-
-#ifndef Takeoff_Distance
-#define Takeoff_Distance 10
+/**
```
flixr: Would make sense to put this description in the doxygen file header so it is put in the detailed description in the generated docs....

flixr:
Except for the dox it looks good to me (without having tested it)....

gautierhattenberger:
http://youtu.be/4RjtJ3CbZ60

PR #1076, opened 2015-01-27
[ocaml] try to live in harmony with the garbage collector

flixr:
Make sure that OCaml can garbage collect unused blocks from the heap when interfacing C code.

With this "fix" it looks like memory usage of messages, etc. doesn't increase so drastically anymore.
There are probably a few more places where we should do the same (e.g. in ivy-ocaml, see https://github.com/flixr/ivy-ocaml/commit/282d68f750bdeaaf66bfcb9c4aa3270e6c2c65dd)

See http://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#sec440 for the OCaml documentation on this.

### DIFF
--- a/sw/lib/ocaml/convert.c
+++ b/sw/lib/ocaml/convert.c
@@ -27,14 +27,16 @@
 #include <fcntl.h>
 #include <sys/termios.h>
 #include <stdio.h>
-#include "caml/mlvalues.h"
-#include "caml/alloc.h"
-#include "inttypes.h"
+#include <caml/mlvalues.h>
+#include <caml/alloc.h>
+#include <caml/memory.h>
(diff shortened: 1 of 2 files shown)


OpenUAS:
Cool if it would be merged asap

flixr:
Still investigating where and how much of a difference it makes...

flixr:
So this doesn't fix the issue of OCaml agents needing increasing memory over time, but already helps a bit.
E.g. link still more or less suddenly needs more than double the memory after approx 1h, seems to happen a bit later with this fix.
Merging as this here should be done in any case...

flixr:
Closed with 1c48e173a0847755176255aaf0f8df1e5ce2117c

Issue #1060, opened 2015-01-02
No option to set BODY TO GPS


OpenUAS:
With the event of more acurate UA position options like Piksi and also using Vicon and optitrac systems there is no setting for the airframe to set what we consider the Mid point of the aircraft a body_to_gps fot xyz or something could be introduced.

As example on my 4m plane there would be a 2 meter difference between DGPS sensor and midwheel landing gear.

PR #1080, opened 2015-01-29
[ins] consider BODY_TO_GPS translation

flixr:
If INS_BODY_TO_GPS_X|Y|Z is defined (in cm!) use it to correct for this offset in the INS gps position update.

start solving #1060

Todo:
- [ ] also consider this offset when (re)setting local origin
- [ ] consider this offset/lever for velocity updates (when angular rates != zero)


OpenUAS:
Nice, good start! s*hitty it was created so fast ,now I really needto repair my big airframe to testfly in spring ;)

PR #1082, opened 2015-01-30
Improvement of Ivy efficiency

gautierhattenberger:
Improving efficiency of Ivy with simpler attached regexp by default (no timestamp), and by only waiting for ALIVE to detect new plane in 'messages'.

When using timespampped messages (or `-local_timestamp` option of link), other programs (GCS, server, messages) have to run with the `-timestamp` option.

When messages is used with something else than an autopilot (or anything sending the `ALIVE` message), but still sending telemetry message, the option `-force` can be used to force waiting on all messages to find new senders.

Tested and working in sim and real link. CPU and network usage slightly improved, even with a single aircraft.

### DIFF
--- a/sw/ground_segment/cockpit/gcs.ml
+++ b/sw/ground_segment/cockpit/gcs.ml
@@ -338,6 +338,7 @@ and display_particules = ref false
 and wid = ref None
 and srtm = ref false
 and hide_fp = ref false
+and timestamp = ref false
 
 let options =
   [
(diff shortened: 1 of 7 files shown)


flixr:
Haven't done any performance statistics, but looks good to me.
Most people don't use the timestamp feature anyway so it makes sense to disable that by default...

I guess where people might find that they don't get any messages anymore (without adding `-force`) is with some of the test progs that don't send ALIVE... but since these were mainly written for testing/debugging and not commonly used it is ok for me.

PR #1087, opened 2015-02-05
[abi] allow to pass variables by value

flixr:
So far the generated ABI callbacks always had a signature with `const type *var` where type was e.g float.
Now the generated signature is simply `type var`.
To pass const pointers again, set it accordingly in abi.xml, e.g. `type="const float *"` instead of `type="float"`

### DIFF
--- a/sw/airborne/modules/sonar/agl_dist.c
+++ b/sw/airborne/modules/sonar/agl_dist.c
@@ -50,7 +50,7 @@ float agl_dist_value_filtered;
 
 abi_event sonar_ev;
 
-static void sonar_cb(uint8_t sender_id, const float *distance);
+static void sonar_cb(uint8_t sender_id, float distance);
 
 void agl_dist_init(void)
 {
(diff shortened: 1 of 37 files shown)


sw/airborne/modules/sonar/agl_dist.c
```diff
@@ -63,10 +63,10 @@ void agl_dist_init(void)
 }
 
 
-static void sonar_cb(uint8_t __attribute__((unused)) sender_id, const float *distance)
+static void sonar_cb(uint8_t __attribute__((unused)) sender_id, float distance)
 {
-  if (*distance < AGL_DIST_SONAR_MAX_RANGE && *distance > AGL_DIST_SONAR_MIN_RANGE) {
-    agl_dist_value = *distance;
+  if (distance < AGL_DIST_SONAR_MAX_RANGE && *distance > AGL_DIST_SONAR_MIN_RANGE) {
```
fvantienen: there is a second *distance here that needs to be removed

flixr:
thx @fvantienen, fixed..

gautierhattenberger:
Looks fine for me

PR #1088, opened 2015-02-05
Added new RC USB type joystick from Hobbyking

OpenUAS:


flixr:
renamed to hobbyking in #1229

Issue #780, opened 2014-07-26
spektrum binding not working reliably on F4


flixr:
Seems that the binding procedure for spektrum radio control is not reliably working on STM32F4.
Seems to work OK on apogee (with additional polarity inverter on the spektrum input which is set to non-invert however).
Looks like there is a short pulse on the pin when after the expected binding sequence pulse train is sent. Probably when switching the pin from output back to input with gpio_setup_pin_af.

PR #1090, opened 2015-02-09
[stm32f4_spektrum] Fixes the additional glitch pulse.

esden:
If you set the alternate function before setting the pin to input AF.
The GPIO will pull down the line down very briefly. This will cause an
additional pulse to be detected by a spektrum satellite if it is being
reconfigured as a UART RX after sending bind mode pulses.

This was tested on Lisa/MX 2.1. It is unlikely but might cause issues for other STM32F4 autopilots. This is why I am sending this in as a pull request instead of pushing the patch directly.


fvantienen:
Noce fix!

flixr:
Nice, should fix #780
I can probably test this tonight, but also only on a Lisa/MX (2.0).
Don't really expect this to cause any issues either...

fvantienen:
I will test this on a navstik also.

flixr:
Ok, works nicely on my Lisa/MX 2.0
@fvantienen do you still want to test this on navstik before we merge it to master?

PR #1092, opened 2015-02-10
Meteo stick

gautierhattenberger:
Update meteo stick module.

The main purpose of the pull request is to cross-check that the changes in SPI input/ouput_length index (from uint8 to uint16) is okay.

### DIFF
--- a/conf/modules/meteo_stick.xml
+++ b/conf/modules/meteo_stick.xml
@@ -10,10 +10,11 @@
       - Temperature in ADC (ADS1220)
       - Humidity in ticks (period of the input signal)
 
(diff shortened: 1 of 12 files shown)


flixr:
I don't see why it would hurt and SPI communication with IMU still works as expected on my Lisa/MX.
Might have been nice to have the `cond` feature for modules defines in a separate commit with description though...

gautierhattenberger:
true, I realized it afterwards...
Can I still merge when Travis is done ?

flixr:
I would prefer to pick it out into a separate commit...

Issue #1079, opened 2015-01-29
cleanup UDP interface/implementation


flixr:
We have:
- arch/linux/mcu_periph/udp_arch.[ch]
- fms/fms_network.[ch]
- modules/computer_vision/lib/udp/socket.[ch]

fvantienen:
From the top of my head only "arch/linux/mcu_periph/udp_arch" is used now in the airborne code. Except for some computer vision modules. Fms can be removed totally if I'm correct.
I think the "arch/linux/mcu_periph/udp_arch" is the way to go, because it has the same interface as in UART etc. So you can switch between those 2 easily.

Next to this we should also check own socket implementations, which don't use any of these and replace those to make it nicer.

flixr:
@fvantienen fms_network is still used in natnet2ivy...
I tried to convert to use udp_socket instead with https://github.com/flixr/paparazzi/commit/638fc0f2a7fcef3cd23c3df6780497f90346c963, would be nice if you could confirm that natnet2ivy still works...

OpenUAS:
Needs testing in Cyberzoo. Will re-assign or report if not working

flixr:
should be finally closed with merge of #1162 which removed the fms stuff

PR #1094, opened 2015-02-11
Linux video update

dewagter:
- [x] opticflow
 - [x] viewvideo / videorecord
 - [x] last mutex problem: possible lost image
 - [ ] opticflow use new functions
 - [x] image_nc_send use new functions
 - [ ] fix more settings for the viewvideo module to be changed at runtime
 - [ ] fix bebop CMOS configuration failing sometimes

This also removes some of #1079 which was inside the vision framework.


gautierhattenberger:
the file image_nc_send also need to be updated to use the new API, it should be very similar to viewvideo.c

dewagter:
any objection to merge this?

flixr:
Is your third point "last mutex problem: possible lost image" fixed/done?
Without looking at the code I have no idea what the intended changes are, would have been nice to write that in the pull request ;-)

fvantienen:
It was in there. But it is during the image free, you set the index to 255 (which means none dequeued, define should be nice) if the current index is the only one dequeued. Problem is that during this check the image thread can change this idx, and you will lose this buffer.

flixr:
@dewagter your call... if you want to merge it with this last point unresolved, go ahead...

fvantienen:
I will fix it tomorrow with a simple extra mutex lock. Was searching for a nicer solution, but couldn't find it.
But since this lock can only happen at very little occasions I think it shouldn't be a problem to just use another mutex.

fvantienen:
Maybe we should also rename the different modules?
We now have names like "cv_opticflow.xml", "video_rtp_stream.xml" and "image_nc_send.xml". Maybe start everything with cv_?

fvantienen:
Currently the CMOS configuration of the Bebop sometimes fails. Don't know how to fix it yet, but hope to have more information soon.
Next to that I changed a lot of the jpeg/rtp function names to avoid double function names.

OpenUAS:
while cv_ is a fine name change from our programmers perspective, however out of experience with students and others coding, a lot of them will confuse it thas to do with Open CV, I would propose something else if only I had some beter ideas anyhow, e.g. iavh_ (**i**mage **a**nd **v**ideo **h**andling) , whatever.... at least something distinctive even fvt_ , cdw_ ;) then others can make up the real meaning of the characters

fvantienen:
I removed the nc_send_image module, because it was 99% the same as the viewvideo module. Instead I made a setting "VIEWVIDEO_USE_NC" that does the same as the nc_send_image module.

flixr:
Can you please add the VIEWVIDEO_USE_NC flag to the description in the module xml?
Maybe calling it VIEWVIDEO_USE_NETCAT would also be a bit more clear?

fvantienen:
I changed the define, if you got other comments please tell them ;) It's still a WIP, but I want to merge this  as soon as possible, because this will help a lot of people in our computer vision course.

flixr:
I think we can merge it after #1117 is (good to be) merged...

flixr:
Actually I don't think it makes sense to use udp_periph in `rtp_send` or any of these modules that don't actually use any of the UART like periph functions...

dewagter:
cv was for computer vision. for me streaming video does not need to be
called computer vision but I would not mind a lot either.

-Christophe

PR #1095, opened 2015-02-11
Logger modules refactor

dewagter:
make modules more consistent


flixr:
Please set the proper name in the logger xml files and set dir="loggers"

gautierhattenberger:
At some point, we'll need a proper generic log interface. Which is already the case for uart and SDIO with chibios on Apogee, so you can write pprzlog messages directly like you would do with telemetry messages.

flixr:
Airframes which currently have openlog.xml loaded need to be changed to use logger_telemetry instead...

But I wonder if this wouldn't rather be named something like `telemetry_timestamp.xml` or `send_timestamp.xml`, as it doesn't do any logging directly.

Also instead of incrementing the timestamp in the periodic_2Hz_openlog, it would probably make more sense to simply use `get_sys_time_msec()`.
We could maybe even remove this module completely and simply add a periodic message for this... but maybe the module is still more convenient?

flixr:
What's the difference between direct_memory_logger.xml and high_speed_logger_direct_memory.xml (now renamed to logger_dataflash.xml)?
Also navstik is failing with the change to logger_dataflash.xml...

dewagter:
Hmmm... ik only know the logger that writes directly to a SPI dataflash
chip.

-Christophe

flixr:
@fvantienen do you have some input on this?

dewagter:
isn't direct_memory_logger.c by @fvantienen  not just an efficient rewrite of the original high_speed_logger_direct_memory.c by C.Roblot?

To me it should be called ```logger_dataflash(_chip)```

dewagter:
apparently the difference is that they use a different dataflash chip

fvantienen:
Mine was indeed a sort of rewrite and made use of the peripheral structure. So it is best to rewrite ```high_speed_logger_direct_memory``` from C. Roblot also inside of a peripheral, so that they can use the same module. I don't have the time to do this.
But what would be more useful is to make it easier to define/change what you want to log inside the different loggers. Because now you have to change the c and header files for that, which isn't really a nice solution. So maybe a simple xml file, which is excluded from git, or something like that? Inside the computer_vision folder we already got another copy, with different values.

PR #1097, opened 2015-02-11
Dynamic Ad-Hoc networking with OLSR on Ardrone2

gautierhattenberger:
Add an extra wifi mode with necessary tools to use OLSR (http://www.olsr.org/) on ARDrone2 platforms.


flixr:
I guess the MODEM_HOST should be adapted for bebop as well...

gautierhattenberger:
argh, forgot this one

PR #1099, opened 2015-02-12
changes to airspeed_ets module

alonsoac:
-allow using eagletree airspeed sensor in third-party mode
-store airspeed to state when MEASURE_AIRSPEED is defined

### DIFF
--- a/sw/airborne/modules/sensors/airspeed_ets.c
+++ b/sw/airborne/modules/sensors/airspeed_ets.c
@@ -156,6 +156,7 @@ void airspeed_ets_read_event(void)
 
   // Continue only if a new airspeed value was received
   if (airspeed_ets_valid) {
+#ifndef AIRSPEED_ETS_RAW      
     // Calculate offset average if not done already
     if (!airspeed_ets_offset_init) {
       --airspeed_ets_cnt;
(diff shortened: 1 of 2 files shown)


flixr:
The `AIRSPEED_ETS_RAW`flag would probably be better named `AIRSPEED_ETS_3RD_PARTY_MODE` or so and the docs need to be updated then.

We definitely don't want to set the airspeed in the state interface if the airspeed from the ETS is not used for control (as you might easily have another sensor that provides this to the state interface).
That is in fact the only reason why there is the `USE_AIRSPEED_ETS` flag.
See also the generated docs: http://docs.paparazziuav.org/latest/module__airspeed_ets.html

alonsoac:
I got confused by the MEASURE_AIRSPEED constant in fixedwing/autopilot.c. This doesn't seem to be used anywhere and I don't see when it would make sense to use it.

alonsoac:
I think the change is complete now.

flixr:
I already removed the MEASURE_AIRSPEED flag, it was basically just used to not send an AIRSPEED message if you don't have that defined.
But if you really don't have an airspeed sensor and want to save bandwidth, it makes more sense to just remove that message from the telemetry.xml file...

sw/airborne/modules/sensors/airspeed_ets.c
```diff
@@ -156,6 +156,7 @@ void airspeed_ets_read_event(void)
 
   // Continue only if a new airspeed value was received
   if (airspeed_ets_valid) {
+#ifndef AIRSPEED_ETS_3RD_PARTY_MODE
```
flixr: it makes more senses to check flags like this with `#if !AIRSPEED_ETS_3RD_PARTY_MODE`

PR #1102, opened 2015-02-12
Dual dual pwm

dewagter:
developments that should eventually make it to master...


sw/airborne/arch/stm32/subsystems/actuators/actuators_dualpwm_arch.c
```diff
@@ -153,13 +154,13 @@ void dual_pwm_isr(void)
   if (num_pulse == 1) {
 
     set_dual_pwm_timer_s_period(ratio_16ms);
-    set_dual_pwm_timer_s_oc(actuators_dualpwm_values[FIRST_DUAL_PWM_SERVO]);
+    set_dual_pwm_timer_s_oc(actuators_dualpwm_values[2],actuators_dualpwm_values[4]);
```
flixr: Seems to be hardcoded now... probably not what we want...
If you really want to keep it hardcoded, at least remove the FIRST_DUAL_PWM_SERVO and SECOND_DUAL_PWM_SERVO defines...

dewagter:
changed ```FIRST_DUAL_PWM_SERVO``` to ```DUAL_PWM_SERVO_5_P1``` and ```DUAL_PWM_SERVO_5_P2```

This should better indicate that the dualpwm pulse comes out of lisa-s servo marked ```servo5```

Unfortunately it is currently not possible to put this defines for servo channels in your airframe file: see #1108

So one should lookup in 'lisa-s.h' that index 0 and 1 pulses come out of servo5 and index 2 and 3 pulses come out of servo6

conf/airframes/BR/DelFlyDualPWMservo.xml
```diff
@@ -0,0 +1,276 @@
+<!-- this is a LadyBird quadrotor frame equiped with Lisa/S 1.0 -->
```
flixr: guess it would make sense to provide proper description here (at least remove the ladybird stuff that is not correct)

flixr:
What about the `FIRST_DUAL_PWM_SERVO` defines for LisaM?

dewagter:
```FIRST_DUAL_PWM_SERVO``` defines were removed. They do not fit in the airframe.xml either and they are non-descriptive. Use ```DUAL_PWM_SERVO5_P1``` en ```P2``` instead.

flixr:
I meant the lisa_m_common.h file which still has the old defines...

dewagter:
Does anybody know where the ```FIRST_DUAL_PWM_SERVO``` defines were ever used?

flixr:
Seems they were added with #678, no idea if they were actually ever used on LisaM...

flixr:
Could you plz update the description in DelFlyDualPWMservo.xml, otherwise it looks good to merge to me...

PR #1103, opened 2015-02-13
[rotorcraft] fix attitude_set_rpy_setpoint

flixr:
Use quat_of_eulers instead of quat_from_rpy_cmd to match what users expect from attitude_set_rpy_setpoint.
Specifically actually make the attitude primitive in the flight plan work as expected.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_float.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_float.c
@@ -203,7 +203,7 @@ void stabilization_attitude_set_rpy_setpoint_i(struct Int32Eulers *rpy)
   // copy euler setpoint for debugging
   EULERS_FLOAT_OF_BFP(stab_att_sp_euler, *rpy);
 
-  quat_from_rpy_cmd_f(&stab_att_sp_quat, &stab_att_sp_euler);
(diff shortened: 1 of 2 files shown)


flixr:
@dewagter @fvantienen is the current ardrone2_opticflow_hover.xml airframe working properly?
Since the opticflow hover module also uses attitude_set_rpy_setpoint, I strongly suspect that it doesn't work properly without this fix...
Or am I missing something?

EwoudSmeur:
Shouldn't the function now be renamed to stabilization_attitude_set_euler_setpoint_i?
What happens when the command is not a correct euler combination, e.g. pitch > 90 degrees?

flixr:
I didn't want to rename the function to to euler, because it wasn't supposed to take "real" euler angles as input.
But calling `quat_of_eulers` instead of `quat_from_rpy_cmd` at least works for the normal cases.
Properly fixed `quat_from_rpy_cmd` would be appreciated ;-)

flixr:
@dewagter @OpenUAS @gautierhattenberger (or anyone who can test the ardrone2_opticflow_hover) do you have any feedback here?

flixr:
Feedback timeout: merging into master as this clearly fixes usage of the attitude flight plan primitive and except from the ardrone2 opticflow hover doesn't seem to be used anywhere else.
I guess we should backport this for v5.4 as well...

PR #1107, opened 2015-02-14
Add ChibiOS HAL

podhrmic:
sw/airborne/arch/chibios
- no makefiles at this point (so it is structured)

### DIFF
--- a/conf/boards/lisa_mx_rt.makefile
+++ b/conf/boards/lisa_mx_rt.makefile
@@ -1,0 +1,102 @@
+#
+# Copyright (C) 2013 AggieAir, A Remote Sensing Unmanned Aerial System for Scientific Applications
+# Utah State University, http://aggieair.usu.edu/
+#
+# Michal Podhradsky (michal.podhradsky@aggiemail.usu.edu)
+# Calvin Coopmans (c.r.coopmans@ieee.org)
+#
+# credits to ENAC team for initial work on ChibiOs and Paparazzi integration
+#
+# This file is part of paparazzi.
+#
+# paparazzi is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2, or (at your option)
+# any later version.
+#
+# paparazzi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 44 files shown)


flixr:
Seems like the updated ChibiOS has some issues...
Do some of the added files in the HAL require this version?

podhrmic:
Apparently some minor bug related to changes in ChibiOS/build process/board settings - probably not important right now, as the changes are not really needed. What about now?

flixr:
You can also force push to the branch and simply remove the commit updating the submodule...

flixr:
I think it would make sense to add the makefiles and e.g. a LisaMX board file as well, so that we can run some of the test progs using ChibiOS....

flixr:
Source and header files should not go into conf...

conf/boards/lisa_mx_rt.makefile
```diff
@@ -0,0 +1,102 @@
...
+# along with paparazzi; see the file COPYING.  If not, write to
+# the Free Software Foundation, 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+#
+
+# Board definition
+BOARD=lisa_mx
+BOARD_CFG=\"../../conf/boards/$(BOARD)/board.h\"#actual chibios defines
+# Arch definition
+ARCH=chibios# for backwards compatibility with the paparazzi make process
```
flixr: should be just `ARCH=chibios`

conf/boards/lisa_mx_rt.makefile
```diff
@@ -0,0 +1,102 @@
...
+CHIBIOS_BOARD_PORT = ARMCMx/STM32F4xx/port.mk
+CHIBIOS_BOARD_LINKER = STM32F407xG.ld
+
+##############################################################################
+# Compiler settings
+#
+MCU  = cortex-m4
+
+# -----------------------------------------------------------------------
+# include Makefile.chibios instead of Makefile.stm32
```
flixr: Not needed, if `ARCH` is correctly set to `chibios`

flixr:
Don't see where the added baro_board.h would be needed...

flixr:
Created the rt_chibios branch for integration, merged your stuff in there (and moved the board files from conf to sw/airborne). Also ran fix_code_style.sh on the arch/chibios files...
This should make it easier to run some tests and lets anyone make pull requests against this rt_chibios branch before we merge it to master...

PR #1109, opened 2015-02-15
implement persistent settings for linux arch

### LINKED ISSUE #969: persistent settings for omap/linux arch
Add support for persistent settings for the omap/linux architecture (e.g. for Parrot ARDrone2, Bebop).

`persistent_write` and `persistent_read` are currently dummy no-ops for that arch.

flixr:
Write persistent settings to a binary file (by default `pprz_persistent_settings.binary` in same directory).

On loading it also check the file size matches the PersistentSettings struct size to detect if persistent settings were added/removed). While this doesn't catch all errors (like replacing a setting of same size), it should be good enough for most cases...

Should close #969


gautierhattenberger:
I looks good to me (although I can't test it this week).

flixr:
Tested and works on the BBB...
Originally I wanted to write a human readable/editable settings file, but this was the easiest for now without adding extra dependencies...
Merging, if someone comes up with a better validation procedure we can still add it later.

PR #1110, opened 2015-02-16
[arch/linux] fix uart driver

### LINKED ISSUE #743: UART missing messages in the ARDrone
The UART seams to be missing around half of the data when using the regular paparazzi library. When trying to send data with an arduino connected to the USB port of the ARDrone I can get around 2s of data, then I lose 2s of data, then back 2S. The value of 2s is what I was able to measure in my case. I also got a lot of picks in the data.

To get those result I was using an arduino that send a counter that was incremented whenever a message was send. The resulting telemetry measurements are those :

I can go around that issue by using directly the serial port as a file, using code taken from the fireswarm project. In that case the results are as expected :

The test code is available as a module (Arduino2Ardrone2) in the branch UARTbug in my fork of papaprazzi : karlito139/paparazzi. The code used in the arduino is available in the repo : karlito139/arduino2ardrone2 in the branch counter.

flixr:
Make sure the output flags are properly reset for raw mode.
Specifically OPOST, clear some others as well to be safe...

might also fix #743
@karlito139 @OpenUAS would be great if you could also test this...

### DIFF
--- a/sw/airborne/arch/linux/mcu_periph/uart_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/uart_arch.c
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <errno.h>
 
 #include "serial_port.h"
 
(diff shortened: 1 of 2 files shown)


flixr:
Merging as it definitely fixes some issues discovered running the tests on the BBB...

PR #1115, opened 2015-02-18
[chibios] Added test_serial and test_shell

podhrmic:
- plus a minor cleanup again


podhrmic:
OK, so here is the error: https://travis-ci.org/paparazzi/paparazzi/jobs/51302659#L2213

Basically, I added USE_CHIBIOS_RTOS defines into mcu_periph/uart.c (https://github.com/podhrmic/paparazzi/commit/744e07d49d3e0ba2a27d351bf3388034baa78131#diff-1cc57dd8ba43fe546e5c21c2636c846dR225)
in order to keep the generic uart interface (```uart_getch()```). The problem is, that the *chibios-libopencm3* version doesn't have that, so now is colision in defines.

Suggested solutions:
* dont change the uart.c file, and use ```sdGet()``` in the flight code instead (see example https://github.com/podhrmic/paparazzi/commit/744e07d49d3e0ba2a27d351bf3388034baa78131#diff-95b7b048aa847135884a5d219462d30aR88)
Drawback - less unified code (ChibiOS vs other archs)
* use different defines for Chibios (USE_CHIBIOS_RTOS) and chibios-libopencm3 (USE_CHIBIOS_LIBOPENCM3) - drawbacks: a lot more ```#if else #enedif``` defines

flixr:
I think it is easier (and cleaner) to just move the uart_getch function into the respective arch...

flixr:
Not a fan of creating symlinks to avoid having the same filename twice.
This is a stupid build rule in the makefile, rather put them into the appropriate subdirs.

conf/Makefile.chibios
```diff
@@ -137,6 +138,8 @@ CSRC = $(PORTSRC) \
        $(BOARDSRC) \
        $(CHIBIOS)/os/various/evtimer.c \
        $(CHIBIOS)/os/various/syscalls.c \
+       $(CHIBIOS)/os/various/chprintf.c \
+       $(CHIBIOS)/os/various/shell.c \
        $(CHIBIOS_BOARD_MAIN) \
        $($(TARGET).srcs)
```
flixr: instead of adding `$($(TARGET).srcs)` to CSRC, add them to `ECSRC` like in the Makefile.chibios-libopencm3
Then you don't need to make symlink to have different filenames...

flixr:
The chibios_test_shell.c would probably also be nicer directly under test instead of test/mcu_periph, as it is not part of mcu_periph...

flixr:
Thx, squashed into two commits and merged...

flixr:
Merged with 888c9243f1dc77ab2e6fd3339cc2b01b5e9063ad and 3f3b9376a11d47318b0dff923784f385b8e88287

PR #1116, opened 2015-02-20
[Setup Actuators] Fix telemetry and add more boards.

esden:
The firmware was crashing because the transport was not being
initialized. Added example airframe files for Lisa/MX V2.0 and
V2.1. As well as appropriate example sessions.

### DIFF
--- a/conf/airframes/examples/setup_lisamx.xml
+++ b/conf/airframes/examples/setup_lisamx.xml
@@ -1,15 +1,15 @@
 <!DOCTYPE airframe SYSTEM "../airframe.dtd">
 
-<!-- this airframe file contains setup firmwares for Lisa/M V2.0 -->
+<!-- this airframe file contains setup firmwares for Lisa/MX V2.1 -->
 
-<airframe name="setup_lisam2">
+<airframe name="setup_lisamx">
 
   <firmware name="setup">
-    <target name="tunnel"           board="lisa_m_2.0"/>
-    <target name="usb_tunnel"       board="lisa_m_2.0">
+    <target name="tunnel"           board="lisa_mx_2.1"/>
+    <target name="usb_tunnel"       board="lisa_mx_2.1">
       <configure name="TUNNEL_PORT" value="UART3"/>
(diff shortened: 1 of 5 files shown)


flixr:
Committed the downlink_init() fix (and also added the missing udp/uart_event which is needed to run on the linux arch) with 5af2f84f72e12ffd819ffc846a9caec271681408

Is it really useful to duplicate the setup_lisam file it twice with only the board changed?
And to have three example sessions for something that most people will never need/use?
IMHO it is not...

esden:
I think most people actually do need it. It is necessary to configure, teach and test your motor controller setup what minimum and maximum values are, testing the rotational direction and which motor is connected where. Some of these things can be done using a servo tester but not everyone has them or wants to use them. Teaching the min max values using a servo tester can be very tedious too. Resulting from that almost everyone who has a quadcopter with pwm esc "needs" this. It actually needs a better interface, probably a dedicated tool to make that setup easier, but these sessions are for now the "easiest" way of setting up your ESCs.

flixr:
Hm... There is no nice way to add the same target (`setup_actuators`) for multiple boards in one airframe file as it would require board specific target names...
The only way of not duplicating the setup airframe files is to handle it like we did so far: require the user to change the board according to what they actually use...

@gautierhattenberger do you have any better ideas?

gautierhattenberger:
I don't really see a better solution using a dedicated airframe/firmware.
I guess we could imagine a module loadable in a regular firmware that can temporarily disable the control loops and mixing to set the actuators by hand. This solution is a bit dangerous (you don't want this to do something during the flight), but at least you can directly work with your airframe file.

flixr:
@esden do you think it would be an ok solution to just keep one setup_lisam.xml airframe file and one session? As an end user you then only have to change the board type as appropriate and can otherwise always use the same file/session...

flixr:
@esden can we close this one, or is there still something here that you would like to see merged?

esden:
@flixr no I think we have it all. :)

PR #1117, opened 2015-02-23
[arch/linux] sys time thread

flixr:
Since using SIGALRM to call the sys_tick_handler (via `setitimer`) can cause problems like interrupting some other syscalls (i.e. read,write, ioctl), setup a separate thread and use timer_fd instead.
That is a Linux-specific set of functions that presents POSIX timers as file descriptors rather than signals.

If the thread runs with a normal priority it can happen that some timer events are missed (and hence sys_time not updated in time). So added a function to set a higher prio for that thread (needs root rights or properly set up limits for the user).
In some simple tests no events were missed when running with prio 29 via SCHED_FIFO, even if events would be missed, the sys_tick_handler is called the appropriate number of times (so that the time is correct again after)...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Paparazzi 5.5_devel
 
 currently ongoing development, changes so far (no particular order, nor complete)
 
+- arch/linux: change the sys timer to a multi threaded implementation
+  [#1117] (https://github.com/paparazzi/paparazzi/pull/1117)
 - python: generate paparazzi math wrappers with SWIG
   [#1066] (https://github.com/paparazzi/paparazzi/pull/1066)
 - ground segment: use pkg-config for ivy-c if available
--- a/conf/Makefile.linux
+++ b/conf/Makefile.linux
@@ -56,7 +56,7 @@ CFLAGS += -Wredundant-decls -Wreturn-type -Wshadow -Wunused
 #CFLAGS += -Wa,-adhlns=$(OBJDIR)/$(notdir $(subst $(suffix $<),.lst,$<))
 #CFLAGS += $(patsubst %,-I%,$(EXTRAINCDIRS))
 
-CFLAGS += -lm
+CFLAGS += -lm -pthread
 
 # flags only for C
 CFLAGS += -Wstrict-prototypes -Wmissing-declarations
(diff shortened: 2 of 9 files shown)


flixr:
Feedback welcome and tests on ARDrone2 and Bebop needed...
Also maybe we should move the `get_rt_prio` function to a c file (e.g. mcu_arch.c)?

fvantienen:
This works better, but shows us that there are some problems with implementation the IMU of the ardrone2. The navboard of the ardrone2 actually doesn't support "NON_BLOCKING" so must be run in a separate thread. This again adds some multi-threading buffer problems, which I'm working on atm.

The problem that the navboard is actually blocking also means that currently the event loop of paparazzi slows down to the sys_time frequency(by default 1024 hz). Because the reading of the navboard only gets interrupted by the sys_time SIGALRM at the moment.

Also some problems with the navboard reset/sleeping have come to show itself because of this. For example an sleep previously had a maximum of 1/1024 second and then it was interrupted and didn't restart. Now the sleeps are actually the correct time and maybe some stuff could react differently to this.

There could actually be more hidden blocked readings inside the ardrone2 implementation, and maybe some testing of the frequency of the event loop is needed to find them.

fvantienen:
Tested it on both the ardrone2 and bebop and works :)
So for me it is ok to merge this..

fvantienen:
Can we merge this, because the video update pull request relies on this?

flixr:
@gautierhattenberger could you plz also have a quick look at this?

gautierhattenberger:
It seems to work for me (on my desk, not in flight).
Only strange thing: I got a IMU_LOST during one of the test (which is not so unusual) but after a few seconds, it came back to life (first time it actually happens). It is a nice thing (previously it would have be gone for good), but on the other hand the AHRS was stuck for too long to be safe.
Still good to merge I think.

flixr:
The handling of the navdata buffer is not thread-safe yet.
It could happen that you read new input in the `navdata_read` thread while it is currently used in the main thread in `navdata_update`.
I think the parsing and checksum check should also be done in the `navdata_read` thread and then the copied over "atomically" in the event loop if there is new (valid) data indicated by a flag marked as volatile.
Or any other thread safe implementation...

fvantienen:
It currently is thread safe because it reads only in the uart thread when the idx is smaller than the navdata size. The parsing is only done when it is exactly navdata size, and the buffer is also exactly navdata size.
But moving the checksum to the thread would be nice.

flixr:
Rewrote the navdata_read thread and added a condition variable to wait until the buffer has been copied to measure in navdata_update.
Please test again...

fvantienen:
Will test this tomorrow ;)

PR #1118, opened 2015-02-23
[nps] update nps_fdm_jsbsim to be compatible with latest jsbsim

gautierhattenberger:
not the best implementation, but should work with current
paparazzi-jsbsim package (from 2012) and latest jsbsim version


flixr:
Jep, still works fine here with the paparazzi-jsbsim package version 1.4.

PR #1122, opened 2015-03-01
Single udp socket

flixr:
Only create one socket for sending and receiving, instead of two where you use one only to send and the other only to receive.
There should be no reason/need to create two separate ones...

Also add udp_socket.[ch] to make it easier to use only the UDP sockets without the UART like mcu_periph/udp interface around it.
Provides the functions:
- `udp_socket_create`
- `udp_socket_send`
- `udp_socket_recv`
- `udp_socket_recv_dontwait`

Which should make it easy to replace the other udp socket wrappers used in e.g. modules/computer_vision and hence address #1079

### DIFF
--- a/conf/firmwares/subsystems/shared/udp.makefile
+++ b/conf/firmwares/subsystems/shared/udp.makefile
@@ -4,9 +4,11 @@ ifndef UDP_INCLUDED
 
 UDP_INCLUDED = 1
 
-#generic spi master driver
 UDP_CFLAGS = -DUSE_UDP
 UDP_SRCS = mcu_periph/udp.c $(SRC_ARCH)/mcu_periph/udp_arch.c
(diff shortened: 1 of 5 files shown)


flixr:
So it works on the BBB and seems to also work on the ARDrone2.
One remaining question: should we keep the `UdpNetwork` name for the struct or change it to `UdpSocket` or so? (Since it now is only one socket anyway and e.g. the socket.h in the computer vision lib calls it UdpSocket).

gautierhattenberger:
tested and working on ardrone2

lethargi:
dont know if it is related with this, but i am compiling with v5.13 and i get 

```modules/computer_vision/viewvideo.c:113:61: warning: declaration of 'socket' shadows a global declaration [-Wshadow]```

PR #1124, opened 2015-03-02
[gcs] periodically request AIRCRAFTS

### LINKED ISSUE #1078: GCS Window is blank sometimes in simulation
I am working with updated master. Sometimes when I do a simulation I can see messages in the Message window but the GCS stays blank. Closing gcs and running it again always solves the problem.

flixr:
periodically probe for new AIRCRAFTS until we got at least one...
This should fix #1078

### DIFF
--- a/sw/ground_segment/cockpit/live.ml
+++ b/sw/ground_segment/cockpit/live.ml
@@ -41,6 +41,9 @@ let track_size = ref 500
 
 let _auto_hide_fp = ref false
 
+(* request AIRCRAFTS list until got first answer from server *)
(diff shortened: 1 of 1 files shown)


flixr:
Hm... maybe it would be better to stop probing after the first AIRCRAFTS message was received, so you don't keep on probing if server is already up, but no aircraft connected?

flixr:
@gautierhattenberger Maybe more like this now?

gautierhattenberger:
I don't think we can avoid the global ref, so it is good for me

PR #1125, opened 2015-03-02
[rotorcraft] fix arming with kill switch

### LINKED ISSUE #1041: kill_switch_arming not perfect yet:
- on GPS loss -> SAFE -> LAND-DETECT -> KILL: does not unarm -> 3D Fix takes off again
 - ... ?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7056665-kill_switch_arming-not-perfect-yet?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flixr:
Untested, but should fix #1041

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot_arming_switch.h
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_arming_switch.h
@@ -59,7 +59,7 @@ static inline void autopilot_arming_set(bool_t motors_on)
     if (autopilot_arming_state == STATE_MOTORS_ON) {
       autopilot_arming_state = STATE_STARTABLE;
       /* if turned off in an AUTO mode, remember it so it can be turned on again in AUTO */
-      if (autopilot_mode != MODE_MANUAL) {
+      if (autopilot_mode != MODE_MANUAL && autopilot_mode != AP_MODE_KILL) {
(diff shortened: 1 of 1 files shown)


flixr:
@dewagter can you please test if this solves the problem and you get the behaviour you expect?

OpenUAS:
Weather permitting tests performed in weekend

flixr:
@dewagter What kind of behaviour do we want in the case when it goes to FAILSAFE but does't KILL on landing (e.g. because you haven't enabled that)?
The currently proposed fix allow immediate rearming from FAILSAFE, should we also prevent it that case?

dewagter:
If the condition that triggered the failsafe is still present, then it
should remain failsafe, otherwise immediate rearming is acceptable.

-Christophe

flixr:
Checking which condition triggered failsafe would be a pain, but if the problem doesn't exist anymore (e.g. GPS lost), then the quad should go into the mode as set on the RC again...
So I added a check for the FAILSAFE mode.

PR #1128, opened 2015-03-03
[arch/linux] sys_time: get time from CLOCK_MONOTONIC

flixr:
Instead of simply adding up the sys_time ticks, seconds, get current time from CLOCK_MONOTONIC and directly set sys_time from that (difference to clock monotonic time at startup).


gautierhattenberger:
tested and working on ardrone2

Issue #1046, opened 2014-12-17
GCS Speech spams GPS accuracy


EwoudSmeur:
When the GPS accuracy is between ~15 - 25m the GCS speech program will tell everytime the accuracy changes a bit. This results in spamming of the first words of the sentence, not even hearing what the accuracy even is. Requiring a change of more than ~5m could solve this issue.

The underlying issue is that the speech program repeats messages before it has finished speaking (same with low battery, it spams the name of the vehicle without ever reaching the battery voltage statement)

flixr:
Would probably require to save the last announced pacc per aircraft and then check that in `get_svinfo` in live.ml

flixr:
@EwoudSmeur @OpenUAS if you are happy with the solution of #1129 we can backport it to v5.4 and close this issue...

flixr:
backported to v5.4 with 3e4169c

PR #1129, opened 2015-03-04
GPS accuracy speech less verbose

flixr:
Addresses #1046

GPS accuracy is considered ok if better than 10m, very bad if worse than 20m and simply low in between.
Now only if the accuracy falls into a different range than previously reported.


flixr:
Merging as it should solve the problem.
@EwoudSmeur @OpenUAS if you are happy with this, we can backport it to v5.4 stable

EwoudSmeur:
Looks good to me!

PR #1130, opened 2015-03-06
Refactor AHRS/INS with ABI (and provide dual AHRS support)

flixr:
Continue refactoring of AHRS and INS:
- add GEO_MAG ABI message
- add GPS ABI message
- remove explicit calling of update_gps functions (replaced by GPS ABI callbacks)
- each ins has it's own unique struct/function names
- send STATE_FILTER_STATUS from each implementation (with added id if you run multiple ones)
- use stateIsAttitudeValid() instead of AHRS.is_aligned
  - The state interface is initialized with invalid attitude.
    It becomes valid as soon as an attitude is set (via one of the stateSetNedToBodyX functions).
    This should be only done by an AHRS/INS after it is aligned.
- add possibility to run multiple (currently two) AHRS implementations and switch which one should push the output to the state interface during runtime

`ahrs_init()` and `ins_init()` are now sort of "dispatcher" functions, that init/register the actually used implementation.

As an example using `float_mlkf` as the `PRIMARY_AHRS` and `int_cmpl_quat` as `SECONDARY_AHRS`, in your firmware section of the airframe file:
```
  <subsystem name="ahrs"          type="float_mlkf"/>
  <subsystem name="ahrs"          type="int_cmpl_quat">
    <configure name="SECONDARY_AHRS" value="int_cmpl_quat"/>
  </subsystem>
```

In `ahrs_init()` it calls the ahrs_x_register functions for both. Each of those calls their own ahrs_foo_init and then `ahrs_register_impl(enable_output_function)`, then do the ABI binding and register periodic telemetry functions.

The "enable_output_function" pointer is used to keep a reference to the implementation functions to switch on/off publishing of the output to the state interface.
Hence you can call `ahrs_switch(idx)` at runtime to switch the output of the AHRS implementations.
E.g. with idx=1 switch to the output of the secondary AHRS (with 0 being the primary)
Add the `conf/settings/estimators/ahrs_secondary.xml` settings file to switch them via settings.

Tested on a Lisa/MX 2.0 for those two and seems to work nicely...

TODO:
- [x] provide the enable functions for each implementation
  - [x] int_cmpl_quat
  - [x] float_mlkf
  - [x] int_cmpl_euler
  - [x] float_cmpl
  - [x] float_dcm

Edit: Moved to  #1145
-  ~~implement the same for INS~~
- ~~deal with the remaining "global" ins_reset_x functions (`ins_reset_local_origin`, `ins_reset_altitude_ref`, ...)~~
- ~~get rid of `ins_periodic` for those impls which need it (probably when we start converting them to modules)~~


flixr:
@gautierhattenberger @dewagter feedback on the general design (e.g. the ahrs/ins_register_impl, etc) wanted...

gautierhattenberger:
Concerning the filter selection, maybe this can be a separated issue, since it seems we need to think/discuss a bit more on this point.
I would suggest to test/merge this branch in its current state without more features (we can even keep the current basic selection) so it doesn't stay behind master for too long.

flixr:
Ok, so we also leave the INS parts as they are for now?
Can you test this on e.g. apogee?

PR #1131, opened 2015-03-09
[NPS] fix some JSBSim FDM issues for fixedwing sim

flixr:
- always run fdm (even when not launched yet) to get proper readings
- sensor_params: imu sensors: take sign into account (so that it works if there are negative signs defined in airframe file, like for jp.xml)
- if NPS_ACTUATOR_NAMES is not specified, use the standard throttle,roll,pitch (and if used in airframe yaw) commands for throttle, aileron, elevator and rudder in JSBSim model.


martinmm:
it works in principle, but az is strange. it decreases acceleration (to like 0.5g) instead of increasing if I pull the aircraft up/climb

PR #1134, opened 2015-03-12
Possibilty to specify conf.xml file for building aircrafts

flixr:
Mostly interesting for running the tests (e.g. compiling all aircrafts in a given conf.xml).

If the CONF_XML environment variable is set, use that conf.xml file instead of the standard $PAPARAZZI_HOME/conf/conf.xml
Meaning you can run tests for a different conf without changing the conf.xml symlink, e.g.
`CONF_XML=$PAPARAZZI_HOME/conf/conf_tests.xml prove tests/examples`


gautierhattenberger:
It looks okay to me

PR #1139, opened 2015-03-19
remove myGtkInit.ml

flixr:
and rather call Main.init directly in the code.
See also http://lablgtk.forge.ocamlcore.org/README.html#_how_to_compile_and_link

This is part of the lib/ocaml cleanup, in the end only the pprz package should remain in there...


gautierhattenberger:
working like a charm here

PR #1142, opened 2015-03-22
[ocaml] Fixed compilation of the sim target on OS X

esden:
Sim target could not find the lib-pprz.cma.

### DIFF
--- a/conf/Makefile.sim
+++ b/conf/Makefile.sim
@@ -31,7 +31,7 @@ include $(PAPARAZZI_SRC)/sw/Makefile.ocaml
 
 CC = gcc
 SIMDIR = $(PAPARAZZI_SRC)/sw/simulator
-CAMLINCLUDES = -I $(LIBPPRZDIR) -I $(LIBPPRZDIR)/pprz -I $(SIMDIR) -I $(OBJDIR)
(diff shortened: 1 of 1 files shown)


flixr:
If we really want to link it statically for OSX, it should still use `-package` in the build rule instead of directly specifying the cmas...

esden:
Yeah, I wish I could switch it over to dynamic linking. I tried to just use the same process as on linux but use dylib instead of so for the autopilot library. Sadly when this command is being run:

```

flixr:
But even when using static linking, we should use the `-package pprz` etc arguments (instead of explicitly providing the include paths and cma files)...

esden:
I found a solution to do dynamic linking. :)

esden:
I have now tested the solution on linux as well as on OS X. It works for me. :)

flixr:
I really don't like all these special cases... it makes maintenance painful.
How about rather simply always statically link? Then we also don't need to add the dllpaths (since they are not in any standard path).
See my ocamlbuild_static branch: https://github.com/flixr/paparazzi/commit/24ef86f799a6e1b55af729a2ea3aa2182a1f8916

flixr:
Merged for now with 6a950da6f423be13969da108a013afd24527b609...

PR #1144, opened 2015-03-23
Flight recorder

gautierhattenberger:
Allow to define the messages to be logged from the telemetry file.
Working with SD log on Apogee, but should also work with data logger over uart.

The periodic telemetry structures had to be merged in a single one, so messages are registering for all telemetries.


flixr:
you forgot to add the flight_recorder.xml module file...

flixr:
Mostly looks good to me...
By default now everything uses the `DefaultPeriodic` which is `pprz_telemetry`, so you have to register messages only once.
But you can still create another pprz_telemetry struct and add different callbacks there if desired.

Naming wise, can we change some of them to be more clear?
E.g. rename pprz_telemetry to periodic_telemetry or so?

PR #1146, opened 2015-03-24
GPS NMEA parser improvements

flixr:
- parse GSV message to get satellite information
  - only populating svinfos with 12 channels of GPS sat info (not saving GLONASS sat info)
- get geoid separation and calculate gps altitude (above ellipsoid) accordingly


flixr:
Seems to work as expected on the Bebop...

PR #1150, opened 2015-03-26
[NPS] add possibility to simulate loss of datalink/downlink

### LINKED ISSUE #934: NPS should be able to simulate loss of datalink
While the OCaml sim can do that since #631, this is still missing for the NPS sim.

You could disable the downlink telemetry by adding
``` xml
<dl_setting var="ivy_dl_enabled" min="0" step="1" max="1" module="ivy_transport" shortname="datalink" values="OFF|ON"/>
```
to conf/settings/nps.xml

But that is probably not really helpful, as the uplink/datalink still exists (and you can't disable that via settings as that uses the datalink).
So we would need to unbind/register the ivy callbacks in NPS.

Using settings to change stuff in NPS is a bit of a hack anyway and we should really have something external (like the aircraft window for the OCaml sim) that communicates with NPS to change simulation parameters.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5838244-nps-should-be-able-to-simulate-loss-of-datalink?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flixr:
It is a bit of a hack, especially we can't disable changing settings if datalink is disabled,
since we need to parse that to enable the datalink via settings again.
Also previously datalink_time was never reset to 0...

Nevertheless it should suffice for most cases and close #934

### DIFF
--- a/conf/settings/nps.xml
+++ b/conf/settings/nps.xml
@@ -7,6 +7,8 @@
        <dl_setting var="nps_bypass_ahrs" min="0" step="1" max="1" module="nps/nps_autopilot" shortname="bypass_ahrs" values="No|Yes"/>
        <dl_setting var="nps_bypass_ins" min="0" step="1" max="1" module="nps/nps_autopilot" shortname="bypass_ins" values="No|Yes"/>
        <dl_setting var="gps_has_fix" min="0" step="1" max="1" module="subsystems/gps/gps_sim_nps" shortname="gps_fix" values="No|Yes"/>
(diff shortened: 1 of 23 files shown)


gautierhattenberger:
It is already a good start, I think we can merge.

PR #1151, opened 2015-03-27
[telemetry] store periodic telemetry msg names as const to save ram

flixr:
Make sure that the msg names are stored in rom (.text) and not ram (.data) by making them const.
Since the callbacks are not const, replace the telemetry_msg struct by typedef for only the name and separate callback array.

This saves ~ 2.5k of RAM with a standard telemetry file, which is really nice if you try to run Paparazzi on a low end board like the CC3D with only 20k RAM.


gautierhattenberger:
I'm fine with this

PR #1152, opened 2015-03-30
Replace NO_RC_THRUST_LIMIT with RC_THRUST_LIMIT

EwoudSmeur:
Replace NO_RC_THRUST_LIMIT with RC_THRUST_LIMIT

The fact that default the thrust is limited by rc in autonomous flight is something that has caused a lot of problems for us, because it is very counter intuitive. In my opinion, when I flip the switch to autonomous mode, the drone should be in control, not me.

Typical issues were: 
- Why is the drone not climbing?? It is almost crashing!!! oh wait didn't define NO_RC_THRUST_LIMIT
- *Put rc_thrust to full in autonomous mode* Oh no something is going wrong! switch to rc mode, with full thrust -> stabilization fails -> crash

I understand that in some cases it can be useful, so I changed it to a define RC_THRUST_LIMIT, so it can still be defined and used for anyone who explicitly wants to use it.

### DIFF
--- a/conf/airframes/BR/DelFlyDualPWMservo.xml
+++ b/conf/airframes/BR/DelFlyDualPWMservo.xml
@@ -104,7 +104,6 @@
    <define name="MODE_AUTO2" value="AP_MODE_NAV"/>
 
     <define name="USE_THROTTLE_FOR_MOTOR_ARMING" value="TRUE"/>
-    <define name="NO_RC_THRUST_LIMIT" value="TRUE"/>
  </section>
 
  <section name="BAT">
--- a/conf/airframes/BR/DreamCacher_bart.xml
+++ b/conf/airframes/BR/DreamCacher_bart.xml
@@ -98,7 +98,6 @@
    <define name="MODE_AUTO2" value="AP_MODE_NAV"/>
 
     <define name="USE_THROTTLE_FOR_MOTOR_ARMING" value="TRUE"/>
-    <define name="NO_RC_THRUST_LIMIT" value="TRUE"/>
  </section>
 
  <section name="BAT">
--- a/conf/airframes/BR/ladybird_kit_bart.xml
+++ b/conf/airframes/BR/ladybird_kit_bart.xml
@@ -107,7 +107,6 @@
    <define name="MODE_AUTO2" value="AP_MODE_NAV"/>
 
     <define name="USE_THROTTLE_FOR_MOTOR_ARMING" value="TRUE"/>
-    <define name="NO_RC_THRUST_LIMIT" value="TRUE"/>
  </section>
 
  <section name="BAT">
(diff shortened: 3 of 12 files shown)


flixr:
While personally and from the technical side I agree here, I'm really not sure if it is a good idea to change this default.
It is documented like this and we had this feature forever...
Changing it will mean that users who are relying on this safety feature will suddenly (and silently) not have this any more after an update.

This is a general problem with otherwise minor changes like this... and it will make the documentation even more annoying: Before version xy this, after yz that...

Maybe some others can chip in here as well, if most people disable this anyway, then I'm fine with changing it... and if someone properly updates the docs everywhere then...

side note: if stabilization fails at full thrust, that is an issue that needs to be addressed separately!

podhrmic:
I dont have a strong opinion about this, but we were using the RC_THRUST_LIMIT for out autonomous flights, and it was actually very useful for tuning the airframes. The safey pilot just has to remember that he has to lower the thrust while going out of auto mode.

dewagter:
It's quite often been "a bit ennoying".
But a few times it would have been terrible not to have it.

EwoudSmeur:
@flixr If I search on the wiki I find no documentation about it. Would a warning if you don't define RC_TRHUST_LIMIT (true or false) be enough information?

@podhrmic I agree that it can be useful indeed for tuning, but then it shouldn't be enabled by default. When you are tuning an airframe you can still enable it

@dewagter Why is that? If you just switch to manual you have all the thrust control you want on the transmitter.

flixr:
@EwoudSmeur https://wiki.paparazziuav.org/wiki/Rotorcraft_Configuration#Autopilot_modes

EwoudSmeur:
I am perfectly willing to edit that piece of text

IHaveADrone:
I my experience, I think it would be better as is.  I've had a hexcopter try to fly up +300m higher than is was supposed to and the safety pilot saved it by backing off the throttle and letting it float downward.

gautierhattenberger:
@EwoudSmeur switch to manual when you are really high means you also "loose" the position holding provided by the navigation. I already had the same kind of issue that @IHaveADrone described (not that high), and it would have been much more difficult to fly safely in attitude mode.
I know this default behavior is maybe not the best, and that some "safety" pilots don't like it (my colleague is usually complaining about this). But speaking for myself, it helped me much more in the past than the opposite.

flixr:
As there are good arguments for and against using this feature and as we already had it for quite a long time like this, I vote to keep it as is.
IMHO there is not enough benefit in changing the default...

EwoudSmeur:
It seems indeed that there are a lot of people against this change, so it might be better to let it be. However, I do want to make the side note that this is one of the obstacles new users of paparazzi will encounter when starting out with paparazzi, simply because it is not intuitive.

flixr:
Yeah, mentioning this in some sort of multicopter "getting started/first flight" tutorial would be nice...

flixr:
Closing this as WONTFIX then...

bmw330i:
Just as perspective only but I heard that same exact argument used by the windows users who said paparazzi running on Linux would be an issue for new users and even Paparazzi being widely used. 
Those same people now are moving their platforms to Linux today.
Resistance is sometimes just a good way to remind us to question why we feel strongly and others feel differently. 
Paparazzi is a very mature system and a community with unparalleled openness, acceptance and flexibility. If this truly is a big issue for new people we will see new people bring it up again and I'm sure the community will openly consider it.
David

PR #1153, opened 2015-03-30
[gps] directly send ABI messages from implementation

flixr:
and get rid of the on_gps_solution callback in main


gautierhattenberger:
It looks good, I'll test this as soon as possible (same for #1154)

PR #1154, opened 2015-03-31
[IMU] directly send ABI messages from implementation

flixr:
and remove on_gyro|accel|mag_event from main.

Based on #1153 which already does this for GPS.

AHRS aligner subscribes to gyro message but still accesses  the global imu struct for now...

### DIFF
--- a/sw/airborne/arch/sim/sim_gps.c
+++ b/sw/airborne/arch/sim/sim_gps.c
@@ -74,7 +74,7 @@ value sim_use_gps_pos(value x, value y, value z, value c, value a, value s, valu
   //gps_verbose_downlink = !launch;
   //gps_downlink();
 
-  gps_available = TRUE;
+  gps_sim_publish();
 
   return Val_unit;
 }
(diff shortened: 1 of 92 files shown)


flixr:
Should we keep separate ABI sender ids for each IMU implementation that is tied to a specific board?
Or use an IMU_BOARD_ID for all of them?

gautierhattenberger:
The number of IMU with have is really crazy...
I think using a generic ID for board integrated IMUs make sense, since there is no point in requesting for instance APOGEE_IMU data on an other board. But knowing immediately that you are using the internal IMU (or that you're not using it) is probably more helpful.

flixr:
consolidated some IDs in a common IMU_BOARD_ID...

flixr:
@gautierhattenberger is this good to merge from your side now?

gautierhattenberger:
It looks fine for me. I couldn't test it before leaving for vacations thought, I'll be back in 10 days.

PR #1155, opened 2015-04-01
[air_data] use CAS/EAS for airspeed and send TAS in message

gautierhattenberger:
On piloted aircraft, the CAS (actually IAS) is used as a flight indication and thus is the controlled airspeed.
I suggest to use EAS (very close to CAS for us and easier to compute) for airspeed (the value updating the state interface) and compute TAS (the "real" airspeed) as an extra indication, especially in message instead of tas_factor.
One of the reason is that to get a correct TAS you need P, T and P_diff, while EAS/CAS only requires P_diff. Today I flew with an uncalibrated temperature sensor (it was their but not needed) and the result is that my airspeed was completely wrong.

http://en.wikipedia.org/wiki/Calibrated_airspeed
http://en.wikipedia.org/wiki/Indicated_airspeed
http://en.wikipedia.org/wiki/Equivalent_airspeed

### DIFF
--- a/conf/messages.xml
+++ b/conf/messages.xml
@@ -1942,13 +1942,13 @@
   </message>
 
   <message name="AIR_DATA" id="222">
-    <field name="pressure" type="float" unit="Pa"/>
-    <field name="diff_p" type="float" unit="Pa"/>
-    <field name="temp" type="float" unit="deg celcius"/>
-    <field name="qnh" type="float" unit="hPa"/>
-    <field name="amsl_baro" type="float" unit="m"/>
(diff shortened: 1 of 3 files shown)


flixr:
Looks good and sensible to me. 
Should then also disable tas factor calculation by default?

gautierhattenberger:
I don't think so as long as TAS is not feeding the state interface
If you want TAS, you need to look for it explicitly, so we can assume that you ever calibrate P and T sensors or you don't have them at all (so TAS = EAS)

flixr:
Will add a check to compute tas only if airspeed > 0

PR #1157, opened 2015-04-02
[settings] add optional target attribute to settings

flixr:
If not target attribute is given, the settings are added for all targets like before.

This makes it easier to add settings files to the conf that are only valid for some targets like nps or some that you don't want when using the test progs.

Beware of: if you compile a target that unloads some settings or modules (e.g. the sim) after you uploaded to the ap, the `var/aircrafts/<ac>/settings.xml` file that has all usable settings combined will change ~~~and the hence the wrong ones used in GCS and settings prog~~~.

Now the calculation of the MD5sum is taking unloaded settings into account, meaning it will change and warn the user if the last compiled target has different settings.

### DIFF
--- a/conf/settings/estimation/body_to_imu.xml
+++ b/conf/settings/estimation/body_to_imu.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE settings SYSTEM "../settings.dtd">
 
-<settings target="ap|nps">
+<settings target="!sim">
   <dl_settings>
     <dl_settings NAME="body2imu">
       <dl_setting MAX="90" MIN="-90" STEP="0.5" VAR="imu.body_to_imu.eulers_f.phi" shortname="b2i phi" module="subsystems/imu" param="IMU_BODY_TO_IMU_PHI" unit="rad" alt_unit="deg" handler="SetBodyToImuPhi"/>
--- a/conf/settings/estimation/body_to_imu.xml
+++ b/conf/settings/estimation/body_to_imu.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE settings SYSTEM "../settings.dtd">
 
-<settings>
+<settings target="ap|nps">
   <dl_settings>
     <dl_settings NAME="body2imu">
       <dl_setting MAX="90" MIN="-90" STEP="0.5" VAR="imu.body_to_imu.eulers_f.phi" shortname="b2i phi" module="subsystems/imu" param="IMU_BODY_TO_IMU_PHI" unit="rad" alt_unit="deg" handler="SetBodyToImuPhi"/>
--- a/conf/settings/control/rotorcraft_guidance.xml
+++ b/conf/settings/control/rotorcraft_guidance.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE settings SYSTEM "../settings.dtd">
 
-<settings>
+<settings target="ap|nps">
   <dl_settings>
 
     <dl_settings NAME="Vert Loop">
--- a/conf/settings/control/stabilization_att_float_euler.xml
+++ b/conf/settings/control/stabilization_att_float_euler.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE settings SYSTEM "../settings.dtd">
 
-<settings>
+<settings target="ap|nps">
   <dl_settings>
 
     <dl_settings NAME="Att Loop">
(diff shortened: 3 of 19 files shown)


conf/settings/estimation/body_to_imu.xml
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE settings SYSTEM "../settings.dtd">
 
-<settings>
+<settings target="ap|nps">
```
flixr: why restrict this to ap and nps?
That means you can't use it with demo and test targets...

gautierhattenberger: Because it fails with sim target where imu.c is not compiled. If you want to keep it without a specific target, I suggest to fix sim then.

flixr: Right, how about something like https://github.com/flixr/paparazzi/commit/493fde03de35bf68c38302603b7a4809af962d61 to add the possibility to exclude a list of targets with ```target="!foo|bar"```

gautierhattenberger: The `try` is incomplete, String.get may return `Invalid_argument`

gautierhattenberger:
do we merge this one ?

flixr:
It is ok to merge for me.
But we definitely need to better handle having a different settings.xml file after compiling another target.
At the very least something like a check for the number of settings?

flixr:
Should we merge this befoer v5.6 regardless of the missing checks if you had unloaded settings/modules?

gautierhattenberger:
It is very useful but sadly it is not possible to make just one line insensitive in the settings selection box. At least we could add a warning when unloading settings ?

flixr:
I'm not so much concerned about the selection box, that seems to "only" be an inconvenience to me.
As said before I don't like the increased probability of mismatched settings with potentially even catastrophic outcome if you unknowingly change the wrong one.

What about adding another MD5sum for only the settings and sending that with ALIVE?
The we could check this when loading settings or the GCS to see if they match...

flixr:
Or better, also do the filtering for used settings (unloaded modules with settings, targets in settings) when creating `var/aircrafts/foo/conf/conf_aircraft.xml` of which the md5sum is computed.

I think that would be the best solution, or am I missing something obvious here?

gautierhattenberger:
It seems that we are currently only removing the settings that are not selected in the pprz center. I guess it should also work if we correctly unload them when the module is not valid for the target.

flixr:
Yes and the settings that are directly added but don't match the target, could you do that in this branch?

flixr:
The filtering now only works correctly from within paparazi center if you don't enable unloaded setting again.
As you said this is is partially a problem with the the list element.
IMHO the filtering should be done properly in gen_aircraft.ml so that the generated conf_aircraft.xml will be always correct and it also works if called directly from the command line....

gautierhattenberger:
I don't get it. If I force the selection of an unwanted settings file (like nps.xml for ap target), it will be removed from conf_aircraft.xml and will not appear in the generated settings as well.
So conf_aircraft.xml really hold the conf used for the last build.

flixr:
Ok, nevermind my previous comment.... that indeed works as expected now.

However I think it would be better to not "uncheck" the setttings file in the paparazzi center if it is unloaded.
Otherwise you suddenly have the settings file disabled if you change to a target that supports it again...

flixr:
Actually I think the unloaded should not change the conf_aircraft.xml (and hence the md5sum) unless they have settings.
If they don't have settings, there is not reason to have a different md5sum as the settings.xml will be the same.

flixr:
Nevermind, seems to work as expected.

flightenabledOrg:
Hi, I'm just getting back into paparazzi. I tried to follow the mac
paparazzi install instructions but they appear to not work, is there an
up-to-date mac install setup I can follow?

flixr:
@flightenabledOrg please use the mailing list for such questions and not some totally unrelated issue...
Also please try to include as much details as possible (OS version, paparazzi version, errors you got, etc). "does not appear to work" is to unspecific to give any real help

flightenabledOrg:
Whoops, I apologize.

@flightenabledOrg <https://github.com/flightenabledOrg> please use the
mailing list for such questions and not some totally unrelated issue...
Also please try to include as much details as possible (OS version,
paparazzi version, errors you got, etc). "does not appear to work" is to
unspecific to give any real help

—
Reply to this email directly or view it on GitHub
<https://github.com/paparazzi/paparazzi/pull/1157#issuecomment-127056345>.

PR #1158, opened 2015-04-14
[modules] viewvideo: use UdpSocket instead of udp_periph

flixr:
I don't see any advantage in using udp_periph for this as it doesn't use any of the periph functions nor the buffer and we never need the input port...


flixr:
@fvantienen would be nice if you could test this...

fvantienen:
Tested on the Bebop and ARDrone and it works.

PR #1161, opened 2015-04-18
Mpu9250 in SPI mode. Test with GCS, not test  in real flight.

MosesWangHao:
1. Mag data read from AK8963 use master I2C mode. 400khz
2. I also add a baro_board.c for lisamx board with BMP085.


sw/airborne/boards/lisa_mx/baro_board.c
```diff
@@ -0,0 +1,86 @@
...
+
+void baro_init(void)
+{
+  bmp085_init(&baro_bmp085, &i2c2, BMP085_SLAVE_ADDR);
+
+  /* setup eoc check function */
+  baro_bmp085.eoc = &baro_eoc;
+
+  gpio_clear(GPIOB, GPIO0);
+  //gpio_set_mode(GPIOB, GPIO_MODE_INPUT,
```
flixr: Should probably be `gpio_setup_input_pulldown(GPIOB, GPIO0);`

sw/airborne/peripherals/mpu9250.c
```diff
@@ -110,6 +114,27 @@ void mpu9250_send_config(Mpu9250ConfigSet mpu_set, void *mpu, struct Mpu9250Conf
       mpu_set(mpu, MPU9250_REG_INT_ENABLE, (config->drdy_int_enable << 0));
       config->init_status++;
       break;
+
```
flixr: All this should not be needed here at all: the setup for i2c slaves is done in `mpu9250_configure_i2c_slaves` and (should) properly use the mst_clock and mst_delay set in the config

sw/airborne/subsystems/imu/imu_mpu9250_spi.c
```diff
@@ -225,19 +236,20 @@ bool_t imu_mpu9250_configure_mag_slave(Mpu9250ConfigSet mpu_set, void *mpu)
...
-  mpu_set_and_wait(mpu_set, mpu, MPU9250_REG_I2C_SLV0_ADDR, (AK8963_ADDR >> 1) | MPU9250_SPI_READ);
+    
+  //Config SLV0 for continus Read 
+  mpu_set_and_wait(mpu_set, mpu, MPU9250_REG_I2C_SLV0_ADDR, (MPU9250_MAG_ADDR >> 1) | MPU9250_SPI_READ);
   mpu_set_and_wait(mpu_set, mpu, MPU9250_REG_I2C_SLV0_REG, AK8963_REG_HXL);
   // Put the enable command as last.
   mpu_set_and_wait(mpu_set, mpu, MPU9250_REG_I2C_SLV0_CTRL,
                    (1 << 7) |    // Slave 0 enable
-                   (7 << 0));    // Read 7 bytes (mag x,y,z + status)
+                   (8 << 0));    // Read 8 bytes (mag x,y,z + status, cntl1)
```
flixr: any reason to read the cntl1 reg as well? no values from it are used anywhere...

sw/airborne/subsystems/imu/imu_mpu9250_spi.c
```diff
@@ -194,6 +204,7 @@ void imu_mpu9250_event(void)
     imu_scale_mag(&imu);
     AbiSendMsgIMU_MAG_INT32(IMU_MPU9250_ID, now_ts, &imu.mag);
```
flixr: Since you check for the overflow status bit above, the measurement should also not be sent via ABI if not valid...

MosesWangHao: Yes, You are right, I test all in V5.4.  but when i commit, I find the Master use AbiSendMsg instead of some Valid flag. 
BTW: should i pull Request again?

sw/airborne/subsystems/imu/imu_mpu9250_spi.c
MosesWangHao: 1. I want use it to check the valid in future. MPU9250 can bind 4 slave IIC devices. As I get from the data sheet,  I afraid it maybe confuse when i add a baro senser.
2. Int16FromBuf is easy to understand in imu_mpu9250_spi.c.

sw/airborne/peripherals/mpu9250.c
MosesWangHao: Sorry,  Is your means move all follow to mpu9250_configure_i2c_slaves?  The follow code is executed after mpu9250_configure_i2c_slaves?

sw/airborne/subsystems/imu/imu_mpu9250_spi.c
flixr: CNTL1 doesn't contain any status/valid flags, only mode settings...
If you just want to check the HOFL bit:
```
if (!bit_is_set(imu_mpu9250.mpu.data_ext[6], 3)) {
  // HOFL overflow bit not set
}
```

flixr:
You don't need to make a new pull request, just push to your mpu9250_spi branch and the changes/added commits will show up here...

sw/airborne/subsystems/imu/imu_mpu9250_spi.c
MosesWangHao: ok, use your code,  let future to future.

sw/airborne/peripherals/mpu9250.c
flixr: The i2c master clock, delay and enable is already configured in mpu9250_configure_i2c_slaves of mpu9250_spi.c, so I don't see why you would do it again afterwards...
Also you are trying to set clock and p_nsr in user_ctrl reg instead of I2C_MST_CTRL

If the P_NSR is needed (stop instead of restart between reads) it should be added as an option to the config struct and set in mpu9250_configure_i2c_slaves together with the clock.

sw/airborne/boards/lisa_mx/baro_board.c
MosesWangHao: ok.

sw/airborne/peripherals/mpu9250.c
MosesWangHao: will test it first, if it go smoothly as your said, delete all

flixr:
Merged the baro_board.c for Lisa/MX with commit f3c98accaa0d7227a0d916e17b7b6e38796d9d58
and merged the relevant and cleaned up changes for mpu9250_spi with 395b77e93543254c5dffb25d1c7c25b025c6ea0d

PR #1162, opened 2015-04-23
remove unused/unmaintained beth and fms code

flixr:
and changed natnet2ivy to use udp_socket instead of fms.


flixr:
@fvantienen could you guys please test if the rewritten natnet2ivy works?

gautierhattenberger:
@fvantienen @OpenUAS can you confirm that natnet is not broken with this ?

flixr:
Merging due to feedback timeout and it should (tm) work...

PR #1164, opened 2015-04-30
[build] use -ggdb3 DEBUG_FLAGS by default

flixr:
-ggdb3 make slightly bigger .elf files, but allows gdb to understand
macros, which paparazzi uses somewhat extensively.
Make this the default, since it only impacts the size of the debug sections and
not the size of the flashed binary (in the lpc21/stm32 case).
On the linux system we really don't care about these few more bytes...

### DIFF
--- a/conf/Makefile.linux
+++ b/conf/Makefile.linux
@@ -34,7 +34,8 @@ LD  ?= $(CC)
 Q=@
 
 OPT ?= 3
-DEBUG_INFO ?= n
+# Slightly bigger .elf files but gains the ability to decode macros
+DEBUG_FLAGS ?= -ggdb3
 CSTANDARD ?= -std=gnu99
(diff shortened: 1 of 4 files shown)


podhrmic:
Can it be used for ChiboOS arch too?

flixr:
Sure, I don't see why not... but rt_chibios branch is not merged yet...

podhrmic:
I though about the Chibios-libopencm3 version, but I guess that is covered in stm32 arch dir.

flixr:
Actually seems to be already set for chibios-libopencm3: https://github.com/paparazzi/paparazzi/blob/master/conf/Makefile.chibios-libopencm3#L73

gautierhattenberger:
yes, I have set gdb3 in rt_chibios branch and it works like a charm.

PR #1165, opened 2015-04-30
[conf] update udev rules

flixr:
Add some udev rules that create convenience symlinks:
- floss jtag -> /dev/jtag and /dev/jtag-serial
- Black Magic Probe -> /dev/bmp-gdb and /dev/bmp-serial
- FTDI serial cables -> /dev/ftdi-serial
- Paparazzi STM32 autopilots usb serial interface -> /dev/paparazzi/stm32-usb-serial


flixr:
These are the rules I have been using for a while already...
Naming of the symlinks could be discussed...

gautierhattenberger:
closed with cd548f17d3797eaaf519b4dbcfa2a78351ecaa67

Issue #385, opened 2013-03-03
improve motor mixing saturation/priority handling


flixr:
This is the current state for the motor mixing of the rotorcraft firmware:
The desired attitude has higher priority than the (individual) thrust/rpm of each motor.
If the desired commands cannot be achieved due to saturation of min/max motor commands (motor can't spin slower than idle or faster than full speed), the other motors are "offset" to achieve the same relative commands (basically offset total thrust to get higher or lower than commanded thrust).

Take the example of rather low commanded thrust, but a significant (positive) pitch command:
- If the back motor can't run any slower, the thrust of all motors is increased by the difference so that they are correct again in relation to each other an the desired attitude achieved. As a result the rotorcraft will have a higher total thrust than commanded.

So:
- For roll/pitch this is desired, as you want that to have a higher priority on these commands so your rotorcraft doesn't fly away in some direction.
- For yaw this is normally not desired, as it is not critical. If a desired yaw command can't be achieved without increasing/decreasing overall throttle it should not be done. Right now your quad might e.g. shoot into the sky...

Several things that should be improved:
- Different priorities for roll/pitch and yaw:
  desired priority order (increasing): yaw, thrust, roll/pitch
- limit the max offset, so that in case of saturation thrust is not increased/decreased by the maximum possible, but only by some amount. Meaning that up to some point the attitude will be maintained by changing overall thrust, but at some point the desired attitude won't be reached anymore but your thrust won't be increased to max to avoid your quad shooting up into the sky.

flixr:
Closing this as it should (tm) be solved with the merge of #1170

PR #1170, opened 2015-05-03
[rotorcraft] try to improve motor mixing priorities

flixr:
Calculate the amount of yaw authority left after applying roll, pitch and throttle commands.
Then limit the yaw command to that.

Simple first shot at improving the priorities in the saturation case, see #385.

Calculating the mean_cmd should not not actually be necessary for symmetric airframes without any trims set.
But it shouldn't hurt....

Should (tm) result in zero saturation failures like this, but might severly limit yaw authority.
Maybe find some sort of compromise, e.g. use MOTOR_MIXING_MAX_SATURATION_OFFSET when bounding yaw or allow to give less throttle to get more yaw authority....

Not tested in real flight, meant as a basis for discussion/tests...

### DIFF
--- a/sw/airborne/subsystems/actuators/motor_mixing.c
+++ b/sw/airborne/subsystems/actuators/motor_mixing.c
@@ -109,7 +109,7 @@ void motor_mixing_init(void)
   motor_mixing.nb_saturation = 0;
 }
 
-__attribute__((always_inline)) static inline void offset_commands(int32_t offset)
+static void offset_commands(int32_t offset)
 {
   uint8_t j;
   for (j = 0; j < MOTOR_MIXING_NB_MOTOR; j++) {
@@ -117,17 +117,17 @@ __attribute__((always_inline)) static inline void offset_commands(int32_t offset
   }
 }
 
-__attribute__((always_inline)) static inline void bound_commands(void)
+static void bound_commands(void)
 {
   uint8_t j;
-  for (j = 0; j < MOTOR_MIXING_NB_MOTOR; j++)
-    Bound(motor_mixing.commands[j],
-          MOTOR_MIXING_MIN_MOTOR, MOTOR_MIXING_MAX_MOTOR);
+  for (j = 0; j < MOTOR_MIXING_NB_MOTOR; j++) {
+    Bound(motor_mixing.commands[j], MOTOR_MIXING_MIN_MOTOR, MOTOR_MIXING_MAX_MOTOR);
+  }
 }
 
-#ifdef MOTOR_MIXING_USE_MAX_MOTOR_STEP_BINDING
-__attribute__((always_inline)) static inline void bound_commands_step(void)
+static void bound_commands_step(void)
 {
+#ifdef MOTOR_MIXING_USE_MAX_MOTOR_STEP_BINDING
   uint8_t j;
   static int32_t prev_commands[MOTOR_MIXING_NB_MOTOR];
   static uint8_t initialized = 0;
@@ -146,12 +146,8 @@ __attribute__((always_inline)) static inline void bound_commands_step(void)
   for (j = 0; j < MOTOR_MIXING_NB_MOTOR; j++) {
     prev_commands[j] = motor_mixing.commands[j];
   }
-}
-#else
-__attribute__((always_inline)) static inline void bound_commands_step(void)
-{
-}
 #endif
+}
 
 void motor_mixing_run_spinup(uint32_t counter, uint32_t max_counter)
 {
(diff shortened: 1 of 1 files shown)


podhrmic:
Since it is related: we noticed a while ago that flying quad and hexarotors with same mixing gains for yaw and roll/pitch makes the flight very non-smooth (as described in #385 ). Having yaw mixing gains at 128 while the rest was at max value (256) resulted in much smoother flight in both auto and manual.
In other words - the yaw mixing gains (for example https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml#L68) should be only around a half of roll/pitch/thrust mixing gains (at least for the airframes we tested).
Although it is not really a fix, maybe a fair warning in wiki could "solve" #385?

flixr:
Yes, I used a lower yaw scale on my quad too, and it probably would have made sense to document it like that... please feel free to add this to the wiki!
The drawback is that if you start changing the scale there, you will have to adjust your gains accordingly...

flixr:
Do we need to make this even more configurable (like disabling the yaw bound) for cases like the quadshot?
@ewoudsmeur any input on this?
I guess it would be even better to write a custom mixer for quadshot?

See also #1175 for (probably sensible) common multicopter motor mixing configurations...

hooperfly:
Flew multiple flights with my hexacopter with the patch. Yaw is much improved and the aircraft is more stable.

OpenUAS:
It is that our multirotors are in "create anew or in repair state", sorry would love to test since this is one very welcomed improvement, but first need to repair something

EwoudSmeur:
@flixr somehow I missed that I was tagged in this... Anyway, I think this is a good idea, also for the quadshot. 

However, I think the direction of the yaw command also matters, as I think the yaw command could 'relief' the motor that is closest to one of the bounds. I thought of an example:

high thrust and a pitch command (for a quad in a plus config) -> one motor will be close to max while the opposite one is in the medium range. Now there can be a yaw command that spins both these rotors down and the other two up that is more than the difference between the first rotor and max.... hope I am still making sense :)

A way to incorporate this could be to, after applying the pitch, roll and thrust commands, check what part of the yaw command can be applied for each motor before it saturates. The minimum of this can then be applied to every motor

flixr:
@EwoudSmeur: line 229 should do exactly that
```
int32_t yaw_authority = Min(ABS(min_cmd), (MAX_PPRZ - max_cmd));
```

EwoudSmeur:
@flixr I disagree. What is calculated in 229 is the minimum leeway there is for each motor to the closest bound, whether it be up or down. My point is that the yaw commando might relieve the motor closest to one of its bounds, resulting in a larger yaw authority. But this depends on the direction the yaw commando is in.

Now that I am thinking about it, it also depends on the yaw_coef. This is not yet taken into account. Take for example the quadshot, that has differently scaled yaw coefficients: <define name="YAW_COEF"    value="{ -256,  256,  128, -128 }"/>
The quadshot would suffer from too strict yaw_authority bounds even more, as the yaw_authority is calculated considering only a 256 scale

flixr:
@EwoudSmeur, right, now I understand what you meant...

As I wrote above this is a simple first shot that can certainly be improved to get the maximum yaw authority possible...
Right now it is basically a simple "conservative" authority limit that should prevent saturations.
You are absolutely right, calculating it per motor taking the coef and yaw command direction into account would enable you to really exhaust the full available motor command range.

Nevertheless, as this currently already works better than before and I probably won't have time to work on it in the coming two weeks, I propose to merge it as is and we can go on to improve it from there...
That is unless you have time to do that now.

flixr:
Ok, I actually did find a bit of time... updated it with a smarter yaw_authority calculation.
Should work better, but is untested...

flixr:
Unless someone can do more tests in the next couple of days and report that, I would propose to merge this as in it's current state as it already improves the situation.

hooperfly:
@flixr I do see some strange behavior in `nps` related to a diverging aircraft upon simulation startup. It doesn't impact the real-world flight. It seems to be related to the simulation. Not sure if it's related to this change or not. The aircraft icon starts to wonder off the screen, whereas before the simulation was "stable".

flixr:
@hooperfly could you please describe this more precisely? With which aircraft, what do you mean by "diverging upon startup"?

hooperfly:
screen shot 2015-06-19 at 12 42 33 pm

<define name="YAW_COEF"    value="{ -128,  128, -128,  128 }"/>

hooperfly:
screen shot 2015-06-19 at 12 43 47 pm

hooperfly:
The TeensyFly Quad has the following yaw values:

```
<define name="YAW_COEF"    value="{ -256,  256, -256,  256 }"/>
```

The RacerPEX Quad's yaw values:

```
<define name="YAW_COEF"    value="{ -128,  128, -128,  128 }"/>
```

hooperfly:
Look at the AGL info in both images. They are way off.

flixr:
Sorry, I can not reproduce that...
what did you do prior to when it went into home mode...

hooperfly:
@flixr Hmmm, I just startup the simulation and I get a wandering multirotor. No input on my part. Did you test with my branch?

flixr:
I don't see how it can have anything to do with the motor mixing if the motors are not even armed and running... and yes, I tested with your branch...

hooperfly:
I thought maybe it was related to either changes in `NPS` or possibly something related to overflow/underflow of a value in the simulation?

hooperfly:
I think you can go ahead and merge the changes. As you suggested, the motors should not be impacting the simulation as they are not started. Something related to geo_init in the simulation is more likely to be the culprit.

flixr:
I'll have another closer look tomorrow...

Issue #641, opened 2014-02-16
GPS cleanup and better LTP origin handling (UTM vs. NED) 


flixr:
- add flags/bitfield to the global GpsState struct to indicate which representations are actually provided by the gps driver
- no need to always calculate UTM pos in GPS drives with GPS_USE_LATLONG, should be handled by the state interface (especially considering that only the old ublox provide position in UTM directly)
- less direct dependencies on GPS (e.g. in common_nav)

  e.g. nav_reset_reference should probably just read the current position from the state interface and reset the reference according to that, instead of explicitly using lla or utm from gps to call stateSetLocalUtmOrigin_f.

This probably also means that in the state interface the ltp_def (origin of local frame) should be set in stateSetLocalUtmOrigin_f as well (and vice versa)

PR #1171, opened 2015-05-04
[state interface] fix LLA calculation if only UTM origin initialized

flixr:
as is the normal case in the fixedwing firmware.

While this fixes the immediate issue of making stateGetPositionLla_i() usable in fixedwing firmware,
we really need to properly fix/implement UTM support in the state interface.
That means removing the HACK that ENU = UTM!

See also #455 and #641


gautierhattenberger:
the fix is fine for me

PR #1173, opened 2015-05-04
[gcs] add -ac_ids option to specify list of displayed aircrafts

### LINKED ISSUE #1056: It should be possible to select which of multiple a/c are seen in each GCS window

flixr:
the parameter needs to be a comma separated list of aircraft ids, e.g.

$ ground_segment/cockpit/gcs -ac_ids 11,22
$ ground_segment/cockpit/gcs -ac_ids 33

to display the Aircrafts with IDs 11 and 22 in first GCS, and only the aircraft with ID 33 in the second one.

Should close #1056

### DIFF
--- a/sw/ground_segment/cockpit/gcs.ml
+++ b/sw/ground_segment/cockpit/gcs.ml
@@ -375,6 +375,7 @@ let options =
     "-zoom", Arg.Set_float zoom, "Initial zoom";
     "-auto_hide_fp", Arg.Unit (fun () -> Live.auto_hide_fp true; hide_fp := true), "Automatically hide flight plans of unselected aircraft";
     "-timestamp", Arg.Set timestamp, "Bind on timestampped telemetry messages";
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
I'm fine for this one

PR #1174, opened 2015-05-04
[rotorcraft] heading integration protection in RC-event

### LINKED ISSUE #452: heading integration protection in RC-event
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.c#L105

The PSI += .... is highly depending on the RC-frame rate. With most RC this is not a problem but with variable message rates (e.g. rc_datalink dropping messages or joysticks sending too many frames) this line becomes dangerous.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/925828-heading-integration-protection-in-rc-event?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flixr:
The PSI += .... was highly depending on the RC-frame rate.
With most RC this is not a problem but with variable message rates (e.g. rc_datalink dropping messages or joysticks sending too many frames) this line becomes dangerous.

So measure the time between to calls to read_rc_setpoint function instead of using a fixed RC_UPDATE_FREQ.
To prevent really weird things from happening, bound dt to 0.5s.

Don't partuculariliy like that this pulls in the sys_time dependency here,
but couldn't think of something better...

should close #452

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.c
@@ -31,10 +31,7 @@
 #include "firmwares/rotorcraft/guidance/guidance_h.h"
 #include "firmwares/rotorcraft/stabilization/stabilization_attitude.h"
 #include "firmwares/rotorcraft/autopilot_rc_helpers.h"
-
-#ifndef RC_UPDATE_FREQ
-#define RC_UPDATE_FREQ 40
-#endif
+#include "mcu_periph/sys_time.h"
 
 #ifndef STABILIZATION_ATTITUDE_DEADBAND_A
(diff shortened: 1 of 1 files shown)


OpenUAS:
Ah, great so much welcomed, thx!

flixr:

gautierhattenberger:
The only addition I can see is to keep the possibility of a fixed `dt`, in case your very confident and want to avoid jitter (can this really happen ?)

flixr:
Keeping possibility for fixed DT could be done, but I would say we add this only if someone actually needs it....

PR #1175, opened 2015-05-05
[rotorcraft] predefined motor_mixing for common configurations

flixr:
The intention is to make it easier to configure for common multitrotor configurations.

Motor counting/numbering order is always clockwise, starting with the front motor (for + frame) or left front motor (x frame). This first motor always turns CW, and alternates from there...

Also define `MOTOR_<position>` (e.g. `MOTOR_FRONT_LEFT`) for each motor to be able to use that in the command laws.

### DIFF
--- a/conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
+++ b/conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
@@ -57,24 +57,16 @@
   </commands>
 
   <section name="MIXING" prefix="MOTOR_MIXING_">
-    <define name="TRIM_ROLL" value="0"/>
-    <define name="TRIM_PITCH" value="0"/>
-    <define name="TRIM_YAW" value="0"/>
-    <define name="NB_MOTOR" value="4"/>
-    <define name="SCALE" value="256"/>
     <!-- front/back turning CW, right/left CCW -->
-    <define name="ROLL_COEF"   value="{    0,    0, -256,  256 }"/>
-    <define name="PITCH_COEF"  value="{  256, -256,    0,    0 }"/>
-    <define name="YAW_COEF"    value="{ -256, -256,  256,  256 }"/>
-    <define name="THRUST_COEF" value="{  256,  256,  256,  256 }"/>
+    <define name="TYPE" value="QUAD_PLUS"/>
   </section>
 
   <command_laws>
     <call fun="motor_mixing_run(autopilot_motors_on,FALSE,values)"/>
-    <set servo="FRONT"  value="motor_mixing.commands[0]"/>
-    <set servo="BACK"   value="motor_mixing.commands[1]"/>
-    <set servo="RIGHT"  value="motor_mixing.commands[2]"/>
-    <set servo="LEFT"   value="motor_mixing.commands[3]"/>
+    <set servo="FRONT" value="motor_mixing.commands[MOTOR_FRONT]"/>
+    <set servo="RIGHT" value="motor_mixing.commands[MOTOR_RIGHT]"/>
+    <set servo="BACK"  value="motor_mixing.commands[MOTOR_BACK]"/>
+    <set servo="LEFT"  value="motor_mixing.commands[MOTOR_LEFT]"/>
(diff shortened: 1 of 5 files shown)


conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
```diff
@@ -57,24 +57,16 @@
...
-    <define name="TRIM_PITCH" value="0"/>
-    <define name="TRIM_YAW" value="0"/>
-    <define name="NB_MOTOR" value="4"/>
-    <define name="SCALE" value="256"/>
     <!-- front/back turning CW, right/left CCW -->
-    <define name="ROLL_COEF"   value="{    0,    0, -256,  256 }"/>
-    <define name="PITCH_COEF"  value="{  256, -256,    0,    0 }"/>
-    <define name="YAW_COEF"    value="{ -256, -256,  256,  256 }"/>
-    <define name="THRUST_COEF" value="{  256,  256,  256,  256 }"/>
+    <define name="TYPE" value="QUAD_PLUS"/>
```
podhrmic: is it still possible to define my own mixing config? I.e. what happens if I dont define TYPE, but just leave the mixing section there?

conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
```diff
@@ -57,24 +57,16 @@
   </commands>
 
   <section name="MIXING" prefix="MOTOR_MIXING_">
-    <define name="TRIM_ROLL" value="0"/>
```
podhrmic: is the trim not used anymore?

conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
flixr: Sure, either you define the type to get a common default or you define it yourself exactly like before.

conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
flixr: Sure it is, but the default for the trim is zero, so you might as well omit it...

conf/airframes/examples/quadrotor_lisa_m_2_pwm_spektrum.xml
hooperfly: Integrated the feature into my [hooperfly_motor_mixing] branch to exercise the HEXA_X code. Identified an issue with the MOTOR_MIXING_ROLL_COEF rotation direction in sw/airborne/subsystems/actuators/motor_mixing_types.h on line 82:

#define MOTOR_MIXING_ROLL_COEF   {  181, -181, -256,  181, -181,  256 }
should be
#define MOTOR_MIXING_ROLL_COEF   {  181, -181, -256,  -181, 181,  256 }

Will report back after field testing of the HEXA_X default settings as defined in sw/airborne/subsystems/actuators/motor_mixing_types.h

flixr: thx, fixed

hooperfly:
Synched with all the updates and flew six batteries worth of test flights with the motor_mixing updates. Completed all blocks including rectangle survey and CAM. It seems stable enough to merge.

OpenUAS:
A <define name="TYPE" value="QUAD_PLUS_CCW"/>  "QUAD_CROSS_CCW could be beneficial since we have rotorcraft that "push" obstacles or long grass to outside insted of grabbing it in...
Where CCW is upperleft of front motor direction seen from top.

flixr:
I actually omitted the configurations that just change rotation direction on purpose.
So that you have some standard configs with easier to follow configuration and documentation....
IMHO that makes more sense than adding pre-defined ones for every possible combination.

gautierhattenberger:
And what about a flag to reverse all the motors direction of the basic types ? So it doubles the number of available configurations at the lowest price.

agressiva:
nice ideia

PR #1179, opened 2015-05-12
[sdcard] direct spi sd-card peripheral, sd logger module and test driven development framework

bartslinger:
The pull request consists of three parts:
* An sd-card spi peripheral, for an SD card that is directly wired to a SPI port. It has functions for reading/writing blocks
* A logger module that uses the SD card connected to the SPI port. A python application is included which starts/stops/downloads recordings through telemetry messages.
* An implementation of a framework for Test Driven Development, using unity and cmock.

The benefit of this is that a small autopilot such as a Lisa/S can be equipped with a light-weight micro-SD card for high speed logging. The micro-SD card can be soldered directly to SPI pins on the autopilot board. We are now able to log IMU data at high speed, for the development of better control loops on small helicopters.

sdlogger_download_app_screenshot

sdcard_thumb


OpenUAS:
Bravo!

conf/modules/sd_logger_spi_direct.xml
```diff
@@ -0,0 +1,33 @@
...
+  <header>
+    <file name="sd_logger_spi_direct.h"/>
+  </header>
+  <init fun="sd_logger_start()"/>
+  <periodic fun="sd_logger_periodic()" autorun="TRUE"/>
+  <datalink message="SETTING" fun="sd_logger_command()"/>
+  <makefile target="ap">
+    <define name="SPI_MASTER" value="1" />
+    <define name="USE_SPI2" value="1" />
+    <define name="SD_LOGGER_SPI_LINK_DEVICE" value="spi2" />
```
gautierhattenberger: all SPI device and slave definition should not be hardcoded

conf/modules/sd_logger_spi_direct.xml
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="loggers">
```
gautierhattenberger: Should be something like  `name="sd_logger_spi_direct" dir="loggers"`.
Also, other logger module files are named `logger_<something>.xml`, so it should probably be `logger_sd_spi_direct.xml` here.

sw/airborne/modules/loggers/sd_logger_spi_direct.c
```diff
@@ -0,0 +1,388 @@
...
+#include "subsystems/datalink/telemetry.h"
+
+#ifdef TEST
+/* The original messages.h uses inline functions, which are incompatible with cmock. TEST is defined by the unittest framework */
+#include "messages_testable.h"
+#endif // TEST
+
+#include "subsystems/imu.h"
+#include "subsystems/actuators/actuators_pwm_arch.h"
+#include "sd_logger_spi_direct.h"
```
gautierhattenberger: better use complete path `modules/loggers/sd_logger_spi_direct.h`

sw/airborne/modules/loggers/sd_logger_spi_direct.c
```diff
@@ -0,0 +1,388 @@
...
+      }
+      /* else: state unchanged */
+      break;
+
+      /* Logging data, write data to buffer */
+    case SdLogger_Logging:
+      sdlogger.packet_count++;
+      sd_logger_uint32_to_buffer(sdlogger.packet_count, &sdcard1.output_buf[SD_LOGGER_BUFFER_OFFSET + sdlogger.buffer_addr]);
+      sd_logger_int32_to_buffer(imu.accel_unscaled.x,
+                                &sdcard1.output_buf[SD_LOGGER_BUFFER_OFFSET + sdlogger.buffer_addr + 4]);
```
gautierhattenberger: All this part is very user specific. Can't we have something more generic with a buffer filled by an external user module ?

sw/ground_segment/python/sdlog_download_app/sdlogdownloadframe.py
```diff
@@ -0,0 +1,220 @@
...
+            self.unique_id = message_values[3]
+            self.statusDataLabel.SetLabel(message)
+            # Set the unique ID before reading, otherwise zeros will be returned
+            setting_index = self.settings.name_lookup["sdlogger.unique_id"].index
+            IvySendMsg("dl DL_SETTING %s %s %s" % (self.ac_id, setting_index, self.unique_id))
+            #self.RequestNextPacket() do this in callback instead
+        elif self.last_command == 57:
+            self.download_timer.cancel()
+            self.inDataLabel.SetLabel(message)
+            fh = open("logfile.txt", "a")
```
gautierhattenberger: Would be better to have the log file name as a parameter, or even better, chosen from a selection box.

gautierhattenberger:
If I get it correctly, what you get in your log file is the end are the logged value in binary (or a list of LOG_DATAPACKET messages, which is the same), but with no start/stop delimiter, so you have to know the format of the data and be sure that have not missed any of them. Am I correct ?

My overall suggestion is to make the sd_logger module independent of the data you want to log (i.e. logging data from an input buffer if there is something to log in there). Then, implement the generic device interface that is used for uart, udp or sdlog on Apogee boards using ChibiOS. With this, you can directly write data on the SD using the standard message functions, by choosing the proper device, and with the pprzlog transport, so they have timestamps, start byte, checksum and so on.

Look at this files for ideas:
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/datalink/pprzlog_transport.h
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/datalink/pprzlog_transport.c
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/mcu_periph/uart.c#L202
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/chibios-libopencm3/chibios_sdlog.c#L86
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/loggers/flight_recorder.c

An other job will be to make a common API for the different loggers...

martinmm:
The first blocks of an SD card (usually holding the FAT table) seems to be organized differently than the rest of the card. I had seen smaller block sizes leading to lower speeds / more delay when writing at the beginning of a card for logging. Maybe you start at some offset (not 0) to use the "real" data sectors?

This suggests 4MiB:

http://3gfp.com/wp/2014/07/formatting-sd-cards-for-speed-and-lifetime/

conf/modules/sd_logger_spi_direct.xml
bartslinger: What do you mean exactly? I used the same structure as logger_spi_link.xml

sw/airborne/modules/loggers/sd_logger_spi_direct.c
bartslinger: I agree and will think about it, but am afraid it will be too much work for me.

bartslinger:
I agree with the comments.
@gautierhattenberger You are correct. I'm not sure about your suggestion. I am currently running the logger on a Lisa/S, not using ChibiOS.

gautierhattenberger:
I know you don't have ChibiOS, but the "generic" device interface is not related to this.
I understand you may not have much time to spend on this, so I'll wait for some other core devs feedback to decide if we integrate this with specific message and interface and improve later or if we try to do it correctly from the start.

conf/modules/sd_logger_spi_direct.xml
gautierhattenberger: logger_spi_link is not correct either...
at least have a look at conf/modules/logger_dataflash.xml which is a bit more generic even if it has default values.

flixr:
Nice!
But I agree with @gautierhattenberger that we should rather implement this properly...
Also if you make pull requests, please try not to put everything into one commit, but rather commit logical units of change. e.g. one for the peripheral, one for the module, one for tests (especially since this introduces new dependencies).

sw/airborne/modules/loggers/sd_logger_spi_direct.h
```diff
@@ -0,0 +1,105 @@
...
+
+extern struct SdLogger sdlogger;
+
+/* Public functions */
+extern void sd_logger_start(void);
+extern void sd_logger_periodic(void);
+extern void sd_logger_command(void);
+extern void sd_logger_stop(void);
+
+/* Private functions */
```
flixr: why expose the private functions in the header?

flixr:
Regarding the unit test framework: that would be best in a separate pull request...
Also all our other tests output in [TAP](http://testanything.org/) format for better integration, see e.g. #958 where the simple libtap is used for C.
At a glance, unity does not seem to provide much more than [libtap](https://github.com/zorgnax/libtap).
CMock looks nice, but again pulls in another dependency: ruby (with gems, rake, ..)

flixr:
Why not use an array for the data in the LOG_DATAPACKET message instead of 12 separate fields?

sw/airborne/modules/loggers/sd_logger_spi_direct.h
```diff
@@ -0,0 +1,105 @@
...
+ *  @brief Module for real time logging using an SD Card in SPI mode.
+ */
+
+#ifndef SD_LOGGER_H_
+#define SD_LOGGER_H_
+
+#include "peripherals/sdcard_spi.h"
+
+#define SD_LOGGER_BUFFER_OFFSET 1           /**< Byte reserved for the data token of a multiwrite packet, only applicable to output buffer */
+#define SD_LOGGER_PACKET_SIZE 52            /**< Number of bytes in the LogPacket struct */
```
flixr: would make sense to use sizeof(struct LogPacket) instead of counting and defining it by hand....

flixr:
The sd_card peripheral looks good to me...

flixr:
@bartslinger do you have time to address the points mentioned in the discussion so far?

bartslinger:
I will issue a new pull request with just the sdcard perhipheral and look a bit more into the logger part

PR #1182, opened 2015-05-20
[bebop] Sonar

fvantienen:
Basic Bebop Sonar implementation. Still needs a better algorithm to determine the distance in more cases, but this basic algorithm works most of the time.

### DIFF
--- a/conf/boards/bebop.makefile
+++ b/conf/boards/bebop.makefile
@@ -29,7 +29,7 @@ GPS_PORT           ?= UART1
 GPS_BAUD           ?= B230400
 
 # handle linux signals by hand
-$(TARGET).CFLAGS += -DUSE_LINUX_SIGNAL
+$(TARGET).CFLAGS += -DUSE_LINUX_SIGNAL -D_GNU_SOURCE
 
 # Compile the video specific parts
 $(TARGET).srcs +=  $(SRC_BOARD)/video.c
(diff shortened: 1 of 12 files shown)


gautierhattenberger:
tested and working
closed with 835d9d0da3929ba7f000cf180ae284b7efa318c6

kirkscheper:
Apparently this doesn't work well <0.5m. The ARDrone was accurate till much lower so I would like to look at this again. @fvantienen do you have a data sheet for this that you can send to me?

fvantienen:
This doesn''t work with a datasheet and works indeed worse than on the ARDRONE. This has several reasons, for example the hardware for the Bebop is worse since the transmitter and receiver is the same device. Next to that this new version is fully analog and you have to implement an algorithm yourself.

So what you could improve is the algorithm for peek detection of the returning pulse.

kirkscheper:
Ok, will do. Do you have any formal description of how the sonar/adc work for the bebop? Would be nice to know how you chose some of the values that are hardcoded in the driver.

fvantienen:
Not really, but the `SPI0_MAX_SPEED_HZ` basically generates the sound itself. So you send a spi message which is transmitted to the sonar itself as a pulse to send out. The ADC works just like a normal ADC which just gathers the pulse coming back from the sonar at a certain frequency. So it is basically the intensity of the reflection that comes back.

But please look into: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RangeFinder/AP_RangeFinder_Bebop.cpp
And prolly c+p this, since this is made by Parrot.

PR #1187, opened 2015-05-26
Gcs icon themes

hooperfly:
Add a new set of icons for use with the GCS. The current classic icons are sometimes difficult to see against a black GCS screen(i.e. papgets). The new `flat theme` hopefully improves readability and adds some new icons related to rotorcraft usage vs. fixed wing. Another goal was to introduce new icons without impacting existing users flight plan icon paths and their use of existing `classic` icons.

The new flat theme icons are localized in their own directory named `flat_theme` in `$PAPARAZZI_HOME/data/pictures/gcs_icons`. Along with a new set of icons, this directory contains the master GIMP file named `flat_theme.xcf`.

In addition to the new `flat_theme` directory, the existing set of icons are mirrored in a `classic_theme` directory under  the `$PAPARAZZI_HOME/data/pictures/gcs_icons` directory. By mirroring the data, it provides a migration path from the existing monolithic icon name space to a theme based naming model. So for completeness, a `classic_theme` directory was needed.

Currently, the GCS derives it's icons location from the `gcs_icons_path` variable set in  `$PAPARAZZI_HOME/sw/lib/ocaml/env.ml`

```
let gcs_icons_path = paparazzi_home // "data" // "pictures" // "gcs_icons"
```
Eventually, the `gcs_icons_path` could be overlayed via the aircraft configuration xml file or other template. Whatever the solution for setting the `gcs_icon_path`, it might be good to have an approach that allows for custom icons at the individual aircraft level. 

The approach/solution for setting a theme is open for discussion. There are multiple approaches for solving this issue. In the mean time, by encapsulating theme based icons in their own directories, simple file overlaying techniques can be used to get new icon behaviors.

Note that flight plans and GCS layout files have dependencies on the `gcs_icons_path`. The `strip_icon` attribute in flight plans and the `icon` property in GCS layout files assume icon files are rooted at the location defined by `gcs_icons_path`. Minimizing negative impacts to existing user's flight plans and GCS layout files is important. Hopefully, it would be "transparent".

This is an image of the new "flat_theme" used for strip buttons and as papgets.
screen shot 2015-05-18 at 10 47 02 pm


sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
```diff
@@ -47,7 +47,7 @@
 #define ONE_MHZ 1000000
 
 /* Number of low pulses sent to satellite receivers */
-#define MASTER_RECEIVER_PULSES 5
+#define MASTER_RECEIVER_PULSES 4     // RAB : Changed from 5 to 4 to enable the R100 satellite rx to bind successfully
```
gautierhattenberger: it should be go in this pull request

gautierhattenberger:
Changing the icon theme for the complete GCS and not for each A/C is much easier. We could also avoid duplicating icons for the default theme.
Compatibility issues will mostly arise when people have already custom icons. We'll probably need to keep looking in several folders an echo warnings in the console.

sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
hooperfly: @gautierhattenberger this was a modification recommended by @esden when we were having problems binding the specktrum RX. I looked through the mod history for the file and didn't see any change by @esden. I think it's needed.

hooperfly:
@gautierhattenberger are you suggesting we make the `gcs_icons_path` point to the `classic_theme` folder and remove the GCS icons at the existing location? 

```
let gcs_icons_path = paparazzi_home // "data" // "pictures" // "gcs_icons" // "classic_theme"
```
That gets rid of the duplication but might impact existing users embedded icon pathing inside of their flight plans and/or GCS layout files. I left them in the existing directory as part of a migration plan.

hooperfly:
@gautierhattenberger with respect to the `gcs_icons_path` binding, do you think adding a new property to the GCS section of the aircraft definition to override the `gcs_icons_path` is a good strategy? Something like:

```
  <section name="GCS">
   <define name="ICONS_THEME" value="flat_theme">
    <define name="ALT_SHIFT_PLUS_PLUS" value="30"/>
    <define name="ALT_SHIFT_PLUS"      value="5"/>
    <define name="ALT_SHIFT_MINUS"     value="-5"/>
    <define name="SPEECH_NAME"         value="Teensy Fly Hexa"/>
    <define name="AC_ICON"             value="hexarotor_x"/>
  </section>
```

sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
gautierhattenberger: I know, I was just saying that this modification should be evaluated in an other pull request since it is not related to the main topic

hooperfly: What's the best way to "split" this out so that it gets pulled into `master`? This is some code that didn't get pushed as part of a prior pull. Mostly due to my inexperience with git. Thanks.

hooperfly: @gautierhattenberger looking for guidance related to the above note. Thanks.

hooperfly:
@gautierhattenberger spoke with @esden about cleaning up this pull request. I will  move the HooperFly specific files to a new pull request as well as the "setup_lisam(x)" specific files to their own pull request.  Leaving only the gcs_icon_theme related files for this pull request.

hooperfly:
I will close this pull request and open a "clean" one. Thanks.

PR #1188, opened 2015-05-27
[settings] separated settings files

TobiasMue:
since we have the opportunity to select more than just one setting file in GCS and it is really simple to select/deselect each, there is no need to have everything in one file for every possible configuration
therefor I stated separating those files
next step should be to delete the old files where everything is mixed up
this should make it much easier to keep them up to date and if someone wants to make his own one he can easily look up what is needed or available for each setting block
in addition I found a small typo inane of the files...

### DIFF
--- a/conf/conf_example.xml
+++ b/conf/conf_example.xml
@@ -171,7 +171,7 @@
    radio="radios/Corona_24_DIY.xml"
    telemetry="telemetry/default_fixedwing_imu.xml"
    flight_plan="flight_plans/basic.xml"
-   settings="settings/fixedwing_basic.xml settings/control/ctl_energyadaptive.xml"
+   settings="settings/fixedwing_basic.xml settings/control/ctl_energy.xml settings/control/ctl_adaptive.xml settings/control/ctl_nav.xml"
    settings_modules="modules/air_data.xml modules/digital_cam.xml"
    gui_color="blue"
   />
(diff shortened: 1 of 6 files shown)


gautierhattenberger:
I completely agree on the principal. The issue is that for people using these settings files (my case), the configuration will not be broken but things will be missing.
I'd like to have more feedbacks from other users/devs.

flixr:
Before the end of the year we will hopefully have "merged" subsystem and module configurations with support for auto loading settings for subsystems like for modules... then it will be solved in a nicer way...
I have no strong opinion in how we handle it until then...

TobiasMue:
maybe we could merge these files anyway and just rename them to something like plane_ctrl_nav.xml or plane_ctrl_energy.xml etc…. this won't brake anything and no one is missing anything and the name should speak for it self. could be a nice transition until self loading really exists.

OpenUAS:
@TobiasMue or other devs, feel free to set to 5.8 release if really needed.

dewagter:
- changing splitting the settings sound fine
 - module-settings are better
 - in case settings are split, this requires a user change then combine with a release and mailinglist item.
 - since there are many special changes requiring user action like this, we could argue that the next release is 6.0

podhrmic:
Is this still a thing? IMHO the functionality already exists in the master...?

flixr:
Closing, as most of these settings will be moved into the respective modules soon anyway....

PR #1190, opened 2015-05-27
[stabilization] yaw damper and active lift ctrl

TobiasMue:
something up for testing (it already passed some real world test flying)
- adaptive ctrl also for pitch
- a yaw damper with full turn compensation
- active trim of by for straight aerodynamic flights
- active lift control for those with four flaps on the wing

the yaw damper is for fixed wings with small stability problems in the yaw axis or for every one else who just don t want to guess anymore about this line
    set servo="RUDDER" value="@YAW + @ROLL*COMBI_SWITCH"/>
just change it to 
    set servo="RUDDER" value="@YAW"/>
and the yaw damper will take care of the rest
I did some testing with two different planes and it looks much better than without but anyone is invited to share is experience!

the active lift loop was only meant to be an idea but I already  filled in quiet a part...
so far I only did bench testing since I haven't a plane with four flap configuration - should work never the less. 

if some one wants to test please setup you airframe file for the servos and H_CTL section like this:

    <set servo="RUDDER" value="@YAW"/>
    <set servo="FLAP" value="@CL"/>
    <define name="YAW_LOOP" value="TRUE"/> 
    <define name="YAW_TRIM_BY" value="TRUE"/>
    <define name="YAW_DGAIN" value="5000."/>
    <define name="YAW_BY_IGAIN" value="500."/>
    <define name="CL_LOOP" value="TRUE"/>
    <define name="CL_LOOP_USE_AIRSPEED_SETPOINT" value="TRUE"/> 
    <define name="CL_FLAPS_STALL" value="0.66"/>
    <define name="CL_FLAPS_NOMINAL" value="0."/>
    <define name="CL_FLAPS_RACE" value="-0.2"/>
    <define name="CL_DEADBAND" value="0."/>


gautierhattenberger:
the gpio_timer module shouldn't be in this pull request

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
```diff
@@ -489,6 +553,17 @@ inline static void h_ctl_pitch_loop(void)
   h_ctl_ref.pitch_accel = 0.;
 #endif
 
+#ifdef USE_KFF_UPDATE
```
gautierhattenberger: Are you really using this ? I did this a really long time ago but never really finished it...

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
```diff
@@ -520,3 +596,103 @@ inline static void h_ctl_pitch_loop(void)
...
+ #else
+  float by = 0;
+ #endif
+
+  if (pprz_mode == PPRZ_MODE_MANUAL || launch == 0) {
+    h_ctl_yaw_by_sum_err = 0.;
+  } else {
+     if (h_ctl_pitch_igain > 0.){
+//     only update wehn: phi<60degrees and by<2g
+       if(fabs(stateGetNedToBodyEulers_f()->phi) < 1.05 &&
```
gautierhattenberger: use `fabsf` for operation on float

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
```diff
@@ -520,3 +596,103 @@ inline static void h_ctl_pitch_loop(void)
...
+//     max half rudder deflection for trim
+       BoundAbs(h_ctl_yaw_by_sum_err, MAX_PPRZ/(2* h_ctl_yaw_by_igain));
+       }
+   } else {
+      h_ctl_pitch_sum_err = 0.;
+    }
+  }
+#endif
+
+float Vo = *stateGetAirspeed_f();
```
gautierhattenberger: If you activate the yaw control without the airspeed, it may result in bad value for `Vo`, most likely a saturation to STALL_AIRSPEED.
Maybe a constant "nominal" airspeed should be used if airspeed is not available.

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
```diff
@@ -520,3 +596,103 @@ inline static void h_ctl_pitch_loop(void)
...
+#if H_CTL_YAW_LOOP
+inline static void h_ctl_yaw_loop(void)
+{
+ 
+#if H_CTL_YAW_TRIM_BY
+  // Actual Acceleration from IMU:
+ #ifndef SITL
+  struct Int32Vect3 accel_meas_body;
+  struct Int32RMat *body_to_imu_rmat = orientationGetRMat_i(&imu.body_to_imu);
+  int32_rmat_transp_vmult(&accel_meas_body, body_to_imu_rmat, &imu.accel);
```
gautierhattenberger: IMU data (at least the `accel` here) should not be accessed directly in control loops.
You have the solution to get the NedAccel and NedToBodyRMat from state interface, or bind to sensor message like ins/ahrs filters.
Here, I suggest solution 1.

sw/airborne/firmwares/fixedwing/main_ap.c
```diff
@@ -425,6 +425,10 @@ static inline void telecommand_task(void)
 
     /** Pitch is bounded between [-AUTO1_MAX_PITCH;AUTO1_MAX_PITCH] */
     h_ctl_pitch_setpoint = FLOAT_OF_PPRZ(fbw_state->channels[RADIO_PITCH], 0., AUTO1_MAX_PITCH);
+#if H_CTL_YAW_LOOP
+    /** Yaw is bounded between [-AUTO1_MAX_YAW_RATE;AUTO1_MAX_YAW_RATE] */
+    h_ctl_yaw_rate_setpoint = FLOAT_OF_PPRZ(fbw_state->channels[RADIO_YAW], 0., AUTO1_MAX_YAW_RATE);
```
gautierhattenberger: I would suggest a dead band on rate setpoint to prevent a divergence of the integral term.

gautierhattenberger:
The indentation should be fixed at some places.

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
TobiasMue: just started to experiment with it, at least it is not doing any harm so far. deftly need more testing on this. but I noticed that it was only implemented for roll so added it for pitch as well.
you are not using it at all? any drawbacks from you're side?
maybe we can separate both loops one for roll and one for pitch? two flags, so everyone can choose if he wants to use both or just one...

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
TobiasMue: fixed

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
TobiasMue: fixed
it is now using NOMINAL_AIRSPEED in case USE_AIRSPEED is not defined

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
TobiasMue: thanks for the tip - but I really don't know how to use/implement this…
could you be so kind and contribute the better implementation

sw/airborne/firmwares/fixedwing/main_ap.c
TobiasMue: hmm… there is no integral therm depending on this. the yaw damper just has an D-term 
the I-Term you are probably referring to is only using "by" which is not directly affected by the stick on you're RC

TobiasMue:
I m not sure if know what you mean by indentation
its a start! and it is more than we had before… of course we can always go deeper into the theory. but lets do it step by step?!

sw/airborne/firmwares/fixedwing/main_ap.c
```diff
@@ -425,6 +425,10 @@ static inline void telecommand_task(void)
 
     /** Pitch is bounded between [-AUTO1_MAX_PITCH;AUTO1_MAX_PITCH] */
     h_ctl_pitch_setpoint = FLOAT_OF_PPRZ(fbw_state->channels[RADIO_PITCH], 0., AUTO1_MAX_PITCH);
+#if H_CTL_YAW_LOOP
+    /** Yaw is bounded between [-AUTO1_MAX_YAW_RATE;AUTO1_MAX_YAW_RATE] */
+    h_ctl_yaw_rate_setpoint = 10 * FLOAT_OF_PPRZ(fbw_state->channels[RADIO_YAW], 0., AUTO1_MAX_YAW_RATE);
```
gautierhattenberger: why is it multiply by 10 ?

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
```diff
@@ -520,3 +596,107 @@ inline static void h_ctl_pitch_loop(void)
...
+/* CL with Flaps - direct lift control */
+#if H_CTL_CL_LOOP
+inline static void h_ctl_cl_loop(void)
+{
+
+#ifndef SITL
+  struct Int32Vect3 accel_meas_body;
+  struct Int32RMat *body_to_imu_rmat = orientationGetRMat_i(&imu.body_to_imu);
+  int32_rmat_transp_vmult(&accel_meas_body, body_to_imu_rmat, &imu.accel);
+  float bx = ACCEL_FLOAT_OF_BFP(accel_meas_body.z);
```
gautierhattenberger: Not sure I get what you want to do. Is it x or z ?
Can you explain a little bit the CL control ?

gautierhattenberger:
Can you try the stabilization branch from the paparazzi repository ?
The last part of the CL control is probably wrong because with my changes, the accels over z is centered on 0, not 9.81 (imu returns accel - gravity, state interface only returns accel).
Some comments and formulas to explain the control would be really helpful.

sw/airborne/firmwares/fixedwing/main_ap.c
TobiasMue: since it is a rate and not an angle like the other ones it was just to low to make a difference. so I peeked into the rate loop of stabilization_attitude.c (line 353)
thats working much better - never the less 10 might not be the final solution but I really don't want to make it configurable.

sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c
TobiasMue: the general idea was that for a 4 flap fixed wing Flaps could be moved by the Cl loop depending on the airspeed
if you are using it only in auto 1 you probably want to use the IAS to fly always with the best flap setting
if you plan to use auto 2 it is probably better to use the set point as this reduces actuator control movement and variation from the airspeed set point should be small 
in addition if you are flying circles with lets say 60 degrees you will need more lift (*sqrt g) at the same speed than in horizontal flight - that where bz is used to virtually reduce the IAS to get more Flap movement. I´ll correct the wrong bx its a typo
but as I stated, this is more an idea which I can't really test without a proper 4 flap plane.
I already asked Rolf if he could put this into his new glider to do a proper analysis if it really increases performance - and he will as soon as time permits.
its working fine on the bench though (tester with my ailerons set as flaps)

TobiasMue:
your are right - with bz centered on 0 it wont work… as well as the bound on +- 9.81 of bz which doesn't make sense.
the idea is that the stall speed as well as the needed Lift in depending on the acceleration in z direction. so if stall speed is increasing by Vstall*sqrt(g) I m reducing the measured IAS virtually by the difference to force the CL loop to command more flaps.
the bound I set was between 0 and 2 g to prevent negative flap movement du to acc as well as unreasonable high acc in turns
maybe we could just add the 9.81 to bz now it is centered 0?
if you like we could discuss this on Skype?!

sw/airborne/firmwares/fixedwing/main_ap.c
gautierhattenberger: The `FLOAT_OF_PPRZ` macro is scaling the radio input the `[-AUTO1_MAX_YAW_RATE; AUTO1_MAX_YAW_RATE]`, and it is compared afterwards to the yaw rate, so it should be with the same scaling.
So if the AUTO1_MAX_YAW_RATE value is really the maximum yaw rate you want to apply (in radian), there is no reason to multiply it.

gautierhattenberger:
Is `bz` the normal load factor ? In France we usually write it `nz`, that's why I was confused. Of course, you can add the gravity to the accel (in inertial frame ?) to get the value you where expecting.
I'll have an other look with a colleague that might be interested and we'll see if we plan a skype meeting.

sw/airborne/firmwares/fixedwing/main_ap.c
TobiasMue: thats wired… without the *10 its definitely way too low (not noticeable at all) and with it its just fine… checked it again in flight today and it is working.  
and the max yaw defined (150 deg/s) seems rather too high then too low
any idea?

gautierhattenberger:
I have updated the stabilization branch from paparazzi repository. I changed a little bit your formula. If you think I made a mistake or misunderstood what you wanted to do, we'll try to have a skype meeting if you want.

TobiasMue:
Thank you very much Gautier! 
I m using normally nz as well but walking trough the code I found bx already used so I went with this definition and made up a bz. 

I tried testing it to verify it is working as aspected but could t compile :-( 
I get this error message trying:
<command-line>:0:0: note: this is the location of the previous definition
firmwares/fixedwing/stabilization/stabilization_adaptive.c: In function 'h_ctl_cl_loop':
firmwares/fixedwing/stabilization/stabilization_adaptive.c:661:12: error: request for member 'z' in something not a structure or union
   accel_ned.z -= ACCEL_BFP_OF_REAL(9.81f);

although I m familiar with the coordinate systems used in aviation and how to get from one to another I seem to have lost it totally in ppz. 
Is nz really defined in the airframe fixed coordinate system?

I just made a little addition to the branch - I think since this seems to be more and more an academic approach and might result in constant heavy actuator moving it should be explicitly defined if someone wants to use it.

skype is always welcome, just write a mail when it suits you and I ll try to find a internet connection that is stable enough ;-)

gautierhattenberger:
Sorry for the bug, my fault, it should be fixed in paparazzi/stabilization branch. If I'm not wrong, the load factor is correctly computed in the body axis.
Also, you don't need to make a new pull request for each addition, just update your branch (stabilization in this case) and push it again on github (use --force option if needed). This will also update the pull request.

TobiasMue:
updated! but still unable to compile…

error:

<command-line>:0:0: note: this is the location of the previous definition
firmwares/fixedwing/stabilization/stabilization_adaptive.c: In function 'h_ctl_cl_loop':
firmwares/fixedwing/stabilization/stabilization_adaptive.c:661:41: error: conversion to non-scalar type requested
   struct Int32Vect3 accel_ned = (struct Int32Vect3)(*stateGetAccelNed_i());
                                         ^
firmwares/fixedwing/stabilization/stabilization_adaptive.c:663:3: error: incompatible type for argument 3 of 'int32_rmat_vmult'
   int32_rmat_vmult(&accel_meas_body, ned_to_body_rmat, accel_ned);
   ^
In file included from ./state.h:34:0,
                 from firmwares/fixedwing/stabilization/stabilization_adaptive.c:33:
./math/pprz_algebra_int.h:365:13: note: expected 'struct Int32Vect3 *' but argument is of type 'struct Int32Vect3'
 extern void int32_rmat_vmult(struct Int32Vect3 *vb, struct Int32RMat *m_a2b,
             ^

TobiasMue:
i did some more testing. started to concentrate on trim-by, since we haven't changed very much of the code on that one. seems it is trimming - somehow but rather very very slow… so I started to check on by (this should be the acceleration into y direction in body coordinates) but when I look at the values for by, it doesn't make sense. it is somehow related to acc in .y but the scale seems to be off (not only by a factor of 10)  this is really wired! have you checked on the values when you changed it? or could you pleas do so. maybe I m missing something.

flixr:
Just a general remark: I'm not a fan of adding evermore ifdefs and extra loops to already existing algorithms.
This makes it increasingly harder to read/understand/maintain and use it.

gautierhattenberger:
how do you suggest to add this then ?

flixr:
I'm not quite sure what the best way is and I didn't look into this addition in detail...
Just voicing my general opinion on the cluttered and increasingly harder to follow fixedwing control loops.
I guess at some point it would probably make sense to split it out into a separate loop or so...

TobiasMue:
@flixr 
I totally agree on the mess we have in fixed wing ctrl 
but this is just a loop like roll or pitch… now we have yaw as well,
its just - not everyone has a rudder or wants to actively control it. 
I don't think we should make a new stabilization for rudder controlled planes to avoid some ifdefs.
@gautierhattenberger 
I found a few things and already changed them! 
- in by the gravity was missing what totally led me into the wrong direction since the small errors in the old by were enough to trim the rudder anyway (just very slow)
this is now working as expected!
and in the CL_LOOP we had a wrong sign. I now changed it so that nz will read +1 if the plane is rested with bank 0 - is that ok with you and your definition?

almost there :-)

TobiasMue:
ok, I did some excessive testing and I think its almost ready. 
had to change the airspeed since airspeed_controlled is not global and constantly changing… the setpoint however should always be lower and is therefor to the safe site with little to no changes at all.

missing:
- the control of the yaw rate in auto1 is too low

gautierhattenberger:
I just pushed a few changes in paparazzi/stabilization branch. I suggest to use the usual lateral load factor `ny`, mostly to be consistent with the CL loop. This may change the gain your currently using.

I'll try to test the AUTO1 with NPS tomorrow in the lab. What is the value of your YAW_DGAIN ?

TobiasMue:
kind of strange to have a load factor in y direction but it looks nice and it is working! 
what I used so far is 

    <define name="YAW_LOOP" value="TRUE"/>
    <define name="YAW_TRIM_NY" value="TRUE"/>
    <define name="YAW_DGAIN" value="5000."/>
    <define name="YAW_NY_IGAIN" value="5000."/>

this is already corrected for the changes you made.

gautierhattenberger:
I just pushed a fix in paparazzi/stabilization branch.

After testing with NPS and a RC as joystick input, I think everything is working, including the yaw rate from RC. Well, probably not as you expected initially.
Maybe, what you were probably expecting was to see the plane gently turn at the desired yaw rate when apply a RC input on yaw only. But this is not going to append if you don't bank the plane. The plane will just slowly turn with the wing leveled. When you were multiplying the input command by 10, it was just like setting a huge gain that was making the turn a bit faster.
Even worse, if you are using some NY_IGAIN, your yaw RC input will be canceled after some time by the integrator that tries to nullify the lateral acceleration (which exists when you try to turn while keeping the wing horizontal).
So, both gains are useful, and the RC input can even be nice to help the turn a little bit, but you can't really use it to turn. It is not a big deal, because this control loop mostly aims at performing nice coordinated turns, without "caring" about the yaw command if I got it right.

Do you agree with this analysis ?

TobiasMue:
I totally agree with you!
But looking from a practical point - I think it is nice to have somehow a relationship between RC-input and rudder deflection in Auto1. This is handy if you land in Auto1 and you have to decrap the plane before touchdown or just want to make small adjustments during approach. With the setup we have now a full RC-YAW command results with steady plane in a rudder deflection of less than 10percent (barely noticeable but moving) this has no use at all! 
Thats probably the reason why in stabilization_attitude.c line 353 they multiplied the same way I did…
I did some test flights and with a much higher input from RC you can easily turn the plane wings level to correct for small adjustments. And as long as you're not forcing the plane to turn for more than a few seconds the influence of the integrator should stay small.

so again, from a mathematical point of view you are totally right 
but for practical reasons…
one way I could think of is to put an RC-rudder-comand on top or to make a factor depending on the gain…
what do you think?

gautierhattenberger:
Concerning the line 353, it is a bit different situation because it is using the roll attitude input to command the roll rate. So you really need a coefficient in there. Hard coding 10 is not really the best solution, but if you look into the history, this has been done in 2006 and untouched (preferably unused as well) since that time.
I understand your point about having something that actually works. But we have already have too much mathematical "approximations" in our code...
What do you think about an other solution: if we consider that we don't really care about control the yaw rate to a specific value (which is usually not possible anyway), I suggest to keep the IGAIN on NY and PGAIN on desired rate computed from the bank angle, but directly add the RC input as an offset on the rudder command. This way you'll have a fast and reliable way do correct your trajectory on the short term.

TobiasMue:
its working now! 
thx to some questions that I got I finally found it. 
The reason was a mistake in my auto1 section - resulting in a way to low max_yaw_rate
so from my side its fine, already added some documentation and cleaned again for unused parameters.
anything that is missing from your side? @gautierhattenberger

gautierhattenberger:
If you're happy like this, we can merge it.

TobiasMue:
:-) :-) :-)

OpenUAS:
Very nice addition!

TobiasMue:
@OpenUAS  thx for the credit - if you need any help to test this addition, it would be a pleasure for me to help.

podhrmic:
@TobiasMue out of curiosity, have you tried to fly with in NPS and have you noticed any significant improvements in terms of flight performance?

TobiasMue:
@podhrmic  no I haven't :-( I m using the old (standard) ppz sim, since I m using it only for flight plan verification. but I have tested the yaw damper extensively in real flight. - for stable planes it results in really good coordinated turns and in no acc in y direction in steady flight. for unstable aircraft it should be able to reduce yaw oscillations tremendously. more testing welcome!    the flaps part is more a theoretical approach with some options so everyone can do tests and can try to improve performance. But I didn't had time to collect enough data to do this myself so far. if you want to try I would be happy to guide you through the first steps ;-)

PR #1191, opened 2015-05-27
[amsys] airspeed sensor

TobiasMue:
- changed default freq to 60

just assuming that if there is an airspeed sensor installed I probably what to use it for auto2
so it should have at least the frequency of the control loops….

### DIFF
--- a/conf/modules/airspeed_amsys.xml
+++ b/conf/modules/airspeed_amsys.xml
@@ -30,7 +30,7 @@
     <file name="airspeed_amsys.h"/>
   </header>
   <init fun="airspeed_amsys_init()"/>
-  <periodic fun="airspeed_amsys_read_periodic()" freq="10."/>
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
Do you know if there is a max refresh rate for this sensor ? Did you already fly with this setting ?

TobiasMue:
from the data sheet:
http://www.analogmicro.de/_pages/sens/ams5812/ams5812_data_sheet.pdf

…these calculated 15-bit digital values (pressure and temperature values) are written to the ASIC output registers and continuously updated (typically every 0.5ms)….

round about 2000Hz… so there should be some safety margin for oversampling and 60Hz is not a problem.
I have it running on the desk at this rate - I noticed that after my last flight.
Do you think there might be a problem with other i2c peripherals?

gautierhattenberger:
closed with 5e1aa2f70d4bfbd0014346eb1dcd17831471ad22

Issue #564, opened 2013-10-22
improve defining of arrays in the airframe file


podhrmic:
In an example airframe quadrotor_lisa_m_2_pwm_spektrum.xml
if I change the stabilization to "float_quat" (and change the settings from "stabilization_att_int.xml" to "stabilization_att_float_quat.xml"), I get the following error:
firmwares/rotorcraft/stabilization/stabilization_attitude_ref_quat_float.c:55:1: error: invalid initializer (see https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_ref_quat_float.c#L54)

Obviously I am initializing an array. However, if I change the airfame config file to include curly braces (define name="REF_OMEGA_P"  value="{800}" unit="deg/s") I get this error:
Error: Unit conversion of parameter REF_OMEGA_P impossible because '{800}' is not a float

(the ocaml parser can't handle it). If I remove the unit attribute from the tag (define name="REF_OMEGA_Q"  value="800" ) it will pass, however then I have to input rad/sec. This has to be done for all attitude gains.

Suggested fix:
1) change the ocaml parser
2) change the float_quat so it doesn't expect an array of values (does anybody besides Transition Robotics uses gain scheduling anywa?)

gautierhattenberger:
for now, closing with #1192 
a new issue might be opened for the settings issue if needed

PR #1192, opened 2015-06-01
[airframe] allow to declare the type of defined value

gautierhattenberger:
- support basic types: int, float, string
- support array of basic types (declared with comma separated values)
- automatic conversion should keep working
- backward compatible
- see [#564]


gautierhattenberger:
closed with 8fa140e83bb2ee19c8ccbdd60fed68becf1b7d7c

flixr:
Hmm.... why use different type names than we have e.g. in messages.xml?
IMHO it would be better to be consistent...

gautierhattenberger:
good point, done with 09e3ba2947c1277fa38572f0993763301ff93f6e

PR #1196, opened 2015-06-03
[fg] Updated flight gear interface and removed obsolete HITL files

podhrmic:
flight_gear.h updated to new version to be compatible with flight gear net_FDM
Removed obsolete HITL files that stayed here just for legacy reason.


flixr:
why also delete the start_fg.sh script?

gautierhattenberger:
It was exactly doing what was expected when trying to load the mikrokopter model. I agree we should have a script to easily start flightgear, but can we try to make it better first ? Like looking into the airframe file for the correct model if a name is passed as argument.

podhrmic:
No program used it directly (like a "launch" button in pprzi), and it wasn't mentioned in the wiki either - so I assumed it was obsolete. But I agree with @gautierhattenberger that we should have something more flexible than that.

flixr:
For which Flight Gear versions was this change needed? What is the minimum FG version now?

podhrmic:
3.4. The difference is in NetFDM interface, and has been there at least sine fg 2.6 
NetGui (what NPS is currently using) has not changed.

flixr:
ok, thanks for clearing that up

PR #1197, opened 2015-06-03
Updated libopencm3 to the latest version. 

podhrmic:
Prerequisite for Lisa MX DFU bootloader.


podhrmic:
I believe the Travis error has nothing to do with the commit ```The command "sudo apt-get install paparazzi-dev paparazzi-jsbsim gcc-arm-none-eabi libipc-run-perl" failed and exited with 100 during .```, can somebody look into it?

podhrmic:
@gautierhattenberger any comments on this?

flixr:
I don't see any harm in updating to latest libopencm3 (although I haven't tested it so far, the spi changes should be ok).
But I'm wondering what change exactly is a prerequisite for the Lisa MX DFU bootloader....?

podhrmic:
I ll have to migrate the SM32F4 bootloader (currently here https://github.com/podhrmic/aggieair-bootloader) to new libopencm3 which is compatible with paparazzi. So why not use the newest version... Plus I remember there were some issues with the older libopencm3 version that paparazzi was using - hence the upgrade.

flixr:
I'm not aware of any issues related to libopencm3 (the version we are currently using), neither could I see any changes that should be relevant for the bootloader...
Need to confirm that SPI is still working properly with the SPI mode setup changes in latest libopencm3...

podhrmic:
I had some issues when I tried to port the bootloader last time, but didn't have enough time to investigate. If it doesn't break anything then it shouldn't hurt to switch to a newer version...

flixr:
merged with 831cde70c12836eeea2b918f4bf0f1a83b40468c

PR #1198, opened 2015-06-03
Chibios update to 2.6.9.

podhrmic:
Update ChibiOS to 2.6.9 version, contains some bugfixes and cleanups. 
Had to remove the -Werror flag, because the ChibiOS USB driver implementation is missing a return statement in a thread which never terminates:
```
paparazzi/sw/ext/chibios/os/hal/platforms/STM32/OTGv1/usb_lld.c: In function 'usb_lld_pump':
paparazzi/sw/ext/chibios/os/hal/platforms/STM32/OTGv1/usb_lld.c:769:1: error: no return statement in function returning non-void [-Werror=return-type]
cc1: all warnings being treated as errors
```
It is not great to have errors like this, but in this case it seems to be fine. Possible solution is to ask for a fix at ChibiOS forum.

### DIFF
--- a/conf/Makefile.chibios-libopencm3
+++ b/conf/Makefile.chibios-libopencm3
@@ -74,7 +74,7 @@ ifeq ($(USE_OPT),)
 
   USE_OPT = -std=gnu11 -O$(CH_OPT) \
 	-falign-functions=16  -fno-omit-frame-pointer\
-	-W -Wall -Werror -Wno-error=unused-variable -Wno-error=format \
+	-W -Wall -Wno-error=unused-variable -Wno-error=format \
 	-Wno-error=unused-function -Wno-error=unused-parameter \
 	$(PPRZ_DEFINITION)
 endif


podhrmic:
@gautierhattenberger any feedback on these?

gautierhattenberger:
I was giving lectures away from my lab so I couldn't really test anything. I'll look at this tomorrow or next week. (Same for libopencm3)

podhrmic:
@gautierhattenberger did you have a chance to test?

gautierhattenberger:
Sorry, but it is not working with libopencm3/chibios integration we are using for SDlog. I will not have time to search for the bug before some time.
Except for the bug fix (which may not affect us because we haven't seen big problems so far), do you have a particular reason to update chibios ?

podhrmic:
With the current version of ChibOS:
- USB-Serial is not working
- ADC driver is not working

So moving to version 2.6.9 will hopefully fix the issue (haven't had time to test it) so we will have full set of drivers. 

Just to make sure  commit #1197 didn't break SDlog, but this one does?

gautierhattenberger:
Actually #1197 have been merged before before I could test the log, but the SDIO driver we are using is the one of ChibiOS, so updating libopencm3 should not change anything.
This pull request is not only breaking the SDlog, it is just crashing the complete autopilot after 1 or 2 seconds.
Also, what is not working with ADC ? I think the ADC driver based on ChibiOS in the rt_chibios branch is working. I don't know for the USB-serial driver.

podhrmic:
I will look into it more then in next few days hopefully and find out what is the difference.

OpenUAS:
@podhrmic , next few days are a bit over ;) is it still an issue? I could ask someone student to look at it ince we've already discussed it and his interest in PPRZ icm ChibiOS.

podhrmic:
@OpenUAS since it is not going to be included in v5.6 it has lower priority. I ll get to it eventually, but you know - priorities:)

podhrmic:
Closing for now.

podhrmic:
Should be marked for v5.8

@gautierhattenberger can you look at where is the problem with the apogee board? All tests compile, and I dont have an Apogee board to do a live test.

podhrmic:
Also, instead of changing the compiler settings (-Werror) I ll submit a patch to ChibiOS.

flixr:
Should we keep this open as you guys are working on integrating chibios 3.0?

OpenUAS:
For next PPRZ Chibi OS v3.x would make indeed more sense IMHO, Ill vote for it..

gautierhattenberger:
Integration of CHibiOS 3 is already ongoing in the rt_paparazzi development branch, so I'm closing this PR and focus on the other branch.

Issue #643, opened 2014-02-24
dynamic sectors


flixr:
Sectors (and their graphic representation in the GCS) should automatically update with waypoint changes that define their corners.

For that we also need a dynamic InsideX function, probably similar to the (unused) flightzone.[ch] but with an additional extension to full 3D (i.e. not only 2D polygon but including alittude).

PR #1204, opened 2015-06-09
Dynamic sectors

gautierhattenberger:
Allow dynamic sectors (#643)

Currently, it is activated by adding the attribute `type="dynamic"` since it is very experimental code.
But it should work with concave hulls.


flixr:
Nice!
Update of the sector borders and the inside function seems to works as expected here.

gautierhattenberger:
Can I merge this ? It is optional for now anyway.

flixr:
jep, good to merge

PR #1205, opened 2015-06-10
stabilization quat_int IERROR scale fix

flixr:
This should (tm) result in less of a quantization error for the integral part (less static offset).
Reduce the GAIN_PRESCALER_I accordingly so that the gains set in the airframe file keep same scale.

While here also simplify the fb command prescalers/scaling...

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_int.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_int.c
@@ -60,11 +60,11 @@ struct Int32Quat stabilization_att_sum_err_quat;
 int32_t stabilization_att_fb_cmd[COMMANDS_NB];
 int32_t stabilization_att_ff_cmd[COMMANDS_NB];
 
-#define IERROR_SCALE 1024
+#define IERROR_SCALE 128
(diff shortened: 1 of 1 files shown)


flixr:
Only tested this in NPS, it should be correct.
@gautierhattenberger would be great if you could cross check the prescalers...

gautierhattenberger:
On the IGAIN, the `sum_err` was divided by1024, then multiplied by 48 and divided 2, so if you have 1024 as input, output is : 1024* (1/1024)*48*(1/2) = 24.
Now you have a division by 128, then multiply by 4, so for the same input you have: 1024*(1/128)*4 = 32.
If I'm correct, by multiplying by 3 instead of 4, you should have the same final value than before.

flixr:
Correct, thanks for checking again :-)
Amended the commit...
I guess should we also increase the INTEGRATOR_BOUND according to the new IERROR_SCALE?

gautierhattenberger:
I guess so. It seems it should be 8 times bigger

flixr:
Actually I think it makes sense to keep the limit, which would translate to just under 180deg (in float representation of the quat int) if would look at it like this...

gautierhattenberger:
I never pay much attention to the `sum_err` value, but if you think it is already high enough, I'm fine.

PR #1207, opened 2015-06-11
INDI simplified

EwoudSmeur:
This is a slightly simplified version of the Incremental Nonlinear Dynamic Inversion (INDI) controller I made for quadrotor attitude control. I tried to condense it to a form where it can be useful and understandable for others.

I submitted a paper to the journal of guidance, control and dynamics which covers all the details. In short: By measuring the angular acceleration (which is obtained by deriving and filtering it from the gyroscope) we can measure how much the actual angular acceleration differs from the desired angular acceleration. By multiplying this with the inverse of the actuator effectiveness we get the correct increment of the input that will result in the desired increment in angular acceleration. The main benefit is that disturbances can be rapidly compensated and nonlinear dynamics are controlled better.

It also contains an adaptive algorithm, that will update the actuator effectiveness in flight. You can take off with a drone without knowing the actuator effectiveness and it will learn this during the flight, removing the need for tuning. The only thing you need is a constant for the first order approximation of the motor dynamics. This can be measured with a step response, and a value for the ARDrone is included. I will upload more airframes later.

I hope it will be easy to apply to other airframes, but it is hard to guarantee that the adaptive least mean squares algorithm will converge to the right values, because it depends on the amount of filtering and the adaptation rate (these are linked). In general: if the airframe is well-damped it will work better.

In the airframe file I kept the normal attitude stabilization section, to be able to switch easily. A small problem is that the setpoint defines need the prefix STABILIZATION_ATTITUDE,  so they are taken from the STABILIZATION_ATTITUDE section instead of the STABILIZATION_INDI section. I wasn't sure how to get around this.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
@@ -1,0 +1,313 @@
+/*
+ * Copyright (C) Ewoud Smeur <ewoud_smeur@msn.com>
+ * MAVLab Delft University of Technology
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/** @file stabilization_attitude_quat_indi.c
+ * MAVLab Delft University of Technology
+ * This control algorithm is Incremental Nonlinear Dynamic Inversion (INDI)
+ *
+ * This is a simplified implementation of the (soon to be) publication in the
+ * journal of Control Guidance and Dynamics: Adaptive Incremental Nonlinear
+ * Dynamic Inversion for Attitude Control of Micro Aerial Vehicles
+ */
+
+#include "firmwares/rotorcraft/stabilization/stabilization_attitude.h"
+#include "firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.h"
+#include "firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.h"
+
+#include "state.h"
+#include "generated/airframe.h"
+#include "paparazzi.h"
+
+#if !defined(STABILIZATION_INDI_ACT_DYN_P) && !defined(STABILIZATION_INDI_ACT_DYN_Q) && !defined(STABILIZATION_INDI_ACT_DYN_R)
+#error You have to define the first order time constant of the actuator dynamics!
+#endif
+
+int32_t stabilization_att_indi_cmd[COMMANDS_NB];
+
+struct FloatRates g1 = {STABILIZATION_INDI_G1_P, STABILIZATION_INDI_G1_Q, STABILIZATION_INDI_G1_R};
+float g2 = STABILIZATION_INDI_G2_R;
+struct ReferenceSystem reference_acceleration = {STABILIZATION_INDI_REF_ERR_P,
+STABILIZATION_INDI_REF_ERR_Q,
+STABILIZATION_INDI_REF_ERR_R,
+STABILIZATION_INDI_REF_RATE_P,
+STABILIZATION_INDI_REF_RATE_Q,
+STABILIZATION_INDI_REF_RATE_R,
+};
+
+struct FloatRates filtered_rate = {0., 0., 0.};
+struct FloatRates filtered_rate_deriv = {0., 0., 0.};
+struct FloatRates filtered_rate_2deriv = {0., 0., 0.};
+struct FloatRates angular_accel_ref = {0., 0., 0.};
+
+struct FloatRates indi_du = {0., 0., 0.};
+struct FloatRates u_act_dyn = {0., 0., 0.};
+struct FloatRates u_in = {0., 0., 0.};
+struct FloatRates indi_u = {0., 0., 0.};
+struct FloatRates udot = {0., 0., 0.};
+struct FloatRates udotdot = {0., 0., 0.};
+
+// Variables for adaptation
+struct FloatRates filtered_rate_estimation = {0., 0., 0.};
+struct FloatRates filtered_rate_deriv_estimation  = {0., 0., 0.};
+struct FloatRates filtered_rate_2deriv_estimation  = {0., 0., 0.};
+struct FloatRates indi_u_estimation  = {0., 0., 0.};
+struct FloatRates udot_estimation  = {0., 0., 0.};
+struct FloatRates udotdot_estimation  = {0., 0., 0.};
+#define SCALE 0.001 //The G values are scaled to avoid numerical problems during the estimation
+struct FloatRates g_est = {STABILIZATION_INDI_G1_P/SCALE, STABILIZATION_INDI_G1_Q/SCALE, STABILIZATION_INDI_G1_R/SCALE};
+float g2_est = STABILIZATION_INDI_G2_R/SCALE;
+float mu = STABILIZATION_INDI_ADAPTIVE_MU;
+
+#ifndef STABILIZATION_INDI_FILT_OMEGA
+#define STABILIZATION_INDI_FILT_OMEGA 50.0
+#endif
+
+#ifndef STABILIZATION_INDI_FILT_ZETA
+#define STABILIZATION_INDI_FILT_ZETA 0.55
+#endif
+
+#define STABILIZATION_INDI_FILT_OMEGA2 (STABILIZATION_INDI_FILT_OMEGA*STABILIZATION_INDI_FILT_OMEGA)
+
+#ifndef STABILIZATION_INDI_FILT_OMEGA_R
+#define STABILIZATION_INDI_FILT_OMEGA_R STABILIZATION_INDI_FILT_OMEGA
+#define STABILIZATION_INDI_FILT_ZETA_R STABILIZATION_INDI_FILT_ZETA
+#endif
+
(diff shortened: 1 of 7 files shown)


flixr:
Looking forward to testing it out :-)

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -0,0 +1,322 @@
...
+#include "state.h"
+#include "generated/airframe.h"
+#include "subsystems/actuators/motor_mixing.h"
+
+#if !defined(STABILIZATION_INDI_ACT_DYN_P) && !defined(STABILIZATION_INDI_ACT_DYN_Q) && !defined(STABILIZATION_INDI_ACT_DYN_R)
+#error You have to define the first order time constant of the actuator dynamics!
+#endif
+
+// dummy gains for systems expecting attitude gains
+struct Int32AttitudeGains stabilization_gains = {
```
flixr: why should these be needed?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.h
```diff
@@ -0,0 +1,66 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#ifndef STABILIZATION_ATTITUDE_QUAT_INDI_H
+#define STABILIZATION_ATTITUDE_QUAT_INDI_H
+
+#include "firmwares/rotorcraft/stabilization/stabilization_attitude_common_int.h"
```
flixr: why include this? especially since indi is implemented in float...

flixr:
Some comments in the code that explain what is happening would really help ;-)

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -0,0 +1,322 @@
...
+}
+
+#define SECOND_ORDER_ACCELERATION(_input, _ddx, _dx, _x, _omega, _omega2, _zeta, _omega_r, _omega2_r) { \
+  _ddx->p = -_dx->p * 2*_zeta*_omega   + ( _input->p - _x->p)*_omega2;    \
+  _ddx->q = -_dx->q * 2*_zeta*_omega   + ( _input->q - _x->q)*_omega2;    \
+  _ddx->r = -_dx->r * 2*_zeta*_omega_r + ( _input->r - _x->r)*_omega2_r;  \
+}
+
+void stabilization_indi_second_order_filter(struct FloatRates *input, struct FloatRates *filter_ddx, struct FloatRates *filter_dx, struct FloatRates *filter_x, float omega, float zeta, float omega_r) {
+  INTEGRATE_RATES(filter_x, filter_dx, PERIODIC_FREQUENCY);
```
flixr: why not use the existing `float_rates_integrate_fi` function here?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -0,0 +1,322 @@
...
+  QUAT_BFP_OF_REAL(stab_att_sp_quat, q_sp);
+}
+
+#define INTEGRATE_RATES(_rates, _derivative, _freq) { \
+_rates->p = _rates->p + _derivative->p/_freq; \
+_rates->q = _rates->q + _derivative->q/_freq; \
+_rates->r = _rates->r + _derivative->r/_freq; \
+}
+
+#define SECOND_ORDER_ACCELERATION(_input, _ddx, _dx, _x, _omega, _omega2, _zeta, _omega_r, _omega2_r) { \
```
flixr: unless you need this to be a macro, please use a static inline function or directly write it in the second_order_filter as it is only used once anyway...

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -0,0 +1,322 @@
...
+}
+
+void stabilization_attitude_run(bool_t enable_integrator) {
+
+  /* Propagate the second order filter on the gyroscopes */
+  struct FloatRates *body_rates = stateGetBodyRates_f();
+  stabilization_indi_second_order_filter(body_rates, &filtered_rate_2deriv, &filtered_rate_deriv, &filtered_rate, STABILIZATION_INDI_FILT_OMEGA, STABILIZATION_INDI_FILT_ZETA, STABILIZATION_INDI_FILT_OMEGA_R);
+
+  /* attitude error                          */
+  struct Int32Quat att_err;
```
flixr: any particular reason to use the fixed point representation here, while the reset of INDI is in float?

flixr:
Please also add doxygen file headers (and probably best move the short description to the doxygen header so it shows up in the generated docs).

flixr:
Since the file logger is used for e.g. the ardrone (without INDI), I think it would make sense to remove it from this pull request.
It should be made configurable anyway instead of using hardcoded variables to log....

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
EwoudSmeur: I intended to write it in fixed point at first, but at some point I decided it was easier to develop in floating point... Though this part I just copied, so I guess it is faster on fixed point processors... should I then change it?

flixr: I think it would be nicer to have this in float as well, you convert it to float again anyway... it's not like one quaternion multiplication in fixed point will save you a lot here...
Also please use the math functions instead of the deprecated macros (which are only kept for backwards compatibility).

flixr:
I also think that adding a stabilization_indi.xml telemetry file is not needed.
You could just add the STAB_ATTITUDE_INDI message to the attitude loop mode of the default_rotorcraft file.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -0,0 +1,305 @@
...
+  indi_du.q = 1.0/g1.q * (angular_accel_ref.q - filtered_rate_deriv.q);
+  indi_du.r = 1.0/(g1.r+g2) * (angular_accel_ref.r - filtered_rate_deriv.r + g2*indi_du.r);
+
+  //add the increment to the total control input
+  u_in.p = indi_u.p + indi_du.p;
+  u_in.q = indi_u.q + indi_du.q;
+  u_in.r = indi_u.r + indi_du.r;
+
+  //bound the total control input
+  BOUND_CONTROLS(u_in.p, -4500, 4500);
```
flixr: could use the Bound macro that already exists...

flixr:
Would also make sense to add the airframe file param names in the settings xml file...
And it would probably be nice to put the variables into a indi struct so we don't have stuff like `udot` in the global namespace

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.h
```diff
@@ -0,0 +1,65 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#ifndef STABILIZATION_ATTITUDE_QUAT_INDI_H
+#define STABILIZATION_ATTITUDE_QUAT_INDI_H
+
+#include "firmwares/rotorcraft/stabilization/stabilization_attitude_ref_quat_int.h"
```
flixr: not needed in header, include in c file instead

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include "math/pprz_algebra_int.h"
```
flixr: should be `#include "math/pprz_algebra_float.h"`

EwoudSmeur:
Thanks for you comments Felix! 
But regarding the fixed point/floating point stuff, I also used fixed point setpoint functions.... is this really and issue? It does not really need to be all in one or the other right?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.h
EwoudSmeur: firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.c:374:51: error: 'stab_att_sp_euler' undeclared (first use in this function)

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
EwoudSmeur: I couldn't find the definition of BOUND anywhere, I think that is why I chose to just write one myself... just to be sure

flixr:
It doesn't really need to be all in the same format, but it makes the code easier to read and understand....
Also in the case of computing the attitude error quat, you probably have less of a quantization error if you do it in float.

hooperfly:
Planning to test this with my HooperFly frames over the weekend. Looking forward to seeing it in action.

EwoudSmeur:
@hooperfly I am very happy that you are interested in using the code! However, I am realizing that I should probably write something on the wiki about how to use this algorithm... So it might be good to wait until next weekend to fly ;)

hooperfly:
@EwoudSmeur looking forward to your wiki entries. :+1:

flixr:
@EwoudSmeur do you regard this as ready to merge now?
I won't be able to really test INDI until the camp :-(
But it seems to work in simulation :-)

Before we merge it, can you plz run it through our code style fixer?

flixr:
One thing I'm wondering about is how got the actuator effectiveness estimation will work in the presence of actuator saturations (especially taking into account that we might limit the yaw authority)....

EwoudSmeur:
I think it is ready to merge now. With the bound on the estimated effectiveness the adaptive part is a lot more robust.

Regarding the saturations, we are actually investigating how to handle this correctly. For the estimation it is no problem if the input bounds within INDI are reached, but if they are afterwards more strictly bounded in the motor mixing this is not taken into account at the moment. But I think this is something to solve later in a nice way.

PR #1211, opened 2015-06-22
[linux] use 8 bit I2C address scheme for all drivers

### LINKED ISSUE #1210: Use 8 bit I2C address scheme for Linux drivers
The imu_mpu60x0_i2c subsystem does not work with STM32 targets. The address gets shifted by one bit in in the i2c init call

https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/imu/imu_mpu60x0_i2c.c#L74

as in other Linux drivers, e.g. Bebop

https://github.com/paparazzi/paparazzi/blob/15f8c182544d6197b21f4da2c88c5b177453abc3/sw/airborne/subsystems/imu/imu_bebop.c#L82

Seems the Paparazzi 8 bit I2C address convention collides with the 7 bit convention in Linux.

ioctl(file, I2C_SLAVE, t->slave_addr);

https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/arch/linux/mcu_periph/i2c_arch.c#L56

ioctl(file, I2C_SLAVE, long addr)
  Change slave address. The address is passed in the 7 lower bits of the
  argument

http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/i2c/dev-interface#n103

Wouldn't it be better to do this bit shift once in the Linux i2c_arch.c rather than in the individual (Linux-only) drivers to keep the code clean and consistent or do I miss something?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22492562-use-8-bit-i2c-address-scheme-for-linux-drivers?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

martinmm:
fixes #1210
needs to be tested as I do not have any Linux target (Bebop, BBB)

### DIFF
--- a/sw/airborne/arch/linux/mcu_periph/i2c_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/i2c_arch.c
@@ -52,8 +52,8 @@ bool_t i2c_submit(struct i2c_periph *p, struct i2c_transaction *t)
 {
   int file = (int)p->reg_addr;
 
-  // Set the slave address
-  ioctl(file, I2C_SLAVE, t->slave_addr);
+  // Set the slave address, converted to 7 bit
+  ioctl(file, I2C_SLAVE, t->slave_addr >> 1);
 
   // Switch the different transaction types
   switch (t->type) {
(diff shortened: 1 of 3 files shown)


martinmm:
fixes #1211 
needs to be tested as I do not have any Linux target (Bebop, BBB)

flixr:
Thx! Works as expected on the bebop.

Issue #1215, opened 2015-06-24
[paparazzi-center] Generic Airframe drop-down dependent sessions


esden:
Currently there is now way to create sessions in the paparazzi-center that are aware which airframe is selected in the drop-down menu. The only session that is aware of that is the '''Simulation''' session.

The '''Simulation session''' is hardcoded in ocaml at the moment.

Motivation:
There is no way to have for example settings application run without creating specific sessions for it as the program takes the ```-ac``` parameter.

Proposal:
Create a "magic" variable that will be replaced by the ocaml code with the selected airframe file in the session command line.

Example:
If you have a session that contains a commandline like:
```
sw/ground_segment/tmtc/settings -ac #AIRCRAFT#
```
it and you have selected the A/C called ```MySuperQuad``` the resulting command call will be:
```
sw/ground_segment/tmtc/settings  -ac MySuperQuad
```

flixr:
Solved with merge of #1217

PR #1217, opened 2015-06-25
[PPRZ Center] Add #AIRCRAFT# placeholder constant.

esden:
When you use #AIRCRAFT# as a constant parameter to a program in your
session, it will be automatically replaced by the name selected in the
A/C dropdown combo box of the Paparazzi Center. This makes it possible
to create generic sessions that start programs which need to know which
aircraft is of interest. For example settings.

This is a possible solution to the #1215 issue.

### DIFF
--- a/conf/control_panel_example.xml
+++ b/conf/control_panel_example.xml
@@ -181,6 +181,19 @@
       <program name="Messages"/>
     </session>
 
+    <session name="Messages and Settings">
+      <program name="Data Link">
+        <arg flag="-d" constant="/dev/ttyUSB0"/>
+        <arg flag="-s" constant="57600"/>
+      </program>
+      <program name="Server">
+        <arg flag="-n"/>
+      </program>
+      <program name="Messages"/>
+      <program name="Settings">
+        <arg flag="-ac" constant="@AIRCRAFT"/>
+      </program>
+    </session>
 
   </section>
 
--- a/sw/supervision/pc_control_panel.ml
+++ b/sw/supervision/pc_control_panel.ml
@@ -244,7 +244,9 @@ let supervision = fun ?file gui log (ac_combo : Gtk_tools.combo) (target_combo :
 	  (fun arg ->
 	    let constant =
 	      try double_quote (Xml.attrib arg "constant") with _ -> "" in
-	    p := sprintf "%s %s %s" !p (ExtXml.attrib arg "flag") constant)
+	    let real_constant =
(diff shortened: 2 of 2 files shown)


gautierhattenberger:
I would prefer `@AIRCRAFT` as key word to be consistent with other variable parameters in airframe file.

esden:
I agree, I have amended the patch to reflect that. :)

fvantienen:
Coul we maybe also add an aircraft id? Since some programs use that (like natnet2ivy).

esden:
Can you point me to an example session that uses natnet2ivy, so I can add it too?

gautierhattenberger:
I know I'm being annoying :), but we usually use AC_ID to point to the aircraft ID

fvantienen:
Because natnet2ivy is written in c and doesnt parse the xml files.
I wanted to be consistent at the start but didn't want to put time in a xml parser in c.

Example session:
natnet2ivy -ac @AC_ID 1

esden:
@gautierhattenberger So we use @AIRCRAFT and @AC_ID? That is not very consistent euther... Unless we use @AC and @AC_ID.

@fvantienen This is off topic of this pull request, but I think you should use a small xml parser still, and maybe create a library that implements getters for simple items. I am sure there will be more C programs in the future, so creating facilities for those is a very good idea. (libpprz.so) We use the name of the aircraft throughout the ground system to stay consistent, so we should stay in that habit I think.

gautierhattenberger:
It is consistent in the way that we use `AIRCRAFT` to give the name of plane to the makefile rule in order to compile the airborne code and we use `AC_ID` to get the numerical id in the airborne code itself.
It is not like this choice is the best, but I don't think changing it is a good idea now.

flixr:
I agree with @gautierhattenberger here...
Also wondering how the added session should be named, it is called "setup actuators" now, but it not in the least specific to that. It's just messages and settings...

esden:
Raw Settings?
No GCS Settings?
Raw Messages and Settings?
...

flixr:
@esden could you plz change AIRCRAFT_ID to AC_ID as @gautierhattenberger suggested?
Also maybe disable the logging in server for the "setup actuators" session?
I would vote to name the session "Messages and Settings" or something like that...

esden:
@flixr sorry, I was convinced that I already did fix the AC_ID part. None the less I have made all the suggested changes and pushed an updated patch set.

PR #1218, opened 2015-06-25
[rotorcraft] make the heading privitive usable for rotorcrafts

flixr:
This make the `heading` flight plan primitive usable for rotorcrafts.
It has a different behaviour than for fixedwings (where you basically specify a course to fly).
For rotorcrafts this sets the heading/yaw.

### DIFF
--- a/conf/flight_plans/rotorcraft_basic.xml
+++ b/conf/flight_plans/rotorcraft_basic.xml
@@ -61,6 +61,13 @@
       <go from="p4" hmode="route" wp="p1"/>
       <deroute block="stay_p1"/>
     </block>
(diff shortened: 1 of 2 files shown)


flixr:
@gautierhattenberger can we merge this as is? Should we keep the `test yaw` block in the rotorcraft_basic flight plan?

esden:
s/privitive/primitive/ ... otherwise I think it is a useful tool for yaw performance testing

gautierhattenberger:
I think we can merge the heading primitive, but for the flight plan I suggest to increase the time to 3 or 4 seconds for each step.

PR #1219, opened 2015-06-26
[NPS] support gaia environment simulator

flixr:
Listen to WORLD_ENV message from gaia environment simulator.

So you can use the gaia gui to adjust the wind dir/speed, gps_availability and time factor in the NPS sim.
This means you can now also more easily adjust the time factor during runtime (without needing to interrupt NPS with CTRL-Z)
Also added a slider for wind up to the gaia gui.


gautierhattenberger:
Looks good to me

PR #1221, opened 2015-06-30
Mavlink pprzservices

LodewijkSikkel:
[modules] datalink/mavlink: Added functionality to be able to communicate with a Android-based GCS. 

An additional folder called the "mission lib" has also been added and contains the functionality to send and receive calls and changes with respect to the blocks and waypoints defined in the flight plan.


LodewijkSikkel:
I'm currently implementing the local waypoint implementation, those changes will be added to another pull request.

podhrmic:
btw wouldn't make more sense to have this as another submodule?

PR #1224, opened 2015-07-07
[fix] battery voltage reading while using spi sd card logger

EwoudSmeur:
I was never able to get a battery voltage reading while using the sd card logger module.
Freek and Christophe fixed it for me.

### DIFF
--- a/conf/modules/logger_spi_link.xml
+++ b/conf/modules/logger_spi_link.xml
@@ -13,8 +13,8 @@
     <define name="SPI_MASTER" value="1" />
     <define name="USE_SPI1" value="1" />
     <define name="HIGH_SPEED_LOGGER_SPI_LINK_DEVICE" value="spi1" />
-    <define name="USE_SPI_SLAVE5" value="1" />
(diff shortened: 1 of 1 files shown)


flixr:
For what board is this?
The SPI peripheral to use should really not be set in the module xml file itself, but rather configurable from the airframe file...

EwoudSmeur:
but it is also nice to have a default no? now i can load the module without additional settings. I don't know how to check if USE_SPI_SLAVE# has been defined, where # is some number.

At least this works for Lisa/m now, I don't even know if anyone else uses this (on other boards)

flixr:
Sure it is nice to have a default, but you should be able to change that from your airframe file, see how it's done for e.g. [direct_memory_logger.xml] (https://github.com/paparazzi/paparazzi/blob/master/conf/modules/direct_memory_logger.xml)

EwoudSmeur:
I actually tried for quite a while, but I am not figuring this out without more help...  No experience with this kind of thing. I tried copying things from the direct_memory_logger.xml, but it is merely guessing and nothing happens as I expect it.

PR #1230, opened 2015-07-08
added bebop indi airframe file

EwoudSmeur:
added an airframe file for the bebop with indi

### DIFF
--- a/conf/airframes/bebop_indi.xml
+++ b/conf/airframes/bebop_indi.xml
@@ -1,0 +1,217 @@
+<!DOCTYPE airframe SYSTEM "airframe.dtd">
+
+<airframe name="bebop">
+
+  <firmware name="rotorcraft">
+    <target name="ap" board="bebop"/>
+
+    <target name="nps" board="pc">
+      <subsystem name="fdm" type="jsbsim"/>
+    </target>
+
+    <!--define name="USE_SONAR" value="TRUE"/-->
+
+	<!-- Subsystem section -->
+    <subsystem name="telemetry" type="transparent_udp"/>
+    <subsystem name="radio_control" type="datalink"/>
+    <subsystem name="motor_mixing"/>
+    <subsystem name="actuators" type="bebop"/>
+    <subsystem name="imu" type="bebop"/>
+    <subsystem name="gps" type="furuno"/>
+    <subsystem name="stabilization" type="indi"/>
+    <subsystem name="ahrs"          type="float_mlkf"/>
+    <subsystem name="ahrs" type="int_cmpl_quat">
+      <configure name="USE_MAGNETOMETER" value="FALSE"/>
+      <define name="AHRS_USE_GPS_HEADING" value="FALSE"/>
+<!--     </subsystem> -->
+      <configure name="SECONDARY_AHRS" value="int_cmpl_quat"/>
+    </subsystem>
+    <subsystem name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
+    <load name="geo_mag.xml"/>
+    <load name="air_data.xml"/>
+    <load name="send_imu_mag_current.xml"/>
+    <load name="logger_file.xml">
+      <define name="FILE_LOGGER_PATH" value="/data/ftp/internal_000"/>
+    </load>
+  </modules>
+
+  <commands>
+    <axis name="PITCH" failsafe_value="0"/>
+    <axis name="ROLL" failsafe_value="0"/>
+    <axis name="YAW" failsafe_value="0"/>
+    <axis name="THRUST" failsafe_value="6000"/>
+  </commands>
+
+  <servos driver="Default">
+    <servo name="TOP_RIGHT" no="0" min="3000" neutral="3000" max="12000"/>
+    <servo name="TOP_LEFT" no="1" min="3000" neutral="3000" max="12000"/>
+    <servo name="BOTTOM_LEFT" no="2" min="3000" neutral="3000" max="12000"/>
+    <servo name="BOTTOM_RIGHT" no="3" min="3000" neutral="3000" max="12000"/>
+  </servos>
+
+  <section name="MIXING" prefix="MOTOR_MIXING_">
+    <define name="TRIM_ROLL" value="0"/>
+    <define name="TRIM_PITCH" value="0"/>
+    <define name="TRIM_YAW" value="0"/>
+    <define name="TYPE" value="QUAD_X"/>
+  </section>
+
+  <command_laws>
+    <call fun="motor_mixing_run(autopilot_motors_on,FALSE,values)"/>
+    <set servo="TOP_LEFT" value="motor_mixing.commands[MOTOR_FRONT_LEFT]"/>
+    <set servo="TOP_RIGHT" value="motor_mixing.commands[MOTOR_FRONT_RIGHT]"/>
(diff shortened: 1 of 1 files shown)


conf/airframes/bebop_indi.xml
```diff
@@ -0,0 +1,268 @@
...
+      <configure name="USE_MAGNETOMETER" value="FALSE"/>
+      <define name="AHRS_USE_GPS_HEADING" value="FALSE"/>
+    </subsystem>-->
+<!--       <configure name="SECONDARY_AHRS" value="int_cmpl_quat"/> -->
+    </subsystem>
+    <subsystem name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
+<!--     <load name="geo_mag.xml"/> -->
```
flixr: why comment out these useful modules?

conf/airframes/bebop_indi.xml
```diff
@@ -0,0 +1,268 @@
...
+
+  <section name="MIXING" prefix="MOTOR_MIXING_">
+    <define name="TRIM_ROLL" value="0"/>
+    <define name="TRIM_PITCH" value="0"/>
+    <define name="TRIM_YAW" value="0"/>
+    <define name="NB_MOTOR" value="4"/>
+    <define name="SCALE" value="255"/>
+    <define name="MAX_SATURATION_OFFSET" value="3*MAX_PPRZ"/>
+
+    <!-- Time cross layout (X), with order NW (CW), NE (CCW), SE (CW), SW (CCW) -->
```
flixr: We should use the "correct" motor mixing for these x-type quads. See https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/actuators/motor_mixing_types.h#L54
Even better, directly set this common type here...

flixr:
What exactly is the reason that you have set a MAX_SATURATION_OFFSET that basically disables this limit?
Is that really still needed after #1170?

See also #1231 for using the standard mixing types....

flixr:
Also is there any specific reason to use the int_cmpl_quat AHRS instead of float_mlkf?
The Kalman filter version should be superior to the complementary filter and it's not like the bebop is short on computing power for that...

EwoudSmeur:
I had problems in the latest master, for which I made an issue and you fixed... at first I figured I should use a much of the subsystems and settings I was using before but that didn't help. Never took the effort to put them back.

EwoudSmeur:
I actually copied this file from bebop.xml, which had the MAX_SATURATION_OFFSET

flixr:
Could you plz test it with the mixing as per #1231 and without setting the MAX_SATURATION_OFFSET?

EwoudSmeur:
It still flies, but the different scaling is quite annoying (I had to scale the gains), what is the thought behind this? First I thought it was to easily change between + and x quads without changing the gains, but then I think it is incorrect... The scaling factor between + and X is: 256/181 = sqrt(2) But in X configuration you have two actuators:   2*1/sqrt2) != 1...  So it actually should be half = 128. Or am I missing something?

flixr:
Of course you have to adjust the gains, but the updated mixing makes way more sense:
You basically can think of it as scaling/saturating on a ring now, where as before it was a rectangle.
I.e. think of the case with an X quad where you give a full command exactly along one axis like 45deg to the front left. There you basically want the full range on the motors on this axis, just as if you have a + quad and give full pitch.
And yes, you will have the same gains for x and +...
The lever arm is of course smaller wrt the x or y axis on a X quad by a factor of 1/sqrt(2).
So you have 2 * 1/sqrt(2) * 1/sqrt(2) = 1

Also added script that will calc the mixing for configurations: https://github.com/paparazzi/paparazzi/blob/master/sw/tools/motor_mixing.py

EwoudSmeur:
Ah I didn't take into account that the axis transformation also gives a 1/sqrt(2). 

Cool script, that should save some time in setting up new drones!

Can we merge this now though?

conf/airframes/bebop_indi.xml
```diff
@@ -0,0 +1,261 @@
...
+    <servo name="TOP_LEFT" no="1" min="3000" neutral="3000" max="12000"/>
+    <servo name="BOTTOM_LEFT" no="2" min="3000" neutral="3000" max="12000"/>
+    <servo name="BOTTOM_RIGHT" no="3" min="3000" neutral="3000" max="12000"/>
+  </servos>
+
+  <section name="MIXING" prefix="MOTOR_MIXING_">
+    <define name="TRIM_ROLL" value="0"/>
+    <define name="TRIM_PITCH" value="0"/>
+    <define name="TRIM_YAW" value="0"/>
+    <define name="MOTOR_MIXING_TYPE" value="QUAD_X"/>
```
flixr: plz remove

conf/airframes/bebop_indi.xml
```diff
@@ -0,0 +1,261 @@
...
+
+    <!-- feedforward -->
+    <define name="PHI_DDGAIN" value="0"/>
+    <define name="THETA_DDGAIN" value="0"/>
+    <define name="PSI_DDGAIN" value="0"/>
+  </section>
+
+  <section name="STABILIZATION_ATTITUDE_INDI" prefix="STABILIZATION_INDI_">
+    <!-- setpoints -->
+    <define name="SP_MAX_PHI" value="60." unit="deg"/>
```
flixr: Seems these setpoint and deadband defines are never used anywhere or am I missing something?

flixr:
Also could you plz remove the defines from the attitude stabilization section that are not used?

EwoudSmeur:
Indeed the setpoint defines are not used. Which is kind of a pity... Can I make a separate section for the setpoint defines, so that they can be set irrespective of the attitude controller? 
Because with the prefix I gave to the STABILIZATION_ATTITUDE_INDI  section I can not set the defines for the max setpoint  and deadband values, and they are located in the setpoint files and not in the stabilization control files.

flixr:
You can split the sections however you like as long as the defines have the correct names later (PREFIX +  define name).
A section is just there to give a better overview of which defines belong to what (and to save typing if you use a prefix).
But since you use the reference model and setpoints from the common attitude control, why not simply keep the defines that are used in `STABILIZATION_ATTITUDE_` section and the ones that are specific to indi in that section.

EwoudSmeur:
Ah yes of course, I like it better with a separate section, because that keeps all the PID stuff and INDI stuff each in one section, without defines that are relevant to both.

EwoudSmeur:
Correct, why would I take those out? now you can change between pid or indi...

flixr:
IMHO we shouldn't have any defines in an airframe file that are then not used.
This only leads to confusion on what is needed/relevant for that configuration (e.g. people trying to change the PGAIN, but using INDI, etc..)
If you want to use the pid loops, choose the "normal" bebop airframe file.
If you really want to keep them, then at the very least comment them out with an added explanation.

EwoudSmeur:
And then you need to copy all your other settings you might change to the other airframe file... Whatever, have it your way

flixr:
Thanks for the updates.

Just for the record: I'm not the only one with the opinion that the airframe file should only contain defines relevant for that configuration.
Mentioned this in the dev meeting and e.g. @gautierhattenberger and @esden also think this way.

flixr:
thx, merged with 37f2e532ef43e24e7585b6dcd483466301b4be9a

PR #1231, opened 2015-07-08
[airframes] use standard motor mixing types for some quads

flixr:
This will change the effect of the gains (depending on the type), but should be finally correct.
Especially the yaw commands have about half the effect now...

Needs testing on ardrone and bebop, and probably some adjustment of the gains.

### DIFF
--- a/conf/airframes/ardrone2.xml
+++ b/conf/airframes/ardrone2.xml
@@ -159,22 +159,22 @@
     <define name="REF_MAX_RDOT" value="RadOfDeg(4000.)"/>
 
     <!-- feedback -->
-    <define name="PHI_PGAIN" value="592"/>
-    <define name="PHI_DGAIN" value="303"/>
+    <define name="PHI_PGAIN" value="850"/>
+    <define name="PHI_DGAIN" value="425"/>
     <define name="PHI_IGAIN" value="0"/>
 
-    <define name="THETA_PGAIN" value="606"/>
-    <define name="THETA_DGAIN" value="303"/>
+    <define name="THETA_PGAIN" value="850"/>
+    <define name="THETA_DGAIN" value="425"/>
     <define name="THETA_IGAIN" value="0"/>
 
-    <define name="PSI_PGAIN" value="529"/>
-    <define name="PSI_DGAIN" value="353"/>
+    <define name="PSI_PGAIN" value="1000"/>
+    <define name="PSI_DGAIN" value="700"/>
     <define name="PSI_IGAIN" value="0"/>
 
     <!-- feedforward -->
     <define name="PHI_DDGAIN" value="0"/>
     <define name="THETA_DDGAIN" value="0"/>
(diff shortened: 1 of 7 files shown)


flixr:
@gautierhattenberger @kevindehecker @fvantienen @EwoudSmeur 
Would be great if you guys could test the updated ardrone and bebop airframe files.
Btw, I was wondering anyhow that none of those have an integral term set...

Also the ardrone2_optitrack.xml (which I didn't change here) has different gains, which I don't quite understand as it is the same controller and airframe.

gautierhattenberger:
Tested and it seems to be working as expected

flixr:
You tested both ardrone2 and bebop?
The behaviour (with these updated gains and mixing) seems to be the same as before?

kevindehecker:
I tested ARDrone2 in ATT and NAV mode with the indi airframe file, which seems to work.

flixr:
The question is if the updated gains match the new motor mixing scaling.
Since INDI does not use these gains, it needs to be tested with the "classic" controllers....

Also I'm still curious on why neither ardrone2 nor bebop have any igains set...
Neither why the optitrack airframe uses different gains...

gautierhattenberger:
I tested both, it looks the same. I just don't like some of the parameters for the ardrone2 (max yaw rate, gains a bit higher than what I'm used to). Otherwise it is flyable like this. Only did ATT so far.

EwoudSmeur:
It does also matter for the INDI controller, as the actuator effectiveness is sqrt(2) lower... Though Kevin was not flying with INDI but with quat_int

flixr:
Yes, but the indi airframe was not changed in this PR... And the question
was specifically if the updated gains match the mixing...
Am 22.07.2015 11:06 schrieb "Ewoud Smeur" <notifications@github.com>:

flixr:
Ups, sorry... I did change the ardrone indi airframe... Wanted to omit that one first...

kevindehecker:
But now I actually have flown with adaptive indi, and it works great! 

And as @EwoudSmeur mentioned, apparently I flied with standard stuff op to now, and that also seems to work. (in both cases for ARDrone2 in optitrack enviroment that is)

flixr:
And what is up with the optitrack airframe and it's totally different gains?

kevindehecker:
That I don't know. I use the standard indi gains with the optitrack systen, except for the h&v P and D guidance gains, which I have set much higher since there is perfect gps. Using those I get almost no position error, and pretty stable flight otherwise.

flixr:
Could you please discuss this with @fvantienen, you guys are the only ones using optitrack, if the ardrone2_optitrack.xml airframe file is outdated, then plz update it to what you are using...
If it is not used at all, plz remove it...

kevindehecker:
@fvantienen I think my setup for optitrack runs pretty smooth by now. So, how about taking this one #1257 ?

PR #1232, opened 2015-07-09
[ground segment] Added geometry param to messages and settings agents.

esden:
You can now provide for example ```-g 300x400-0-200``` to settings and messages agents to set the position and size of the window on launch. Useful if you want to recreate a session exactly. Useful for sessions with a lot of windows like the Raw Sensors and Scaled Sensor sessions.


esden:
Here is the layout of the Raw Sensors session in the above commit. :)
screenshot from 2015-07-08 17 11 09

Issue #977, opened 2014-11-21
ARdrone2 drops out of sky in att zhold


kevindehecker:
Using Master v5.2_stable I experience sudden drops out of the sky with the ARdrone2 during hovering in zhold. After the drop, it tends to recover to the correct height. An example can be viewed in this [video](https://drive.google.com/file/d/0B_SZaF2Rt7FreVBIRDJZTnlibXM/view?usp=sharing) at 1:50 - 1:58. A log from a similar event is available from [here](https://drive.google.com/file/d/0B_SZaF2Rt7FrdUY5VHZUYVM0TnM/view?usp=sharing)

I used the standard ardrone2_raw.xml airframe, only disabled the magneto. (so both sonar and baro should be enabled).

OpenUAS:
With PPRZ > 5.4 needs a re-test, lost of improvements of sensor readings. Helping with at least > 200+ flights in CyberZoo I did not encounter this behaviour anymore ( but lots of other issue)  >;-)

gautierhattenberger:
Is this still an issue ?

gautierhattenberger:
closing for feedback timeout
re-open if needed

PR #1237, opened 2015-07-14
Optitrack fixes

kevindehecker:
These are the changes I needed to make, to fly the ARDrone2 in the Cyberzoo, using the optitrack system for altitude and yaw estimation. Without it, several bugs such as #977 appear on quasi-regular basis.

Basically, this disables the baro+magneto (sonar was already disabled) and fixes some code that kind of ignored those settings or caused problems for it.

### DIFF
--- a/conf/airframes/ardrone2_optitrack.xml
+++ b/conf/airframes/ardrone2_optitrack.xml
@@ -5,6 +5,8 @@
   <firmware name="rotorcraft">
     <target name="ap" board="ardrone2">
       <define name="FAILSAFE_DESCENT_SPEED" value="0.5"/>
+      <define name="USE_SONAR" value="0"/>      
+      <configure name="USE_BARO_BOARD" value="FALSE"/>
       <configure name="USE_MAGNETOMETER" value="0"/>
     </target>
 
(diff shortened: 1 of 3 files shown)


flixr:
Thx. The last two commits look good to me.
The `#if USE_BARO_BOARD` check in ins_int.c is not what we want to use here however...
E.g. you might have an external baro connected.
However I think that we can simply always propagate the vff and skip the check for `ins_int.baro_initialized` completely.
@gautierhattenberger do you see any reason to only propagate the filter if the baro is not yet initialized?

gautierhattenberger:
I think the issue is that if the baro takes some time to be initialized (with was the case for the analog baro of Booz boards, so maybe not relevant anymore...), the ins will diverge during this time.
It is probably not a really big issue, especially on recent boards with digital barometers.

flixr:
It will not diverge if you have gps or sonar measurements during that time.
I think a better solution than only checking if the baro is initialized is to simply have some sort of counter that resets with every update and if you don't get any updates it times out and stops the propagation.

flixr:
I merged everything except the check "propagation fix" via `#if USE_BARO_BOARD`:
ecd3b1a75f0c19860ccbc5d011626f107d7a2a30
20f8670ba9ae4efe10a4e7e1565a50079e033f90
c2a4b2b3ed2c9178678ffd0ce721be72211ed40e

flixr:
finally closed with #1241

PR #1238, opened 2015-07-14
Added nav copy function again for usage in flightplan

kevindehecker:
I couldn't use my flightplan anymore, because this function was missing. (I know @EwoudSmeur also sometimes uses it).


flixr:
This does not properly handle a waypoint with global coordinates or even update them...

PR #1239, opened 2015-07-14
Fixed 100% cpu ap issue on linux based systems

### LINKED ISSUE #1233: Autopilot uses 100% cpu on ARdrone2
Using newest master, ap.elf uses 100% cpu on the ARdrone2. Also tested with 5.4.2 stable tag, which does not have the problem (around 2% according to top).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24087799-autopilot-uses-100-cpu-on-ardrone2?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kevindehecker:
Maybe not the nicest solution, but fixes #1233

### DIFF
--- a/conf/Makefile.linux
+++ b/conf/Makefile.linux
@@ -56,6 +56,7 @@ CFLAGS += -Wredundant-decls -Wreturn-type -Wshadow -Wunused
 #CFLAGS += $(patsubst %,-I%,$(EXTRAINCDIRS))
 
 CFLAGS += -lm -pthread
+CFLAGS += -DUSING_OS
 
 # flags only for C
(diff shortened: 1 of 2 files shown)


flixr:
Thx, I think this is good enough as a workaround for now...
Could you please also add a comment in the code explaining the reason for the usleep?

PR #1240, opened 2015-07-15
[rotorcraft][linux] limit main/event loop to 1kHz

### LINKED ISSUE #1233: Autopilot uses 100% cpu on ARdrone2

flixr:
This is a kludge until we can better leverage threads and have real events.
Without this limit the event flags will constantly polled as fast as possible,
resulting on 100% cpu load on boards with an (RT)OS.
On bare metal boards this is not an issue, as you have nothing else running anyway.

Enable this for the bebop and ardrone2 by default.

Alternative to #1239 and should fix #1233

### DIFF
--- a/conf/boards/ardrone2.makefile
+++ b/conf/boards/ardrone2.makefile
@@ -42,6 +42,9 @@ $(TARGET).CFLAGS += -DUSE_LINUX_SIGNAL -D_GNU_SOURCE
 $(TARGET).CFLAGS += -DLINUX_LINK_STATIC
 $(TARGET).LDFLAGS += -static
 
+# limit main loop to 1kHz so ap doesn't need 100% cpu
+$(TARGET).CFLAGS += -DLIMIT_EVENT_POLLING
+
 # -----------------------------------------------------------------------
 
 # default LED configuration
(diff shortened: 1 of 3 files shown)


flixr:
@kevindehecker @gautierhattenberger @fvantienen 
what do you think about this solution instead?

podhrmic:
IMHO we will need a separate main for RTOS and bare metal (i.e. main.c and main_rtos.c), because once you start splitting the tasks the code will be too different. As an example main_chibios.c that we are flying here: https://github.com/podhrmic/paparazzi/blob/rt_paparazzi/sw/airborne/firmwares/rotorcraft/main_chibios.c#L645 (flight tested)

kevindehecker:
Just tested this, works great: ~1% cpu usage on the ARDRone2. Same as with #1239, but this implementation is definitely better because maximum delay is decreased.

flixr:
Disabled this again with fb91702596b310d77342a0c55ab3d98bdb8a8f9d as it delays the mcu_event too much and makes problems with the uart driver....

PR #1241, opened 2015-07-15
[ins] prevent propagation if there are no measurement updates

flixr:
To prevent the position and velocity estimation from running away while you only have accel measurements (done in the propagation),
Propagate only if we got any measurement during the last INS_MAX_PROPAGATION_STEPS.
Otherwise halt the propagation to not diverge and only set the acceleration.
This should only be relevant in the startup phase when the baro is not yet initialized
and there is no gps fix yet...

Should fix the last remaining part of #1237 in a more generic and robust way.


flixr:
@kevindehecker would be great if you could test this with your optitrack setup.

kevindehecker:
Yep, seems to work and fly. Optitrack gps gets propagated properly.

Issue #1011, opened 2014-12-06
If ARDrone2 does not take off, GCS provides no information on the why


kevindehecker:
During my tests it regularly happens that the drone will not start engines. Sometimes the flight plan log will show that it went to mayday - kill, sometimes it just indicates start engine was successful but it wasn't.

In many cases, I actually did something wrong, in other cases I still don't know what the reason may have been (I reboot until it works again). 

In both cases it would be very nice to know why a particular action was not performed by e.g. mentioning it in the flight plan log in GCS. At least in case of a user mistake this should be doable, I reckon that the following causes (that happened to me) can be easily recognized:
-no link 
-gps initialization not ready
-AHRS initialization not ready  
-joystick switch still in ATT instead of NAV mode
-no rc

I suggest the GCS flight plan log, because this would be a central place in which all these problems can be shown. Otherwise, the user has to know where to look, which for novice users is unworkable. (e.g., no link is already shown in GCS by this tiny red bar.... which took me some months to discover)
##

OpenUAS:
Software improvment could help. Also an instructional document and video like "GCS for starters" explaining pitfalls would be welcomed.

Note that in a flightplan block the current line executed is highlighted and shows where it can not continue.

gautierhattenberger:
closing, ardrone2 had motors issue, but this frame is outdated now

Issue #1010, opened 2014-12-06
Disarming ARDrone2 works awkwardly in stressed situations


kevindehecker:
The Ardrone2 props are (dis)armed by max yaw-ing. However, it regularly happens that I undesirably re-arm the drone after it crashed to the ground sometimes due to know fault of my own.

When I get the Ardrone2 into a spot which I anticipate will be problematic to get out of, I try to immediately land and or disarm the drone. But in many cases I'm too late or the landing is a bit rough and the ARDrone2 righteously disarms as a safety feature. But because at the same time I'm also disarming manually, as a result it will immediately arm again. (in many cases related to #1009) It takes several seconds for me to disarm manually again, damaging the drone and the experimental setup I carefully placed on the drone.  

I think there should be a delay of at least 5s after a automated disarm happened, before the user can arm the props again.

kevindehecker:
Closed with 6aa736409887bd4e8be3a6ba1865542a3403aa3d

PR #1242, opened 2015-07-15
Fixed unsanctioned yaw unkill problem with uninitialised joysticks

kevindehecker:
Atleast the HobbyKing RC initializes often in such a way that all axis are maxed out. This causes the drone to repeatedly unkill (because of full yaw), and to dance around violently (because full pitch and roll). 

Today I got bitten twice by this problem. Luckily it cost me only 1 finger. This pull request at least prevents the unkill part. (which happens very unexpectedly, just after uploading from center).

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
@@ -41,6 +41,9 @@
 
 /// Motors ON check state machine states
 enum arming_state {
+  STATUS_INITIALISE_RC,
+  STATUS_MOTORS_AUTOMATICALLY_OFF,
+  STATUS_MOTORS_AUTOMATICALLY_OFF_SAFETY_WAIT,
   STATUS_MOTORS_OFF,
   STATUS_M_OFF_STICK_PUSHED,
   STATUS_START_MOTORS,
@@ -56,7 +59,7 @@ enum arming_state autopilot_check_motor_status;
 static inline void autopilot_arming_init(void)
 {
   autopilot_motors_on_counter = 0;
-  autopilot_check_motor_status = STATUS_MOTORS_OFF;
+  autopilot_check_motor_status = STATUS_INITIALISE_RC;
 }
 
 
@@ -67,7 +70,7 @@ static inline void autopilot_arming_set(bool_t motors_on)
   if (motors_on) {
     autopilot_check_motor_status = STATUS_MOTORS_ON;
   } else {
-    autopilot_check_motor_status = STATUS_MOTORS_OFF;
+    autopilot_check_motor_status = STATUS_MOTORS_AUTOMATICALLY_OFF;
   }
 }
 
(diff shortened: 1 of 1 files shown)


podhrmic:
That sounds pretty serious. Maybe we should add some warning against using HobbyKing RCs and recommend reliable ones such as Taranis or Turnigy? Do you know which model it was?

kevindehecker:
It just says Hobbyking and a bunch of features (6 channels, USB, etc), no number. We at TUDelft have a bunch of those (I guess they are cheap).

podhrmic:
can you post the link to the product so we can add a proper warning?

kevindehecker:
http://www.hobbyking.com/hobbyking/store/__20951__Hobbyking_6CH_RC_Flight_Simulator_System_Mode_2_.html

podhrmic:
wow, that is an incredible price...!

kevindehecker:
Haha, and there I was mistaking your interest for concern :P

podhrmic:
well the question is how much do you value your fingers...:-P

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -30,6 +30,7 @@
 #define AUTOPILOT_ARMING_YAW_H
 
 #include "autopilot_rc_helpers.h"
+#include "subsystems/radio_control/rc_datalink.h"
```
flixr: We can't have any implementation specific radio control stuff in here...

kevindehecker:
@flixr how about this then? (Tomorrow I will test on drone)

flixr:
I think it would be better to add another state that is entered on startup and only allow transition to the `MOTORS_OFF` state if throttle is low and the yaw is centered...
In the end this really needs to be handled in input2ivy...

kevindehecker:
OK good idea, in the state machine it is. While I was at it, also added a fix for #1010

(Untested, will do that hopefully tomorrow)

flixr:
You at least need to set the inital state to STATUS_INITIALISE_RC as well....

kevindehecker:
Tnx :)

kevindehecker:
OK, it works on a drone.

One remark: if the sticks are not properly centered (e.g. because the joystick needs to be initialized, or calibrated), the motors won't start. Which I think is a good thing. But no feedback to the user is provided that the joystick is the culprit. (#1011)

flixr:
I'm also not sure if we should require the roll and pitch sticks to be centred....

kevindehecker:
I am strong proponent of that, simply because currently <i>all</i> the sticks need to be touched before they are initialized. Separately.

If you unkill with just an initialized yaw in ATT mode, but non-initialized pitch/roll, the drone will flip instantly.

flixr:
Well, you are using a joystick, most people have a proper RC.
And then you seem to be the first one to actually use such a crappy joystick, I never had any problems with the 3 different ones I have ;-)

I just remember that we had this discussion quite some time ago and some people explicitly didn't want the requirement of having all sticks centred...
Only want to prevent people from wondering why the suddenly can't arm anymore... but maybe it is better anyway, because it forces people to at least roughly calibrate their input devices....

@gautierhattenberger do you remember that discussion? what do you think?

kevindehecker:
I think  this is the kind of issue (chance of actual accidents) in which "yes the problem exist, but not always" is not a reason to ignore it.

(I do agree about the wondering part, but that is something that can should fixed in addition.)

gautierhattenberger:
What I remember is that sometimes, the initial values returned by sdl are stupid until you have moved your joystick to its full positions. Which should be done on the ground before powering the plane anyway...
I not against more safety and it is probably not a big deal to at least center / throttle down your RC once before flight.
Maybe it is a problem on some PWM ESC that needs to go full throttle and then down for arming ? Not sure if this problem is relevant here.
Why the delay before re-arming is actually needed ?

kevindehecker:
Re-arming delay: #1010 -> after a ardrone2 crash, drone is automatically killed, but at the same time user is trying to kill it ending up in a re-arming.

flixr:
It is fine to merge for me....

gautierhattenberger:
same here

flixr:
merged with 6aa736409887bd4e8be3a6ba1865542a3403aa3d

PR #1244, opened 2015-07-15
Add a flag to default motor_mixing types to reverse direction

gautierhattenberger:
Define `MOTOR_MIXING_REVERSE` to TRUE to reverse the yaw (motor rotation directions)
- flag works on all types
- removing X_CCW type

### DIFF
--- a/sw/airborne/subsystems/actuators/motor_mixing_types.h
+++ b/sw/airborne/subsystems/actuators/motor_mixing_types.h
@@ -29,11 +29,17 @@
 /* already defined common configurations*/
 #define QUAD_PLUS   1
 #define QUAD_X      2
-#define QUAD_X_CCW  3
-#define HEXA_X      4
-#define HEXA_PLUS   5
-#define OCTO_X      6
-#define OCTO_PLUS   7
+#define HEXA_X      3
+#define HEXA_PLUS   4
+#define OCTO_X      5
+#define OCTO_PLUS   6
+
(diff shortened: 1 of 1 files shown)


flixr:
Looks good, just wondering if we can find a better name than `MOTOR_MIXING_REVERSED`...

flixr:
maybe `MOTOR_MIXING_REVERSE_YAW` is a bit more clear?

gautierhattenberger:
I'm not really happy with my name either. I feared that having `YAW` in the name makes it more difficult to understand for new users (they might not see that changing the rotation direction is only changing the yaw coeffs).
`MOTOR_MIXING_REVERSE_DIRECTION` ?
As long as it is properly documented...

flixr:
maybe the long version `MOTOR_MIXING_REVERSE_MOTORS_DIRECTION` is more clear, but it is really long...

gautierhattenberger:
since motor mixing section usually have a prefix, it makes reasonably long

esden:
Why not ```MOTOR_MIXING_REVERSE``` ?

flixr:
If it's clear enough I'm fine with `MOTOR_MIXING_REVERSE` (better than `MOTOR_MIXING_REVERSED`).
Was just wondering if it is clear that the motor directions are reversed... but I guess so...

gautierhattenberger:
I think it is fine like this

PR #1247, opened 2015-07-16
[pprz center] improve handling of programs from control_panel.xml

flixr:
Add possibility to launch global commands from tools menu.
Prefixing a program command in control_panel.xml with `$` will strip the `$` and then execute the command.

This makes it possible to start programs without knowing their absolute path.
Already used to start a simple http server from python.

Also parse the args from the programs section of control_panel.xml when starting them from the tools menu.
This makes it possible to specify default args.

### DIFF
--- a/conf/control_panel_example.xml
+++ b/conf/control_panel_example.xml
@@ -2,100 +2,52 @@
 
 <control_panel name="paparazzi control panel">
 
-  <section name="variables">
-    <variable name="downlink_serial_port" value="/dev/ttyUSB0"/>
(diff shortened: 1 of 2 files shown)


flixr:
Hm... only one problem: if you save the session like this, it saves the args separately and adds them twice upon starting the session again.
But specifying the args in program doesn't seem to work..

gautierhattenberger:
I'm not sure their is an other choice than making a local script that set the default options

flixr:
Updated, now you can specify args in the programs section and they are actually used when starting from the tools menu...

gautierhattenberger:
tested and working

PR #1248, opened 2015-07-17
[server] set http port: replace -kml_port option with -port

flixr:
Use the port and hostname for serving config xml files via http.


gautierhattenberger:
Good to merge for me

Issue #1236, opened 2015-07-14
Prepare flight tested airframes + select the unsoported ones


podhrmic:

flixr:
see also #888 and #974

podhrmic:
So I did a quick check of which configurations/airfames are actually compiling. A summary is here: https://docs.google.com/spreadsheets/d/1JK6Di9Z9cGQN0RycM6JxvBHufjcdzLZuKnN8s68IXP4/edit?usp=sharing

I am not sure what should be the next step - IMHO I should delete the non-working configurations and move the related airframes into some "obsolete" folder.

As for the control panels - I did some tests too (on the second page of the spreadsheet) and I think we should keep only the ``control_panel_example.xml``, the rest should go to obsolete folder as well (``airframe/TUDeldf/control_panel.xml`` and ``airframes/OpenUAS/opensuas_controL_panel_obc2014.xml``)

Let me know if you agree.

flixr:
At least TwoSeas should compile (hopefully) work again after fixing ins xsens700 in 3d41040278f724cffe9122bf37eebbc8bc6f3c71

For the personal/institution confs/airframes, I would leave it to @gautierhattenberger and @dewagter to fix/move/delete them as they see fit.

IMHO the more important part is checking/moving/fixing all the airframe files in examples that are not in any conf.
In the end all that work should be in a conf that can be tested...

podhrmic:
OK, so I did check for which aiframes are actually used in any conf file. From total 176 aiframe config files, there is 96 used and 85 unused. Please see this spreadsheet for details: https://docs.google.com/spreadsheets/d/1JK6Di9Z9cGQN0RycM6JxvBHufjcdzLZuKnN8s68IXP4/edit?usp=sharing

Please look through the list, I will make a pull request with the unused airframes being moved to the obsolete folder.

podhrmic:
@flixr @gautierhattenberger any feedback on this?

flixr:
Thanks a lot for going through all those!

As I said before, I would leave it up to the respective people to cleanup their own subfolders.
For the files in examples, I would say we move all that are not part of a conf to an `unmaintained` or `untested` folder.

OpenUAS:
Repairing, Preparing etc. some airframes

PR #1249, opened 2015-07-20
Aiframe cleanup

podhrmic:
unused airframe xml files fronm conf_example and conf_test.xml were moved to obselete folder, see #1236 for more info.


flixr:
Should we really call the folder `obsolete`? Maybe we can find a better name...
Could you plz also update the conf_tests.xml?

OpenUAS:
I vote for untested, gives less of expectancy for the outcome of using these airframes.

podhrmic:
Changed the name to "untested". Also the conf_tests.xml is updated.

flixr:
One more thing that we should probably consider:
while these airframes are untested (in the sense that they are not part of the example or test conf), most of them should still work just fine.
But since they are moved one directory level lower now all of their relative paths to the airframe.dtd will be wrong.
Maybe it would make more sense to make the directory structure rather flat and similar between `examples` and `untested`, i.e.:
```
conf/airframes
   |
   |---- examples
   |    |
   |    |---- foo.xml
   |    |---- bar.xml
   |
   |---- untested
   |    |
   |    |---- baz.xml 
```

Then we could also more easily move a file from examples to untested and vice versa...

podhrmic:
I already did that.

flixr:
@podhrmic Uhm... no...
e.g. `conf/airframes/examples/beagle_bone_black.xml` was renamed to `conf/airframes/untested/examples/beagle_bone_black.xml` without changes...
So adding another directory level...

podhrmic:
yes, so then you can move it from ``conf/airframes/untested/examples/beagle_bone_black.xml`` to ``conf/airframes/examples/beagle_bone_black.xml`` without any change. That is what we want, no?

flixr:
But I want to be able to just compile/test an airframe file from the untested dir without having to do any changes or moving it to examples first.
So that is why I propose to have a flat tree, with no more subdirs under examples or untested...

podhrmic:
OK. Do you want all files in one folder (i.e. no subdirs in ``untested``) or have 

untested
untested-> examples
untested-> TestHardware

?

flixr:
As I said I would vote for a flat tree, so no subfolders...
And you can delete the empty airframes from TestHardware...

podhrmic:
This should do it...

PR #1250, opened 2015-07-21
Fixed ardrone2 battery checks

kevindehecker:
For some reason, the ARDrone2 did not check battery levels. Instead, on catastropic level, it just dropped dead out of the sky. Closer inspection showed the proper values were not updated, because electrical.c was replaced by the ARDrone2 specific implementation.


flixr:
How about a cleaner approach: convert the `boards/ardrone/electrical.c` functionality to a module that only updates `electrical.vsupply`?
That way we re-use electrical.c as it is and the setup is exactly the same as using any other external voltage/current sensor when you don't use the internal ADC of the board...

flixr:
closed with #1252

PR #1251, opened 2015-07-21
[boards] bebop: use standard electrical and init in actuators

flixr:
We should have never added an electrical "dummy" for the bebop in the first place.
Now we should have the electrical.bat_low and electrical.bat_critical flags again...


flixr:
We could also think about adding a `board_init()` or something like that where you would do stuff like done for the bebop: killing the dragon prog, init cmos sensors...

In any case, there is no reason here to not use the normal electrical.c subsystem....

flixr:
Closed with #1253

PR #1252, opened 2015-07-21
Moved electrical ardrone2 to module, which fixes bat checks

kevindehecker:
Moved electrical ardrone2 to module, hopefully cleaner implementation than #1250

### DIFF
--- a/conf/firmwares/rotorcraft.makefile
+++ b/conf/firmwares/rotorcraft.makefile
@@ -123,9 +123,7 @@ $(TARGET).srcs += $(SRC_ARCH)/mcu_periph/i2c_arch.c
 #
 $(TARGET).CFLAGS += -DUSE_ADC
 $(TARGET).srcs   += $(SRC_ARCH)/mcu_periph/adc_arch.c
-ifneq ($(ARCH), linux)
 $(TARGET).srcs   += subsystems/electrical.c
-endif
 
 
 ######################################################################
@@ -137,8 +135,6 @@ endif
 ifeq ($(BOARD), booz)
 ns_CFLAGS += -DUSE_DAC
 ns_srcs   += $(SRC_ARCH)/mcu_periph/dac_arch.c
-else ifeq ($(BOARD), ardrone)
-ns_srcs   += $(SRC_BOARD)/electrical.c
 else ifeq ($(BOARD), bebop)
 ns_srcs   += $(SRC_BOARD)/electrical.c
 endif
--- a/conf/modules/electric_ardrone2.xml
+++ b/conf/modules/electric_ardrone2.xml
@@ -1,0 +1,15 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="electric_ardrone2" dir="sensors">
+  <doc>
+    <description>Voltage and current sensors for ARDrone2</description>
+  </doc>
+  <header>
+    <file name="electric_ardrone2.h"/>
+  </header>
+  <init fun="electric_ardrone2_init()"/>
+  <periodic fun="electric_ardrone2_periodic()" freq="10"/>
+  <makefile target="ap">
+    <file name="electric_ardrone2.c"/>
+  </makefile>
+</module>
--- a/sw/airborne/modules/sensors/electric_ardrone2.c
+++ b/sw/airborne/modules/sensors/electric_ardrone2.c
@@ -43,30 +43,6 @@
 
 void electrical_ardrone2_setup(void);
 
-#if defined ADC_CHANNEL_VSUPPLY || defined ADC_CHANNEL_CURRENT || defined MILLIAMP_AT_FULL_THROTTLE
-static struct {
-#ifdef ADC_CHANNEL_VSUPPLY
-  struct adc_buf vsupply_adc_buf;
-#endif
-#ifdef ADC_CHANNEL_CURRENT
-  struct adc_buf current_adc_buf;
-#endif
-#ifdef MILLIAMP_AT_FULL_THROTTLE
-  float nonlin_factor;
-#endif
-} electrical_priv;
-#endif
-
-#if defined COMMAND_THROTTLE
-#define COMMAND_CURRENT_ESTIMATION COMMAND_THROTTLE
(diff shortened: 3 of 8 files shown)


kevindehecker:
Question: currently ap.elf is killed in the init function. Probably something that needs to happen quite early on in the program for other stuff as well...? By moving it in this module, I don't know what happens with that. (at least it does not seem the best location to do this)

conf/firmwares/rotorcraft.makefile
```diff
@@ -123,9 +123,7 @@ $(TARGET).srcs += $(SRC_ARCH)/mcu_periph/i2c_arch.c
 #
 $(TARGET).CFLAGS += -DUSE_ADC
 $(TARGET).srcs   += $(SRC_ARCH)/mcu_periph/adc_arch.c
-ifneq ($(ARCH), linux)
```
kevindehecker: This change is connected to #1251

sw/airborne/modules/sensors/electric_ardrone2.c
```diff
@@ -39,7 +41,7 @@
 #include "subsystems/commands.h"
 #include "generated/airframe.h"
 
-struct Electrical electrical;
+void electrical_ardrone2_setup(void);
 
 #if defined ADC_CHANNEL_VSUPPLY || defined ADC_CHANNEL_CURRENT || defined MILLIAMP_AT_FULL_THROTTLE
```
flixr: you can remove this stuff as it is not used by the ardrone2...

conf/modules/electric_ardrone2.xml
```diff
@@ -0,0 +1,15 @@
...
+
+<module name="electric_ardrone2" dir="sensors">
+  <doc>
+    <description>Voltage and current sensors for ARDrone2</description>
+  </doc>
+  <header>
+    <file name="electric_ardrone2.h"/>
+  </header>
+  <init fun="electric_ardrone2_init()"/>
+  <periodic fun="electric_ardrone2_periodic()" freq="10"/>
```
kevindehecker: Should this maybe be set to the same freq as the electric subsystem?

sw/airborne/modules/sensors/electric_ardrone2.c
```diff
@@ -79,11 +81,11 @@ void electrical_init(void)
     fprintf(stderr, "Failed to set slave address: %m\n");
   }
 
-  electrical_setup();
+  electrical_ardrone2_setup();
   electrical_priv.nonlin_factor = CURRENT_ESTIMATION_NONLINEARITY;
```
flixr: this is already done in electrical_init

sw/airborne/modules/sensors/electric_ardrone2.c
```diff
@@ -144,4 +146,10 @@ void electrical_periodic(void)
    */
   electrical.current = b - pow((pow(b, electrical_priv.nonlin_factor) - pow((b * x), electrical_priv.nonlin_factor)),
```
flixr: This is already calculated in electrical_periodic, you should only need to update `electrical.vsupply` here...

flixr:
@kevindehecker did you check if it works as expected?

Should we introduce some sort of board_init function now where we can put system commands like killing running progs on the Parrot drones?

kevindehecker:
No, it is untested. Currently doing some science (which uses my other fix which works). Will test later tonight or tomorrow.

gautierhattenberger:
I think if we have a weak function for that it should be easy to implement for the boards that needs it.

flixr:
Should we keep the name `electric_ardrone2.xml` for the module? Maybe `vsupply_ardrone2.xml` or something like that would be more clear?

OpenUAS:
First I was a bit thrown of by the  module approach, however I see if well implemented it poses benefit. But to make it an optional feature, where we expect it to be those values always available, how to deal with that? Well more about it tomorrow after a talk with Kevin.

flixr:
Yeah, not sure if a module is really the nicest for the user, we don't have a concept of mandatory modules or so...
But it is definitely a clean implementation, never quite understood why there was a separate electrical made for ardrone and bebop...

flixr:
On a sidenote, plz set a valid git name and email....

flixr:
Resolved the conflicts and minor cleanup in https://github.com/flixr/paparazzi/commit/e63fb9144ff6c09e914bd2e7f1c6e770f3060171

Could someone with an ARDrone2 plz test this?
Also should we do this as a module (and keep the `electric_ardrone2.xml` name)?
@gautierhattenberger any better ideas?

gautierhattenberger:
The issue is that if you don't have it in your airframe file, you don't have the battery level at all.
option 1: keep ardrone electrical and duplicate a little bit of code
option 2: use a module and expect people will see their is something wrong when upgrading
option 3: do the reading (periodic call) in a `board_periodic` function (since we already have a `board_init` now)

Option 3 is not the best long term solution, but it looks cleaner than option 1 and safer than option 2...

flixr:
If the only worry is that someone might forget to load the module, it would be easy to add a compile warning/error for that case:
Simply add a check in boards/ardrone/board.c if  `ELECTRIC_ARDRONE2_PERIODIC_FREQ` is defined and warn/error otherwise...

flixr:
e.g. https://github.com/flixr/paparazzi/commit/be1403da20db4f656b3a578b8b283570da5a7957

gautierhattenberger:
sure, this is nice option as well

kevindehecker:
I just tested this branch, vsupply works and with a correctly configured flightplan, the drone will land when battery is low. (so the checks also work). I did not check the program.elf kill stuff, other then the fact that this branch flies.

Hopefully tonight I will change the module name to vsupply. I don't have an opinion on which pull request is better (this one or #1250). They both work.

kevindehecker:
Since it is closed already, I'm gonna assume no name changes are necessary.

flixr:
I took the liberty to change it to `bat_voltage_ardrone2.xml` already...

kevindehecker:
Ah, perfect. Tnx!

PR #1253, opened 2015-07-22
[boards] add an optional board_init() function

flixr:
and call it at the end of mcu_init

This should be used to init board functions like killing other processes on the Parrot drones
or initializing other onboard peripherals like the cmos cam chips.

So we use this instead of killing progs in electrical_init.

Replaces #1251 and is related to #1252


gautierhattenberger:
seems to work as expected on the Bebop

PR #1255, opened 2015-07-23
Fixed INDI start up crash bug

kevindehecker:
Using the in_flight variable often results in crashes during take off. Apparently it is to slow or something. The thrust variable does not show that problem.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
@@ -220,7 +220,7 @@ static void attitude_run_indi(int32_t indi_commands[], struct Int32Quat *att_err
                                          STABILIZATION_INDI_FILT_OMEGA, STABILIZATION_INDI_FILT_ZETA, STABILIZATION_INDI_FILT_OMEGA_R);
 
   //Don't increment if thrust is off
-  if (!in_flight) {
+  if (stabilization_cmd[COMMAND_THRUST] < 300) {
     FLOAT_RATES_ZERO(indi.u);
     FLOAT_RATES_ZERO(indi.du);
     FLOAT_RATES_ZERO(indi.u_act_dyn);


gautierhattenberger:
Only using the thrust might result in a loss of control in flight if a fast descent is commanded (which was the reason to "improve" the in_flight checking).
Is it an option to keep the fast ground -> in_flight detection you suggest, but properly use the `in_flight` flag to stop the control?

flixr:
Current condition for `in_flight` to become true at takeoff is:
- thrust > AUTOPILOT_IN_FLIGHT_MIN_THRUST for more than AUTOPILOT_IN_FLIGHT_TIME

flixr:
Also I'm wondering if it really makes sense to set all the indi variables to zero while not `in_flight`...
@EwoudSmeur wouldn't it be sufficient to only run the lms_estimation if you are already in flight (without the reset)?

EwoudSmeur:
Maybe I can clarify the problem a bit and explain why this is how I originally did it:

First off: The issue here is NOT about the adaptive part, lms_estimation (even though that should also not run on ground).

@gautierhattenberger I think the in_flight check was properly used, but as Kevin points out, it is very slow in reacting due to the built in timer. You will run into problems in the following case:

The (armed) drone is on the ground, but in_flight true with low throttle. This can happen when you leave low (non-idle) throttle on for a while, or if you land and shortly after want to take-off again. Due to the low throttle the controller has not enough authority to attain the attitude that it wants (because if standing on the ground, there is always a very small attitude error). The INDI controller is designed such that it will increment its control output as fast as it can, so in a very short time it will give the maximum command. If at this point you suddenly increase the thrust and the control does become effective, the command is way too large. 

If, on the other hand, you gradually increase throttle, you will likely not see a flip, because then at some thrust the quad will slowly start to rotate and the command will be reduced.

@flixr This is also why it does make sense to put all variables to zero when not in flight, because they can (and will) increment until saturation when on the ground. When quickly putting the thrust to take-off, this will go berserk.

To wrap this story up: I used the thrust because it is a lot faster than the in_flight check. I never experienced the fast-descend-no-control problem, but I (and others) do experience this problem. I suggest merging this now, and if we come up with something that works for every case we merge that later.

flixr:
Ok, mergin thid for now, and @kevindehecker please please write actually useful/informative commit messages....
It doesn't hurt to put the description on why you changed this into the commit message itself.
If you make a pull request with only one commit, that will become the pull request description as well so you don't have to type it twice...

EwoudSmeur:
@flixr If that is the case than that should have happened with the magneto pull request I did earlier today. Only one commit with a message, with a message. However, in the pull request: No description provided.
How does this work?

flixr:
well, the first line of the commit message becomes the pull request "topic", then you should leave one blank line, the rest is the description body...
Ideall a commit message should stand for itself anyway without needing to consult the pull request or issue...

PR #1256, opened 2015-07-23
Reduced default ahrs align delay

kevindehecker:
Could also set this from airframe file of course, but everybody I talk to seems to be annoyed by the delay.

On ARDrone2 this still works fine.


gautierhattenberger:
I'm afraid the problem is that the aligner is crappy and mostly useless on recent board/ahrs. I personally don't mind reducing this time, but re-writing the aligner would be even better!

flixr:
I also don't mind reducing the delay...
@gautierhattenberger what problems do you see with the current aligner?

PR #1260, opened 2015-07-24
[ABI] special ID to DISABLE callback

flixr:
- change ABI_BROADCAST from 0 to 255
- add ABI_DISABLE with id 0
- separate ABI IDs for binding mag message

This makes it easier to explictitly not use an ABI binding by only defining the ID to ABI_DISABLE (zero) without further change of the code.
The idea is that you can use it to e.g. disable a mag callback for one AHRS without globally setting USE_MAGNETOMETER to FALSE.   
Also make it easy to use a different mag than the one on the IMU.
However the mag still has to have the same alignment as the IMU (still same BODY_TO_IMU used).


gautierhattenberger:
I think it is a good idea.

PR #1261, opened 2015-07-26
Python pprz msg

gautierhattenberger:
Add some utilities and a serial link interface to decode/encode PPRZ messages over a serial link in Python

### DIFF
--- a/sw/lib/python/pprz_msg/pprz_transport.py
+++ b/sw/lib/python/pprz_msg/pprz_transport.py
@@ -96,8 +96,8 @@ class PprzTransport(object):
         data = msg.payload_to_binary()
         # STX + length + sender_id + msg_id + data + ck_a + ck_b
         length = 6 + len(data)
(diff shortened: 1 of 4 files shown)


sw/lib/python/pprz_msg/pprz_transport.py
```diff
@@ -0,0 +1,103 @@
...
+        for c in msg[1:]:
+            ck_a = (ck_a + ord(c)) % 256
+            ck_b = (ck_b + ck_a) % 256
+        return (ck_a, ck_b)
+
+    def pack_pprz_msg(self, sender, msg):
+        data = msg.payload_to_binary()
+        # STX + length + sender_id + msg_id + data + ck_a + ck_b
+        length = 6 + len(data)
+        msg = struct.pack("=BBBB", STX, length, sender, msg.msg_id) + data
```
flixr: We should not not use native byte order, but rather set a fixed byte order: `<` for little endian in our case

PR #1262, opened 2015-07-27
Pullrequestbranch - addition of simple algorithm for divergence calculation

guidoAI:
This branch adds a simple algorithm to determine optical flow divergence to the optical flow code in Paparazzi. It takes the optical flow vectors and evaluates how the distances between corners change over time. This allows to determine the divergence, which is 0 in hover, negative when going up and positive when going down. 

The divergence can be used, e.g., for constant divergence landings, which permit safe landing on surfaces which are not at 0 altitude (roof top for instance).

Main change is the addition of size_divergence.c / h for the calculation of the divergence. The other modifications call the algorithm or pass its values on to the Paparazzi ground station.

### DIFF
--- a/sw/airborne/modules/computer_vision/opticflow/size_divergence.c
+++ b/sw/airborne/modules/computer_vision/opticflow/size_divergence.c
@@ -23,7 +23,7 @@
  */
 
 /**
- * @file modules/computer_vision/opticflow/size_divergence.h
+ * @file modules/computer_vision/opticflow/size_divergence.c
  * @brief Calculate divergence from flow vectors by looking at line sizes beteween the points.
  *
  * Uses optical flow vectors as determined with a corner tracker and Lucas Kanade to estimate divergence.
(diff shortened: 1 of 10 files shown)


flixr:
Nice, can you please also add a copyright and doxgen header to size_divergence.c
and make sure that the files have a newline at the end?

guidoAI:
Done.

sw/airborne/modules/computer_vision/opticflow/size_divergence.c
```diff
@@ -0,0 +1,146 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file modules/computer_vision/opticflow/size_divergence.h
```
flixr: please correct the path to the file...

guidoAI: Corrected

PR #1264, opened 2015-07-27
STM32 I2C driver fix

softsr:
Improves stability in case of the bus stuck.

### DIFF
--- a/sw/airborne/arch/stm32/mcu_periph/i2c_arch.c
+++ b/sw/airborne/arch/stm32/mcu_periph/i2c_arch.c
@@ -176,7 +176,8 @@ static inline void PPRZ_I2C_SEND_START(struct i2c_periph *periph)
 
   // Reset the buffer pointer to the first byte
   periph->idx_buf = 0;
-
+  periph->watchdog = 0;
+  
 #ifdef I2C_DEBUG_LED
   LED_SHOW_ACTIVE_BITS(regs);
 
@@ -976,7 +977,7 @@ void i2c1_ev_isr(void)
   uint32_t i2c = (uint32_t) i2c1.reg_addr;
   i2c_disable_interrupt(i2c, I2C_CR2_ITERREN);
   i2c1.watchdog = 0; // restart watchdog
-  i2c_irq(&i2c1);;
+  i2c_irq(&i2c1);
   i2c_enable_interrupt(i2c, I2C_CR2_ITERREN);
 }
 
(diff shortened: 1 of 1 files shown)


softsr:
How can it be that the code compiles on the computer but gives errors with the the tests?

gautierhattenberger:
you have defined `WD_DELAY` somewhere, probably the header, but didn't commit it

softsr:
The WD_DELAY defined in the same file:
https://github.com/softsr/paparazzi/blob/stm32_i2c_fix/sw/airborne/arch/stm32/mcu_periph/i2c_arch.c#L1301

gautierhattenberger:
true, I thought it was a new variable, that's why I didn't see it in the diff

softsr:
What could be done in this case?

gautierhattenberger:
No idea, it is compiling on my laptop. I just restarted the CI to see.

flixr:
You have `WD_DELAY` defined inside a `#if USE_I2C1 || USE_I2C2 || USE_I2C3` block.
So if you don't have one of those enabled, compilation will fail.

Can you please also run the file through the code style fixer?
`./fix_code_style.sh sw/airborne/arch/stm32/mcu_periph/i2c_arch.c`

softsr:
But what it then wrong with having it inside? It will be defined if using any of the I2C peripherals.

flixr:
Because the i2c code is compiled by default, even if you then don't have any devices actually using it...

flixr:
@gautierhattenberger did you test this?
I'm not able to test this right now, shall we merge it as is or does it need more testing?

podhrmic:
I ll test it today, please don't merge without testing.

softsr:
Any kind of tests or feedback are very appreciated.

OpenUAS:
If Merged I can test with some I2C devices, no timeframe guarantee sadly (weather etc permitting) most likely the Eagletree airspeed sensor first

softsr:
I tested it with the hmc5883 sensor, but would be very important to check if the eagletree airspeed sensor work correctly.

podhrmic:
I tested it on Lisa MX and tried to abuse the bus really hard, but it always recovered. Recommended for merge.

flixr:
Great, thx guys!

flixr:
Also see #1268 for better error reports...

OpenUAS:
+1: Super, time to un-dust some of our eagletree sensors and find an airframe to test on.

PR #1268, opened 2015-07-28
[I2C] cleanup error counters and reports

flixr:
- remove unused i2c error vars and actually count entry of i2c_error irq on STM32
- I2C: add watchdog reset counter to i2c_errors
  - on STM32 increment wd_reset_cnt when the watchdog resets the bus
  - send this counter in I2C_ERRORS
- arch/linux: inc error counters if read/write failed (better use some existing ones with not quite matching names than no reports)


gautierhattenberger:
looks good

PR #1270, opened 2015-07-28
Update Geo Mag module and math Data to WMM 2015 data

OpenUAS:
Update to the WMM2015 data, Tested in Bebop. More outside testing in different parts of the world welcomed


conf/modules/geo_mag.xml
```diff
@@ -4,8 +4,14 @@
   <doc>
     <description>
       GeoMagnetic field vector.
-      Calculation of the normalized geomagnetic field vector (saved to ahrs_impl.mag_h) at startup using GPS fix.
-      Based on the WMM2010 model (http://www.ngdc.noaa.gov/geomag/models.shtml).
+      Calculation of the normalized geomagnetic field vector (saved to ahrs_impl.mag_h) at startup as soon as a valid location is aquired.
+      Method uses is based on the the ngdc noaa DODWMM model and data (http://www.ngdc.noaa.gov/geomag/WMM/DoDWMM.shtml).       
+      To obtain the a magnetic heading to be used by the autopilot, the program must know the magnetic declination at all the points en route, at the indicated flight altitude.
```
flixr: It's not only the declination, but the complete vector....

OpenUAS: Yes, my description skills need improvement. It was more as a brainwash for self to my still PPRZ TODO list to improve all module description to something more informative than some module description.

conf/modules/geo_mag.xml
```diff
@@ -14,11 +20,14 @@
   <init fun="geo_mag_init()"/>
   <periodic fun="geo_mag_periodic()" freq="1"/>
   <event fun="geo_mag_event()"/>
-  <makefile target="ap|nps">
+  <makefile target="ap|nps|sim">
```
flixr: why add the sim target? There is no ahrs running in this sim anyway....

sw/airborne/modules/geo_mag/geo_mag.c
```diff
@@ -49,6 +51,7 @@ void geo_mag_init(void)
 
 void geo_mag_periodic(void)
 {
+  //FIXME: kill_throttle has no place  in a geomag module
```
flixr: This was a deliberate "feature" so that it doesn't calc the data while you are already in flight (as on the older boards it can take quite a while and will delay the control loops during that time...)

PR #1271, opened 2015-07-28
Update Piksi driver

gautierhattenberger:
Some updates for the Piksi driver. Major improvements are:
- better handling of the the baseline messages
- use the libsbp as a submodule

Also try the following operations:
- compute the base "position" from the global position and the local baseline
- send the `BASE_POS` message to module so he can produce pseudo-global DGPS position
- send the antenna frontend mode via sbp message

The sending of message to the module is not properly tested and may not work currently. These changes are already improving and should be used instead of the current one. Some help for testing and finishing the dynamic configuration would be much appreciated.


flixr:
Can't test this as I don't have a piksi...

gautierhattenberger:
It needs more works, but the current driver is really doing incorrect things (actually using the baseline as an ecef position). I don't have time to make it better right know but I still suggest to merge it. I think only TUDelft currently own a Piksi module and they probably won't have time to work on this either.

agressiva:
Hi Gautier, i have a piksi gps and can do tests.
What you need ?

Date: Thu, 30 Jul 2015 13:42:20 -0700

gautierhattenberger:
When you want a pseudo-global RTK position, you need to set a fixed position of one of the module that will broadcast this information allowing the mobile receiver to compute fixed global base position + baseline.
The idea here is to set the base position from the autopilot itself by computing current global position - baseline.
Receiving the position data from Piksi should work, but I couldn't properly test sending information to the module and checking that it has the desired effect.
I also tried to change some settings (here the antenna frontend to use the internal patch antenna) but don't if it really works either.

OpenUAS:
We have a Piksi and intend to test after 20150825

fvantienen:
This implementation doesn't use the DGPS through the datalink of Paparazzi if I'm correct?
I will test this and see if I can add that feature ;)

fvantienen:
Currently rewriting some parts.. but got a question about what to do with libsbp.

The libsbp has includes like "navigation.h" which we already have and causes for errors because it has the same include. I now solved this by including a directory above and changing the libsbp(made an issue for it).
But for this to be resolved in their repository we could either link it statically(like they do in the piksi_firmware), or make a copy of their repo fix it and then make a pull request. Which one do we prefer?

gautierhattenberger:
Maybe the static linking is a bit more annoying, we have to it for all platforms (arm-none, arm-linux). You want to use their makefile or do it yourself with our own build system ?

fvantienen:
Her is my current work in progress: https://github.com/fvantienen/paparazzi/tree/piksi

flixr:
Merged with #1349

PR #1272, opened 2015-07-29
compile/run paparazzi in a docker container

flixr:
Add docker files and scripts to easily compile and run paparazzi in a Docker container.

cd ~/$PAPARAZZI_HOME/docker
`make pull`
then you should have the docker images...
Run a shell in docker:
`make bash`
or
`make terminator`
Then you are in the docker container and you can compile/start paparazzi as usual...
If you already compiled paparazzi within docker, you can also directly run paparazzi the next time:
`make paparazzi`
or
`make start`

/dev/ttyx and /dev/ACMx devices are also automatically forwarded to docker and you can flash and use telemetry from within docker.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -8,6 +8,7 @@ help:
 	@echo "   1. make pull             - pull all pprz images"
 	@echo "   1. make clean            - remove all pprz images"
 	@echo "   2. make bash             - run bash on pprz-dev"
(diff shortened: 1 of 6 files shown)


flixr:
If you guys are fine that the pre-built docker images are in my repo and only updated when I update the docker branch in my repo, then I gues we can merge it as is...

OpenUAS:
Docker is an interesting concept indeed, yes let merge.

PR #1273, opened 2015-07-29
[imu] possibility to disable mag on aspirin 2

flixr:
Use `<define name="ASPIRIN_2_DISABLE_MAG value="TRUE"/>` to disable the mag.


gautierhattenberger:
merged with 0b7a6ebc9512896737463e136772d0c623f22633

PR #1277, opened 2015-07-31
[linux] replace mcu event polling with threads

flixr:
On the linux arch, instead of polling for data to read/write in the uart and udp event functions, start a thread that handles incoming data.

This should make it possible to limit the main event loop to 1kHz as per #1240.

uart_transmit should always be able to directly write the single byte to the serial device, and hence no need to use the tx queue.
Since the uart threads calls select on the fds to check when there is something to read, it will not wake up if there is data in the tx buffer...

Using the tx buffer was previously done because the system timer was generating signals and interrupted the write. Since this is no longer the case it should be fine like this (unless someone adds something that will interrupt the writes again).
    
Seems to work fine on the bebop and beagle bone black.


OpenUAS:
Wow, yes, yes plz this is PPRZ most wanted :+1:  . First today some "old" code testflight, than later on the day this.

OpenUAS:
Testsflew it , looks quite OK and if it where not for a broken GPS of a previous flight I'd say, merge it. Trying to see today to manage another testflight with Bebop and ARdrone2

flixr:
Did you also test it in combination with enabling the event loop limiting again (i.e. reverse https://github.com/paparazzi/paparazzi/commit/fb91702596b310d77342a0c55ab3d98bdb8a8f9d)

OpenUAS:
Thx, will bring discussion to gitter...

PR #1279, opened 2015-08-02
[link] try to fix sending LINK_REPORT

### LINKED ISSUE #1278: Link does not send LINK_REPORT
The `LINK_REPORT` message doesn't seem to be sent by link and hence the link is not properly updated/displayed in the GCS.

The `send_status_msg` callback that is supposed to send this seems to raise an exception:
```(process:22635): LablGTK-CRITICAL **: GSourceFunc: callback raised an exception```

Printing the exception:
```Scanf.Scan_failure("format read '%Lu' does not match specification '%d'")```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25289191-link-does-not-send-link_report?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flixr:
this type stuff in OCaml is such a pain...

Fix sending of LINK_REPORT message in link so that server can properly report TELEMETRY_STATUS for each link.

Also actually fill `tx_msgs` and add `datalink_nb_msgs`.

should fix #1278
~~~ping times still seem to be wrong...~~~

### DIFF
--- a/conf/messages.xml
+++ b/conf/messages.xml
@@ -2774,7 +2774,7 @@
   <message name="LINK_REPORT" id="36">
     <description>Datalink status reported by Link for the Server</description>
     <field name="ac_id" type="string"/>
-    <field name="link_id" type="uint8"/>
+    <field name="link_id" type="string"/>
     <field name="run_time" type="uint32" unit="s"/>
     <field name="rx_lost_time" type="uint32" unit="s"/>
     <field name="rx_bytes" type="uint32"/>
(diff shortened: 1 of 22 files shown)


flixr:
@gautierhattenberger could you plz check this?

flixr:
I also don't seem to get any `PONG` messages...

flixr:
So not getting any PONG messages and hence wrong ping times was a problem with the serial port of the BMP....

PR #1293, opened 2015-08-07
[vision] Now required to indicate that you want to send video using RPT

rmeertens:
Now you can turn off the RTP stream and still log images. By default RTP stream is still on to not confuse RTP stream users.

### DIFF
--- a/sw/airborne/modules/computer_vision/video_usb_logger.c
+++ b/sw/airborne/modules/computer_vision/video_usb_logger.c
@@ -31,7 +31,7 @@
 
 /** Set the default File logger path to the USB drive */
 #ifndef VIDEO_USB_LOGGER_PATH
-#define VIDEO_USB_LOGGER_PATH /data/video/usb
+#define VIDEO_USB_LOGGER_PATH "/data/video/usb/"
 #endif
 
 /** The file pointer */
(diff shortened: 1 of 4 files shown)


flixr:
Thanks :-)

But RTP is now off by default... and the define should be `VIEWVIDEO_USE_RTP` not `VIEWVIDEO_USE_RPT`
The last else is superfluous, you can just leave it out...

I guess it would be even nicer to make this changeable during runtime (so add a variable/flag that you can change via settings) instead of only making it a compile time option.

flixr:
@rmeertens are you willing to update this PR accordingly?

rmeertens:
Fixed it. However, it is on a private branch. Any idea how I can create a pull request with files on a private repository?

Also: is anybody using netcat? It does not compile when I enable it.

flixr:
Just push it to the branch you already used for this pull request (tempfix in your public repo) and it will show up here.

rmeertens:
Updated! Somebody has to check/implement the netcat, as I don't know what it is.

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -44,12 +44,12 @@ void video_usb_logger_start(void)
   char filename[512];
 
   // Check for available files
-  sprintf(filename, "%s/%05d.csv", STRINGIFY(VIDEO_USB_LOGGER_PATH), counter);
+  sprintf(filename, "%s%05d.csv", VIDEO_USB_LOGGER_PATH, counter);
```
flixr: why change this?

flixr:
@fvantienen what is up with the netcat option? Is that used/needed? If yes, can you plz fix it or I'll remove it....

flixr:
pushed only the (cleaned up) changes for disabling the rtp stream...

PR #1294, opened 2015-08-08
Elle0

esden:
Adding support for the ELLE0 V1.0 autopilot. This autopilot is still in beta phase, but for easier testing and further integration it would be nice if it was part of the mainline. :)

### DIFF
--- a/sw/airborne/boards/elle0/baro_board.h
+++ b/sw/airborne/boards/elle0/baro_board.h
@@ -1,0 +1,17 @@
+/*
+ * board specific functions for the elle0 board
+ *
+ */
+
+#ifndef BOARDS_ELLE0_BARO_H
+#define BOARDS_ELLE0_BARO_H
+
+// only for printing the baro type during compilation
+#ifndef BARO_BOARD
+#define BARO_BOARD BARO_MS5611_SPI
+#endif
+
+extern void baro_event(void);
+#define BaroEvent baro_event
+
+#endif /* BOARDS_ELLE0_BARO_H */
--- a/sw/airborne/boards/elle0_1.0.h
+++ b/sw/airborne/boards/elle0_1.0.h
@@ -1,0 +1,443 @@
+/*
+ * Copyright (C) 2015 Piotr Esden-Tempski <piotr@esden.net>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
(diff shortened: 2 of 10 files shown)


sw/airborne/boards/elle0/baro_board.h
```diff
@@ -0,0 +1,18 @@
+
+/*
+ * board specific functions for the lisa_mx board
+ *
+ */
+
+#ifndef BOARDS_LISA_MX_BARO_H
```
fvantienen: Forgot to rename something?

sw/airborne/boards/elle0/baro_board.h
```diff
@@ -0,0 +1,18 @@
+
+/*
+ * board specific functions for the lisa_mx board
```
fvantienen: also here...

sw/airborne/boards/elle0_1.0.h
```diff
@@ -0,0 +1,446 @@
...
+#define ACTUATORS_PWM_NB 8
+#define USE_PWM7 1
+#define USE_PWM8 1
+#define PWM_USE_TIM4 1
+#endif
+#else
+#define ACTUATORS_PWM_NB 6
+#endif
+
+// Servo numbering on LisaM silkscreen/docs starts with 1
```
fvantienen: Lisam?

sw/airborne/boards/elle0_1.0.h
```diff
@@ -0,0 +1,446 @@
+/*
+ * Copyright (C) 2013 Piotr Esden-Tempski <piotr@esden.net>
```
fvantienen: You live in the past :smile:

fvantienen:
Cool.. is it already working a bit?

esden:
@fvantienen of course otherwise I would not be sending in a pull request ;)

https://www.youtube.com/watch?v=QBs-y-IvI9E

flixr:
The board files look good :-)

Regarding the airframe files:
- We said that we would only keep airframes in conf/examples that are actually part of the example of test conf...
- it would make sense to use the correct motor mixing
- not so sure about the value of the setup_elle0 file

esden:
Regarding the setup. It would be great to find a solution of compiling the setup firmwares without having to put it into either the airframe file or having a "dummy" airframefiles for the specific autopilot. The best thing would be to have those targets available for every aircraft that uses certain autopilot. But that is a fix that needs to be done separately. I use those setup firmwares every time for debugging and configuration purposes. That is why I like the fact that they are separate "aircraft" a better solution would be nice though...

Regarding your other comments I am working on a patch and I will push it in a few.

hooperfly:
@esden maybe the motor layouts for the examples should reflect the ordering @flixr introduced for default motor layouts. In the case of a quad-x configuration.

```
<set servo="TOP_LEFT" value="motor_mixing.commands[MOTOR_FRONT_LEFT]"/>
<set servo="TOP_RIGHT" value="motor_mixing.commands[MOTOR_FRONT_RIGHT]"/>
<set servo="BOTTOM_RIGHT" value="motor_mixing.commands[MOTOR_BACK_RIGHT]"/>
<set servo="BOTTOM_LEFT" value="motor_mixing.commands[MOTOR_BACK_LEFT]"/>
```

esden:
The setup I have put into the example airframe file is the one I always had on all of my rotorcraft since forever. That is why I have that in this frame as it matches my test airframe. :) I can change that if this is the desired setup for the example airframes but that also means that I will not be using it on my test airframe...

esden:
Ok. I have cleaned up the patches. I added back the ```USE_SERVOS_7AND8``` because we can use it to make them available for the upcomming WS2812 module as well as auxiliary I2C1 interface.

PR #1298, opened 2015-08-11
Transparent subsystems

podhrmic:
Add transparent ("dummy") subsystems, so we can compile the airframe even without a specific subsystem (imu, ahrs , gps). Useful when one subsystem takes the function of multiple subsystems.

An example is an external INS sensor - such as MTI-G or Vectornav. Using such an INS we dont need GPS, IMU and AHRS subsystems, but we still have to include them so the AP code compiles properly. 

Or if we are using GX3 (and AHRS module), we dont need IMU subsystem and so on.

### DIFF
--- a/conf/firmwares/subsystems/fixedwing/gps_transparent.makefile
+++ b/conf/firmwares/subsystems/fixedwing/gps_transparent.makefile
@@ -11,3 +11,8 @@ ap.CFLAGS += -DUSE_GPS
 ifneq ($(GPS_LED),none)
   ap.CFLAGS += -DGPS_LED=$(GPS_LED)
 endif
+
+nps.CFLAGS += -DUSE_GPS
+nps.srcs += $(SRC_SUBSYSTEMS)/gps.c
+nps.CFLAGS += -DGPS_TYPE_H=\"subsystems/gps/gps_transparent.h\"
+nps.srcs += $(SRC_SUBSYSTEMS)/gps/gps_transparent.c
--- a/conf/firmwares/subsystems/fixedwing/gps_transparent.makefile
+++ b/conf/firmwares/subsystems/fixedwing/gps_transparent.makefile
@@ -1,0 +1,13 @@
+# GPS Transparent
+# 2014, Michal Podhradsky, michal.podhradsky@aggiemail.usu.edu
+# Utah State University, http://aggieair.usu.edu/
+GPS_LED ?= none
+
+ap.srcs += $(SRC_SUBSYSTEMS)/gps.c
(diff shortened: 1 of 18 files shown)


flixr:
I don't quite understand the motivation for this...
A properly written INS subsystem that includes IMU/AHRS functionality shouldn't require any dummy IMU/AHRS subsystems...
If it does, it's a bug and you should let us know...

podhrmic:
The motivation is that instead of having all the defines for all subsystems (in case of having an external INS) in the ins file (which will make it large and cluttered), I can define a minimal "dummy" subsystem and don't worry about it (and focus on INS implementation).

Also, in case I have lets say two different INS systems, A and B, each would have to have its own defines for IMU, AHRS etc which will lead to code duplication.

I guess one improvement would be to merge the makefiles, so you can either just define 
```    <subsystem name="ins" type="ext_A"/>```
and nothing else, or (if you are testing something) you can still use 
```     <subsystem name="imu" type="transparent"/>```

Does it makes more sense now?

podhrmic:
Plus this way you can test it in NPS too with the same functionality...

podhrmic:
What do you think @flixr can we merge it as is?

flixr:
I still don't quite see the necessity for this, will take a proper look when I have some quiet minutes... so please bear with me...
If you could provide a really concrete (code) example where this is needed it might help me to understand why you want this added, but in general my feeling is that this should not be needed...

podhrmic:
I just added a driver example for Vectornav VN-200 (http://www.vectornav.com/products/vn200-rugged) - it is not finished and needs some polishing, but you can see the basic functionality from it.

Let me know what you think.

flixr:
Well, I still don't see why we would need dummy subsystem in order to use something like the vectornav INS.
We also don't need them for XSens (ok, that module/subsystem is currently not a good example)...

If there is something annoying about having an INS subsystem that also provides GPS, AHRS and IMU functionality, please make notes of that...
So IMHO we should rather fix/clean it at the root instead of adding more hacks like the dummy subsystems...

podhrmic:
OK, I will post a separate pull request with just the driver.

gautierhattenberger:
Sorry, I'm a bit late on the discussion, but I also think we should have a solution where dummy subsystems are not needed.
Especially, using ABI messages and the state interface (and finish the conversion of subsystems to modules) should make it pretty easy.
I also don't see any issue when a subsystem is providing functionality below his own "level" (like an INS providing ahrs, imu, gps, baro,...). Only the opposite should be avoided (like a GPS directly pretending to be an INS).

PR #1301, opened 2015-08-16
[BMPM] Tell Black Magic probe to use the reset pin before scanning

esden:
This should fix situations where it is impossible to program a target
using JTAG or SWD as the target can not be found.

Also added comments and a check if the load was successful.

### DIFF
--- a/sw/tools/flash_scripts/bmp_jtag_flash.scr
+++ b/sw/tools/flash_scripts/bmp_jtag_flash.scr
@@ -1,4 +1,15 @@
+# Print BMPM version
+mon version
+# To make sure the target is not in a "strange" mode we tell BMPM to reset the
+# target using the reset pin.
+mon connect_srst enable
+# Find the target
 monitor jtag_scan
+# Attach to the first device
 attach 1
+# Load the binary
 load
+# Check if the flash matches the binary
+compare-sections
+# Reset and exit
 kill
--- a/sw/tools/flash_scripts/bmp_swd_flash.scr
+++ b/sw/tools/flash_scripts/bmp_swd_flash.scr
@@ -1,5 +1,15 @@
+# Print BMPM version
 monitor version
+# To make sure the target is not in a "strange" mode we tell BMPM to reset the
+# target using the reset pin.
+mon connect_srst enable
+# Scan for devices using SWD interface
 monitor swdp_scan
+# Attach to the first device
 attach 1
+# Load the binary into the flash
 load
+# Check if the flash matches the binary
+compare-sections
+# Reset and exit
 kill


hooperfly:
Trying to use lastest paparazzi/master with thai patch to upload new firmware. I must need to modify my target in some way based on the error messge?

```
cd sw/airborne; /Applications/Xcode.app/Contents/Developer/usr/bin/make TARGET=ap upload
Using "arch/stm32/lisa-mx.ld" as ldscript for target "ap".
Assuming luftboot bootloader: no
Using Black Magic Probe with SWD on BMP_PORT /dev/cu.usbmodemE2B9BDC1
Using GDB = arm-none-eabi-gdb
  BMP    /Users/richardaburton/paparazzi/var/aircrafts/Teensy_Fly_Quad/ap/ap.elf
Black Magic Probe (Firmware 1.5, build 20130323)
Copyright (C) 2011  Black Sphere Technologies Ltd.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

/Users/richardaburton/paparazzi/sw/tools/flash_scripts/bmp_swd_flash.scr:5: Error in sourced command file:
Target does not support this command.
```

flixr:
Hum... works for me here with `Black Magic Probe (Firmware 1.5-00233-ge5624a7-dirty, build 20140416)`

flixr:
Can you remove the `compare_sections` command in the flash script and test if it works for you?
Maybe we should find a way to conditionally verify the image?

hooperfly:
Error message complains about line 5 in the script:

`mon connect_srst enable`

hooperfly:
So I commented out line 5 and line 13 to get my load to work.

esden:
You should upgrade the firmware on your BMP this is the correct way of "fixing" the problem. We need the srst to make the flashing more robust. :)

hooperfly:
Making everyone upgrade their BMP firmware for an enhancement seems extreme. I can still do work without using this upgraded script. I like the idea of giving a warning to the user and that they will need to upgrade their firmware to take advantage of enhancements while still supporting the existing feature/target set for a given BMP version. If we expect folks to upgrade their BMP firmware, there should be `good` documentation to walk them through the process for Linux and MacOS, the two operating systems PaparazziUAV supports. My 2 cents.

hooperfly:
Updated my BMP to the latest firmware. I will try to `beef up` the BMP related pages on the wiki pages for building and installing BMP firmware with Linux and MacOS X.

flixr:
I guess ideally we would check the BMP version first and then execute that command or not...
@esden in what cases was the reset needed/beneficial?

gautierhattenberger:
It is a bit sad, but the best instructions I could find for upgrading a BMP are on the pixhawk wiki: https://pixhawk.org/dev/jtag/black_magic_probe#instructions_for_linux
I couldn't really find it in BMP wiki or readme files...

dewagter:
or in paparazzi:

cd ./sw/ext/blackmagic/
make

-Christophe

gautierhattenberger:
True, but this one is not in the wiki I think, only the manual version (I should have checked our own wiki first...).
Still, I think BMP guys should improve the readme or at least add an upgrade target in the makefile, it would be easier for newbies.

hooperfly:
@gautierhattenberger, yes the Pixhawk page was the most informative. I snagged the MacOS and Linux related instructions and cached them on my local page.

http://wiki.paparazziuav.org/wiki/User:Earthpatrol#Black_Magic_Probe.28BMP.29

I'm hoping to get some time to update the BMP related wiki topics at some point.

PR #1302, opened 2015-08-17
[STM32F4] Added support for persistent settings on STM32F4.

### LINKED ISSUE #600: persistent settings for STM32F4 [$10]
pflash_program_bytes is a no-op on F4

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/6530252-persistent-settings-for-stm32f4?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

esden:
This should fix #600

To make things simple we always use the last page of the flash. We also assume for the F4 cores that all pages are the same size. This means the current code will likely brake for very small flash sizes under 256kb. Which we are not very likely to use but it needs to be kept in mind.

### DIFF
--- a/sw/airborne/arch/stm32/apogee.ld
+++ b/sw/airborne/arch/stm32/apogee.ld
@@ -26,8 +26,8 @@ MEMORY
 {
 	/* only 128K (SRAM1 and SRAM2) are accessible by all AHB masters. */
 	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 128K
-	/* Leaving 2k of space at the end of rom for stored settings */
-	rom (rx) : ORIGIN = 0x08000000, LENGTH = 1022K
+	/* Leaving 128k of space at the end of rom for persistent settings */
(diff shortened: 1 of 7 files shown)


flixr:
All F4 linker scripts still only reserve 2k at the end of ROM for the persistent settings...

flixr:
Seems it is getting the wrong flash size and then writes to the wrong page thus overwriting application code...

flixr:
Thx, works as expected now here...
I guess we should probably still overwrite the page with the persistent settings when we re-flash...?

flixr:
Added support for 2MB flash devices with 3dc65f6da2ad80689e9997213ff142c63848f0d6

PR #1303, opened 2015-08-17
[telemetry] BlueGiga (bluetooth) Telemetry and SPI slave on STM3

dewagter:


flixr:
IMHO the bluegiga_usb_dongle should be a submodule...

dewagter:
Could become a submodule if it were used by any other project

flixr:
why is there a binary blob (kernel driver) in there? For what arch, kernel? Why is that needed if it is just a cdc-acm driver?

conf/firmwares/subsystems/rotorcraft/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,17 @@
...
+# The bluegiga module as telemetry downlink/uplink
+#
+#
+ap.CFLAGS += -DUSE_$(MODEM_PORT)
+ap.CFLAGS += -D$(MODEM_PORT)_BAUD=$(MODEM_BAUD)
+
+ap.CFLAGS += -DDOWNLINK -DPERIODIC_TELEMETRY -DDOWNLINK_DEVICE=bluegiga_p
+ap.CFLAGS += -DDOWNLINK_TRANSPORT=pprz_tp -DDATALINK=BLUEGIGA
+#ap.CFLAGS += -DDefaultPeriodic='&telemetry_Main'
+ap.CFLAGS += -DUSE_SPI2_SLAVE -DSPI_SLAVE
```
flixr: should not be hardcoded

conf/firmwares/subsystems/rotorcraft/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,17 @@
...
+
+# for some reason the following line required to get GPS over datalink working
+ap.CFLAGS += -DUSE_UART3 -DUART3_BAUD=B38400
```
flixr: ????

sw/airborne/subsystems/datalink/bluegiga.c
```diff
@@ -0,0 +1,245 @@
...
+}
+static void trans_cb(struct spi_transaction *trans __attribute__((unused)))
+{
+  bluegiga_receive();
+}
+
+void bluegiga_init(void)
+{
+
+  LED_INIT(3);
```
flixr: do not hardcode leds!

flixr:
Btw, if Kirk wrote all that code, why didn't he make the pull request?
Would also make sense if he would be the author of the commits...

flixr:
Also when adding support for a new device like this, please please do not use one global var for the device struct, but rather pass a pointer to it with every function, so that you can have multiple ones...
So basically split the blue_giga device driver from the datalink wrapper that just uses blue_giga...

kirkscheper:
I have never made a pull request before so Christophe was showing me how it was done (I thought that would be more time efficient). In the future I will make my own pull requests.

The kernel driver is there cause we can also run the program on the ARDrone with a bluetooth dongle in the usb port. I needed the CDC-ACM driver to mount the dongle. I am sure there are other ways to do it but this is the way that worked for me. It was a lot of work to get the driver made so I didn't want to the separate it and chance that it gets lost.

This push is based on the superbit_rf telemetry files hence the SPI definition and single global var. I assume the structure has changed since I started developing this, I will take a look at how it is defined now and update the code.

I will move the UART3 definition to a different pull request for GPS over datalink.

I will leave the decision about being a subproject to you guys, my only comment is that you can't really separate the bluetooth temeletry and this program as it is currently the only way to communicate with the bluetooth usb dongles.

kirkscheper:
I don't really get the comment of splitting the device peripheral from the datalink wrapper, can you elaborate? Currently, the bluegiga.c is a sort of device driver which handles all the spi transfers and recoding the spi msgs and the h file I assume is what you are referring as a datalink wrapper. How do you want me to split them? I also am not sure how you can have multiple peripherals with the same datalink.

dewagter:
Could this be merged as is or do you have a suggestion?

conf/firmwares/subsystems/rotorcraft/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,19 @@
...
+
+ap.CFLAGS += -DDOWNLINK -DPERIODIC_TELEMETRY -DDOWNLINK_DEVICE=bluegiga_p
+ap.CFLAGS += -DDOWNLINK_TRANSPORT=pprz_tp -DDATALINK=BLUEGIGA
+
+BLUEGIGA_PORT ?= SPI2
+BLUEGIGA_PORT_LOWER=$(shell echo $(BLUEGIGA_PORT) | tr A-Z a-z)
+ap.CFLAGS += -DUSE_$(BLUEGIGA_PORT)_SLAVE -DSPI_SLAVE -DBLUEGIGA_SPI_DEV=$(BLUEGIGA_PORT_LOWER)
+
+MODEM_LED ?= none
+ap.CFLAGS += -DMODEM_LED=$(MODEM_LED)
```
flixr: missing a `ifneq ($(MODEM_LED),none)` check

Makefile
```diff
@@ -281,6 +283,7 @@ clean:
 	$(Q)rm -f  $(GEN_HEADERS)
 	$(Q)find . -mindepth 2 -name Makefile -a ! -path "./sw/ext/*" -exec sh -c 'echo "Cleaning {}"; $(MAKE) -C `dirname {}` $@' \;
 	$(Q)$(MAKE) -C $(EXT) clean
+	$(Q)$(MAKE) -C $(TOOLS)/bluegiga_usb_dongle clean
```
flixr: not needed, already cleaned by the line two above

flixr:
There should be a `telemetry_bluegiga.makefile` in `conf/firmwares/subsystems/shared` which is then included in fixedwing and rotorcraft variants as it is done with the other telemetry subsystems.

flixr:
Please at least add a Readme to the bluegiga directory explaining what it does and especially document what kind of architecture and kernel version the cdc-acm module is for....
Would also be good to have an example airframe that uses bluegiga that we could add to conf_tests.xml so it will be built on the CI.

sw/airborne/subsystems/datalink/bluegiga.c
```diff
@@ -0,0 +1,249 @@
...
+  for (int i = 0; i < bluegiga_spi.output_length; i++) {
+    bluegiga_p.work_tx[i] = 0;
+  }
+  for (int i = 0; i < 255; i++) {
+    bluegiga_rssi[i] = 127;
+  }
+
+  // Configure generic device
+  bluegiga_p.device.periph    = (void *)(&bluegiga_p);
+  bluegiga_p.device.check_free_space = (check_free_space_t) true_function;
```
flixr: why always return true instead of making a proper check_free_space function?

flixr:
Doesn't even seem to compile here without further changes...

sw/airborne/arch/stm32/mcu_periph/spi_arch.c
```diff
@@ -1311,6 +1311,104 @@ void dma1_channel2_isr(void) {
...
+  spi2_dma.rx_nvic_irq = NVIC_DMA1_CHANNEL4_IRQ;
+  spi2_dma.tx_nvic_irq = NVIC_DMA1_CHANNEL5_IRQ;
+  spi2_dma.tx_dummy_buf = 0;
+  spi2_dma.tx_extra_dummy_dma = FALSE;
+  spi2_dma.rx_dummy_buf = 0;
+  spi2_dma.rx_extra_dummy_dma = FALSE;
+
+  // set the default configuration
+  set_default_comm_config(&spi2_dma.comm);
+  spi1_dma.comm_sig = get_comm_signature(&spi2_dma.comm);
```
flixr: should be spi2...

sw/airborne/subsystems/datalink/bluegiga.h
```diff
@@ -0,0 +1,111 @@
...
+#endif
+      pprz_parse_payload(t);
+      t->trans_rx.msg_received = FALSE;
+    }
+    bluegiga_increment_buf(&bluegiga_p.rx_extract_idx, c);
+  } while (BlueGigaChAvailable()); // continue till all messages read
+}
+
+// transmit previous data in buffer and parse data received
+#define BlueGigaCheckAndParse(_dev,_trans) {      \
```
flixr: would be better to make this a static inline function and use the passed dev and trans (like e.g. in w5100)

flixr:
> I don't really get the comment of splitting the device peripheral from the datalink wrapper, can you elaborate? Currently, the bluegiga.c is a sort of device driver which handles all the spi transfers and recoding the spi msgs and the h file I assume is what you are referring as a datalink wrapper. How do you want me to split them? I also am not sure how you can have multiple peripherals with the same datalink.

In general it is much better to not assume one global var for device drivers, instead pass the bluegiga periph to the bluegiga_init/transmit/receive/... functions as an argument. Also makes it possible to factor it out into a lib...
Then in a datalink/downlink wrapper you add the actual var holding the bluegiga periph for datalink and pass it as pointer...

Maybe we don't want to worry about that for now (it is a bit unlikely that you have multiple bluegiga modules connected at the same time).
But in general when adding new device drivers, they should be written in a clean way to enable maximum re-use. And it is normally easier to do it properly from the start then changing that later....

dewagter:
- fixed many useful comments: tnx
 - added test airframe in conf_tests.xml (compiles on our machines)
 - no makefile in shared is on purpose as this is now rotorcraft only
 - not possible to run several bluegiga's for now, which is why the easy short global var was used
 - wiki/readme still in progress (@kirkscheper?)
 - 'do it properly from the start': this code is > 0.5 year old and back then this was the standard so you probably mean we should have pull requested it earlier? I agree :+1:

sw/airborne/subsystems/datalink/bluegiga.c
```diff
@@ -0,0 +1,249 @@
...
+#include "subsystems/datalink/bluegiga.h"
+#include "mcu_periph/spi.h"
+#include "mcu_periph/gpio.h"
+
+#ifndef BLUEGIGA_SPI_DEV
+#error "bluegiga: must define a BLUEGIGA_SPI_DEV"
+#endif
+
+#ifndef BLUEGIGA_SLAVE_IDX
+#define BLUEGIGA_SLAVE_IDX SPI_SLAVE2
```
flixr: This does not make any sense, if I see it correctly, then the bluegiga_spi.slave_idx is not even used anywhere in the slave code.
Better completely remove this and set slave_idx to zero below, so that it works if SPI_SLAVE2 is not defined

flixr:
> added test airframe in conf_tests.xml (compiles on our machines)

thx, that one compiles because it already includes spi master somewhere else..

> no makefile in shared is on purpose as this is now rotorcraft only

Why? There is nothing rotorcraft specific about this.... also not having it in shared means that you can't use it in other firmwares, e.g. setup or test_progs...

What about the check_free_space function?

sw/airborne/subsystems/datalink/bluegiga.h
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * @file subsystems/datalink/bluegiga.h
+ * Bluegiga Bluetooth chip I/O
+ */
+
+#ifndef BLUEGIGA_DATA_LINK_H
+#define BLUEGIGA_DATA_LINK_H
+
+#include "mcu_periph/link_device.h"
+#include "generated/airframe.h"
```
flixr: why include this here?

conf/firmwares/subsystems/rotorcraft/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,21 @@
...
+# The bluegiga module as telemetry downlink/uplink
+#
+#
+ap.CFLAGS += -DUSE_$(MODEM_PORT)
+ap.CFLAGS += -D$(MODEM_PORT)_BAUD=$(MODEM_BAUD)
+
+ap.CFLAGS += -DDOWNLINK -DPERIODIC_TELEMETRY -DDOWNLINK_DEVICE=bluegiga_p
+ap.CFLAGS += -DDOWNLINK_TRANSPORT=pprz_tp -DDATALINK=BLUEGIGA
+
+BLUEGIGA_PORT ?= SPI2
```
flixr: Should be called BLUEGIGA_SPI_DEV for consistency...

conf/firmwares/subsystems/rotorcraft/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,21 @@
+#
+# The bluegiga module as telemetry downlink/uplink
+#
+#
+ap.CFLAGS += -DUSE_$(MODEM_PORT)
```
flixr: These should go...

conf/firmwares/subsystems/shared/telemetry_bluegiga.makefile
```diff
@@ -0,0 +1,25 @@
...
+
+# LED
+MODEM_LED ?= none
+ifneq ($(MODEM_LED),none)
+ap.CFLAGS += -DMODEM_LED=$(MODEM_LED)
+endif
+
+ap.srcs += subsystems/datalink/downlink.c subsystems/datalink/bluegiga.c
+ap.srcs += subsystems/datalink/pprz_transport.c subsystems/datalink/telemetry.c
+ap.srcs += $(SRC_FIRMWARE)/datalink.c $(SRC_FIRMWARE)/rotorcraft_telemetry.c
```
flixr: using `rotorcraft_telemetry.c` obviously only works for rotorcraft, the shared one should contain the common parts without that and the fixedwing and rotorcraft specific ones include the shared one while adding the specific files.
See all other telemetry subsystems...

flixr:
The firmware specific `$(SRC_FIRMWARE)/datalink.c` should not be in shared, but in fixedwing and rotorcraft... so you can the shared subsystem for other targets like test_progs, etc. as well...
But I can fix that...

But I'm still wondering what is up with the check_free_space function?

flixr:
Using the link device was fine, I meant why assign a function that always returning true instead of actually implementing `bluegiga_check_free_space` (the declaration already exists in the header) which basically would return what `BlueGigaCheckFreeSpace()` does...
And then also use that function instead of the macro in the other places...

dewagter:
was still busy... tried to make an implementation to allow multiple bluegiga's but got stuck as there is a need for special SPI-SLAVE without NSS and an DR/DY pin per device. This goes way beyond the scope of the intended use.

dewagter:
The BlueGigaCheckFreeSpace macro is a more simple implementation that
checks space for 1 byte.
I implemented a full check space for entire message function.

-Christophe

hooperfly:
Compilation failing after the merge:

```
luftboot.c: In function 'main':
luftboot.c:481:9: note: #pragma message: Luftboot using 12MHz external clock to PLL it to 72MHz.
 #pragma message "Luftboot using 12MHz external clock to PLL it to 72MHz."
         ^
  CC      luftboot.elf
  create  luftboot.bin
  create  luftboot.hex
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C sw/tools/bluegiga_usb_dongle
uart.c:265:25: error: use of undeclared identifier 'B921600'
  cfsetispeed(&options, B921600);
                        ^
uart.c:266:25: error: use of undeclared identifier 'B921600'
  cfsetospeed(&options, B921600);
                        ^
2 errors generated.
make[1]: *** [uart.o] Error 1
make: *** [ext] Error 2
```

flixr:
The bluegiga_usb_dongle is a part of the ground segment and not airborne...
Must be an OSX problem as it compiles under LInux....

hooperfly:
@flixr right, ground not air. Thanks. I searched the code base for B921600 definitions/declarations and only found them in the following two files:

```
./sw/airborne/arch/linux/mcu_periph/uart_arch.h
./sw/airborne/arch/stm32/mcu_periph/uart_arch.h
```
Am I missing some other file that might be defining `B921600`?

flixr:
The baudrates are normally defined in `termios.h` provided by the system...
Seems that it is not necessarily the case for higher baud rates on OSX: https://discussions.apple.com/thread/1587991?start=0&tstart=0

hooperfly:
@flixr @dewagter can we kludge or fix this in some reasonable way so that every time I sync with the mainline, I don't have to patch this file? Here's a possible kludge for now:

```
/* KLUDGE: Faster baudates not defined in MacOS X */
#ifdef __APPLE__
  cfsetispeed(&options, 921600);
  cfsetospeed(&options, 921600);
#else
  cfsetispeed(&options, B921600);
  cfsetospeed(&options, B921600);
#endif
```

flixr:
@dewagter does it have to be that baudrate?
@hooperfly what is the largest baudrate available on osx, then I would simply set that for now, unless you want to do the ioctl?

hooperfly:
@flixr @dewagter Here's a snippet from /usr/include/sys/termios.h on Mac OS X:

```
/*
 * Standard speeds
 */
#define B0	0
#define B50	50
#define B75	75
#define B110	110
#define B134	134
#define B150	150
#define B200	200
#define B300	300
#define B600	600
#define B1200	1200
#define	B1800	1800
#define B2400	2400
#define B4800	4800
#define B9600	9600
#define B19200	19200
#define B38400	38400
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
#define B7200	7200
#define B14400	14400
#define B28800	28800
#define B57600	57600
#define B76800	76800
#define B115200	115200
#define B230400	230400
#define EXTA	19200
#define EXTB	38400
#endif  /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */
```

I tested the `B230400` value, the fastest speed in the list and it compiles fine. Here's the code I used for the modification:

```
/* KLUDGE: Faster baudates not defined in MacOS X */
#ifdef __APPLE__
  cfsetispeed(&options, B230400);
  cfsetospeed(&options, B230400);
#else
  cfsetispeed(&options, B921600);
  cfsetospeed(&options, B921600);
#endif
```

flixr:
@hooperfly thx, committed the kludge/fix
@dewagter would still be good to know what baud rates make sense for bluegiga...

hooperfly:
@dewagter @flixr here's a link to the data sheet for the BLE113. In the UART section, page 10 of the document, is a list of baud settings. 230400 is the fastest baudrate listed.

http://www.mouser.com/ds/2/52/BLE113_Datasheet-224874.pdf

PR #1305, opened 2015-08-17
[INS] Vectornav VN-200 driver

podhrmic:
* Allows user to use Vectornav VN-200 as an external IMU for either fixedwing or rotorctraft.
* No other subsystems (IMU, AHRS, GPS) are needed, all is self-contained
* Compatible with NPS, so the airframe with Vectornav can be compiled for both AP and NPS targets

**Note:** Part of the pull request is re-addition of float IMU, since we are getting all data from Vectornav as float, so it makes sense to keep it that way.

### DIFF
--- a/sw/airborne/peripherals/vn200_serial.c
+++ b/sw/airborne/peripherals/vn200_serial.c
@@ -28,12 +28,51 @@
  */
 #include "peripherals/vn200_serial.h"
 
+
+/**
+ * Calculates the 16-bit CRC for the given ASCII or binary message.
+ * The CRC is calculated over the packet starting just after the sync byte (not including the sync byte)
+ * and ending at the end of payload.
+ */
+static inline unsigned short calculateCRC(unsigned char data[], unsigned int length)
+{
+  unsigned int i;
+  unsigned short crc = 0;
+  for (i = 0; i < length; i++) {
+    crc = (unsigned char)(crc >> 8) | (crc << 8);
+    crc ^= data[i];
+    crc ^= (unsigned char)(crc & 0xff) >> 4;
+    crc ^= crc << 12;
+    crc ^= (crc & 0x00ff) << 5;
+  }
+  return crc;
+}
+
+
+/**
+ * Verify checksum
+ */
+static inline bool verify_chk(unsigned char data[], unsigned int length, uint16_t *calc_chk, uint16_t *rec_chk)
+{
+  unsigned short calc_crc = calculateCRC(data, length);
+  unsigned short rec_crc = (unsigned short)(data[length] << 8 | data[length + 1]);
+  *calc_chk = (uint16_t) calc_crc;
+  *rec_chk = (uint16_t) rec_crc;
+  if (calc_crc == rec_crc) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
+
+
 static inline void vn200_read_buffer(struct VNPacket *vnp)
 {
   while (uart_char_available(&VN_PORT) && !(vnp->msg_available)) {
     vn200_parse(vnp, uart_getch(&VN_PORT));
   }
 }
+
 
 void vn200_event(struct VNPacket *vnp)
 {
@@ -93,7 +132,7 @@ void vn200_parse(struct VNPacket *vnp, uint8_t c)
         vnp->status = VNMsgSync;
       }
       break;
-    default:;
+    default:
       vnp->status = VNMsgSync;
       vnp->msg_idx = 0;
       break;
--- a/sw/airborne/peripherals/vn200_serial.c
+++ b/sw/airborne/peripherals/vn200_serial.c
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2015 Michal Podhradsky, michal.podhradsky@aggiemail.usu.edu
+ * Utah State University, http://aggieair.usu.edu/
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+/**
+ * @file vn200_serial.c
+ *
+ * Vectornav VN-200 INS subsystem
+ *
+ * @author Michal Podhradsky <michal.podhradsky@aggiemail.usu.edu>
+ */
+#include "peripherals/vn200_serial.h"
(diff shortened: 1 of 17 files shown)


flixr:
I'm not in favour of re-adding the global imuf struct. We want to get rid of the global imu struct anyway...

podhrmic:
What do you recommend then? Having the accel and gyro data stored internally in VN struct and just send out the ABI messages instead?

flixr:
Basically, yes...

With the global IMU struct we can't support multiple IMUs, but it is currently still used to send the IMU RAW/SCALED messages from one common place (so that we don't have to do that from every IMU driver)...

So if you want the IMU msgs, either send them directly from the vectornav driver/wrapper or convert to the fixedpoint imu for the time being (if you want to have the IMU messages) until we find a nicer solution for sending the IMU msgs.

sw/airborne/peripherals/vn200_serial.h
```diff
@@ -0,0 +1,119 @@
...
+ */
+#define VN_SYNC 0xFA
+#define VN_OUTPUT_GROUP 0x39
+#define VN_GROUP_BYTES 8
+
+#define VN_BUFFER_SIZE 512
+#define VN_HEADER_SIZE 9
+#define VN_PAYLOAD_SIZE 144
+
+#define DEG_TO_RAD 0.017453292519943
```
flixr: Why not use `RadOfDeg` from std.h that you have included already anyway?

sw/airborne/peripherals/vn200_serial.c
```diff
@@ -0,0 +1,101 @@
...
+          vnp->msg_available = TRUE;
+          vnp->counter++;
+        } else {
+          vnp->msg_available = FALSE;
+          vnp->chksm_error++;
+        }
+        vnp->status = VNMsgSync;
+      }
+      break;
+    default:;
```
flixr: extra semicolon

sw/airborne/peripherals/vn200_serial.h
```diff
@@ -0,0 +1,119 @@
...
+void vn200_event(struct VNPacket *vnp);
+void vn200_read_message(void);
+void vn200_parse(struct VNPacket *vnp, uint8_t c);
+
+/**
+ * Calculates the 16-bit CRC for the given ASCII or binary message.
+ * The CRC is calculated over the packet starting just after the sync byte (not including the sync byte)
+ * and ending at the end of payload.
+ */
+static inline unsigned short calculateCRC(unsigned char data[], unsigned int length)
```
flixr: why make these static inline functions rather than just normal static functions in the c file?

sw/airborne/subsystems/ins/ins_vectornav.h
```diff
@@ -0,0 +1,184 @@
...
+
+#if !USE_NPS
+#define ImuEvent imu_vectornav_event
+#endif /* Only for non-NPS targets */
+
+/*
+ * IMU scalling defaults - still needed, although not used
+ * Filled in with dummy values (ones)
+ */
+#if !defined IMU_GYRO_P_SENS & !defined IMU_GYRO_Q_SENS & !defined IMU_GYRO_R_SENS
```
flixr: not needed if you don't add imu.c and register the IMU msgs from this implementation.

flixr:
In general also please try to follow the naming conventions (lower_case names for variables)....

podhrmic:
Sorry for delay, I ll get you the fixes in next few days.

sw/airborne/peripherals/vn200_serial.h
podhrmic: Fixed

sw/airborne/peripherals/vn200_serial.c
podhrmic: Fixed;

sw/airborne/peripherals/vn200_serial.h
podhrmic: Moved to the c file. Motivation for inline is performance -> fewer function calls, although at 400Hz it probably doesn't matter that much.

podhrmic:
Take two, what do you think?

flixr:
I still think it makes more sense to not use the imu struct (and imu.c) in this driver at all and directly register the IMU messages from the vectornav driver... 
Or do you see a reason not to do that?

podhrmic:
I guess I still need the orientation conversion struct - for rotating from IMU to body frame, and since that is already properly initialized in imu.h/c it seemed to me easier to just reuse it, rather than duplicating it in the vectornav driver. What do you think?

flixr:
I think it would be nicer to do it in the vectornav driver to enable running vectornav alongside some other IMU/AHRS/INS and make it more future proof.

Basically add body_to_imu to the vectornav struct (like ins_invariant already has).
And since we want to have the possibility for multiple IMUs/INSs it would probably make sense to add new defines for INS_VN_BODY_TO_IMU (or so).

And only send the IMU_GYRO|ACCEL|MAG messages and skip the conversion back to fixed point... hm... or we want to have that conversion in order to send it via ABI, so maybe still convert for that?

Also it would be probably better to put the ABI bindings and periodic messages into the ins_vectornav wrapper like in ins_invariant...
And I think the wrapper should not register it as an AHRS...

Sorry for the many changes, but trying to get this into a shape where it will be easy to continue the separation/refactoring of the imu/ahrs/ins...

In any case I think it at least makes sense to get rid of the imu subsystem for this driver...

sw/airborne/subsystems/ins/ins_vectornav.h
```diff
@@ -0,0 +1,186 @@
...
+
+
+
+// Ins implementation state (fixed point)
+struct InsVectornav {
+  struct LtpDef_i  ltp_def; // initial position
+  bool_t           ltp_initialized; // status indicator
+
+  // output LTP NED for telemetry messages
+  struct NedCoor_i ltp_pos;
```
flixr: I don't see any particular reason to have these in the InsVectornav struct the way you use those right now...
IMHO it would make more sense to just grab those from the state interface (like you do already anyway) when actually needed in the `send_ins` function. That should even save some conversions...

Ideally you would also not grab those from the state interface, but directly convert the float values from the vectornav INS...
Since hopefully we will have proper support for multiple INS soon and the one sent to the state interface might not be the one from vectornav....
Unfortunately we don't have a component id (or something similar) in the messages right now, so this is still a problem if multiple components send the same message... but better prepare properly for this right now as it doesn't cost much...

podhrmic:
@flixr honestly that is a lot of work - even just removing the imu subsystem and having to redefine all IMU messages and the orientation conversion. I understand the motivation, but I simply don't have resources to do it in the foreseeable future. The driver works as is, so I guess it boils down to - do you want to merge it as is, or not have driver for Vectornav?

sw/airborne/subsystems/ins/ins_vectornav.h
```diff
@@ -0,0 +1,186 @@
...
+  // Packet data
+  struct VNPacket vn_packet;///< Packet struct
+  enum VNStatus vn_status;  ///< VN status
+  float vn_freq;            ///< data frequency
+  uint16_t vn_chksm;         ///< aux variable for checksum
+  uint32_t vn_time;          ///< VN time stamp
+  uint32_t vn_ltime;         ///< aux time stamp
+
+  // Auxilliary data fields
+  float timestamp; // System time [s]
```
flixr: would be nice to make all those comments doxygen comments like the ones aboce...

flixr:
it's not that much work removing imu from the driver... so here we go...
see #1318

Issue #1306, opened 2015-08-18
A.R.Drone 2.0  program.elf.respawner.sh and program.elf process not killed in time


OpenUAS:
Using PPRZ version: v5.6.0_stable-2-g7ce6e8a after an upload of default ardrone2 example the main ap.elf process starts but before these processes are killed:

* /bin/sh /bin/program.elf.respawner.sh -360p.slices 0 -live.tcp 
* /bin/program.elf -360p.slices 

Manually via telnet killing the processes and restarting the ap.elf all is OK
Using v5.5.2_testing ( 609d3a0) this behaviour was not observed.

NOTE: adding a  linux wait or waitpid could be of help

OpenUAS:
Message to self: in mcu_arch_init() the board_init() is after threaded inits and for ARDRone2, default Processes must be killed first(bettereven gracefully)

flixr:
thx, fixed with #1307

PR #1307, opened 2015-08-18
Quickfix for #1306 ARdrone2 ap.elf not starting

OpenUAS:
Quickfix for #1306 ARdrone2 ap.elf not starting. There was no discussion on the issue but will only impact the Parrot UAS.


flixr:
merged with a458414e6fcb6459d9bcc5c3cf781e37259adabf

PR #1309, opened 2015-08-24
[settings] implement settings_clear

flixr:
settings_clear should clear all persistent settings from flash/file.

Tested on STM32F4...


flixr:
Merging as it works on STM32F4 and should (tm) work on LPC21 and Linux...

PR #1310, opened 2015-08-25
IMU testing + test airframe for apogee

honzik666:
The airframe manifest + improved test programs allowed us verifying basic functionality of the board without jumping straight into the aircraft firmware - which maybe undecided for a particular user. 
Let me know if everything complies with paparazzi's standards or should be redone.

Jan

### DIFF
--- a/conf/airframes/examples/setup_apogee.xml
+++ b/conf/airframes/examples/setup_apogee.xml
@@ -1,0 +1,76 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<!-- this is a quadrotor frame equiped with Lisa/M and MKK motor controllers -->
+
+<airframe name="setup_apogee">
+
+  <firmware name="setup">
+    <target name="tunnel"           board="apogee_1.0"/>
+    <target name="usb_tunnel"       board="apogee_1.0">
+      <configure name="TUNNEL_PORT" value="UART6"/>
+    </target>
+    <target name="setup_actuators"  board="apogee_1.0">
+      <subsystem name="actuators" type="pwm"/>
+    </target>
+  </firmware>
+
+  <firmware name="test_progs">
+    <target name="test_manual"      board="apogee_1.0"/>
+    <target name="test_sys_time_timer"      board="apogee_1.0"/>
+    <target name="test_sys_time_usleep"     board="apogee_1.0"/>
+    <target name="test_telemetry"           board="apogee_1.0"/>
+
+    <target name="test_actuators_pwm"       board="apogee_1.0">
+    </target>
+    <target name="test_actuators_pwm_sin"   board="apogee_1.0">
+    </target>
+    <target name="test_baro_board"          board="apogee_1.0">
+      <subsystem name="imu" type="apogee"/>
+      <configure name="IMU_INIT" value="1"/>
+    </target>
+    <target name="test_imu" board="apogee_1.0">
+      <subsystem name="imu" type="apogee"/>
+    </target>
+    <target name="test_adc"                 board="apogee_1.0"/>
+    <target name="test_uart"                board="apogee_1.0">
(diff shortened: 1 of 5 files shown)


conf/airframes/examples/setup_apogee.xml
```diff
@@ -0,0 +1,76 @@
...
+
+  <servos>
+    <servo name="1"   no="0" min="950" neutral="1500" max="2050"/>
+    <servo name="2"   no="1" min="950" neutral="1500" max="2050"/>
+    <servo name="3"   no="2" min="950" neutral="1500" max="2050"/>
+    <servo name="4"   no="3" min="950" neutral="1500" max="2050"/>
+    <servo name="5"   no="4" min="950" neutral="1500" max="2050"/>
+    <servo name="6"   no="5" min="950" neutral="1500" max="2050"/>
+    <servo name="7"   no="6" min="950" neutral="1500" max="2050"/>
+    <servo name="8"   no="7" min="950" neutral="1500" max="2050"/>
```
gautierhattenberger: where is this servos ? there is only 7 servos on an Apogee.

honzik666:
Hi Gautier, I have reworked the commits with 7 servos only. Sorry for the typo, the reason was that I didn't want to change the setup_actuators.xml and other settings - otherwise the build process would complain due to hte mismatch between the settings and actual number of actuators/servos:
```
CC /home/.../paparazzi/var/aircrafts/braiins_setup_apogee/test_actuators_pwm/subsystems/actuators.o
In file included from test/test_actuators_pwm.c:31:0:
test/test_actuators_pwm.c: In function 'dl_parse_msg':
/home/.../paparazzi/var/aircrafts/braiins_setup_apogee/test_actuators_pwm/generated/settings.h:43:22: warning: array subscript is above array bounds [-Warray-bounds]
     case 7: actuators[7] = _value; break;\
                      ^
test/test_actuators_pwm.c:110:7: note: in expansion of macro 'DlSetting'
       DlSetting(i, var);
       ^
/home/.../paparazzi/var/aircrafts/braiins_setup_apogee/test_actuators_pwm/generated/settings.h:74:29: warning: array subscript is above array bounds [-Warray-bounds]
     case 7: return actuators[7];

```

gautierhattenberger:
true, this is a bit annoying, but I don't see any solution for this.
@esden @flixr should we keep the extra servo definition even if it doesn't exist on the board, or remove it and then the settings generate a file where you can go out of bounds ?

flixr:
Um... the setup_actuators.xml settings file should not even be used for the `test_actuators_pwm` target....

flixr:
I added the target to setup_actuators.xml in master, so it should only be loaded for that target...

honzik666:
@flixr You are right, the test_actuators_pwm.xml is the proper setting, I didn't mention. Does it make no harm if it specifies more actuators than the board actual defines and has?

flixr:
well, it will do harm if you try to change a setting for an actuator that you don't have since it will write over the array bounds...
So the correct way to deal with this would be to make a settings file with only the number of actuators you actually have...

honzik666:
Well, I will leave this up to you guys. The learning curve is really steep, so I would find least confusing having the setup_apogee airframe a bit more generous, so that a new comer doesn't have to inspect all these details. I was aware that I have 6 servo sockets + 1 extra servo signal, but the build would complain.

honzik666:
@flixr Any chance the test airframe for apogee + baro testing change would get merged ;-)? Or shall I redo any part of changes?

gautierhattenberger:
Since it is only test program, I guess we can take the risk to have a mismatch between the settings file and the airframe file. @flixr do you agree to merge like this ? I don't see any simple solution that wouldn't be really ugly.

flixr:
The easiest solution would be to use a setup_actuators.xml settings file with only 7 servos for the setup_actuators target....

PR #1311, opened 2015-08-25
Exif header on JPEGS

dewagter:


sw/airborne/modules/computer_vision/lib/exif/exif_module.c
```diff
@@ -0,0 +1,282 @@
+
```
gautierhattenberger: missing Paparazzi header

sw/airborne/modules/computer_vision/lib/exif/exif_module.h
```diff
@@ -0,0 +1,10 @@
+
```
gautierhattenberger: missing Paparazzi header

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -46,6 +46,10 @@
 #include "lib/encoding/rtp.h"
 #include "udp_socket.h"
 
+#ifdef JPEG_WITH_EXIF_HEADER
```
gautierhattenberger: better use `#if` even if it is probably useless ?

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -195,6 +199,13 @@ static void *viewvideo_thread(void *data __attribute__((unused)))
         sprintf(save_name, "%s/img_%05d.jpg", STRINGIFY(VIEWVIDEO_SHOT_PATH), viewvideo.shot_number);
         // Check if file exists or not
         if (access(save_name, F_OK) == -1) {
+#if JPEG_WITH_EXIF_HEADER
+          	double lattitude = 50.788851;
```
gautierhattenberger: not sure you really want to take pictures at the exact same point ;)

gautierhattenberger:
can you resolve conflicts with master ? otherwise it looks good to me.

flixr:
hm... I think it would have been nice to make libexif a submodule to not bload the repo even more...

dewagter:
First of all the lib is modified to not give embedded compiler warnings and
be much lighter. So we need a clone submodule.

Then if not all submodules are downloaded but would only be downloaded on
making the aircraft that would be make submodules have a  benefit. Now a
submodule takes more disc space, is slower, is not updated on git remote
update, requires 2 commits per change etc. For me they are the way to go
when you develop the submodule as well or plan on often using a new version.

If a submodule is better, I can still do that if you revert the merge.

flixr:
now it's too late anyway and it's already part of the tree...

PR #1313, opened 2015-08-26
Bebopfrontcamera

dewagter:
black and white streaming


flixr:
Nice, will be great to finally have access to the front camera!
But the bebop_front_camera.xml module (or rather the use of viewvideo) will conflict with the video_rtp_stream.xml.... so you can't have both at the same time.

dewagter:
The camera interface will still be updated hopefully with ISP processing of the image instead of user space processing, but this already works. Will update naming to make it hopefully compatible (code-wise)

dewagter:
Now you can have both at the same time. Takes quite some processing and bandwidth though. But it compiles and runs nicely on my desk.

gautierhattenberger:
not really the time to test it, but it looks good

sw/airborne/modules/computer_vision/lib/vision/demosaic.h
```diff
@@ -0,0 +1,173 @@
+#ifndef DEMOSAIC_H
+#define DEMOSAIC_H
+
+/**
+ * @brief Bilinear demosaicing
```
flixr: Where is this from? Did you guys write that from scratch?
Just want to make sure you didn't forget some license/copyright header here....

PR #1315, opened 2015-08-26
[flightplan] Navigation parameters belong in the airframe

dewagter:
Navigation parameters are aircraft depending.

We end up with many flightplans that only differ 1 number.

If this approach is good I can change more airframes.

Other option is a set of variables that can also be set/stored in a rotorcraft_performance_parameters.h file.


gautierhattenberger:
This is a good idea, I want to do this for a long time. But I think you can go further, with the default parameters in the navigation.h files and maybe have settings so you can change the value without reflashing.

dewagter:
like this?

conf/flight_plans/rotorcraft_basic.xml
```diff
@@ -3,6 +3,12 @@
 <flight_plan alt="152" ground_alt="147" lat0="43 33 50.83" lon0="1 28 52.61" max_dist_from_home="150" name="Rotorcraft Basic (Enac)" security_height="2">
   <header>
 #include "autopilot.h"
+#ifndef TAKEOFF_CLIMB_SPEED
```
flixr: These are not used anymore and can be removed again...

flixr:
Aparat from the now superflous defines in the flight plan (as already commented), it looks good to me...

hooperfly:
Simulation is not working for any of my aircraft. I tried adding the `NAV`  block to one of my airframes and it made it even worse:

```
+  <section name="NAVIGATION" prefix="NAV_">
+    <define  name="CLIMB_VSPEED" value="1.0" />
+    <define  name="DESCEND_VSPEED" value="-1.0" />
+  </section>  
```

hooperfly:
So I've discovered that I can make the simulation behave correctly as long as I have `ADAPT_THROTTLE_ENABLED` set to `TRUE` in the airframes. I see the problem in simulation when I set it to `FALSE`. Am I missing something else with respect to airframe configuration if I disable the adaptive throttle? I would like to not have to use the adaptive throttle if possible since I'm not carrying variable payloads at this point. Any ideas or thoughts are appreciated.

dewagter:
@hooperfly sorry for the trouble, but with the defines above the default `flare` block should result in a `vmode=climb` with negative speed.

You can adjust and verify this descend rate value via the settings.

It has flown many times already with this commit, so I'm lost to what the cause could be in your case.

dewagter:
Can you first of all check while in simulation in settings that `NAV` `nav_descend_vspeed` is set to a negative value: e.g. -1 (minus 1).

Then, when `ADAPT_THROTTLE` set to `OFF` make sure your `nominal_throttle` matches the value you see when hovering. If the value is too high and the integrator is low, you can make a MAV climb. I just tested this in simulation as well: (bebop -> set nominal_throttle to 0.8 while the sim only needs 0.25. Set `Ki` gain low to 10. Then during flare it climbs dramatically.

The only thing is: it's been like that for a while.

hooperfly:
Hi @dewagter,

I think you are correct in that the problem has been here for a while. I hadn't seen it because I was always using `ADPT_THROTTLE_ENABLED` set to `TRUE`. I think the simulation may not model the actuators as they really are, so that might be part of the issue. I reverted all my airframes back to using the adaptive throttle setting for now. I know that, in the field, I have disabled the attribute in the `Settings` panel and the aircraft seems to work as expected.

By the way, I like the cleanup and consolidation of the NAV attributes, much appreciated. I'll keep thinking and testing to see if I can understand better the difference between the simulation and the live aircraft behavior.

dewagter:
@hooperfly,

You can safely fly without `adapt_throttle` with no problem (if your weight
is not changing too much), just make sure your `nominal_throttle` is more
or less correct for your real airframe, and make the `Ki` gain in the
vertical loop large enough to cope with the remaining error. In order for
the simulation to work well too, you might need to change the
`nominal_throttle` for the simulator to work well. Since its a define it
could be added in the `<target ' defines with a different value for
simulator and for `ap`

-Christophe

hooperfly:
@dewagter I'm going to stick with the `adapt_throttle_enabled` for now. My challenge is to get the takeoffs cleaner/snappier with `adapt_throttle`. I'm going to go to the test field today and see if I get them to hop off the ground in a more predictable way. I'm flying 3 different airframes with different hardware, so it a bit of a combinatorial exercise. :) Might be nice to expose parts of the guidance_v_adapt parameters in the `settings` at some point.

PR #1317, opened 2015-08-27
[module] read in and decode stereo camera messages

kirkscheper:
Added a module to read in and decode messages from a stereo camera using a UART connection. Module is in sw/airborne/modules/sensors. 

Please let me know if I need to move, change update anything.

### DIFF
--- a/sw/airborne/modules/sensors/stereocam.c
+++ b/sw/airborne/modules/sensors/stereocam.c
@@ -1,0 +1,194 @@
+/*
+ * Copyright (C) 2015 Kirk Scheper
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+/** @file modules/stereocam.c
+ *  @brief interface to usb stereocam
+ *  In current implementation, messages limited to 256 characters
+ */
+
+#include "stereocam.h"
+#include "mcu_periph/uart.h"
+#include "subsystems/datalink/telemetry.h"
+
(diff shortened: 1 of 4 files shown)


kirkscheper:
Idea... I can also make the baud rate configurable in the airframe file. Might be useful considering the issues with the previous attempt to push to master.

gautierhattenberger:
that is indeed a good idea to be able to set both port and baudrate

flixr:
Would also make sense to not "hardcode" `SEND_STEREO` to true in the makefile...
Better define it to `TRUE` by default in the .c file and check it with `#if SEND_STEREO`
Or even better make it a flag that you can change at runtime...

Also please add the configure and define options to the documentation section in the module xml.

flixr:
And probably most importantly: what kind of "stereocam" is this module for??

kirkscheper:
The stereocam here is a custom design by Christophe de Wagter that is extensively used at the TU Delft. We have a couple of people working on this camera now in paparazzi, that's why I am trying to push this module now.

flixr:
Thx!
Would be really nice if you could document the configure and define options with their respective xml tags in the module xml.
That way they are properly used in the generated documentation.... (see e.g. the digital cam module):

sw/airborne/modules/sensors/stereocam.c
```diff
@@ -0,0 +1,204 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+/** @file modules/stereocam.c
```
flixr: path incorrect...

sw/airborne/modules/sensors/stereocam.h
```diff
@@ -0,0 +1,43 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+/** @file modules/stereocam.h
```
flixr: same as in c file, please correct to modules/sensors/stereocam.h

sw/airborne/modules/sensors/stereocam.h
```diff
@@ -0,0 +1,43 @@
...
+  uint8_t *data;
+} uint8array;
+
+extern uint8array stereocam_data;
+
+extern void stereocam_start(void);
+extern void stereocam_stop(void);
+extern void stereocam_periodic(void);
+
+#endif /* LISA_STEREO_H_ */
```
flixr: not really important, but since I noticed it: the define is `STEREOCAM_H_` not `LISA_STEREO_H_`

flixr:
Thx!

PR #1319, opened 2015-08-28
Implemented MILLIAMP_AT_ZERO_THROTTLE in current estimation

cfr34k:
This change allows to define MILLIAMP_AT_ZERO_THROTTLE, which is the current estimation when the motors are running, but zero throttle is applied (this is relevant for rotorcrafts). This makes it possible to calibrate for a change in the magnetic field caused by starting the motors.

By default, MILLIAMP_AT_ZERO_THROTTLE is set to 0, so there is no change needed for existing airframe configurations.

Additionally, the range of x, which represents the throttle command, is now limited to [0, 1] in both directions to prevent math errors.

### DIFF
--- a/sw/airborne/subsystems/electrical.c
+++ b/sw/airborne/subsystems/electrical.c
@@ -89,12 +89,12 @@ static struct {
 #define CURRENT_ESTIMATION_NONLINEARITY 1.2
 #endif
 
-#if defined MILLIAMP_AT_FULL_THROTTLE && !defined MILLIAMP_AT_ZERO_THROTTLE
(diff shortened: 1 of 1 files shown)


OpenUAS:
Nice, MILLIAMP_AT_IDLE_THROTTLE would be a more insightful name IMHO, even if it means your throttle is a Zero/minimum"

flixr:
Jep, was about to make the same comment as @OpenUAS

cfr34k:
Yes, that’s indeed more consistent with ..._FULL_THROTTLE. Renamed the constant as suggested.

hooperfly:
Nice! This is a good addition. I will roll it into my current multi-aircraft test cycle.

PR #1323, opened 2015-08-31
[module] digital_cam_video

dewagter:


gautierhattenberger:
looks good for me

conf/modules/digital_cam_video.xml
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+<module name="video_cam_ctrl" dir="digital_cam">
+  <doc>
+    <description>
+      Digital Camera Triggering of an embedded camera on a linux based autopilot.
```
flixr: Would be nice to make the first line of the description the name that is supposed to be displayed in the generated modules page. Then the detailed description below in the next line(s).

PR #1324, opened 2015-09-01
[fixedwing] energy_ctrl: use ABI instead of imu struct

flixr:
energy control should not read anything from the global IMU struct, as that is not always available (e.g. when using CHIMU and is supposed to go away anyway...)

So grab the data directly via ABI instead...


dewagter:
Looks good but I could not test-fly.

OpenUAS:
Superb, will testfly in weekend. Will be based on this https://github.com/OpenUAS/paparazzi/blob/master_with_ou_conf/conf/airframes/OPENUAS/openuas_moksha_etecs.xml

Issue #1145, opened 2015-03-24
run multiple AHRS/INS estimators in parallel


flixr:
The goal is to have the possibility to run multiple state estimation algorithms at the same time and switch the output (to the state interface) between then during runtime.

Basic switching to a secondary AHRS is already possible with #1130.
Moved the remaining TODO list to here.

Open Questions/TODO:
- [ ] implement switching for INS
  - [ ] deal with full INS (e.g. invariant) that also outputs attitude
- [ ] deal with the remaining "global" ins_reset_x functions (`ins_reset_local_origin`, `ins_reset_altitude_ref`, ...)
- [ ] get rid of `ins_periodic` for those impls which need it (probably when we start converting them to modules)
- [ ]  deal with multiple implementations registering the same telemetry message

gautierhattenberger:
Not closing this for now as a reminder, however the proper way to handle this needs to be clarified.

PR #1326, opened 2015-09-02
[ahrs] check if all AHRS are aligned

dewagter:
needed in several automatic flightplans
to be used in arming

but: is this the way to go?

### DIFF
--- a/conf/flight_plans/rotorcraft_basic_superbitrf_from_hand.xml
+++ b/conf/flight_plans/rotorcraft_basic_superbitrf_from_hand.xml
@@ -56,7 +56,7 @@ The goal of this flightplan is to have a safe, simple no-brainer flightplan for 
   <blocks>
     <block name="Init">
       <call fun="NavKillThrottle()"/>
-      <while cond="!GpsFixValid() || gps.pacc > 400 || !(ahrs.status == AHRS_RUNNING)"/>
+      <while cond="!GpsFixValid() || gps.pacc > 400 || !(ahrs_all_aligned())"/>
     </block>
     <block name="Geo init">
       <while cond="LessThan(NavBlockTime(), 8)"/>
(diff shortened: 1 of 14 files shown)


flixr:
Not quite sure if this is the way to go either...

Using this in a flight plan only works if you have an AHRS, so e.g. not for a full INS where you don't need an AHRS... (e.g. vectornav, xsens...)

To just check if the actually used atiitude estimate is valid (the primary AHRS is aligned) it is easier to simply use `stateIsAttitudeValid()`, which will also work for the pure/full INS solutions....

dewagter:
Yes we plan to use dual-imu more and more: e.g. external IMU + aspirin backup. Why did you make it otherwise ;-)

But if you really don't want this code in master, I can understand and you can throw it away, but ideally we should cherry-pick the output-enable for CHIMU.

flixr:
> Yes we plan to use dual-imu more and more: e.g. external IMU + aspirin backup. Why did you make it otherwise ;-)

Me too, but was just wondering if you _already_ use it regularly... since there are still some annoyances (#1145)

> But if you really don't want this code in master, I can understand and you can throw it away, but ideally we should cherry-pick the output-enable for CHIMU.

Well, I was still hoping that someone comes up with a nicer solution on how to register/enable/check multiple AHRS implementations in a way that also harmonizes with full INS solutions that don't have any separate AHRS functions.

Also the name is_aligned is maybe a bit too specific and doesn't exactly apply to all, what we actually want to know if the AHRSs are *ready*.
And we really want to know that not only for the AHRS, but also other subsystems/sensors.
I think for this we should start looking at how to check if all required systems are ready in a more generic way...

That is why I was asking if you are already using multiple AHRS *now* and you really need something like in this PR... if yes we can merge it for lack of something better right now. But in the longer run I don't think this is the way to go.

dewagter:
Yes, every subsystem should add itself to the generic Ready-list. And at
the same time critical things should also have a watchdog.

dewagter:
I saw you added the the CHIMU enable. So shall we delete this code or do you want to use it as base?

flixr:
Let's keep it open for now, maybe we still need it before we can come up with a better solution...

PR #1327, opened 2015-09-02
Qrcode

dewagter:
- zbar (as submodule)
 - tested scanner

### DIFF
--- a/conf/modules/cv_qrcode.xml
+++ b/conf/modules/cv_qrcode.xml
@@ -1,0 +1,55 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="qrcode" dir="computer_vision/qrcode">
+  <doc>
+    <description>
+      QR code reader
+    </description>
+  </doc>
+
+  <header>
+    <file name="qr_code.h"/>
+  </header>
+  <periodic fun="qrcode_periodic()" freq="2"/>
+  <makefile target="ap">
+    <file name="qr_code.c"/>
+
+    <!-- zbar -->
+    <configure name="EXTSW" value="$(PAPARAZZI_SRC)/sw/ext" />
+    <configure name="LIBZBAR" value="$(EXTSW)/libzbar/zbar-0.10" />
+
+    <file name="bch15_5.c" dir="$(LIBZBAR)/zbar/qrcode"/>
+    <file name="binarize.c" dir="$(LIBZBAR)/zbar/qrcode"/>
+    <file name="isaac.c" dir="$(LIBZBAR)/zbar/qrcode"/>
+    <file name="qrdec.c" dir="$(LIBZBAR)/zbar/qrcode"/>
+    <file name="qrdectxt.c" dir="$(LIBZBAR)/zbar/qrcode"/>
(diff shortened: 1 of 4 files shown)


sw/airborne/modules/computer_vision/qrcode/qr_code.c
```diff
@@ -0,0 +1,74 @@
+
```
flixr: plz add a copyright and doxygen header...

sw/airborne/modules/computer_vision/qrcode/qr_code.c
```diff
@@ -0,0 +1,74 @@
...
+#include "qr_code.h"
+
+#include "zbar.h"
+#include <stdio.h>
+
+void qrcode_periodic(void)
```
flixr: if the periodic is empty, simply remove it...

conf/modules/cv_qrcode.xml
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="qrcode" dir="computer_vision/qrcode">
+  <doc>
+    <description>
+      QR code reader
```
flixr: A bit more details on how to use it would be great...

conf/modules/cv_qrcode.xml
```diff
@@ -0,0 +1,55 @@
...
+    <file name="decoder.c" dir="$(LIBZBAR)/zbar"/>
+    <file name="img_scanner.c" dir="$(LIBZBAR)/zbar"/>
+
+    <!-- libexif flags -->
+    <define name="$(LIBZBAR)/zbar" type="include"/>
+    <define name="$(LIBZBAR)/include" type="include"/>
+
+    <!--define name="DEBUG_QR_FINDER" value="1"/-->
+
+    <flag name="CFLAGS" value="Wno-cast-qual"/>
```
flixr: This will disable all these warnings for the whole code as soon as you have this module loaded...

sw/airborne/modules/computer_vision/qrcode/qr_code.c
```diff
@@ -0,0 +1,74 @@
...
+  // extract results
+  const zbar_symbol_t *symbol = zbar_image_first_symbol(image);
+  for (; symbol; symbol = zbar_symbol_next(symbol)) {
+    // do something useful with results
+    zbar_symbol_type_t typ = zbar_symbol_get_type(symbol);
+    char *data = (char *)zbar_symbol_get_data(symbol);
+    printf("decoded %s symbol \"%s\"\n",
+           zbar_get_symbol_name(typ), data);
+
+#if PERIODIC_TELEMETRY
```
flixr: I guess this should be `#if DOWNLINK` instead...

conf/modules/cv_qrcode.xml
dewagter: any idea on how to do this for only the library? It has a massive amount of warnings.

flixr: The only sensible solution would be to compile it as as lib separately with these warnings disabled and then link that lib...

dewagter:
this would plug in #1328 with a cv_add(qr_scan); in the init.

PR #1328, opened 2015-09-03
Computer vision framework

dewagter:
instead of having many new copies of the same, a computer vision framework is initiated:

video_thread

which can run one or more computer vision modules (e.g. rtp_stream or qrcode or opticflow)

### DIFF
--- a/sw/airborne/modules/computer_vision/cv.c
+++ b/sw/airborne/modules/computer_vision/cv.c
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2015
+ *
+ * This file is part of Paparazzi.
+ *
+ * Paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * Paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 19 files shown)


sw/airborne/modules/computer_vision/cv.c
```diff
@@ -0,0 +1,50 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/**
+ * @file modules/computer_vision/cv.h
```
flixr: plz fix file path
also please make use two spaces indent

flixr:
I guess the viewvideo_nps.c could be removed (it doesn't do anything anyway) and the modules disabled for the nps target...

dewagter:
I would like to implement viewvideo_nps instead of removing it.

flixr:
even better, although I wonder a bit what the random image will be good for ;-)

podhrmic:
Maybe you can somehow link it with flight gear and get some kind of video from the "cabin"?

flixr:
Looks good to me, without having had the chance to actually test it....

dewagter:
@flixr obviously the test image is not very useful indeed, but allows checking of the computer vision routines in simulation. Note that the video_thread code from AP can also run in linux, showing your webcam image while simulating. Not sure how this should be done when you have 3 options for the simulation code: 1) (no vision in nps) 2) video_thread_nps (test image or flightgear/ROS/smartuav coupling), 3) webcam.

PR #1330, opened 2015-09-03
[state interface] pass basic types by value

flixr:
Makes calls like `stateGetAirspeed_f()` easier to use as they directly return the float value instead of a pointer that you have to dereference first....

IIRC my original idea was to make it consistent so that all state interface functions take/return pointers, but was the wrong decision...


gautierhattenberger:
Yes, I think it is a good start. We can probably merge this already and continue the rest later.

PR #1333, opened 2015-09-04
Board knows camera settings

dewagter:
@fvantienen this is a start of the idea we discussed

 - [x] put camera settings, format and filters into board files
 - [x] video_thread - define name="CAMERA" value"FRONT|DOWN"
 - [ ] make it possible to start 2x video_thread

@flixr @gautierhattenberger any idea is welcome

### DIFF
--- a/conf/airframes/BR/bebop_default.xml
+++ b/conf/airframes/BR/bebop_default.xml
@@ -33,11 +33,8 @@
       <define name="FILE_LOGGER_PATH" value="/data/ftp/internal_000"/>
     </load-->
     <load name="video_thread.xml">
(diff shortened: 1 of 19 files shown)


flixr:
Definitely sounds like a good idea, maybe the filters should not be part of the device?

flixr:
So it should work now to define `VIDEO_THREAD_CAMERA` to e.g. `bottom_camera` for ARDrone2 and Bebop.
Filters are not added/used in video_thread yet, so you still can't use it for the Bebop front camera (use the bebop_front_camera module instead).

flixr:
Ok, so added the filter, works for the bebop front camera now as well (although the subdev format and size of the debayered image are still hardcoded).
Also made the fps adjustable during runtime via settings.

PR #1340, opened 2015-09-10
 - added integer unit conversion coef attrib of msg members to pprz_m…

x75:
…sg python interface

needed the conversion coef in python land, copied the code for type attrib


flixr:
Thx!

PR #1354, opened 2015-09-15
outer loop INDI control

EwoudSmeur:
Outer loop INDI control.

Controls accelerations of the drone to counteract (aerodynamic) disturbances. This is a quick implementation that only works for the horizontal accelerations, but vertical can be added easily. 

Test flown with hover_z_hold

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
@@ -1,0 +1,228 @@
+/*
+ * Copyright (C) 2015 Ewoud Smeur <ewoud.smeur@gmail.com>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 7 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -0,0 +1,228 @@
...
+    struct FloatQuat q_yaw_diff;
+    float_quat_comp_inv(&q_yaw_diff, &q_yaw_sp, &q_yaw);
+
+    /* compute final setpoint with yaw */
+    float_quat_comp_norm_shortest(&q_sp, &q_rp_sp, &q_yaw_diff);
+  } else {
+    QUAT_COPY(q_sp, q_rp_sp);
+  }
+
+  QUAT_BFP_OF_REAL(stab_att_sp_quat,q_sp);
```
gautierhattenberger: missing include `stabilization/stabilization_attitude_ref_quat_int.h`

and new line at end of file

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -0,0 +1,228 @@
...
+  //Bound euler angles to prevent flipping and keep upright
+  Bound(guidance_euler_cmd.phi, -0.7, 0.7);
+  Bound(guidance_euler_cmd.theta, -0.7, 0.7);
+
+  stabilization_attitude_set_setpoint_rp_quat_f(in_flight);
+
+}
+
+//low pass the accelerometer measurements with a second order filter to remove noise from vibrations
+void guidance_indi_filter_accel(void)
```
flixr: Would probably make sense to use an already existing low pass filter from [sw/airborne/filters/low_pass_filter.h] (https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/filters/low_pass_filter.h)

EwoudSmeur:
the problem was not that I had to include stabilization/stabilization_attitude_ref_quat_int.h, but that if you build with euler than stab_att_sp_quat is nowhere defined... not really sure what to do about this

PR #1357, opened 2015-09-16
Python attitude ref models

flixr:
Python implementations of different (and improved) attitude referene models from Antoine.

There is also a a GUI to play around with the parameters: att_ref_gui.py

Tried to fix up the gui, but it produces garbage when both references are updated at the same time in concurrent threads (e.g. right after startup or after changing the setpoint)...
Maybe someone with more threading experience in python can look into this...


flixr:
Made a workaround: only compute one ref at a time...

PR #1360, opened 2015-09-17
[HITL] initial commit

podhrmic:
This is an initial commit for a new Hardware-In-The-Loop simulator. Not meant to be merged as is, I am looking for feedback and help with improvement.

It was verified in real flight that the simulation is close enough so it can be trusted (we are actually publishing a paper about it soon).

**Features:**
- simulates fixedwing flight with external INS (in this case Vectornav)
- allows full flight simulation on the actual hardware (closest to real flight)
- uses flight gear for visualisation
- saves flight data both in the Ground Station (telemetry low frequency data) and as a csv file (high-frequency data)

**Limitations:**
- fixedwing only
- hardcoded position & model settings
- has to be compiled separately (e.g. from Eclipse, not from Pprz center)
- also requires boost, which some might consider as a limitation

**TO DO**:
- [ ] integrate with paparazzi make process (so it includes the right flight plan data etc)
- [ ] add VTOL option (?)
- [ ] integrate Gaiga Environment simulator and Ivy messages listening (so you can control it easily)

As I said, feedback, suggestions and help implementing the last issue is very much appreciated!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,7 @@ paparazzi.sublime-workspace
 # /sw/simulator/
 /sw/simulator/gaia
 /sw/simulator/simhitl
+/sw/simulator/hitl/hitl.bin
 
 # /sw/supervision/
 /sw/supervision/gtk_process.ml
--- a/sw/simulator/hitl/Makefile
+++ b/sw/simulator/hitl/Makefile
@@ -1,0 +1,15 @@
+# HITL temporary makefile
+#  
+CXX		:= c++
+CXXFLAGS 	:= -lpthread -lrt
+INCLUDES	:= -I /usr/include/boost/ -I /usr/include/JSBSim/ -I ../../airborne/ -I ../../include -I ../nps
+OBJECTS		:=  
+LIBS      := -lboost_thread -lboost_system /usr/lib/libJSBSim.so
+SRCS		  := ../../airborne/math/pprz_geodetic_double.c ../../airborne/math/pprz_geodetic_float.c \
+			   ../../airborne/math/pprz_geodetic_wmm2015.c ../../airborne/math/pprz_algebra_double.c
+# 
+all: main.cpp
+	$(CXX) main.cpp ${SRCS} -o hitl.bin -std=c++11 ${INCLUDES} ${CXXFLAGS} ${LIBS}
+
+clean:
+	rm hitl.bin
--- a/sw/simulator/hitl/jsbsim
+++ b/sw/simulator/hitl/jsbsim
@@ -1,0 +1,1 @@
+../../../conf/simulator/jsbsim
--- a/sw/simulator/hitl/main.cpp
+++ b/sw/simulator/hitl/main.cpp
@@ -1,0 +1,120 @@
+/*
+ * Copyright (C) 2015 Michal Podhradsky, michal.podhradsky@aggiemail.usu.edu
+ * Utah State University, http://aggieair.usu.edu/
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+/**
+ * @file main.cpp
+ *
+ * HITL demo version
+ *
+ * @author Michal Podhradsky <michal.podhradsky@aggiemail.usu.edu>
+ */
+#include <algorithm>
+#include <cstdlib>
+#include <deque>
+#include <iostream>
+#include <list>
+#include <set>
+#include <boost/bind.hpp>
+#include <boost/shared_ptr.hpp>
+#include <boost/enable_shared_from_this.hpp>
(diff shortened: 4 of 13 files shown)


flixr:
IMHO it would make most sense to integrate this into NPS (essentially add a HITL option to NPS), so we can re-use the whole generation of sensor data (including noise/delay, etc), the flight gear viz and FDM abstraction....

Also I'm not so sure it really makes sense to transmit so much ("hardcoded") data back from the ap. Except for the actuator output that you need as input for the FDM.

podhrmic:
Hi Felix, yes I thought about using NPS with HITL, but there are some significant challenges regarding timing - synchronizing 2 asynchronous rx/tx tasks plus running the simulation at variable speed (plus data logging). But I am open to the idea.

The "hardcoded" data that come from the autopilot are to verify that the TX data = RX data, in practice you would run your autopilot with HITL module, and then optionally disabled it before flight (it doesn't cost much to send more data).

What would be a good way to integrate HITL makefile into paparazzi center? I tried to have it as another target using nps makefile as a reference, but I wasn't very successful. Any recommendations?

flixr:
For HITL I would just disable the variable speed feature...

> The "hardcoded" data that come from the autopilot are to verify that the TX data = RX data

IMHO this is not a good idea, as these things are rather specific to the airframe/configuration you use,
and I don't get why or what you want to verify here...

> in practice you would run your autopilot with HITL module, and then optionally disabled it before flight (it doesn't cost much to send more data).

I don't think this will work for other more standard configurations where you would normally read the IMU/baro/etc. directly... it happens to work for you since the vectornav provides everything you need and only reads a uart.

> What would be a good way to integrate HITL makefile into paparazzi center? I tried to have it as another target using nps makefile as a reference, but I wasn't very successful. 

Well, I think that would actually be the way to go if you don't want to integrate it into NPS (right away).
What problems did you have?

podhrmic:
OK:) so when I tried to modify NPS target, I had the problem that most subsystem makefiles had special options for NPS (like nps_sensors etc), but not for hitl target. So I had to add those to hitl target (and that means modifying a lot of makefiles). And then, HITL is in C++ but the NPS makefile I started from is set to use C - how would I change it?


----

> For HITL I would just disable the variable speed feature...

But then you will get a funky behaviour close to the ground, right? I didn't test it personally, but that is what I read in the comments.

> IMHO this is not a good idea, as these things are rather specific to the airframe/configuration you use,
and I don't get why or what you want to verify here...

Yes, they are specific. The idea is to verify that at time *t1* I send to the autopilot that it's roll=10deg, and I want to see that the autopilot got the message at time *t2* and the value of roll was close (or the same) as I sent. It can be optional or not included at all, I don't have a strong opinion about it.

> I don't think this will work for other more standard configurations where you would normally read the IMU/baro/etc. directly... it happens to work for you since the vectornav provides everything you need and only reads a uart.

Of course. There is only so much you can do without having a "fake" SPI and i2c signals on the autopilot. And that is probably not happening any time soon. Realistically - if you happen to use an external INS (or at least AHRS, like GX3) system, then you can use this HITL, that's it. It is like not everybody is flying with ArDrone, yet we still have all the computer vision code...

flixr:
I guess that we will have to update quite a few makefiles regardless of which solultion we take...

Regarding the variable speed, there are two different things:
- running the whole simulation faster/slower than real-time and even pausing it
- "automatically" lowering the dt for the FDM step size when you are close to the ground to prevent this funky behaviour

The former does not make much sense for HITL, while the latter is needed...

I wouldn't go as far as implementing fake SPI and I2C systems, but it should be possible to either send simulated IMU values and run AHRS onboard (if the freq is not too high) or directly set simulated ahrs outputs...
IMHO the (main/only) benefit of HITL is that you can test your actuator outputs... everything else is better tested in SITL....

podhrmic:
I ll keep digging in it and post some updates later this or next week.

The main point of HITL (at least that's how we use it) is to fully test the flight code (Radio, INS, telemetry, etc.) before the flight, so you can verify that the code won't break. Basically it is the only way to know that your plane will fly before actually flying it. Something that can be useful when applying for [COA](http://www.faa.gov/about/office_org/headquarters_offices/ato/service_units/systemops/aaim/organizations/uas/coa/) for example...

flixr:
Yes...
I meant that if you have sensors that are not easily emulated (because the are connected via I2C, SPI, CAN..) then then HITL doesn't help you that much, since you will be running different code.
Then the only thing you can test in HITL is radio/telemetry/actuators (to some limited extent)...

podhrmic:
@flixr @gautierhattenberger I made HITL configurable - so you can build it and launch from GCS, and it works for both rotorcraft and fixedwing (I tested it with Lisa MX).

You still need Vectornav subsystem to be able to use HITL (so you can send your position/attitude data to the autopilot over UART), and it sends data back to the computer over another serial port.

Is there still interest to integrate this feature? We are using it to test our avionics in simulated flight before we send it out to real mission, but unless you are using Vectornav it won't be that useful.

Thanks for your feedback.

flixr:
@podhrmic did you push your additional changes to make it configurable somewhere?

podhrmic:
Not yet, I wanted to know if you were interested into this feature before I put time into updating this pull request.

flixr:
I'm mainly concerned about the maintainability of this and the duplication of all the NPS/JSBSim code...

So if you made improvents in configuration and code de-duplication I wouldn't mind merging this if you can maintain it later on...

masierra:
@podhrmic  Any updates on this? HITL would be really useful.

podhrmic:
@masierra I have a functional paparazzi HITL we are using for testing here at AggieAir, but it is currently depending on boost libraries (mostly for timing and serial port handling), which is not desirable for mainline paparazzi (adding another dependency). It can be probably re-done using C++11 standard (although I am not sure about the serial ports), but I won't have time to do it before March.

flixr:
@podhrmic  any updates here?

podhrmic:
Work in progress - give me a few more weeks.

podhrmic:
FYI it is in progress, I am getting rid of the Boost dependencies.

flixr:
great!

podhrmic:
Good point, we should close this and keep the discussion in #1831

PR #1361, opened 2015-09-18
[generators] gen_flightplan: check waypoint altitude against SRTM

flixr:
Print a warning if a waypoint has an altitude below the ground altitude provide by SRTM (if available).

Could be extended to check if the alt specified in `vmode="alt"` is also ok according to SRTM data.
This requires knowing the waypoint pos in the `output_mode` function...

@gautierhattenberger maybe you want to improve this a bit?

### DIFF
--- a/sw/supervision/paparazzicenter.ml
+++ b/sw/supervision/paparazzicenter.ml
@@ -224,7 +224,7 @@ let () =
       tag#set_property (`BACKGROUND color);
       (color, tag))
       (* since tcl8.6 "green" refers to "darkgreen" and the former "green" is now "lime", but that is not available in older versions, so hardcode the color to #00ff00*)
-      ["red"; "#00ff00"; "orange"; "cyan"] in
+      ["red"; "#00ff00"; "orange"; "cyan"; "yellow"] in
   let tag_table = GText.tag_table () in
   List.iter (fun (_color, tag) -> tag_table#add tag#as_tag) background_tags;
   let buffer = GText.buffer ~tag_table () in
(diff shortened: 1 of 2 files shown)


fvantienen:
Seems like a good idea.. where does this idea come from? :stuck_out_tongue_winking_eye:

gautierhattenberger:
maybe you can do that without the global ref variable, but not sure it is worth the effort

flixr:
@gautierhattenberger I meant that maybe you want to look at how to add a check for alt vs. SRTM in vmode alt, as currently that is not done (as opposed the already existing alt checks) as that requires you to know the associated waypoint position of the vmode block....

dewagter:
@fvantienen :expressionless:

dewagter:
@flixr This is in any case a good idea!

flixr:
Merging as is... still would be nice to extend the altitude checks later...

gautierhattenberger:
This would be a bit complicated actually, since vmode is independent of the horizontal mode, so it can be used with attitude or heading modes without associated WP and in general patterns where the actual trajectory is close but not the point itself.
Also, with the alt attribute you can use C-like expressions (macros, functions, operations, variables,...) that the generator will not be able to understand.
Handling all the hmode cases in vmode is not really nice I think, and the benefit is low.

flixr:
Just thought we would add the SRTM alt checks everywhere where we currently also have alt checks against the ground_alt+security_height defined in flight plan...
One remaining option then would be to use the flight plan location as coordinates for the SRTM check in vmode if we want that added...

PR #1363, opened 2015-09-22
[detect_window] Compile warnings

dewagter:
@kirkscheper 
@guidoAI 

any objection to any of this?


kirkscheper:
I guess it depends on if it is planned to implement detect window size...

If not then its fine. I guess you can always include it later if someone decides its useful. I don't use it...

dewagter:
I presumed someone deliberately removed this code.

guidoAI:
Yes, faster.

G

Issue #1338, opened 2015-09-07
bluegiga has sufficient buffer space for an entire message?


dewagter:
@kirkscheper 

failed to debug this error. since current code did not work it was reset to TRUE

https://github.com/paparazzi/paparazzi/commit/83c2c1a7d0bd7d3bb6e85e2d35f374e7bc105065

kirkscheper:
@dewagter should be fixed with #1365

PR #1365, opened 2015-09-22
[datalink] updated bluegiga code - Implemented check if full message …

kirkscheper:
…fits in buffer, tried to use funciton inputs instead of global in case the plan is to define the periph somewhere else in the future, removed some funcitons that were not being used, moved some funciton definitions to .c. Tested with LadyLisaBartBlueGiga

### DIFF
--- a/sw/airborne/subsystems/datalink/bluegiga.c
+++ b/sw/airborne/subsystems/datalink/bluegiga.c
@@ -27,8 +27,8 @@
 
 #include "mcu_periph/sys_time.h"
 #include "subsystems/datalink/bluegiga.h"
-#include "mcu_periph/spi.h"
 #include "mcu_periph/gpio.h"
+#include "mcu_periph/spi.h"
 
 #ifndef BLUEGIGA_SPI_DEV
 #error "bluegiga: must define a BLUEGIGA_SPI_DEV"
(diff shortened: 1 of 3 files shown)


kirkscheper:
This addresses issue #1338

sw/airborne/subsystems/datalink/bluegiga.c
```diff
@@ -49,34 +49,71 @@ struct spi_transaction bluegiga_spi;
...
-static int dev_check_free_space(struct bluegiga_periph *p __attribute__((unused)), uint8_t len __attribute__((unused)))
+void bluegiga_transmit(struct bluegiga_periph *p, uint8_t data);
+void bluegiga_receive(struct spi_transaction *trans);
+
+// Functions for the generic link device device API
+// TODO: I tried to remove the hard coded bluegiga_p here but I clearly don't understand how the link_device calls are made
+static int dev_check_free_space(struct link_device *dev __attribute__((unused)), uint8_t len)
 {
-  return TRUE;
+  struct bluegiga_periph *p = &bluegiga_p; //(struct bluegiga_periph*)(dev->periph);
```
gautierhattenberger: why the cast is commented ? it's not working ?

kirkscheper:
Yea, in short the cast didn't work. 

It is a bit confusing where the DefaultDevice or DOWNLINK_DEVICE (from downlink.h) is set but I guess it is not set correctly. Do you have any idea where the DOWNLINK_DEVICE should be defined. I can of course define it in the bluegiga.h as the link_device from the local variable 'bluegiga_p.device' but not sure if that is the idea...

gautierhattenberger:
can you my fix: https://github.com/gautierhattenberger/paparazzi/commit/136969563a1422e426ce319526a29fc21dcd9e47 ?

kirkscheper:
I tried this as well, got a seg fault. I will ask @dewagter about this.

gautierhattenberger:
If you run your code with GDB, check that p->device.periph really points to the bluegiga_p structure, because it is this pointer that is passed to the generic link_device functions (see pprz_transport.c file for instance). That's why you should directly have your bluegiga struct pointer as first argument of the functions like dev_check_free_space (hence no cast needed, actually it was really wrong to do that here).

kirkscheper:
I also don't understand that. For example, autopilot.c:205 static void send_fp(struct transport_tx *trans, struct link_device *dev) calls pprz_msg_send_ROTORCRAFT_FP (in messages.h: 2738) which in term send the link_device not the periph to the functions like put_bytes.

I had already briefly checked with GDB, the pointer sent to the put_byte function was not the dev.

kirkscheper:
ah, I see in pprz_transport.c where it is dereferenced. but I still get seg faults. I will look into this tomorrow and post in more detail what is going on.

gautierhattenberger:
in messages.h, only functions from the generic transport interface (see transport.h) are used, and they require a `link_device` struct as second argument (in general). That's why you only have `dev` here.
But in the implementation of the transports (pprz_transport, xbee, ...) where the generic `link_device` functions are actually called, it is `dev->periph` that should be passed.
So dev->periph` should point to your actual device structure (`bluegiga_p` in your case). But in `put_bytes` it should be `bluegiga_p->device`, the generic interface.
I agree that all this stuff is a bit painful to follow...

kirkscheper:
I guess I had a typo somewhere before cause it works now... Better this way then.

PR #1366, opened 2015-09-22
[arch] Fix stm32 ADC bug

fvantienen:
This needs to be tested first, but it solves the problem with having to use ```ADC2_1``` if you want to use a higher ADC channel from ADC2. Same problem occurs with ADC3.


flixr:
Makes sense to me, but can't test right now...

gautierhattenberger:
sorry our boards are only using ADC1, so I can't really test it either

fvantienen:
Tested and works

PR #1369, opened 2015-09-30
[datalink] New: serial message interface for python

bartslinger:
This library can send and receive paparazzi messages over a serial data link, without using Ivy.

Subscribe to messages with the subscribe function. The callback gets a SerialMessage as argument, which will have the requested data in payload_items. Example:
`subscribe("DL_VALUE", myCallbackFunction)`

Send messages using the sendMessage function. Example:
`sendMessage("datalink", "SETTING", (setting_id, aircraft_id, setting_value))`

Takes control of the serial link so it cannot be used simultaneously with the GCS for example.

### DIFF
--- a/sw/lib/python/serialmessagelink.py
+++ b/sw/lib/python/serialmessagelink.py
@@ -1,0 +1,229 @@
+import serial
+import threading
+import sys
+from os import path, getenv
+import struct
+
+"""Import Paparazzi Message library"""
+# if PAPARAZZI_SRC not set, then assume the tree containing this
+# file is a reasonable substitute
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
A serial message interface without Ivy already exists (`sw/lib/python/serial_msg_interface.py`) and a PPRZ message parser is in `sw/lib/python/pprz_msg/pprz_transport.py`. The only missing thing in the existing code from what I can see is that a callback is triggered for any received message, there is no option to subscribe to a unique message.
I don't think it is wise to duplicate the code and functionality, so I suggest to add the subscribe function to the existing code or use it directly by manually filtering the incoming messages.

bartslinger:
I was not aware of its existence

PR #1370, opened 2015-10-01
Opticflow plugs into video thread

dewagter:
no need to redefine the video thread

this should become an example for video processing stuff:

 - plug into one of the video_threads
 - thread safe data copy to paparazzi
 - publish ABI

### DIFF
--- a/conf/airframes/TUDelft/airframes/ardrone2_opticflow.xml
+++ b/conf/airframes/TUDelft/airframes/ardrone2_opticflow.xml
@@ -28,6 +28,10 @@
     <load name="gps_ubx_ucenter.xml"/>
     <load name="send_imu_mag_current.xml"/>
     <load name="logger_file.xml"/>
+    <load name="video_thread.xml">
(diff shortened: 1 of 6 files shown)


flixr:
One of the video_threads?

dewagter:
good point: this means currently we can have opticflow bottom camera and other-computervision on the front via video_thread. After this pull request that is not possible anymore!?

Maybe first need the multi_video_thread....

flixr:
That is correct.

dewagter:
outdated

flixr:
Why is this outdated? I still think it is a good idea, we just need to support two threads in the video_thread module first...

flixr:
@dewagter do you already have updates for stuff like this? Running multiple video threads... running opticflow on different cams?

dewagter:
@flixr I do not have a multi-thread update yet

flixr:
Just wondering why you closed this...

PR #1374, opened 2015-10-03
refactor attitude reference

flixr:
Attitude reference generation implementations have no global vars themselves anymore...


flixr:
@gautierhattenberger @EwoudSmeur any comments on this?

EwoudSmeur:
I like that you can turn the attitude reference off!

flixr:
That part is actually exactly the same as before...

flixr:
So the main idea here was to implement the ref models as sort of "library" functions, where there are no global vars needed by other parts declared in the ref model code.
E.g. before indi also compiled stabilization_attitude_ref_quat_int.c, but all it needed/used from that was the `stab_att_sp_quat` without ever actually running the ref model.

On the potential flip-side of this is that there are no pure "interface" functions like `stabilization_attitude_ref_update` anymore that are shared across all implementations.
Instead you have to choose if/which ref model you use directly in the stabilization algo.

In general I would like to see more changes like this, meaning putting more things in structs and passing those as arguments to the functions instead of relying on global vars.
This also makes it a lot easier to use/test these functions in other contexts, like e.g. run different ref models for comparison and call them from python code, etc...

gautierhattenberger:
I didn't test it yet, but on the general idea it is fine for me

PR #1382, opened 2015-10-12
runtime adjustable attitude reference saturation

flixr:
Make attitude reference saturation runtime adjustable and add default saturation defines.


flixr:
@gautierhattenberger any comments/objections?

gautierhattenberger:
nope

PR #1383, opened 2015-10-13
Fixed all configs for OpenUAS

OpenUAS:
**Refreshed first batch of airframes...**more to come. Compiled AP, NPS, SIM, TUNNEL, FBW where appropriate. Not all testflown but under OpenUAS responsibility. Naming convention of UPPERcase for 3rd party config files an 3rd party organization. this way later on even duming e.g. airframe files in one tested directory will not gave naming conflicts and no owner conflicts.


conf/telemetry/OPENUAS/openuas_magnetodebug_ardrone2.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
```
flixr: why a second identical file?

OpenUAS: Thx for looking at it, THe answer is configs and values in it will be changed to other values asap. there is still a lot of stuff to be merged, If would have done it before a pullreqest, clean up would be postponed and postponed again... Now I have a "clean" start to merge all changes lurking in various branches before the end of the year, then with all in, a new pull request to be expected.

flixr:
You wrote "more to come", so do you want this merged as is, or update it in this PR?

OpenUAS:
Merge as is plz

PR #1385, opened 2015-10-15
[messages] support multiple variable length arrays in telemetry

flixr:
The `DL_<MESSAGE>_<field>` get macros should only be needed for receiving datalink messages on the airborne side.
So only generate those for messages of the datalink class...

This part currently can't handle multiple variable length arrays per message.
In principle it should not be a problem to have multiple variable length arrays in telemetry messages though...

### DIFF
--- a/sw/lib/python/ivy_msg_interface.py
+++ b/sw/lib/python/ivy_msg_interface.py
@@ -55,7 +55,7 @@ class IvyMessagesInterface(object):
             return
 
         # first split on array delimiters
-        l = re.split('([|\"][^|]*[|\"])', larg[0])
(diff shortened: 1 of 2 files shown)


flixr:
@podhrmic what parts of this did you test? E.g. did you also test the messages tool and the python implementations?
Does replay work?

podhrmic:
I made a telemetry message that has multiple array fields, and then I was able to receive in GCS using standard "messages".
Haven't tested replay or any of the python stuff.

flixr:
Fixed parsing in ivy_msg_interface, python messages now also work.

PR #1386, opened 2015-10-15
[gps] gps_piksi: report correct gps fix

masierra:
As the driver sits currently, it does not report the correct fix if it
had previously been RTK/DGPS back to 3D. 

- added get_fix_mode() to report fix based on current/last gps fix
- added timeout to report GPS_FIX_NONE when not receiving messages from piksi at all
- tested using piksi simulation

### DIFF
--- a/sw/airborne/subsystems/gps/gps_piksi.c
+++ b/sw/airborne/subsystems/gps/gps_piksi.c
@@ -48,6 +48,12 @@
 #include <libsbp/settings.h>
 #include <libsbp/piksi.h>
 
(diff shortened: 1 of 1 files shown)


flixr:
Thx, looks good to me.
@gautierhattenberger @fvantienen maybe you want to test this?
Feel free to merge it, I'm off to into the sun...

fvantienen:
Don't have any time to test soon, but this seems fine ;)
This was actually one of the things I still needed to change to make the Single point solution to work, but didn't had the time for it during IMAV. So thanks @masierra

PR #1388, opened 2015-10-19
[datalink] update bluetooth datalink code

kirkscheper:
- updated connection interval and connection speed parameters for coms speedup
- updated various support files for the bluegiga tools program
- added bluegiga debug message to messages.xml
- updated the computation of datalink speeds
- added field in datalink_msg to display total number of received messages per second
- various spell fixes


kirkscheper:
Can someone plz take a look at this before another merge conflict is caused cause the pull request is waiting here?

dewagter:
I do not understand the reason for the 3 empty files 0 4 and 8... 

but except for that it looks good to me and with little interference to other parts so no reason not to merge

kirkscheper:
done

PR #1390, opened 2015-10-22
Unified conf, airframe and simulator files for Hooperfly

OpenUAS:
Modified for more consistent naming for "3rd party" configuration and belonging to XMl documents like airframes or JSBSim aircraft. In the same run removed the execute bit from some PNG files :)

Test compiled all airframes for AP and NPS


podhrmic:
Please remind me again why are we having upper-case folder names now?

dewagter:
Is a personal folder also thrid party?

flixr:
Personally I'm not a big fan of UPPER_CASE file/directory names, but @OpenUAS suggested that all "third-party" and "personal" confs should be in UPPER_CASE to distinguish from the general/supported examples...
I guess it's not such a bad idea...

flixr:
Although I'm not sure why @OpenUAS added an `hooperfly_` prefix to all files...
@hooperfly what do you think?

hooperfly:
@flixr I'm not sure why `hooperfly_` was prefixed to the files? I think it might not be necessary. @OpenUAS is there a reason for the prefix? I was already isolating my development in `HooperFly` folders so changing the name to `HOOPERFLY` should be straight forward.

flixr:
ping @OpenUAS

OpenUAS:
hooperfly was added so we can still dum all the files in one big fat examples or whatnot directory without any naming conflicts etc.

hooperfly:
Hi @OpenUAS, I was wondering if simply renaming my directories in `conf/airframes` and `conf/flight_plans` from `HooperFly` to `HOOPERFLY` meets the needs of the unifying effort? Of course I will need to update any files that have `HooperFly` in the path name as well. It seems we need to resolve this for the 5.8 release. I was hoping to avoid the whole `hooperfly_` prefixes that were added on the branch. Thoughts? Why do we need a flattened name space model? If that's the case, we should abandon directories and put all of the file in one massive directory. It's a bit confusing and seems arbitrary.

hooperfly:
Running the following command agains the `airframes` and `flight_plans` directories gives a nice view of the name distribution:
```
cd $PAPARAZZI_HOME
tree -i conf/airframes | uniq -c
tree -i conf/flight_plans | uniq -c
```

OpenUAS:
Thanks. Yes, you see it right all clean up is is not finished. FYI  2 Years ago, I've agreed ad assigned to clean up the conf and related files (a bigger mess back then) and we agreed all to doing in the way it is changed now. I know it was before you joined and realize it can be a bit of getting used to. But indeed we could opt to change to a flatted model ;) and abandon directories and put all of the files in one directory again. All your airframes where compiled in AP an NPS all where working 100%

And yes there are still name collisions we are busy cleaning up as last also the TU Delft stuff but will before v5.8

OpenUAS:
on to TU Delft, where no doubt the cleaner will first get slapped, then honoured ;)

flixr:
I still don't see the necessity for the file name prefixes, otherwise I'm all for it...

PR #1393, opened 2015-10-28
Sending rotation and height to external device over UART

rmeertens:
As I and other people in the MAVLab are working with computer vision on devices attached to the drone it would be nice to send the rotation and height of the drone to these devices. I added a module that does this.


kirkscheper:
There is some double code here with what is in modules/sensors/stereocam.c and .h. I would be more useful to merge those files with the stereoprotocol files you made so we only have one protocol to maintain.

Also I am not sure if subsystems is the best place for the stereoprotocol files, however I may not be the best person to say where it should go. Maybe subsystems/sensors or modules/sensors.

sw/airborne/subsystems/stereoprotocol.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * stereoprotocol.h
+ *
+ *  Created on: Sep 23, 2015
+ *      Author: roland
+ */
+
+#ifndef SW_AIRBORNE_SUBSYSTEMS_GPS_STEREOPROTOCOL_H_
```
kirkscheper: This isn't the correct header format for this file location

sw/airborne/subsystems/stereoprotocol.c
```diff
@@ -0,0 +1,167 @@
...
+}
+
+//void stereoprot_get_msg_properties(uint8_t *, MsgProperties *, uint16_t,uint16_t);
+
+
+void WritePart(struct link_device *dev,uint8_t* code,uint8_t length){
+
+    for(uint8_t index=0; index < length; index++){
+        uint8_t toWrite = code[index];
+        dev->put_byte(dev->periph,code[index]);
```
kirkscheper: You should check that there is space in the transmit buffer before trying to write to it with 

dev->check_free_space(struct uart_periph *p, uint8_t len);

Currently in the uart_arch ofr stm32 if the buffer is full it simply returns and doesn't give any indication that there is something wrong.

void uart_put_byte(struct uart_periph *p, uint8_t data)
{
  uint16_t temp = (p->tx_insert_idx + 1) % UART_TX_BUFFER_SIZE;

  if (temp == p->tx_extract_idx) {
    return;  // no room
  }
...

rmeertens:
Fixed almost all remarks by Kirk. 
I would prefer to put the protocol in some lib directory, but I couldn't find a suitable one. Anyone else has a preference?

Also: unrelated to the stereoprotocol I fixed some optical flow issues, and drones can now use the pix4flow to hover.

fvantienen:
Some remarks:
- Adding a ```RM``` directory isn't needed, since you don't use any personal airframes. (You are part of the TUDelft)
- Maybe take a look at periodic functions in modules, because the way you implemented the ```write_serial_rot``` function and the ```SERIALRMAT``` message shouldn't be the way to do this. Please remove the telemetry message ```SERIALRMAT``` and chang the ```write_serial_rot``` to a periodic function.
- For the stereoprotocol add the files to the folder of your module . Next to that you can also reuse original PPRZ message parsers and use that protocol(as we use almost everywhere in the lab).

Next to that I will add some remarks in the code itself.

conf/messages.xml
```diff
@@ -2092,7 +2092,9 @@
   </message>
 
   <!-- 245 is free -->
-  <!-- 246 is free -->
+ <message name="SERIALRMAT" id="246">
```
fvantienen: Remove this message, this is misusing the telemetry messages.

conf/modules/uartrotation.xml
```diff
@@ -0,0 +1,22 @@
...
+  </doc>
+  <header>
+    <file name="uartrotation.h"/>
+  </header>
+  <init fun="uart_rotation_init()"/>
+  <makefile>
+    <file name="uartrotation.c"/>
+    <file name="stereoprotocol.c" dir="subsystems"/>
+	<raw>
+	ODROID_GPS_PORT_LOWER=$(shell echo $(GPS_PORT) | tr A-Z a-z)
```
fvantienen: GPS PORT?

conf/modules/uartrotation.xml
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="uartrotation">
+  <doc>
+    <description>Writes rotation to the uart bus</description>
```
fvantienen: Some more descriptive.. for Odroid only? which protocol?

conf/telemetry/default_rotorcraft.xml
```diff
@@ -26,6 +26,7 @@
       <message name="AIR_DATA"               period="1.3"/>
       <message name="SURVEY"                 period="2.5"/>
       <message name="OPTIC_FLOW_EST"         period="0.25"/>
+      <message name="SERIALRMAT"         period="0.25"/>
```
fvantienen: Remove please..

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -40,11 +40,19 @@ bool_t optical_flow_available;
 
 // request struct for mavlink decoder
 struct mavlink_msg_req req;
+#define PIX4FLOWSENDER_ID 1333
 
 // callback function on message reception
 static void decode_optical_flow_msg(struct mavlink_message *msg __attribute__((unused)))
 {
   optical_flow_available = TRUE;
+ // printf("Stabilisation decoded %d %d height: %f \n", optical_flow.flow_x, optical_flow.flow_y,optical_flow.ground_distance);
```
fvantienen: Remove commented out code

sw/airborne/modules/sensors/stereocam.c
```diff
@@ -44,25 +44,16 @@ struct link_device *dev = STEREO_PORT;
 #define StereoSend(_dat,_len) { for (uint8_t i = 0; i< (_len); i++) StereoSend1(_dat[i]); };
```
fvantienen: This is old and shouldn't be nedded anymore

sw/airborne/modules/sensors/stereocam.c
```diff
@@ -44,25 +44,16 @@ struct link_device *dev = STEREO_PORT;
...
 #define StereoUartSetBaudrate(_b) uart_periph_set_baudrate(STEREO_PORT, _b);
```
fvantienen: same here

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+#include "state.h"
+#include "subsystems/gps.h"
+#include "subsystems/ins/ins_int.h"
+#include "firmwares/rotorcraft/autopilot.h"
+#include "mcu_periph/uart.h"
+#include "subsystems/gps.h"
+#include "subsystems/datalink/telemetry.h"
+#include "subsystems/stereoprotocol.h"
+#include "navdata.h"
+int frameNumberSending=0;
```
fvantienen: Use static variables and preferable also structures.

sw/airborne/modules/uartrotation/uartrotation.h
```diff
@@ -0,0 +1,32 @@
...
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/readlocationfromodroid/readlocationfromodroid.h"
+ * @author Roland
+ * reads from the odroid
+ */
+
+#ifndef READLOCATIONFROMODROID_H
```
fvantienen: Not really a correct name

sw/airborne/modules/uartrotation/uartrotation.h
```diff
@@ -0,0 +1,32 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/readlocationfromodroid/readlocationfromodroid.h"
```
fvantienen: Wrong path + file

sw/airborne/subsystems/stereoprotocol.c
```diff
@@ -0,0 +1,164 @@
+/*
+ * stereoprotocol.cpp
```
fvantienen: Needs a correct header with copyright etc.

sw/airborne/subsystems/stereoprotocol.c
```diff
@@ -0,0 +1,164 @@
...
+ *
+ *  Created on: Sep 23, 2015
+ *      Author: roland
```
fvantienen: You're not the only one who made this...

sw/airborne/subsystems/stereoprotocol.c
```diff
@@ -0,0 +1,164 @@
...
+  if (stack[i] == 255 && (stack[stereoprot_add(i, 1,buffer_size)] == 0) && (stack[stereoprot_add(i, 2,buffer_size)] == 0) && stack[stereoprot_add(i, 3,buffer_size)] == 175) {
+    return 1;
+  }
+  return 0;
+}
+
+//void stereoprot_get_msg_properties(uint8_t *, MsgProperties *, uint16_t,uint16_t);
+
+
+void WritePart(struct link_device *dev,uint8_t* code,uint8_t length){
```
fvantienen: Choose better function names please.. this is to generic.

sw/airborne/subsystems/stereoprotocol.h
```diff
@@ -0,0 +1,60 @@
+/*
```
fvantienen: Same for header here

sw/airborne/subsystems/stereoprotocol.h
```diff
@@ -0,0 +1,60 @@
...
+#define SW_AIRBORNE_SUBSYSTEMS_STEREOPROTOCOL_H_
+
+#include <inttypes.h>
+#include "mcu_periph/link_device.h"
+struct MsgProperties {
+  uint16_t positionImageStart;
+  uint8_t width;
+  uint8_t height;
+} ;
+typedef struct MsgProperties MsgProperties;
```
fvantienen: We don't prefer typedefs.. next to that if you do use a typedef use it before the struct.

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#include "modules/uartrotation/uartrotation.h"
+#include "subsystems/abi.h"
+#include <serial_port.h>
+#include <stdio.h>
+#include <inttypes.h>
+#include <errno.h>
+#include "state.h"
+#include "subsystems/gps.h"
```
fvantienen: are all these includes needed????

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/readlocationfromodroid/readlocationfromodroid.c"
+ * @author Roland
+ * reads from the odroid and sends information of the drone to the odroid using JSON messages
```
fvantienen: JSON?

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+	uint8_t ar[lengthArrayInformation];
+	int32_t *pointer = (int32_t*)ar;
+	for(int indexRot = 0; indexRot < 9; indexRot++){
+		pointer[indexRot]=ltp_to_body_mat->m[indexRot];
+	}
+	pointer[9]=(int32_t)(state.alt_agl_f*100); //height above ground level in CM. 
+	pointer[10]=frameNumberSending++;
+	printf("Whoo sending serial\n");
+	stereoprot_sendArray( &((UART_LINK).device),ar, lengthArrayInformation, 1);
+	printf("Whoo sending serial2\n");
```
fvantienen: Remove debug code

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+#include "navdata.h"
+int frameNumberSending=0;
+static abi_event odroid_agl_ev;
+float lastKnownHeight = 0.0;
+int pleaseResetOdroid = 0;
+
+
+static void write_serial_rot(struct transport_tx *trans, struct link_device *devasdf) {
+	struct Int32RMat *ltp_to_body_mat = stateGetNedToBodyRMat_i();
+	int32_t lengthArrayInformation = 11*sizeof(int32_t);
```
fvantienen: why 11? you can just to sizeof(struct Int32RMat) which is more clear. Also make it static.

sw/airborne/modules/uartrotation/uartrotation.c
```diff
@@ -0,0 +1,65 @@
...
+	uint8_t ar[lengthArrayInformation];
+	int32_t *pointer = (int32_t*)ar;
+	for(int indexRot = 0; indexRot < 9; indexRot++){
```
fvantienen: Copying is not needed since the RMAT is already an array in the memory

fvantienen:
Please also squash your commits and use the codestyle formatter

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -40,11 +40,18 @@ bool_t optical_flow_available;
 
 // request struct for mavlink decoder
 struct mavlink_msg_req req;
+#define PIX4FLOWSENDER_ID 1333
```
gautierhattenberger: ABI sender id should be in subsystems/abi_sender_ids.h file and should be between 0 and 255

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -40,11 +40,18 @@ bool_t optical_flow_available;
...
 
 // callback function on message reception
 static void decode_optical_flow_msg(struct mavlink_message *msg __attribute__((unused)))
 {
   optical_flow_available = TRUE;
+
+  // Y negated to get to the body of the drone
+  AbiSendMsgVELOCITY_ESTIMATE(PIX4FLOWSENDER_ID, 0,
+                                  (optical_flow.flow_x/optical_flow.ground_distance),
+								  -1.0*(optical_flow.flow_y/optical_flow.ground_distance),
```
gautierhattenberger: please check indentation

gautierhattenberger:
The files involved in the stereo cam sensor are scattered  in different directories with name that don't always make sense. Would it be possible to gather them at the same place ? Like a `stereo_cam` folder, even if some files could be reused in other situations.

In general, why the standard PPRZ format is not used to encode and send/receive messages ? For the stereo cam, I can agree that a specific sensor use a specific protocol, but when dealing with a generic external CPU like Odroid, why not using regular messages ? We now provide a Python interface and a C library can be extracted from the airborne code (this last stuff might be a bit broken, but not that much).

flixr:
I agree with @gautierhattenberger that unless there is a really good reason to implement a separate protocol the standard paparazzi protocol should be used to send messages to other onboard computers (like the odroid).
Also the protocol should not be limited to UART...

Also it would be nice to split unrelated changes into several pull requests (px4fow and sending attitude to odroid).
And hopefully you can find a better name than uartrotation...

rmeertens:
Fixed the ABI sender ID. 

The reason that the PPRZ protocol was not used was because I did not know it existed. As this protocol was used on our stereo boards for the communication for a long time I decided to add it to paparazzi. 
I still like to switch to the PPRZ protocol, but this will come later (including Felix's suggestion to add other ports than UART). 

Also: @gautierhattenberger , what would you like to have in this stereo_cam folder? The stereocam.h and .h files that are currently in the modules/sensors folder, the uart rotation stuff I added and the protocols?

knmcguire:
This would be very useful for me as well. Right now I'm implementing optical flow and divergences for the stereocam, and it would improve the quality if I could send IMU values to the stereoboard too.

gautierhattenberger:
@rmeertens Yes, the folder should have all files related to the stereo_cam sensor: main code and header, protocol and imu data sending. As @flixr suggested, a better name should be find for uavrotation. I would even suggect to merge it in the main code since their is only one function.
Concerning communication with external CPU, we have better solution using the Paparazzi message system. You can find an example with the `meteo_france_DAQ` module with a bidirectional communication with a Mbed MCU.

OpenUAS:
@rmeertens as Gautier proposed and a quick look at 
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/meteo/meteo_france_DAQ.c

We know, implementing it in such a way sadly is work for you for now but likely will save work in the future. A renewed pull request after that is best. Will help to make it so. 

But first discuss it in gitter with core developers so it is clear that if implemented the next pull request can simply be honored quickly. Maybe something for Volker to fix.

rmeertens:
Moved the files to the stereo_cam folder, fixed the codestyle and tested my modules. In this pull request I am NOT going to add the PPRZ protocol. The code works right now and is useful for others. Although using the PPRZ format over serial is a great idea I would definitely have used it if it was better implemented (i.e. a git submodule that I could add to my other projects) and if I had known this before I implemented this. 
I would still like to use the PPRZ protocol for the communication between our stereo board and our drones, but I am not going to implement it in the coming month (also, @OpenUAS, implementing it is not really @Pold87 's task or expertise).

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -226,4 +226,13 @@
 #define IMU_MPU60X0_ID 16
 #endif
 
+#ifndef UART_ROTATION_ID 
+#define UART_ROTATION_ID 17
```
gautierhattenberger: This ID is not used (and I don't see where it should be used)

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -41,10 +41,17 @@ bool_t optical_flow_available;
...
 
+
 // callback function on message reception
 static void decode_optical_flow_msg(struct mavlink_message *msg __attribute__((unused)))
 {
   optical_flow_available = TRUE;
+
+  // Y negated to get to the body of the drone
+  AbiSendMsgVELOCITY_ESTIMATE(PIX4FLOW_VELOCITY_ID, 0,
+                                  (optical_flow.flow_x/optical_flow.ground_distance),
```
gautierhattenberger: indentation

gautierhattenberger:
We can probably merge this for know, but the uavrotation module is clearly not a long term solution. It shall be removed as soon as we can provide an easier to use API for messages.

PR #1399, opened 2015-11-02
[exif] negative lat/lon/alt

dewagter:
Not tested yet.
@stephen hulme does this work for you?


dewagter:
Could not find documentation, but I think it is done with LAT/LON_REF exif tags, which, if I can thrust google code examples would be 2 byte ASCII values.

Anyone feels like giving this a try in the south/west?

OpenUAS:
http://www.cipa.jp/std/documents/e/DC-008-2012_E.pdf

dewagter:
Rebased, ready to merge.

PR #1400, opened 2015-11-03
[fix] remove saturation information from rpm measurement

EwoudSmeur:
this removes the saturation bit from the rpm values for the bebop, with thanks to @fvantienen


flixr:
Should we also save the saturation information?

flixr:
Also some comments in the code on why that bit is removed would be helpful for the next guy reading it...

EwoudSmeur:
I added a comment. If anyone wants to use it, they can save it...

PR #1401, opened 2015-11-03
[datalink] Extended bluegiga spi message to 20 bytes

kirkscheper:
I finally identified and fixed the error of the missing first byte in the spi message. The dma buffer copies the first byte of the tx buffer into the spi peripheral immediately when the register for slave transaction command is called, that was not immediately clear. I added a comment to the spi_arch to reflect this fact.

After fixing this I found another small "feature" in the send_message function in bluegiga which improves the responsiveness of the tx flow.

I also added code for transactions with the spi3 slave peripheral, compiles but not tested.

I removed the software slave management code. As was implemented if spi2 slave management was set to software all spi slaves would also use software management...

Last and not least I added the firmware needed to run onboard the bluegiga module.

### DIFF
--- a/sw/airborne/arch/stm32/mcu_periph/spi_arch.c
+++ b/sw/airborne/arch/stm32/mcu_periph/spi_arch.c
@@ -100,7 +100,7 @@ struct spi_periph_dma {
   bool_t tx_extra_dummy_dma;       ///< extra tx dummy dma flag for tx_len < rx_len
   uint16_t rx_dummy_buf;           ///< dummy rx buffer for receive only cases
   bool_t rx_extra_dummy_dma;       ///< extra rx dummy dma flag for tx_len > rx_len
-  struct locm3_spi_comm comm;      ///< current communication paramters
+  struct locm3_spi_comm comm;      ///< current communication parameters
   uint8_t  comm_sig;               ///< comm config signature used to check for changes
 };
 
@@ -623,7 +623,7 @@ static void spi_start_dma_transaction(struct spi_periph *periph, struct spi_tran
    * To receive data, the clock must run, so something has to be transmitted.
    * If needed, use a dummy DMA transmitting zeros for the remaining length.
    *
-   * In the reveive only case (output_length == 0),
+   * In the receive only case (output_length == 0),
    * the dummy is used right from the start.
    */
   if (trans->output_length == 0) {
@@ -1204,7 +1204,6 @@ void process_tx_dma_interrupt(struct spi_periph * periph) {
 
 static void process_slave_rx_dma_interrupt(struct spi_periph * periph);
 
-
 // SPI arch slave init
 #if USE_SPI1_SLAVE
 #warning "SPI1 slave: Untested code!"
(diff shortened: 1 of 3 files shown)


flixr:
remind me again, but why should the bluegiga_usb_dongle firmware (including stuff like bleupdate-1.3.6.exe) be part of the main repo?

flixr:
Also why was `SPI2_SLAVE_NO_NSS` added in the first place? What was this needed for?

kirkscheper:
Hey felix,
Tnx for the quick response.

I think the slave no nss was first implemented cause the first Bluetooth
module we tested couldn't toggle the slave select pin but now we fixed that
(a hardware issue). So no longer needed. And I guess it is generally not
needed.

I figure that software control may be interesting to someone but in its
current implementation it is not suitable if you have multiple slaves. I
can put the implementation back but commented if you like, as a reference
for future development.

The Bluetooth dongle has two modes, long story short, we run it in the mode
where a program is running onboard to handle the datalink forwarding as
well as some more functionality I am implementing now (multinode Bluetooth
broadcasting without dedicated connection, rssi recording, ect.). The
driver in paparazzi won't work unless you flash the Bluetooth modules with
this firmware (custom code I wrote).

The .exe is needed if you want to flash the Bluetooth firmware. You have to
make an account at the company that supplies the software to get a copy (a
but annoying). It only works for windows (they say there is some Linux
version somewhere but is experimental, don't know too much about it) and
the company doesn't provide the source code so I figured its not really
suitable for the /ext folder. In the end I figured it was useful to have
everything you need in one place as one won't work without the other.

If you prefer I can remove the .exe and put the link to download it in the
readme. However, like I mentioned, anyone who wants to download it will
have to make an account.

Further more if you don't want the firmware here at all then @dewagter will
need to advise me on where it should be stored.

Let me know on these two issues!

kirkscheper:
Also... If I remember correctly the .exe doesn't work properly with wine and may brick the module... So you really need windows.

II should probably extend the readme

flixr:
I'm totally fine with removing the software slave management for slave mode.

But I really think that the bluegiga dongle firmware and tools should go into a separate repository...

dewagter:
@kirkscheper maybe we can put the bluetooth firmware in here: https://github.com/tudelft/bluegiga

I would keep the blueiga_usb_dongle in paparazzi as this is needed in order to get telemetry and is paparazzi specific.

This way we can quickly accept this pull request.

OpenUAS:
I concur with CDW

flixr:
We could also make a repo in the paparazzi organization for the bluegiga firmware. But should be in a separate repo in any case.
And then plz remove those commits from this pull request before merging it....

kirkscheper:
firmware now at https://github.com/tudelft/bluegiga

flixr:
Could you plz rebase your commits on top of master and leave out the firmware commits, otherwise the firmware will end up in the tree after the merge (even if it was removed again).

dewagter:
Can I merge-squash this directly on master?

kirkscheper:
ok, never tried rebasing before, trying something now... gime a sec

flixr:
I guess merge-squash is not that nice, since you it makes sense to keep the individual that fix different problems...

kirkscheper:
... I think I may have broken it... semaphoreci failed at a really weird place. You ever seen this before?

flixr:
The build fail on semaphore is unrelated, seems they are having trouble with their apt mirror....

flixr:
However you now seem to miss some stuff that was previously in your branch. You might want to check my spi_slave branch

flixr:
@kirkscheper if you reset your spi_slave branch to mine and push again we can easily merge this and have it tracked properly...

kirkscheper:
tnx

PR #1403, opened 2015-11-04
radio-control signs conform paparazzi 

dewagter:
Values on the transmitter screen are the opposite of paparazzi. It should be:

1 - throttle
2 - X (roll right positive)
3 - Y (pitch up positive)
4 - Z (yaw right positive)

Nowadays in 2015 every radio can reverse channels: why have minus signs everywhere in paparazzi.

### DIFF
--- a/sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
+++ b/sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
@@ -35,6 +35,8 @@
 // for timer_get_frequency
 #include "mcu_arch.h"
 
+INFO("Radio-Control now follows PPRZ sign convention: this means you might need to reverese some channels in your transmitter: RollRight / PitchUp / YawRight / FullThrottle / Auto2 are positive deflections")
+
 // for Min macro
 #include "std.h"
 
--- a/sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.h
+++ b/sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.h
@@ -65,7 +65,11 @@
 /* reverse some channels to suit Paparazzi conventions          */
 /* the maximum number of channels a Spektrum can transmit is 12 */
 #ifndef RADIO_CONTROL_SPEKTRUM_SIGNS
-#define RADIO_CONTROL_SPEKTRUM_SIGNS {1,-1,-1,-1,1,-1,1,1,1,1,1,1}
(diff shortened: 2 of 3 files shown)


OpenUAS:
Thx, yes merge plz  I'm all for it... cleaning up the PPRZ steps at the time

flixr:
This will however mean that for people using spektrum (without having `RADIO_CONTROL_SPEKTRUM_SIGNS` defined in their airframe file), the radio control will suddenly stop working properly...

dewagter:
I know, that is why I make a public pull "request".

 - the longer we wait, the more transmitters need to be changed. Ideal was
to do this right from the start.
 - best is to implement this as close as possible to a release so it can be
communicated.
 - keeping the default as-is does not make things better.

Just say to everyone: check on your transmitter screen: right/up are
positive, just like everywhere else in paparazzi.

-Christophe

gautierhattenberger:
I'm not using spektrum, but I tend to agree with @dewagter, better to fix conventions and a release is a good opportunity.

flixr:
So shall we merge this right now or just right before we do the v5.8 release (where we can announce this)?

dewagter:
I'm fine with now, but would add a warning when the affected files are used.

flixr:
Add a warning? How?

dewagter:
Here it is.

flixr:
Hm.. but now you get this warnin _every_ time.... including in all the CI test output...

dewagter:
Changed warning to INFO message.

gautierhattenberger:
Ready to merge ?

flixr:
OK to merge for me. When you merge please also post this change on the
mailing list.
Am 10.11.2015 4:55 nachm. schrieb "Gautier Hattenberger" <
notifications@github.com>:

gautierhattenberger:
@dewagter this was my question, what do user have to change exactly (I don't use spektrum) ? Only reverse some channel on the transmitter ? Every one is impacted or not ?

dewagter:
@gautierhattenberger 

seems to be accepted better than expected: http://doodle.com/poll/fsynmsvhw8yw2f26

Users of Spektrum and superbitRF have to reverse roll, pitch, yaw and channel 6. Throttle and channel5=mode were good already.

Alternatively if they do not accept to do this, they can define RADIO_CONTROL_SPEKTRUM_OLD_SIGNS in their airframe file. 

Best practice: check on your transmitter screen that right/up/right are positive (large pulse)

PR #1404, opened 2015-11-04
use UDP telemetry in NPS sim

flixr:
Get rid off all the datalink parsing "hacks" and use the actual parser from the firmware. That includes that the datalink parse functions of modules are run...

This however means that you need to run link with the -udp option to use the NPS sim.

So I guess we should find a nice way to also automatically start link with the simulation session (but only for type NPS).


flixr:
I also added parsing of DL_SETTING messages directly via ivy again, so that you can turn the `datalink_enabled` setting back on.
What is missing there is a check that only allows you to change that option and none of the others.
@gautierhattenberger do you have a good idea on how to get the settings index for that specifc setting?

flixr:
Or we create another ground message to set all the NPS options and we can completely get rid of the nps settings "hack"....

gautierhattenberger:
we are generating a SETTINGS_NAME macro in settings.h, so can init a table with that and search for the index of the settings you need.

flixr:
When I have some time I'll look into checking the settings names, but I guess that doesn't have to be the focus now... was working like this so far as well...

Any ideas on how to best also start link for NPS sim sessions? Create a separate NPS session?

flixr:
When starting a simulator session with NPS, link should now be started as well...

@gautierhattenberger @dewagter any comments/objections? Otherwise I would like to merge this.

dewagter:
I like the idea that a simulation is as similar to flight as possible: so no problem starting link (and being able to simulate link loss as well).

Probably needs some mention on the mailing list as well.

flixr:
So you need the added `-udp_broadcast` option as well if you simulate multiple aircrafts. Then it work on Linux...
For OSX you also need https://github.com/paparazzi/paparazzi/commit/545883421c0bf10e4798cb5a5fa52d2b9531b482 to be able to listen on the same address/port
Tried to set the default broadcast address for OSX to `224.255.255.255` (just like for Ivy) with https://github.com/paparazzi/paparazzi/commit/fc910c956ab4d8f1e498351436bc6e6372ef4fa2 but that doesn't seem to do the trick.
However using the broadcast address of your actual interface seems to work...

hooperfly:
For Mac OS X, the `-udp_broadcast_addr` argument is needed for multiple aircraft simulation. Reference the wiki for details:
https://wiki.paparazziuav.org/wiki/NPS#Simulating_Multiple_Aircraft

hooperfly:
Tested with Mac OS X `El Capitan` -- Working just fine. Ship it. :+1:

Issue #940, opened 2014-11-11
improve and properly use module dependencies


flixr:
We already have a basic dependency mechanism for modules, but it needs to be improved and used properly.

Right now you can depend on a list of other modules (either using the name or xml file name) or specify a conflict with another module.
This is not quite enough... sometimes we need to have a dependency on at least one module "providing" a specifc functionality/interface.

E.g. for the digital_cam modules we have several implementations:
- digital_cam.xml (using gpio, should probabyl be renamed to digita_cam_gpio.xml anyway)
- digital_cam_uart.xml
- digital_cam_i2c.xml
- digital_cam_servo.xml

The proposed digital_cam_shoot_rc.xml needs to depend on one of these modules. So either we have to set the name of all of those to `digital_cam` so it doesn't matter which one you have added or add the possibilty to specify several different options with some "logical OR" in depend.
I guess it would make sense to allow multiple `depend` nodes and add the possibility to specify OR dependencies similar to debian depends like `<depend require="digital_cam_uart | digital_cam_servo | digital_cam"/>`

Another alternative would be to add something like a `provides="foo"` attribute to modules so you can specify an "abstract" fuctionality/interface that it provides and that other modules can depend on.

Also if a dependency is not satified it currently only prints a warning, when used properly everywhere we should throw an error...
##

flixr:
We should probably also consider ordering the init functions according to the dependencies (first init modules that are dependencies of others).
E.g. the init order is currently crucial for modules calling `cv_add` in init, see also #1539

gautierhattenberger:
there is a working dependency for modules now #2669

PR #1405, opened 2015-11-04
select an optional calibration xml in paparazzi center

dewagter:
Example use case:

10 bebop / ardrones with same firmware, but different calibrations and IP...
 - now have only 1 airframe file
 - put calibrations and IP into the parameters.xml which is appended to the airframe.h

this should allow to reduce the number of almost similar airframes significantly


OpenUAS:
Adding module selection for overnext release wishful

gautierhattenberger:
how is this different than including a calibration file based on AC id ? like https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/ENAC/quadrotor/ard2_basic.xml#L43

dewagter:
It's almost the same except for a few things:

1) you can see from paparazzi center which files are used and quickly swap
2) with the aircraft id include trick you are not free to choose your
calibration names. We have ladybird 1 2 3 and bebop 1 2 3 etc and I have no
idea about their ac/id
2b) bebop ac-id 34 and bebopINDI 35 and bebopoptitrack 36 all use the same
calibration...
3) last but not least: this should make SAVING gains possible as you just
enumerate over the files while includes are inserted in the main file.

I also wanted to implement module enable/disable and gain saving but my
ocml is not very fast so any help or correction is hugely appreciated...

gautierhattenberger:
I agree to merge this for the next release, but I think changing the overall configuration should become a top priority for later versions.
@flixr do you agree ?

flixr:
I'm not a big fan of merging stuff that strongly impacts user side configuration when we know that we will change it for the next release....

dewagter:
@gautierhattenberger Thank you for the feedback. I do wonder however: How would you like the new overall user configuration to be in a new release? What do you have in mind with "the overall configuration"?

@flixr why would this "strongly impact" user configuration?

gautierhattenberger:
I don't have a clear solution yet. I mostly want to avoid the scattering of xml files and produce a unique file that contains all information for parsing/saving/running softwares.

flixr:
Well, suppose someone uses this new parameters feature and we change it in the next release the whole conf.xml might not be valid anymore...
It just sounded like @gautierhattenberger already had some ideas on how to handle this differently for the next release.
So I guess this depends whether we already know that this will be an intermediate solution or not...

Also something to keep in mind: this will further complicate running the server on another host and serving the files per HTTP.
IMHO this would also need to be added to the CONFIG message and handled correctly.

dewagter:
@flixr If you have any suggestion on how to reduce double files bebop1/ bebop2/ bebop3/ bebop_optitrack/ bebop_opticflow/... your suggestions are very welcome. And I'm also interested to know more about this CONFIG message and http server.

flixr:
I'm not saying that we shouldn't merge this or that I have a better idea... just that if we (@gautierhattenberger ) already know that this is not a longer term solution and will be changed in the next release again, then it would make sense to consider this already. But since @gautierhattenberger already said that he also doesn't have a clear idea yet, that point is a bit moot.
Just wanted to avoid that we add something if we already know that we will change it for the next release again and causes upgrade pains.

Some information on the CONFIG message is in the wiki: https://wiki.paparazziuav.org/wiki/Server#Running_server_and_GCS_on_different_hosts
Basically it contains the URLs of the xml files to fetch

gautierhattenberger:
Compatibility is indeed an issue. For most user, we could do something to automatically upgrade the conf file, but for devs that switch branches more often it will be a pain.

dewagter:
@gautierhattenberger I do not see any conf upgrade issues nor dev switch pains:

If you open a conf file without parameters tag, there is no problem: you just have empty parameters. And if you open a conf file which DOES have parameters in an old branch which does not support parameters, there is also no error, but the parameters tag will be removed again from the conf on closing paparazzi center. This could only be annoying if were using parameters a lot. Note however that if you commit your conf files, then you can also swap branches as often as you want and go from with to without with no problem.

gautierhattenberger:
if the new format can read the old one and then save the file with a new xml structure not matching with the previous version, when going back to the old code, your conf will be mostly lost. Of course, keeping two conf file using the symbolic links solves the issue if you do it carefully.

but this doesn't change the main question: do we add this to the next release ?

gautierhattenberger:
@dewagter an other idea, less flexible but also less intrusive for now: we already have the $AC_ID keyword and I understand the limitations, especially regarding the name of the included files. What about also have $AC_NAME ? Then files and the conf.xml entry still have to match, but the included file names are more related to the actual aircraft name.

flixr:
Wie also already have $AIRCRAFT or so...
Am 12.11.2015 18:29 schrieb "Gautier Hattenberger" <notifications@github.com

gautierhattenberger:
true, it's actually `@AIRCRAFT`, I forgot this one...
But the remark stays the same.

dewagter:
I see this is not a very popular one.

I still think after working with is for a week that causes way less pain than any message above anticipates (for me it was zero pain since I have committed confs, and for people not using parameters also zero.).

I would still like to gather your ideas on how to proceed in the future since followng real pains are now not reduced:

 - saving on a file with includes puts the include in the file (which was the main motivation for this pull request). If paparazzi center knows the name of the include: it can save in both files.
 - we have many double airframe files with only tiny differences. When the master copy gets updated, the "copies" often miss that update and get outdated. Most users of ardrone/ladybird/bebop/rollingspider will just want to make a copy of the airframe file to change only the calibration. We need a framework where they can use the up to date (non-copied) original airframe file and just select the file with their gains.
 - something which was not jet solved by this pull request but I was wanting to add if this idea was accepted is to move modules also to conf. We already have this for module settings, but for modules themselves this would also allow to highly reduce the amount of copied airframe files (with just one module difference)

gautierhattenberger:
I'll keep that in mind when think of a new configuration design

dewagter:
Also flightplans should be able to include modules:

 - why have nav_line, nav_circle, nav .... when you don'y use them, of why does do you need to change your airframe to select another flightplan.

Would also save on double airframe: an vision_avoid flightplan could include the vision module. no need for an extra airframe file.

flixr:
That is something that we should consider when improving the dependency mechanism (#940)....
Still, in quite a few cases you still need to add the modules per airframe, as you need to set the correct configuration for it...

Issue #1209, opened 2015-06-18
add copy AC menu item to paparazzi center


flixr:
I think it would be nice to add a "A/C->Copy" menu item to the Paparazzi Center menu additionally to "A/C->New".
Often one starts with one of the existing airframes anyway and goes on to adapt it.
So it would be desirable to have a "copy" option that creates a new A/C entry in conf.xml with the same initial xml files as the currently selected one. Preferably there would also be a dialog to ask if the airframe file of the chosen aircraft should be copied and saved under a new name.

TobiasMue:
great idea!

agressiva:
+1

OpenUAS:
Only if copy is copy it will be handy IMHO, if not the Airframe XML is copied it is more like an alias an will give much confusion(based in hands on with +100 different students using Paparazzi)

OpenUAS:
Squeezing it in for the overnext release because of positive response and usefulness

PR #1406, opened 2015-11-05
[paparazzi center] add copy A/C menu entry

flixr:
Start to solve #1209

Does not yet ask for a new airframe file name and copy the airframe file.

### DIFF
--- a/sw/supervision/paparazzicenter.glade
+++ b/sw/supervision/paparazzicenter.glade
@@ -129,6 +129,18 @@
                       </widget>
                     </child>
                     <child>
+                      <widget class="GtkImageMenuItem" id="menu_item_copy_ac">
+                        <property name="label">gtk-copy</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="use_action_appearance">False</property>
+                        <property name="use_underline">True</property>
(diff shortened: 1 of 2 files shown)


flixr:
There doesn't seem to be an OCaml function to copy a file....?

dewagter:
Nice feature! Use this very often.

gautierhattenberger:
It doesn't seems to exist. Here is an example on how to make it by hand: https://ocaml.github.io/ocamlunix/ocamlunix.html#sec33

flixr:
Yeah, but further up in `write_conf_xml` we already use
```Sys.command (sprintf "cp %s %s"  Utils.conf_xml_file Utils.backup_xml_file)```
, so I guess we can just do the same for this here... although this is not so nice in terms of portability...

gautierhattenberger:
indeed...

gautierhattenberger:
If someone can test this before merging

dewagter:
Just tested: works fine for my.

(although the first time my cursor could not be seen and I could not type in the airframe.xml selection dialog, but since I can not reproduce this anymore.)

flixr:
Works nicely for me. Thanks @gautierhattenberger for fixing the warning and adding the copy of the airframe file.

PR #1407, opened 2015-11-05
[mavlink] Fly with PPRZ-Mavlink direct-AP APP

dewagter:
Direct bebop/ardrone/udp mavlink to drone link without need for paparazzi server.

Example: LS/conf -> BebopMavlink : try the nps target first

For now the procedure is:
 - Put the Phone/tablet IP in the airframe file mavlink module
 - make sure drone/nps and phone are in the same network (can both ping each other)
 - Install: https://play.google.com/store/apps/details?id=com.pprzservices
 - Install: https://play.google.com/store/apps/details?id=com.gcs

Note: the planning is to merge the 2 apps into 1.

### DIFF
--- a/sw/airborne/modules/datalink/missionlib/waypoints.c
+++ b/sw/airborne/modules/datalink/missionlib/waypoints.c
@@ -1,0 +1,239 @@
+/*
+ * Copyright (C) 2015 Lodewijk Sikkel <l.n.c.sikkel@tudelft.nl>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+/** @file modules/datalink/missionlib/waypoints.c
+ *  @brief Improvement of the missionlib implementation of the waypoint protocol,
+ *  truly global waypoints are used such that they will not be relocated after you
+ *  run GeoInit.
+ */
+
+// Include own header
+#include "modules/datalink/missionlib/waypoints.h"
+
+#include <stdio.h>
+#include <string.h>
+
+#include "subsystems/navigation/waypoints.h"
+//#include "subsystems/navigation/common_nav.h" // for fixed-wing aircraft
+
+#include "modules/datalink/mavlink.h"
+#include "modules/datalink/missionlib/mission_manager.h"
+
+static void mavlink_update_wp_list(void)
+{
+  // Store the waypoints in a mission item
+  if (NB_WAYPOINT > 0) {
+//        for (uint8_t i = 0; i < NB_WAYPOINT; i++) {
+//            mavlink_mission_item_t mission_item;
+//
+//            /*
+//             * Convert ENU waypoint to UTM
+//             * May be removed, but nav_move_waypoint() uses UTM coordinates in case of fixed-wing aircraft
+//             */
+//            struct UtmCoor_f utm;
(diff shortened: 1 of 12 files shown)


flixr:
Um, what's up with the chances to the mavlink common message set? Are these changes made by you guys or a newer mavlink version?
Still think that mavlink should be a submodule...

flixr:
Um... not going to merge this like it is... now you can only use UDP. Before it worked with UART, UDP and USB serial...
Also plz fix indentation in the missionlib

flixr:
Also doesn't work for fixedwings anymore...

sw/airborne/modules/datalink/missionlib/waypoints.c
```diff
@@ -0,0 +1,239 @@
...
+//            mission_item.z = lla.alt; // altitude
+//            mission_item.seq = i;
+//
+//            mission_mgr.waypoints[i] = mission_item;
+//        }
+        for (uint8_t i = 0; i < NB_WAYPOINT; i++) {
+            mavlink_mission_item_t mission_item; 
+            // waypoint_set_global_flag(i);
+            if (waypoint_is_global(i))
+                printf("Waypoint(%d): is global\n", i);
```
flixr: No `printf`s! This will not work on non-linux systems...

sw/airborne/modules/datalink/mavlink.c
```diff
@@ -21,553 +21,353 @@
...
+  }
 }
 
-#include "subsystems/electrical.h"
-static inline void mavlink_send_battery_status(void)
+/**
+ * Event loop MAVLink calls
+ * Called at the end of the event loop (runs at maximum allowable frequency)
+ */
+void mavlink_event(void)
```
flixr: What's wrong with the previous way the actual connection (UDP, UART, etc.) was handled? It was working for bare metal and linux systems and uart and not only linux & UDP

flixr:
Also removed quite some messages and setting and reading of params if I see that correctly...

flixr:
Btw, regarding mavlink as submodule, check my [mavlink_submodule branch](https://github.com/flixr/paparazzi/tree/mavlink_submodule)

sw/airborne/modules/datalink/missionlib/waypoints.c
```diff
@@ -0,0 +1,239 @@
...
+            else
+                printf("Waypoint(%d): is NOT global\n", i);
+            waypoint_globalize(i);
+            mission_item.x = (float)waypoint_get_lla(i)->lat*1e-7; // lattitude
```
flixr: You shouldn't send LLA coordinates as float, you loose too much precision.....

flixr:
Could you plz explain the idea behind the mission manager?

flixr:
Looks to me like it would be eaiser/quicker to start re-adding stuff feature/function by feature/function to the existing module...

dewagter:
First of all:
 - I understand this could not be pulled in as is, that was not the idea.
 - The idea is that this is a working module/app while the mavlink module currently in master can not make the drone take-off. This is why this code should be shared and re-integrated in master a.s.a.p. and any help is welcome.

option 1: new module in parallel with current mavlink.xml
option 2: merge new with old ( @fvantienen ?)
option 3: first 1 then 2...

Checklist 
 - [ ] pull request mavlink extra messages in https://github.com/mavlink/mavlink/pull/454
 - [x] mavlink as submodule in paparazzi
 - [x] fix code style
 - [ ] remove debug printf's (also when protected by ifdef's? this is will be needed in the future as well when new groundstations are supported)
 - [ ] restore original mavlink module capabilities uart/udp/usb and settings

flixr:
I'm not in favour of having two mavlink modules... so IMHO only option 2 or 3 make sense...

flixr:
btw, printfs protected by ifdefs are fine (so it will compile for the bare metal systems), but even nicer would be a debug function that you can enable/disable, e.g. like in gps_nmea.c

audaxes:
It would seem the apps are not available in New Zealand if Google Play within Chrome is to be believed.

agressiva:
not available in brasil too

dewagter:
Google is processing the update to be available is a lot more countries.

- Please note it is still beta
- For now you need 2 apps: services and groundstation
- For now it's UDP only

flixr:
Please test #1422 instead of this one...

flixr:
Another question: the `SCRIPT_ITEM` is currently (mis?)used to send the active block, but this is the message to send the available blocks. Shouldn't a separate `SCRIPT_CURRENT` message (analogous to `MISSION_CURRENT` be used for that?

flixr:
@dewagter can we close this in favour of #1422 ?

dewagter:
#1422 overrules this pull request

PR #1409, opened 2015-11-06
build mavlink from paparazzi

dewagter:


dewagter:
@flixr This should be how you proposed to integrate mavlink

When/IF https://github.com/mavlink/mavlink/pull/454 gets accepted we can remove the paparazzi/mavlink fork

PR #1411, opened 2015-11-08
[actuators] Spektrum output actuators

fvantienen:
This can output spektrum to two devices, a primary and secondary output.


sw/airborne/firmwares/rotorcraft/main.c
```diff
@@ -218,6 +218,9 @@ STATIC_INLINE void main_init(void)
   // send body_to_imu from here for now
   AbiSendMsgBODY_TO_IMU_QUAT(1, orientationGetQuat_f(&imu.body_to_imu));
 #endif
+
+  // Do a failsafe check first
+  failsafe_check();
```
dewagter: This change is not related to the actuator driver, but is a more generic problem that in paparazzi the initial signal sent to actuators is not the failsafe value but the driver-initial values.

fvantienen: It is related.. because here the problem occurred ;) This actuator is why I added it, because wrong pulses were send before getting valid RC commands.

PR #1416, opened 2015-11-11
New asctec ESC i2c protocol

EwoudSmeur:
The i2c protocol to communicate with asctec escs has changed. The escs are still called v2, so I am not sure if I should make an actuators_asctec_v2_new.c file with another settings file, or make a define that switches between protocols within the actuators_asctec_v2.c file. What is preferred?


flixr:
Not sure if it is needed or desirable, but if we add the new protocol in separate files/subsystem it will be possible to use both old and new versions of the controllers at the same time...

fvantienen:
I would vote for seperate files/subsystem. Not sure about the naming of it.

Small fix in the code: at switch case 4 of set_addres the variable set_address should be set to 0 again. This fixes the problem of needing to reboot the drone to set each address.

fvantienen:
Seems ok for me to merge

PR #1419, opened 2015-11-12
Simplified rate control

EwoudSmeur:
I suggest removing the filtering of the setpoint commands for rate control. This simplifies the rate control and makes it more predictable. Furthermore, you can rest assure that any command will be executed immediately without delay. 

Opinions?

### DIFF
--- a/conf/airframes/AGGIEAIR/ark_quad_lisa_mx.xml
+++ b/conf/airframes/AGGIEAIR/ark_quad_lisa_mx.xml
@@ -96,7 +96,6 @@
     <define name="DEADBAND_P" value="20"/>
     <define name="DEADBAND_Q" value="20"/>
     <define name="DEADBAND_R" value="200"/>
-    <define name="REF_TAU" value="4"/>
 
     <!-- feedback -->
     <define name="GAIN_P" value="400"/>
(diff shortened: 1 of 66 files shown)


gautierhattenberger:
Maybe disabling by default with `#if` would have been enough.

Also a useful addition could be a derivative term for the control (only a PI currently). It seems that most other project have that in their rate controller and it is indeed helpful if correctly tuned.

dewagter:
@EwoudSmeur is this what flew on the bebop today?

EwoudSmeur:
@gautierhattenberger I totally agree with you. I intend to make an INDI rate controller that makes use of the angular acceleration.

@dewagter yep

OpenUAS:
It's flown with success, looking at Ewoud's track record, I'd say merge it.

flixr:
I would be in favour of either just `#ifdef`-ing it out, or removing the now obsolete variables and defines completely... (DDGAIN defines, settings files, etc)

flixr:
Can you please also update the settings file accordingly and remove the defines from all airframe files:
```
find conf/airframes -name "*.xml" -exec sed -i '/REF_TAU/d' {} +
find conf/airframes -name "*.xml" -exec sed -i '/DDGAIN_/d' {} +
```

PR #1420, opened 2015-11-13
INDI - flightplan attitude fix

dewagter:
Setting attitude from the flightplan was spinning when the heading is not perfectly north.

@EwoudSmeur: is this OK?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
@@ -171,7 +171,7 @@ void stabilization_attitude_set_rpy_setpoint_i(struct Int32Eulers *rpy)
   // stab_att_sp_euler.psi still used in ref..
   stab_att_sp_euler = *rpy;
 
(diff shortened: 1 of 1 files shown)


EwoudSmeur:
that function has as a comment: // only a plug for now... doesn't apply roll/pitch wrt. current yaw angle
in guidance_h.c the function is only called if you set the attitude from the flight plan, would you want roll and pitch to be applied to the current or the supposed-to-be yaw angle?

this was changed in february for quat_int with https://github.com/paparazzi/paparazzi/commit/8634b728ae6fff28acd89e14394b3906180f74b5

We can adopt this change, but I don't see how this would have anything to do with spinning

dewagter:
I did not investigate _why_ but unless ```psi=0``` is commanded, the quad would apply a mix of bank and pitch instead of pure pitch commands. With the spinning I mean the direction (mix of pitch and roll) in which the pitch command was executed to would vary in time when using ```quat_from_rpy_cmd_i```. (like if the heading command was integrated)

So if you know no objection I will apply this pull request because it solves the setting of attitude from the flightplan for non zero headings in INDI.

EwoudSmeur:
ok

OpenUAS:
THX ! (= less repairing...)

PR #1423, opened 2015-11-15
FBW for rotorcraft

fvantienen:
This is still a work in progress and needs to be tested. If I forgot something in the list below feel free to add.

Things that need to be fixed:
- [ ] Fix the NPS simulator
- [ ] Fix Radio control channels order
- [x] Fix code style
- [x] Cleanup the commit messages
- [x] Testing
- [ ] Maybe write a config for Pixhawk


flixr:
We probably could/should use the same `<configure name="SEPARATE_FBW" value="1"/>` technique as in fixedwing...

OpenUAS:
Ditto that

dewagter:
- channel order = spektrum = fixed (make new issue if this is not what you want)
 - sim of heli (without mixing) is a separate issue

Fixed:

 - code style
 - docs
 - cleanup
 - all tested, fixed and flying fine

Ready to merge

audaxes:
@dewagter - "all tested, fixed and flying fine." Brilliant - any chance of a list of servo type and model numbers for 3x cyclic, throttle and tail.

dewagter:
this pull request is flying a 600 sized logo with VBAR still connected and
2 autopilots (1 as FBW with little well checked code which is not often
programmed and needs a full range of safety checks after each programming)
and an AP board which is reprogrammed as usual.

a next pull request will address the cyclic mixing (which is not in this
pull request). This particular pull request is about safety and support for
e.g. dual-mcu pixhawk boards.

-Christophe

audaxes:
@dewagter, thanks for info - I am very interested in the paparazzi heli stuff. I have a few align models which could be put to into service with Lisa/M boards. I  hope it is not an issue to use older spektrum DSM2 satellites  with DX8?

dewagter:
@audaxes be careful with helicopters: vibrations make helicopters very
different than small multicopters. Do not just upload and try to fly but
use proper damping and proper filtering and testing and then rate-only
etc...

-Christophe

PR #1425, opened 2015-11-16
[docker] fix run.sh for OSX with access to X

### LINKED ISSUE #1421: [Docker] Access to X-Server not working on Mac
For using Docker in Paparazzi see https://wiki.paparazziuav.org/wiki/Docker

While this works nicely on Linux, there are some issues with using this on Mac:
The https://github.com/paparazzi/paparazzi/blob/master/docker/run.sh helper script currently can't properly deal with Mac to enable access to the X-Server and forward the correct USB devices.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28252027-docker-access-to-x-server-not-working-on-mac?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10895&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flixr:
- OSX readlink doesn't have the -f or -m options, require correct PAPARAZZI_SRC to be set already for now
- using xauth with docker on OSX doesn't work, so we use socat to forward the socket via TCP
  - see https://github.com/docker/docker/issues/8710
- no (pulse)audio on OSX...
- find on OSX doesn't have the -printf option... so use exec echo instead

this will hopefully fix #1421

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -10,27 +10,67 @@ else
     args="$@"
 fi
 
+# check if running on Linux or OSX
+UNAME=$(uname -s)
+
+
+############################################################
+# share this paparazzi directory with the container
+############################################################
+
 # set PAPARAZZI_SRC to this tree
-SCRIPT=$(readlink -f $0)
-SCRIPT_DIR=$(dirname $(readlink -f $0))
-PAPARAZZI_SRC=$(readlink -m $SCRIPT_DIR/..)
+# on OSX: readlink doesn't have the -f or -m options, try using pwd
+if [ $UNAME == "Linux" ]; then
+    SCRIPT=$(readlink -f $0)
+    SCRIPT_DIR=$(dirname $(readlink -f $0))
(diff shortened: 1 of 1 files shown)


flixr:
@hooperfly would be nice if you can test this... maybe we have a way to circumvent all the xquartz/gtk issues for now...

hooperfly:
Will do. I'll give it a go tonight.

hooperfly:
Also, of note, purchased a Lenovo laptop to run Ubuntu native. It will be nice to be able to develop, compare and contrast differences between our MacOSX version and the Linux version.

hooperfly:
Tested with Mac OS X `El Capitan` -- Working just fine. Ship it. :+1:

Issue #1312, opened 2015-08-25
viewvideo: netcat option is broken


flixr:
The `VIEWVIDEO_USE_NETCAT` option of the `viewvideo_rtp_stream` module is broken, it doesn't even compile...

Plz either fix it or we remove it completely...

See also #1293

gautierhattenberger:
This was actually my doing for last imav. I'll have a look.

OpenUAS:
It's was used and to be honest quite useful...

flixr:
@gautierhattenberger would be nice if you could fix this for the v5.8 release.

flixr:
Should be fixed with #1426

PR #1426, opened 2015-11-17
[cv] fix netcat video module

gautierhattenberger:
it is compiling and "should" run correctly except that the video_tread
module is now too big and the memory is not enough on ardrone2 to fork
the process

see #1312


sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -197,7 +205,7 @@ void viewvideo_init(void)
 
 #if VIEWVIDEO_USE_NETCAT
   // Create an Netcat receiver file for the streaming
-  sprintf(save_name, "%s/netcat-recv.sh", STRINGIFY(VIEWVIDEO_SHOT_PATH));
```
flixr: why remove `STRINGIFY` here? Now you need to make sure it is properly quoted if you want to define this yourself....

gautierhattenberger: It tries to open `"/data/video/images"/netcat-recv.sh` and fails.

flixr: But only because it is already a string, i.e. the default?
What about if you add `<define name="VIEWVIDEO_SHOT_PATH" value="/foo/bar"/>` to your airframe file to set a custom location?

gautierhattenberger:
fixed and tested STRINGIFY

PR #1427, opened 2015-11-18
Flying Bebop2

EwoudSmeur:
With this you can fly a Bebop2. The damping on the roll axis is a bit odd, so if you directly feed back the roll rate in this axis it starts to resonate. If the rate in this axis is filtered it is slightly delayed and this problem does not occur. The control is a bit slower though.


EwoudSmeur:
oh... maybe this breaks the bebop1... @fvantienen?

fvantienen:
Yes it does.. I will fix it

fvantienen:
Closed in favor of #1429

PR #1434, opened 2015-11-21
[mavlink] upgrade

dewagter:
Proposed road forward for the mavlink handling of blocks. (See #1422 for reference)

 - [x] update paparazzi
 - [x] update PPRZServices
 - [x] update PPRZ Groundstation


dewagter:
@flixr is this more or less what you had in mind. I managed to update the app so it could be merged if this is OK.

dewagter:
Yes, I added it explicitly for the longer strings overwriting whatever last character there might be. This solves overflow but does not solve possible identical names when you chop off the end:

VeryLongMissionNameWithMoreThan49CharactersNumber1 and
VeryLongMissionNameWithMoreThan49CharactersNumber2

would end up with the same name, but would show up as 2 items in the block list since they also have a seq nr, so I guess that is OK.

PR #1437, opened 2015-11-23
[helicopter] Add helicopter mixings and throttle curves

fvantienen:
This adds two modules for Helicopters:
- Throttle curves
- Swashplate mixers (which depends on throttle curves)


flixr:
How about using float factors for the swashplate mixing?
I was never a fan of the int implementation with it's discretization errors

fvantienen:
Adjusted and tested

flixr:
Nice :-)

PR #1439, opened 2015-11-24
Some stereocam code

rmeertens:
Hey everybody, 

Last week Sjoerd and I worked on some cool stuff with the 4 gram stereo board. The result is two modules: follow_me and follow_egg. Follow_egg is meant for the "egg" the MAV lab is housed into and controls the heading and velocity of the drone. Follow_me is more general for every quad with a stereoboard. 
Let me know what to change before this can be merged.

### DIFF
--- a/conf/conf_example.xml
+++ b/conf/conf_example.xml
@@ -211,12 +211,12 @@
   <aircraft
    name="ardrone2"
    ac_id="201"
-   airframe="airframes/examples/ardrone2.xml"
+   airframe="airframes/TUDelft/airframes/ardrone2_opticflow_stereo_follow_me.xml"
    radio="radios/dummy.xml"
    telemetry="telemetry/default_rotorcraft.xml"
-   flight_plan="flight_plans/rotorcraft_basic.xml"
+   flight_plan="flight_plans/rotorcraft_follow_me_stereocam.xml"
    settings="settings/rotorcraft_basic.xml settings/control/rotorcraft_guidance.xml settings/control/stabilization_rate.xml settings/control/stabilization_att_int_quat.xml settings/estimation/ahrs_int_cmpl_quat.xml settings/estimation/body_to_imu.xml settings/nps.xml"
-   settings_modules="modules/gps_ubx_ucenter.xml modules/air_data.xml modules/geo_mag.xml"
+   settings_modules="modules/cv_opticflow.xml modules/opticflow_hover.xml"
    gui_color="red"
   />
   <aircraft
--- a/conf/modules/stereocam_follow_egg.xml
+++ b/conf/modules/stereocam_follow_egg.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE module SYSTEM "module.dtd">
 
-<module name="follow_egg" dir="stereocam/follow_egg">
+<module name="stereocam_follow_egg" dir="stereocam/follow_egg">
   <doc>
     <description>follows the egg next to the mavlab</description>
   </doc>
--- a/sw/airborne/modules/stereocam/follow_egg/follow_egg.c
+++ b/sw/airborne/modules/stereocam/follow_egg/follow_egg.c
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) Roland and Sjoerd
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/follow_egg/follow_egg.c"
+ * @author Roland and Sjoerd
+ * follows the egg the MAV lab is housed in. It follows a corridor while assuming that the corridor always goes left.
+ */
+
+#include "modules/stereocam/follow_egg/follow_egg.h"
+#include "modules/stereocam/stereocam.h"
+#include "subsystems/abi.h"
+#include "guidance.h"
+#include "state.h"
+#include "subsystems/datalink/telemetry.h"
+#include "navigation.h"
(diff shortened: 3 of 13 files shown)


rmeertens:
Looks like Travis has some problems installing paparazzi

flixr:
Travis seemed to have a temporary problem: `Unable to connect to ppa.launchpad.net` nothing to worry about

flixr:
@rmeertens it would be great if you could add some more comments and an example flight plan if you want other people to be able to use this as well.

I guess you use the follow_me module by setting the `ref_roll`, `ref_pitch` in an attitude block in the flight plan or something like that?

Regarding the follow_egg module: I don't see how this is supposed to work...
You are settings a velocity reference, but then use it as a position to directly est the `navigation_carrot`?
Also the sending of FOLLOWEGG message should be removed or the message added.
Btw, if you want to send debug info, you can also use the [INFO_MSG](http://docs.paparazziuav.org/latest/paparazzi_messages.html#INFO_MSG)

And please, please remove unused variables/includes... this just unnecessarily raises the maintenance effort needed in the future (apart from making the code harder to read).

flixr:
It would also be nice if the naming of the modules would reflect that it is using the stereocam....

rmeertens:
Fixed the unused variables and imports and renamed the modules. 
One last problem: I get this in my paparazzi: Invalid_argument("Array.make")
While the datalink is on. Any idea what this could be?

flixr:
Sorry, no idea right now....

flixr:
@rmeertens did you figure out the problem? Are these modules working?
As I already mentioned, I don't think that the follow_egg module will work as expected since you are using a velocity as if it was a position...

rmeertens:
Hey Felix. 
As for the problem: have not solved it. Probably changed something deeper down in paparazzi. As for the follow_egg module: the velocity as if it was a position is working given that you set the I-gain to zero. I can show you a video if you like (we also have delflys flying in a corridor, really exciting stuff!). 
The best awesome solution would be:
- adding a mode: velocity_control that takes a velocity reference that any(?) module can set. 
- adding a function to nav such as <velocity forward="1.0" alt="2.0"/> that would try to keep a certain velocity. 

As I'm not really in the nav mode it would probably cost me a lot of time to implement it neatly, and this "hack" works for me right now. Unfortunately I don't have time the coming weeks to implement a nice velocity reference in the navigation mode. 

We could either: 
- merge it as is, with me helping confused people
- merge it with a few changes that make the idea of setting a velocity in this module clear, but without the nav capabilities to react to them

A third option would be that somebody who is more into how the whole of paparazzi works (nav mode, xml interpretation, state estimation) implements a velocity control option in NAV. I really think that this would benefit the whole paparazzi project and would mean more research teams could switch. As I said: this isn't something I have time for in the short period in which I like this pull request to be merged. 

As for merging in general: the files that are changed only concern me and some people on the TU Delft, so I think that merging would solve problems for us, without creating more problems for other people. In fact: thanks to this pull request I'm already collaborating more with people from another university who also want to have awesome obstacle avoidance and corridor following capabilities.

flixr:
Ok. We can merge it, but I really doubt that it works with only the changes in this PR.
You set `navigation_carrot` to a fixed position (1.1m oder -0.1m away from the local origin) via `setVelocityReference`. What should work is to just add the velocity reference to the current position and set that.
And if you are running nav mode it will overwrite `navigation_carrot`, better use `navigation_target`.

flixr:
On a sidenote for later: of course the proper solution is to have a nice API for a velocity mode (that also works with all/most guidance algos).
What already should work is updating `guidance_h.sp.speed` and using HOVER mode...

flixr:
@rmeertens to sum up: it would be great if you could verify that the version of stereocam_follow_egg in this pull request indeed works.
Also can we remove your confs? It's broken anyway as the airframe files don't exist...

conf/conf_example.xml
```diff
@@ -211,12 +211,12 @@
   <aircraft
    name="ardrone2"
    ac_id="201"
-   airframe="airframes/examples/ardrone2.xml"
+   airframe="airframes/TUDelft/airframes/ardrone2_opticflow_stereo_follow_me.xml"
```
gautierhattenberger: The conf_example file should not contain so specific config. Please restore the original version here and eventually add this example to the TUDelft conf so it is tested for compilation.

conf/modules/stereocam_follow_egg.xml
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="stereocam_follow_egg" dir="stereocam/follow_egg">
+  <doc>
+    <description>follows the egg next to the mavlab</description>
```
gautierhattenberger: I think it would help to have a bit more descriptive text here. Especially which sensor is used (dedicated sensor, on-board Parrot camera, ...).
Same thing for the follow_me module.

sw/airborne/modules/stereocam/follow_egg/follow_egg.c
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (C) Roland and Sjoerd
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/follow_egg/follow_egg.c"
```
gautierhattenberger: quotes are not needed

PR #1440, opened 2015-11-24
Register multiple periodic messages

flixr:
So far when multiple modules/subsystem were calling `register_periodic_telemetry` for the same message/string, only the first one would actually register it's callback.

For cases like #1145 we would like to register multiple callbacks, so:
- Allow to add multiple callbacks: up to `TELEMETRY_NB_CBS` slots (4 right now).
- Also return slot index of callback (to make it easier to add an `unregister` function later) or -1 on failure

### DIFF
--- a/sw/airborne/subsystems/datalink/telemetry.c
+++ b/sw/airborne/subsystems/datalink/telemetry.c
@@ -30,10 +30,11 @@
 #include "subsystems/datalink/telemetry_common.h"
 #include "generated/periodic_telemetry.h"
 
-/* Implement global structures from generated header
+/* Implement global structures from generated header.
+ * Can register up to #TELEMETRY_NB_CBS callbacks per periodic message.
  */
 telemetry_msg telemetry_msgs[TELEMETRY_NB_MSG] = TELEMETRY_MSG_NAMES;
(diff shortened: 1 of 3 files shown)


flixr:
@gautierhattenberger any good idea on how to avoid "hardcoding" the number of slots in gen_periodic.ml to generate `TELEMETRY_CBS_NULL`?

gautierhattenberger:
do you think this can work:
```
struct telemetry_cb_slots telemetry_cbs[TELEMETRY_NB_MSG][TELEMETRY_NB_CBS] = { 0 };
```
?

flixr:
No, that doesn't work... what seems to work is not specifying any initializer at all and relying on them to be initialized by zero... but that is not very safe, as it is not guaranteed by the standard.

gautierhattenberger:
and what about generating the macro init like this:
```
#define TELEMETRY_CBS_NULL { \
  {{ 0 }}, \
 ...
  {{ 0 }}, \
}
```
According to Alex, it should initialized each line to 0 regardless of the number of columns.

flixr:
Good idea! That should work

flixr:
So for me this is good to merge now...

gautierhattenberger:
I'm fine with this too

PR #1441, opened 2015-11-25
Add component/filter ID to AHRS messages

flixr:
When running multiple AHRS algos in parallel, each filters can send the same message but with a different ID to distinguish the source.
See also #1145 

This adds a `comp_id` field to the following messages:
- AHRS_EULER
- AHRS_EULER_INT
- AHRS_QUAT_INT
- AHRS_RMAT_INT
- AHRS_GYRO_BIAS_INT
- GEO_MAG

Also register some more message in some of the filters and don't use the value from the state interface to send the attitude in body frame but rather compute it from the actual ahrs output (imu frame).
So it will still be correct if muliple AHRS are running and another one is enabled (sets the attitude in the state interface).


gautierhattenberger:
I guess this one is okay now

PR #1442, opened 2015-11-25
Change register_periodic_telemetry() to use defines instead of string for matching

flixr:
So far `register_periodic_telemetry` takes a string as argument and compares it to the strings in used telelemetry xml files to register the callback.
This makes it possible to have "meta" names for messages, meaning that name does not need to correspond to an actual message.
E.g. if you would register "MYMODULE" and have a "MYMODULE" as message name in the telemetry.xml file, the callback would be registered and you could send any message from within that.

However this requires the strings to be stored in ROM (flash) for comparison...

On small systems (e.g. CC3D, naze32, etc) with very small flash this becomes a major limitation.
The proposal here is to drop the "meta" name functionality and instead require to use an actual message name (actually a generated define), so it can be reduced to a simple numerical id.

So the main question is if we want to sacrifice the flexibility we currently have to reduce ROM size on the very small systems?

### DIFF
--- a/sw/airborne/boards/ardrone/navdata.c
+++ b/sw/airborne/boards/ardrone/navdata.c
@@ -236,7 +236,7 @@ bool_t navdata_init()
   }
 
 #if PERIODIC_TELEMETRY
-  register_periodic_telemetry(DefaultPeriodic, "ARDRONE_NAVDATA", send_navdata);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_ARDRONE_NAVDATA, send_navdata);
 #endif
 
   // Set to initialized
--- a/sw/airborne/boards/bebop/actuators.c
+++ b/sw/airborne/boards/bebop/actuators.c
@@ -66,7 +66,7 @@ void actuators_bebop_init(void)
   actuators_bebop.led = 0;
 
 #if PERIODIC_TELEMETRY
-  register_periodic_telemetry(DefaultPeriodic, "ACTUATORS_BEBOP", send_actuators_bebop);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_ACTUATORS_BEBOP, send_actuators_bebop);
 #endif
 }
 
--- a/sw/airborne/firmwares/demo/demo_ahrs_actuators.c
+++ b/sw/airborne/firmwares/demo/demo_ahrs_actuators.c
@@ -106,10 +106,10 @@ static inline void main_init(void)
 
   downlink_init();
 
-  register_periodic_telemetry(DefaultPeriodic, "AUTOPILOT_VERSION", send_autopilot_version);
-  register_periodic_telemetry(DefaultPeriodic, "ALIVE", send_alive);
-  register_periodic_telemetry(DefaultPeriodic, "COMMANDS", send_commands);
-  register_periodic_telemetry(DefaultPeriodic, "ACTUATORS", send_actuators);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_AUTOPILOT_VERSION, send_autopilot_version);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_ALIVE, send_alive);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_COMMANDS, send_commands);
+  register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_ACTUATORS, send_actuators);
 
   // send body_to_imu from here for now
   AbiSendMsgBODY_TO_IMU_QUAT(1, orientationGetQuat_f(&imu.body_to_imu));
--- a/sw/airborne/firmwares/fixedwing/autopilot.c
+++ b/sw/airborne/firmwares/fixedwing/autopilot.c
@@ -191,18 +191,18 @@ void autopilot_init(void)
 
 #if PERIODIC_TELEMETRY
   /* register some periodic message */
-  register_periodic_telemetry(DefaultPeriodic, "AUTOPILOT_VERSION", send_autopilot_version);
-  register_periodic_telemetry(DefaultPeriodic, "ALIVE", send_alive);
-  register_periodic_telemetry(DefaultPeriodic, "PPRZ_MODE", send_mode);
-  register_periodic_telemetry(DefaultPeriodic, "ATTITUDE", send_attitude);
-  register_periodic_telemetry(DefaultPeriodic, "ESTIMATOR", send_estimator);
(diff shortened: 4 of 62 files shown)


flixr:
@martinmm @gautierhattenberger: question on an implementation detail: This now generates IDs based on the messages in the telemetry.xml file instead "simply" using the actual IDs of the messages.
Any particular reason for that besides simplicity of implementation?

@gautierhattenberger also why is the undef needed?

gautierhattenberger:
probably not needed, I was trying to solve an issue that turns out to be a spelling error. At least this way it doesn't matter if messages.h is included before the periodic ID definition.

Not sure I get your ID problem. In messages.h it generates default (-1) value and in periodic.h you need IDs matching the array index, but not the actual messages ID.

martinmm:
@flixr How I understood it:

At compile time there is a string table created in periodic_telemetry.h (TELEMETRY_MAIN_STRUCT) for each used/usable message. The register_periodic_telemetry() function also contains a string and searches that in the table.

- this registering can fail (i.e. typo) as it is checked only at run time; that is hard to see for a user
- the type of messages that can be sent and the mapping is fixed at compile time, no need for any (string based) dynamics
- it needs several kBytes (and some time)

In the patch the message is mapped to a defined number, no need for an intermediate string. The default "-1" is needed for all the register_periodic_telemetry() code that contains the #define but is never called (in the current implementation the string would be in the register function but not in the periodic table).

This assumes that DefaultPeriodic is set to pprz_telemetry. I could not find any example of that not being the case and do not know how your own DefaultPeriodic *_telemetry struct is created.

dewagter:
Does anyone use this flexibility?

I believe paparazzi airborne code has always been and still is oriented
towards (also) smaller chip support. Otherwise we should have a linux
threaded c++ autopilot.

So I'm in for the proposed change.

(not sure this will be sufficient for CX10 support but every step counts)

-Christophe

podhrmic:
For example we are flying Lisa MX that has plenty of memory and power, so few extra kb don't matter. However, I am personally not too excited that I would have to change our code to conform with new messages.
So overall - neutral.

martinmm:
@podhrmic what do you mean with new messages?

podhrmic:
The change from string values to defined values when registering messages.

martinmm:
@podhrmic you mean the one-time change you have to do for code that is not in the central paparazzi repository? There should not be a difference in the future. Messages are just registered a little different. Or am I missing something?

old:  register_periodic_telemetry(DefaultPeriodic, "GPS", send_gps);
new: register_periodic_telemetry(DefaultPeriodic, TELEMETRY_MSG_GPS, send_gps);

(the large "TELEMETRY_MSG_" could be changed to something shorter)

podhrmic:
Yes, that is correct. Is not that big a deal (hence I am saying I am neutral to the change), but I would imagine that other users will have to do the same, so just to keep it in mind.

flixr:
@gautierhattenberger there is no specific "problem" like in the sense that it doesn't work....
Just wondering if it is the best solution, e.g. considering what @martinmm already pointed out: This only works for the `pprz_telemetry` and it will be difficult to add a second/different one (if the IDs are generated depending on telemetry xml contents).

E.g. it would be nice to use the same register_periodic_telemetry for mavlink messages in the future.
I imagine that if we take the actual message IDs instead of generating different ones it would make things easier?
Would it really cost more to use the actual message ids? It should only be one for loop at register and even save some stuff in the code generation and the troubles/workarounds with include order and undef...
Or am I missing something?

martinmm:
Only messages that are defined to be sent periodically (through xml) are
actually possible to send through register_periodic_telemetry() registered
callbacks. Therefore we need to know what messages are in the xml.

We could add another generated list and some code to map the original message id
number to the (xml dependant) periodic number. Guess that is some more work in
OCAML generation as we have to assign the full message id range to the periodic
id. The #ifndef trick lets us do that easily, I could not work that out in
another way.

gautierhattenberger:
If you want to register with message ID, the only way I see for now is to generate the register function (or a part of it) to directly map the two ID set with a switch. Not sure it really helps to reduce the code size...
But the main question as you said @flixr, do we really want this ? That is some work and I'm a bit busy the coming week, and the overall result is still less flexible than what we have now (but annoying to debug, I agree on this point).

martinmm:
What flexibility do we lose? We are only changing a (memory consuming) string to a #define. The algorithm does not change.

If you add a MYMODULE to the telemetry file

`<message name="MYMODULE" period="23.0"/>`

it will be in periodic_telemetry.h

`#define TELEMETRY_MSG_MYMODULE_ID 47`

and you can use it as you use any meta name now. Having created a new MSG type by that is also a good argument not to use the actual message name to not confuse both of them.

flixr:
True, you would just need to add a
```
#ifndef TELEMETRY_MSG_MYMODULE
#define TELEMETRY_MSG_MYMODULE -1
#endif
```
to the piece of code where you want to register that pseudo message, so it will compile even if it is not in your telemetry file.

IMHO its debatable whether this really has any practical benefits though or if it might not even be better to "enforce" using actually existing telemetry message names to lower confusion and catch simple typos in the telemetry.xml file...
So in that regard I would we could merge this PR as it really reduces size and has basically the same functionality.

But my main concern is that I would like to use `register_periodic_telemetry` for more than just pprz_telemetry: Mavlink, for the "companion computer" use case...
So I'm wondering if there is a nicer way and if using the actual message ids would not make something like this easier and cleaner...

flixr:
See my `register_with_msg_id` branch for an alternative using the msg id and an added possibility to specify a telemetry type for a process (default: pprz) and using that to register mavlink messages in the same way:
https://github.com/paparazzi/paparazzi/compare/master...flixr:register_with_msg_id

The overhead is to add an extra `id` field to the struct which contains the actual message id and looping over all callback structs upon register until the matching id is found.

Since there it is really using the msg id, you loose the ability for registering meta/pseudo messages (currently only really used for `STAB_ATTITUDE` and `STAB_ATTITUDE_REF`).
But if we decide that we don't want/need this functionality, I think it would be nice to check if the messages specified in the xml file actually exist (e.g. `ACTUATORS_BEBOP` vs. `BEBOP_ACTUATORS`).

And I think it would have actually also been possible to add the type extension and use it for mavlink based on this PR (so with the generated defines for consecutive numbers for the messages).

What do you guys think?
In any case I would like to have something that makes it usable for other types of telemetry as well...

gautierhattenberger:
> In any case I would like to have something that makes it usable for other types of telemetry as well...

It was the idea to have a telemetry struct, even if the first implementation didn't really allowed that. So I'm fine with this.
For the ID, I don't have a very strong opinion after all. It is some efforts to support low grade boards, but this is also part of the project philosophy.

flixr:
Closing in favour of #1448

Issue #1446, opened 2015-11-29
Computer gets unresponsive after running Realtime Plotter for a long time


flixr:
From @esden on Ubuntu 14.04:
Computer gets unresponsive when running e.g. the "Raw Sensors" session with 4 realtime plotters for a long time.
The `hud-service` process seems to take up quite a bit of CPU and fills the memory until it gets unresponsive.

This seems to be related to #1445: The HUD service seems to connect to menu events of all apps, and every event changing the avg/stdev values in the curves menu trigger it...

flixr:
Btw, this seems to be Unitiy specific issue (since `hud-service` is from Unity). There doesn't seem to be this problem with the e.g. the Gnome Desktop.

flixr:
Closing, since workaround of disabling the menu updates (as per #1447) works for now....

PR #1447, opened 2015-11-29
[logalizer] plotter: disable update of average and stdev in menu

flixr:
first of all the image/label with the current value is not displayed anymore (at least on Ubuntu 14.04).
And probably more importantly, these constant updates of the menu result in high cpu-load and memory usage of hud-service,
Rendering the computer unusable if left running for a long time (until hud-service is killed and restarted).

Workaround for #1446 by simply disabling the update of these labels/menu entries


flixr:
@esden if this works for you as well, I'll merge it.

PR #1448, opened 2015-11-29
Change register_periodic_telemetry() to use msg ID and support other protocols (mavlink)

flixr:
Alternative to #1442:
- Change register_periodic_telemetry() to use msg ID
- possibility to use register_periodic_telemetry for mavlink and others.
 -  in the telemetry xml file a type can be specified for the process (default is "pprz").
 - You can also set "mavlink" and use mavlink messages, which will be sent as specified if the mavlink module is loaded and that mavlink message is registered there.
- use PPRZ_MSG_ID_x instead of DL_x

This requires an extra unit8_t id field in the struct (so if you have 20 different messages in the telmetry file, this would add an extra 20bytes compared to #1442).
At runtime it will loop over the periodic messages until it finds the matching id (compared to already knowing the index).
Also this really makes it impossible to register "meta/alias" messages... but I would argue that even in #1442 this is not nice since you need to provide the #ifndefs in the code...


flixr:
So I would vote for this solution, so basically dropping support for registering alias messages and requiring the messages to match instead.
Main reasons for me are: makes it easier to use for mavlink or other protocols as well and reduces footprint on the small systems.

@gautierhattenberger @martinmm @dewagter what do you think?

sw/airborne/subsystems/datalink/telemetry_common.h
```diff
@@ -54,23 +56,22 @@ struct telemetry_cb_slots {
...
-  struct telemetry_cb_slots *cbs; ///< array of associated callbacks
+  uint8_t nb;                     ///< number of messages
+  struct telemetry_cb_slots *cbs; ///< array of callbacks defined through TELEMETRY_MSG
 };
 
 /** Register a telemetry callback function.
  * empty implementation is provided if PERIODIC_TELEMETRY is not set or set to FALSE
  * @param _pt periodic telemetry structure to register
- * @param _msg message name (string) as defined in telemetry xml file
+ * @param _msgn message id/number (use DL_<message_name> define)
```
gautierhattenberger: should be `PPRZ_MSG_ID_`

gautierhattenberger:
I'm fine with this one

sw/airborne/subsystems/datalink/telemetry_common.h
flixr: thx, fixed

flixr:
ping @dewagter @martinmm would be great if you could chip in here... would like to merge this ASAP...

flixr:
@dewagter @LodewijkSikkel also changed the mavlink module to send messages as specified in the telemetry xml file (if no mavlink process in telemetry file it will fall back to hardcoded periods).

martinmm:
still fits into the Naze32 if the compressed sine is used, guess the overall size increased over the last 5 months; starts and looks ok
(something is wrong with the compressed sine)

sw/airborne/modules/datalink/mavlink.c
```diff
@@ -107,22 +138,30 @@ void mavlink_init(void)
...
-  RunOnceEvery(21, mavlink_send_battery_status());
-  RunOnceEvery(32, mavlink_send_autopilot_version());
-  RunOnceEvery(33, mavlink_send_gps_global_origin());
-  RunOnceEvery(10, mavlink_send_gps_status());
-  RunOnceEvery(10, mavlink_send_vfr_hud());
+#if PERIODIC_TELEMETRY && defined TELEMETRY_MAVLINK_NB_MSG
+  // send periodic mavlink messages as defined in the Mavlink process of the telemetry xml file
+  // transport and device not used here yet...
+  periodic_telemetry_send_Mavlink(&mavlink_telemetry, NULL, NULL);
+#else
```
dewagter: I think this RunOnceEvery was needed before because there was no other option, but now that you have the telemetry file which is much more flexible, I would argue that this RunOnceEvery can be completely removed.

dewagter:
This looks great and is a milestone in the support of multiple protocols.

Not really about this pull request: but is there no better technical solution than to loop through the list to match ID's on every message? Now that it are uint8_t, can't we have a list of 256 function pointers or is RAM more valuable than a few hundred machine codes per iteration?

sw/airborne/modules/datalink/mavlink.c
flixr: I kept the RunOnceEvery for now so it will still work when you have a telemetry file without the mavlink process... just wanted to avoid breaking existing stuff...
But I wouldn't mind removing those (except for the `send_params`)

flixr:
Of course we could have a list of 256 function pointers, but IMHO this is wasting space.
The for loop is only needed once when registering the message and at maximum it will loop up to the number of different registered messages... so it should not be very costly.

PR #1449, opened 2015-11-29
[tmtc][rotorcraft] set course using east/north velocity

flixr:
Try to set a course to a sensible value for rotorcrafts, by computing it from north and east velocity.
Needed e.g. for ACINFO.


flixr:
No idea why it was set to heading, but also incorrectly converted to degrees first...
@gautierhattenberger seems you added this quite some time ago... any feedback?

gautierhattenberger:
not really, it shouldn't harm to use the "correct" value...

PR #1452, opened 2015-12-01
[math] compressed sine table for flash and RAM, support for 20K RAM boards

flixr:
Possibility to compress the trig tables to save flash and RAM.
Needed for boards like CC3D and naze32 (STM32F103CBT6, 128K flash, 20K RAM).

Also adds support for the naze32 rev4 and rev5 and CC3D boards.


flixr:
@martinmm can you please explain how the compressed table works a bit more?
What is the `USE_REAL_TABLE_12_USE` define for? In the 4 and 8bit variants the equivalent seems to be always done.

Also do you have any test progs that compare/validate the different compressed tables to each other?

martinmm:
> can you please explain how the compressed table works a bit more?

There are two compressions, one for ROM by storing only the difference to the
last value

PPRZ_TRIG_INT_COMPR_FLASH

the other for having it compressed in RAM by using a tree to store the higher
bits, this comes in various qualities

PPRZ_TRIG_INT_COMPR_HIGHEST
PPRZ_TRIG_INT_COMPR_HIGH
PPRZ_TRIG_INT_COMPR_LOW

There is some simple description in the comments

https://github.com/paparazzi/paparazzi/commit/eea7ece4636b5c58213393bc40b36e876c8b9314#diff-a1a067f4abd6fd1ba74f8f3083d85df7R30

> What is the `USE_REAL_TABLE_12_USE` define for? In the 4 and 8bit variants the
> equivalent seems to be always done.

USE_REAL_TABLE_12_USE uses the original table instead of unrolled "if".
Unrolling makes it faster but uses slightly more code. For the higher
compressions it is not faster/creates more code. The option is for
debugging/understanding only.

> Also do you have any test progs that compare/validate the different compressed
> tables to each other?

There is some test code in the commit that checks the four algorithms, it runs
on processors > 128kB, i.e. Apogee

https://github.com/paparazzi/paparazzi/commit/eea7ece4636b5c58213393bc40b36e876c8b9314#diff-a1a067f4abd6fd1ba74f8f3083d85df7R108

flixr:
Ah, thanks!
Overlooked the `test_tables` function, only saw the timing tests...

flixr:
@martinmm can you test this on your boards and add the cjmcu?
UART1_RX is not working on my flip32 so I can't send it commands... but it might just be broken on my board.

martinmm:
> @martinmm can you test this on your boards and add the cjmcu?

ok

> UART1_RX is not working on my flip32 so I can't send it commands... but it
> might just be broken on my board.

is the USB of the board plugged and the TX pins are fighting with each other?

flixr:
> is the USB of the board plugged and the TX pins are fighting with each other?

OMG, totally missed that in the schematics. Yep... that was it, works fine if not powered through USB.

MAVProxyUser:
Will you be adding support for OpenPilot Revo?

flixr:
Don't have one, but pull requests are welcome :-)

MAVProxyUser:
Would donation of a board help the possibilities? Not much time to contribute code atm unfortunately.

flixr:
Sure, it would help... but I can't promise when I'll have time to do it... although it should be fairly straight forward if there is a schematic available, just need to write a board file to assign the correct pins, etc...

MAVProxyUser:
hardware files are here... https://github.com/librepilot/LibrePilot/tree/next/hardware/Production/OpenPilot%20Revolution 
What is the best place to shoot you an email to get your physical address?

flixr:
felix.ruess@gmail.com

PR #1453, opened 2015-12-01
Increase buffer size so larger messages can be sent

podhrmic:
The default buffer size is too small for larger messages needed for example by HITL (~170bytes).

### DIFF
--- a/sw/airborne/mcu_periph/uart.h
+++ b/sw/airborne/mcu_periph/uart.h
@@ -33,10 +33,10 @@
 #include "std.h"
 
 #ifndef UART_RX_BUFFER_SIZE
-#define UART_RX_BUFFER_SIZE 128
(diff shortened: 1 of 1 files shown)


flixr:
Hm.... until there are actually messages that are larger in mainline, I'm a bit reluctant to increase the default buffer size.
You could just define a larger buffer size for your hitl target in the airframe file.

OpenUAS:
@podhrmic, have to agree with Flixr at this point, impact unknown, if needed now you could op to just define UART_RX_BUFFER_SIZE in airframe and you are good to go...

flixr:
Closing this, since this default should work fine for all messages in master and increasing the default size will negatively impact the support for the small devices.
We should totally raise the default once it we have something in mainline where it is not sufficient anymore.

PR #1454, opened 2015-12-02
Avoiding conditional directives that split up parts of statements

lzmths:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

    https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
    https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/sw/airborne/modules/nav/nav_catapult.c
+++ b/sw/airborne/modules/nav/nav_catapult.c
@@ -35,7 +35,7 @@
  */
 
 
-
+#include <stdbool.h>
 #include "generated/airframe.h"
 #include "state.h"
 #include "subsystems/datalink/downlink.h"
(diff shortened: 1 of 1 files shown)


flixr:
Nice.
IMHO it would be even better to choose a more descriptive name than `test` and use `bool_t` (like in the rest of the code) instead of `bool` from stdbool.h.
Also I'm afraid we have quite a few more cases where it's even less readable ;-)

lzmths:
Did you like the name reset_lauch?

flixr:
Jep, thanks!

PR #1455, opened 2015-12-02
Multigps Subsystem

masierra:
Out of the necessity of robustness from the Piksi RTK gps system I've written this subsystem which automatically switches between two gps units. Its completely compatible as a gps subsystem so all one needs to do to use it is set: `<subsystem name="gps"  type="multi"/>`

I've added GCS strip settings to manually select between the two gps's if you just add settings/rotorcraft_basic_multi.xml. New messages MULTI_GPS_INT_PRIMARY and MULTI_GPS_INT_SECONDARY report the status of individual GPS units and PIKSI_HEARTBEAT reports the heartbeat for the piksi (since piksi wont send any kind of position messages without lock).

The subsystem switches when the fix of one gps is higher than that of the other. 

There is currently only support for one Piksi and one Ublox gps.


flixr:
Cool! This is definitely functionality many will appreciate a lot!

But I hope we can do this with less code duplication.
Basically change the "single" ubx and piksi to have their own GpsState and only update the global gps struct on demand like you have done for the "multi" implementations.
Best pass a pointer to the gps struct as argument to the parse function.
The goal is to completely get rid of the global gps stuct at some point, but that will take some time/effort so in the meantime we have to work around that.

flixr:
Also for the extra GPS messages: it would probably be nicer to add a `comp_id` to the existing GPS message (and to the gps struct), similar to https://github.com/paparazzi/paparazzi/pull/1441

flixr:
@masierra the functionality is great, but IMHO this still needs bit of work to avoid code duplication.
Do you have time to work on this? Are the points I mentioned before clear?

masierra:
For sure I'll get right on it.

flixr:
Awesome :-D

OpenUAS:
Nice indeed...

flixr:
@masierra any progress here? Would be great to get multi GPS support ASAP, but want to avoid doing the work twice...
Do you need help on specific points?

masierra:
@flixr just got back from the holidays so progress is back on.

flixr:
@masierra nice, let me know if you have questions or so...

masierra:
Reopening with updated changes

PR #1456, opened 2015-12-02
[spi] SPI slave implemented and tested for STM32F4

kirkscheper:
This expands the spi slave protocol to include the stm32F4. This was tested with a bluegiga bluetooth module and a Discovery board to verify its operation. A couple of comments:

The spi slave works really well with spi3, this was fully tested and verified. For some reason when using spi1, after about 1 second paparazzi stops sending data over the datalink although the spi continues to run(I verified this by sending an incrementing value everytime the spi dma returned), no idea why this is. Although I could clearly see the spi still working I can't say for certain that this works totally for spi1. In my test using spi2 the MISO pin remained low the entire time and didn't work for me, again not sure why, it uses exactly the same code as the other two so...

I added warnings to each spi reflecting the level of testing done thus far.

This will not work out of the box with the discovery board as the spi slave defines in the discovery board are not the alternate function NSS pins as defined by STM. These are true for the LisaM and Lisa MX so this will work for them. I am not sure why the pin definitions are different for these boards. If you like I can push the alternate function definitions as defined by STM. Let me know.

The bluegiga protocol is changed slightly.

The updated spi code still works with the F1 as tested with a lisa/s with bluegiga.

Let me know what you think.


flixr:
Thanks!
But it is indeed strange that spi1 and spi2 don't seem to work properly. Can't see anything wrong either.
@esden maybe you have an idea?

Regarding the `SPI_SELECT_SLAVEx_PORT|PIN` defines: so far these were only used in master mode and as we are using software slave management in master mode you can use any pin you like.

What are the bluegiga changes about? Does anything else need to be changed accordingly as well (i.e. the bluegiga firmware)? If yes, to which version?
It would be nice if you could split stuff like this in two separate commits with description in the future....
I'll split them up this time.

kirkscheper:
Yea, I thought it was weird that the gpio headers for the f1 and f2/4 are very different. What I could do is define SPIX_GPIO_PORT_NSS and SPIX_GPIO_NSS pins in the board files of the discovery board and the lisa_m. In the spi_arch.c I can check if those pins are not yet set, if not set them to the SPI_SELECT_SLAVEX_PORT and SPI_SELECT_SLAVEX_PIN defaults.

I tried to sneak the bluegiga stuff past you :p. There is nothing significant there just yet, it is fully backwards compatible with the previous firmware. I am currently working on expanding the bluetooth protocol to also receive data without a dedicated connection allowing for drone-drone communication along side a dedicated datalink. If you still want I can split the commits, no prob.

 It is a good comment to attach a comment in the bluegiga.c/h to which version of the bluegiga firmware is required for the current version (or to add a version to the bluegiga firmware at all...).

Let me know about the defines. Thanks for the comments.

flixr:
>Yea, I thought it was weird that the gpio headers for the f1 and f2/4 are very different.

The `GPIO_BANK_SPIx_XXX` and `GPIO_SPIx_XXX` for F1 defines are from libopencm3.
Since there are no equivalents defined in libopencm3 for F4, we did that ourselves in the board files as `SPIx_GPIO_PORT_XXX` and `SPIx_GPIO_XXX`.

> What I could do is define SPIX_GPIO_PORT_NSS and SPIX_GPIO_NSS pins in the board files of the discovery board and the lisa_m. 

That would be great!
Thinking about it again, we shouldn't use the `SPI_SELECT_SLAVE1_PORT|PIN` defines for the slave mode at all. These are really only meant for SS lines to peripheral devices when running master mode and the number does not in any way correspond to a specific SPI device.

flixr:
Thx! But I don't think it makes sense to even fallback to the `SPI_SELECT_SLAVEx_PIN` defines if the NSS defines are not there... these are two different things and the number of the SPI slave is not tied to the SPI port used...

kirkscheper:
I didn't that cause in the F4 the slave select pins are the correct nss
pins. But I get it.

I will define the nss pins in the Lisa m common when I'm back at my PC.

Kirk

flixr:
lisa_m_common.h is for the F1 boards, so not needed there. lisa_mx_common would be for the F4 boards...
Also you now removed the defines for the stm32f4_discovery again...

sw/airborne/boards/stm32f4_discovery.h
```diff
@@ -253,6 +253,14 @@
 #define SPI_SELECT_SLAVE2_PORT GPIOE
 #define SPI_SELECT_SLAVE2_PIN GPIO3
```
kirkscheper: aren't these correct for the discovery board?

PR #1461, opened 2015-12-04
Obstacle avoidance by Clint

rmeertens:
Also includes the follow_egg and avoid stuff from my previous request. Shall I remove them?


flixr:
Cool, the obstacle avoidance looks promising!

But can you please make a clean pull request with only the changes relevant for this?
E.g. don't change the mavlink submodule version, not change other messages back, etc.

To do that it would probably be easiest to create a new branch from master and checkout the relevant files from this branch and commit them in change units belonging together.

Another point regarding adding new messages: we are limited in how many we can add, so we have to choose carefully and can't just add a new one for each debug situation.

OpenUAS:
@rmeertens one can use an own crafted message.xml for specific purpose, no need to use default ..after al it is in conf... thus.. configurable oneself :fast_forward:

PR #1462, opened 2015-12-04
All stereo avoidance stuff I did last week

rmeertens:
Improved:
- pocketdrone avoidance flight
- stereo protocol only we use 
- avoidance module that takes distances in meters and angles of these measure devices and outputs a velocity or attitude ref.


flixr:
Better :-) But still doesn't compile, can you please fix your conf and make sure it works?

rmeertens:
Mmm, what doesn't compile for you???

flixr:
The ardrone2_RM airframe in your conf...

rmeertens:
Mmm, strange. For me it compiles.... Any idea what could be wrong?

flixr:
well, first of all the flight plan you set doesn't exist (or isn't added in this PR)
your bebop conf is also broken...

rmeertens:
Added the flight plan, and removed the bebop conf!

flixr:
C`mon Roland, you guys can do better... it should be obvious that we can't merge this as it is:
it's spewing loads of warnings because you are using the old periodic_telemetry_register (which won't work), shadowing variables, have messages in there that don't exist and added stabilization_attitude defines twice in the airframe file.
Also the new files have wrong doxygen headers...

sw/airborne/modules/obstacle_avoidance/guidance_OA.c
```diff
@@ -0,0 +1,353 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file modules/computer_vision/opticflow/stabilization_opticflow.c
```
flixr: please fix the doxygen headers

PR #1466, opened 2015-12-07
[fix] bebop1 motor direction

EwoudSmeur:
The motors were spinning all in the same direction on bebop1.


flixr:
So it looks to me like you could also remove the `actuators_bebop.i2c_trans.buf[1] = 0b00000000;` since it's not sent anymore, right?

EwoudSmeur:
Eh yes I suppose so... Perhaps it is 'better' to send the second byte as well, though it is clearly not correct right know. @fvantienen knows more about this but he is on holiday

flixr:
Ok, merging this now to fix it...
You can hash out the best solution with @fvantienen later..

PR #1469, opened 2015-12-08
Bebop i2c changes

martinmm:
Allow an external I2C bus connected through USB for the Bebop. Use the correct repeated-start for Transceive function for Linux targets as required by some I2C sensors.

### DIFF
--- a/conf/airframes/mm/bebop.xml
+++ b/conf/airframes/mm/bebop.xml
@@ -1,0 +1,234 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="bebop">
+
+  <firmware name="rotorcraft">
+    <target name="ap" board="bebop"/>
+
+    <target name="nps" board="pc">
+      <subsystem name="fdm" type="jsbsim"/>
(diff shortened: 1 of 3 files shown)


flixr:
Looks good to me, but can't test it...
The repeated start should not pose a problem for stuff that already works, right?

martinmm:
Works for me, can see all raw sensor messages. The external I2C shows a correct repeated start on the electrical/logical level. Would be nice if someone could cross check that.

OpenUAS:
I can cross check it, plz add used hardware details, parts, solution hickup and I can test this week.

martinmm:
@OpenUAS Would be great if you can test normal functionality, maybe focused on the sensors that are connected through I2C. We are changing/correcting the low-level I2C protocol for transceive
from START write STOP START read STOP
to START write START read STOP
and we should make sure the hw/sw is fine with that

OpenUAS:
OK will check normal ops

flixr:
@OpenUAS @gautierhattenberger anyone tested this as well?

gautierhattenberger:
It works for me

PR #1471, opened 2015-12-09
[build] Modules can be used like subsystems in airframe files

gautierhattenberger:
First important step towards merging modules and subsystems:
    - modules can be loaded from firmware section like subsystems.
    - last point includes filtering on target.
    - subsystems can be renamed to modules (even if conversion is not done).
    - completely backward compatible (for now).

Now the big question: shall we try to merge this before or after v5.8 stable release ?

My opinion is that before is better: it will let users (both people using stable and dev branches) to prepare their airframes to the new format that may become the standard in later versions.


flixr:
Nice!! This is indeed a great step forward!

Doesn't respect the targets specified in the makefile sections of modules anymore though. So e.g. for the video_thread.xml module, it will not only add the makefile section with target="ap" but also the makefile with nps target.

Another question: how do you specify the modules frequency if you migrate them all to the firmware section?

If we can indeed get this fully backwards compatible and robust, then I agree that it would be great to have in v5.8 already.

gautierhattenberger:
I'm look for a fix for target filtering, we forgot this case in our tests.
For the freq, you can still have an empty node `<modules main_freq="xxx"/>` if you want to set a non-default value.

flixr:
Ok, cool!

Also what happened to `settings_modules` in conf.xml?
Is the functionality to disable settings from modules now gone?
E.g. for the Microjet it is now empty, whereas before it had the modules with settings listed there (and you could see them in the Paparazzi Center).

gautierhattenberger:
??
I can still see the settings from modules and I can enable/disable them

flixr:
Hm... it must have somehow removed it from my conf first, probably because I forgot to close Paparazzi Center in time before rebuilding on the new branch...
So not an issue...

So the only feature subsystems now have over modules is that you can have them for different firmwares (with fallback to shared)...

gautierhattenberger:
It has no consequences for now until subsystems are actually converted. Before it happens, we'll had a `firmwares` attribute to the makefile section of modules on the same basis of targets.

flixr:
So is there any difference to before if you say `subsystem`?

What is the current search order for `module`?
Check if you find a module with that file name? If not fall back to subsystem?
Is the name of the module used anywhere?

flixr:
@dewagter @martinmm what do you guys think? Merge it ASAP?

A more general point: when we convert subsystems to modules, there is not really a distinction between mandatory and optional parts any more. Not sure if that matters regarding clarity for users....

gautierhattenberger:
Some answers:
- if you ask for a subsystem, the result is the same as before, but the subsystem has to exist
- if you ask for a modules with `load` in the `modules` node, same as before and the module has to exist
- if you ask for a module with `module` in `modules` node, same as `load` but with extended format possibilities:
 - `module name="bla_boo.xml"/>`
 - `module name="bla_boo"/>`
 - `module name="bla" type="boo"/>`
- if you ask for a module with `module` in `firmware` (or `target` inside a firmware), it will look for the module and fall-back to subsystem if module file doesn't exist, and you can used the same extended formats like above
- the search (and inclusion) order should be deterministic, but might be inverted, that's easy to fix in case
- the names are not used for now, but they will when we'll start adding the dependency mechanism (which will impact the inclusion order actually, but for a safer results normally)
- right now, how can you tell what is really mandatory anyway ? the dependency system might actually help to tell people what is missing in the end.

flixr:
Thanks for the detailed explanation!

And yes, an improved dependency system will help a lot :-)

For me this is good to merge...
Should we wait for feedback from @dewagter @martinmm or others?

gautierhattenberger:
it can probably wait 1 or 2 days for more comments

dewagter:
Looks great! I'm fine with a merge.
This starts to look more and more like a 6.0 release to me.

 - Do I understand correctly that subsystems (converted to modules) can now also have periodic functions?
 - Wonder how this will affect FBW which does not run modules at the end.

OpenUAS:
@gautierhattenberger If the testsuite is run against it and everything works, one could merge.Great initiative, thx!

gautierhattenberger:
@dewagter the easy conversion is to make a module with only the makefile part so the behavior is the same than before. But it is also possible of course to use the init, periodic and event capabilities of modules.
Concerning FBW, it can stay the same for now (being build from makefile type modules) but at some point everything could be done from modules, with a "simple" generic main (this last point as to be discussed and it will take time anyway)

PR #1472, opened 2015-12-10
fixed pocket stereo demo

rmeertens:
the pocketdrone now uses the normal stereocam library.


flixr:
Plz fix indentation...

rmeertens:
Fixed indentation! Also added my own conf to the TUDelft folder so other people can give these demonstrations here in the future.

OpenUAS:
@rmeertens Thx, note that the naming was changed just after your push, thus better run ./start.py once more on you local laptop (or ise unlink && ln -s on console)

PR #1473, opened 2015-12-10
Follow me based on stereo camera

rmeertens:
Last thing I want to add today: a follow algorithm that uses our stereo camera module. 
Just tested an demonstrated it, whoo.


sw/airborne/modules/stereocam/stereocam_follow_me/follow_me.c
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (C) Roland
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/follow_me/follow_me.c"
```
flixr: plz fix path

flixr: and no quotes needed...

sw/airborne/modules/stereocam/stereocam_follow_me/follow_me.c
```diff
@@ -0,0 +1,127 @@
...
+#include "navigation.h"
+
+#include "subsystems/datalink/telemetry.h"
+
+float ref_pitch = 0.0;
+float ref_roll = 0.0;
+float selfie_alt = 1.0;
+
+
+void follow_me_init()
```
flixr: why have an empty init?

gautierhattenberger:
it looks like the follow_me module is in two pull request with different names, which one should be kept ?
see #1439, some comments also apply for this PR.

rmeertens:
This one should be kept. I believe I fixed all comments right now. Let me know if there are more requests.

sw/airborne/modules/stereocam/stereocam_follow_me/follow_me.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (C) Roland
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/stereocam/stereocam_follow_me/stereocam_follow_me.h"
```
gautierhattenberger: quotes not needed

gautierhattenberger:
Can you add a few more lines to describe the module and merge/solve conflicts with paparazzi/master ?

Beside that it is okay for me

rmeertens:
Added a short description and removed the quotes!

flixr:
You did more than that.... added a `STEREO_VELOCITY` messages and changed the confs and flight plan...

flixr:
Why do we have that message now?

gautierhattenberger:
since it was only TUDelft stuff I assumed they were okay with the changes. If it is not the case, @rmeertens can you provide a valid conf in a new pull request ?

flixr:
That message is not used anywhere...

Please try to only include changes relevant to that specific pull request... check again before you push...
Also it's good practice to rebase/squash them in cases like this to not clutter the history needlessly...

rmeertens:
Welp, sorry guys. Looks like I already pushed the message of a module I planned to push to master next week :s next time I will also rebase. Is it ok for now to leave the message in although the module that uses it will be pushed next week?

flixr:
I'll remove that accidentally added stuff for now.
Also think that we can find a better message that can also be used to report velocity from other sensors...

Also please try to document all options in the module xml (like `STEREOCAM_FOLLOW_ME_USE_OPTITRACK`)...

rmeertens:
Will document it in the next push to master. What message would you like to use? Currently I plot the velocities in this message, and the velocities of optitrack to be able to compare them. I would like to keep the distinction between for example this measurement and the fused INS message.

flixr:
I think it would be nice if we added a slightly more generic message that basically reports a velocity estimate (e.g. also from optic flow or px4flow).
For that we could e.g. add an ID to indicate the "sensor" this is coming from....
Also some shorter field names like e.g. `vx`, `vy` `vz` would make sense (more consistent with other messages).
And lastly add the units and some doc on how to interpret the message (what frame are the velocities in, etc.)

PR #1474, opened 2015-12-11
motor order/spin direction changed

EwoudSmeur:
When updating to the newest firmware, the bebop2 motor order/spin direction changed again. Here is a fix, I flew with this today


flixr:
Would there be a way to check the firmware version and set this accordingly?

fvantienen:
This should be for everyone correct now since we had a prerelease version

martinmm:
works for me with 3.0.2 and 4.2.R.4

PR #1476, opened 2015-12-13
[gps] add fields_valid and remove GPS_USE_LATLONG

### LINKED ISSUE #641: GPS cleanup and better LTP origin handling (UTM vs. NED)

flixr:
add fields_valid bitfiled to gps struct and use that to decide whether a valid UTM pos is availabe instead of always computing UTM if GPS_USE_LATLONG is defined

should close #641


kirkscheper:
Loving this Felix!

flixr:
Tests to confirm all is still working as expected would be appreciated, I could only test the sim...
@gautierhattenberger do you think we should wait with this for after the v5.8 release?

gautierhattenberger:
This is a really nice change, but except testing some ublox module in the lab, I won't be able to do more before leaving for vacations. It doesn't really change much for end-users and it may have some side effects if we missed some stupid bugs, so maybe it is not wise for a stable release...

flixr:
Agreed, waiting for after the release...

PR #1478, opened 2015-12-13
[boards] naze32: default to 6 servos

flixr:
and if chosing RADIO_CONTROL_PPM_PIN as PA0 (RC_CH1):
disable servos 1 and 2, but keep numbering of other servos as on silkscreen

### DIFF
--- a/conf/airframes/untested/hex_naze32.xml
+++ b/conf/airframes/untested/hex_naze32.xml
@@ -12,7 +12,6 @@
 <airframe name="Hexcopter Naze32">
 
   <firmware name="rotorcraft">
-    <define name="USE_SERVOS_5AND6"/>
     <define name="PPRZ_TRIG_INT_COMPR_FLASH"/>
     <define name="PPRZ_TRIG_INT_COMPR_HIGHEST"/>
     <target name="ap" board="naze32_rev5">
--- a/sw/airborne/boards/naze32_common.h
+++ b/sw/airborne/boards/naze32_common.h
@@ -36,14 +36,14 @@
  *
  * sets on which pins the UARTs are connected
  */
-/* UART1 on main port */
+/* UART1 on main port, TX (PA9), RX (PA10) */
 #define UART1_GPIO_AF 0
 #define UART1_GPIO_PORT_RX GPIO_BANK_USART1_RX
 #define UART1_GPIO_RX GPIO_USART1_RX
 #define UART1_GPIO_PORT_TX GPIO_BANK_USART1_TX
 #define UART1_GPIO_TX GPIO_USART1_TX
 
-/* UART2 on RC I/O, tx pin 5/id "3", rx pin 6/id "4") */
+/* UART2 on RC I/O, TX (PA2) pin 5/id "3", RX (PA3) pin 6/id "4" */
 #define UART2_GPIO_AF 0
 #define UART2_GPIO_PORT_RX GPIO_BANK_USART2_RX
 #define UART2_GPIO_RX GPIO_USART2_RX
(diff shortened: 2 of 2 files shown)


flixr:
@martinmm what do you think of this?

martinmm:
Paparazi uses injected ADCs whereas the original Naze32 software does not. That leads to a different configuration. My idea was to have the exact same configuration (servos 1-4, PPM_IN on PA0) as Cleanflight/Baseflight so that you can easily switch between softwares without creating another confusing way of numbering and naming servos. The hex is not the standard configuration so have 6 servos optionally.

flixr:
At least on my board (flip32) the servos are numbered on the silkscreen... is that also the case for the "official" naze32?
If yes, then I think it makes sense to keep that numbering in Paparazzi... regardless of whether all of them are active or not.

So which are the servos 1-4 in cleanflight/baseflight? Also the PPM input defaults to PA7 right now...

I have no strong opinion on what the defaults should be though...

martinmm:
PA0 needs a manual configuration, thought it was done through SERVOS_5AND6. PA0 is the Cleanflight PPM_IN.

For the motors they have two numberings, I used the "servo" one that maps 1-4 to 3-6.

http://www.abusemark.com/downloads/naze32_rev3.pdf (rev.5 user manual)

flixr:
Looks to me like following the "standard mode" numbering will create less confusion as that seems to be the - well - standard...
That would also mean that the servo numbers don't change if you enable 6 output (SERVO_1AND2 according to that numbering).
Also it would better fit with the numbering of some clones like the flip32.

So my suggestion would be to default PPM input to PA0 (as in cleanflight) and have servos 1 and 2 disabled by default.
Then if you `<configure name="RADIO_CONTROL_PPM_PIN" value="PA7"/>` automatically enable servos 1 and 2...

martinmm:
Well, then you have to connect servo output 3 on the Naze32 to motor number 1. Whatever we do, it is confusing...

But right, let's always keep the numbering 1-6 so that we do not have to switch that. Then have some wiki entry that tells you what was in the header file. And PA0 is standard PPM_IN but i would switch that with the USE_SERVO_1AND2 automatically or manually with the PPM_CONFIG.

flixr:
You will actually always have to specify the servo `no` from 0 on in the servos section.
The name you give it in the board file has really no relation to the `no` you set in the airframe file...
And anfortunately you can't use the defines for the `no` in the servos section, since this needs to be a number so the generator can find out the max number...

Since f59bdeb6cc5174e012be0d6930c9998f71f9499e the `PPM_CONFIG` is set according to the `RADIO_CONTROL_PPM_PIN` you configure.
So only a `<configure name="RADIO_CONTROL_PPM_PIN" value="PA7"/>` would be enough to enable the servos 1 and 2... or do you want two ways on how to enable them (and remap the PPM input)?

martinmm:
There is no good reason to switch the PPM_IN from PA0 to something else in the first place. You might want to use 6 servos and then it has to be PA7 as PA0 cannot be used in that case.
The only reason might be that you switch to PA7 anyway, then a manual possibility to switch to PA7 helps.

flixr:
So are you suggesting we remove the options to configure the PPM_PIN (added with f59bdeb6cc5174e012be0d6930c9998f71f9499e) again?

Or make a configure option `USE_SERVOS_1AND2` that will then automatically configure the PPM_PIN to PA7?
That would mean you have two ways to end up with the same result:
`<configure name="RADIO_CONTROL_PPM_PIN" value="PA7"/>` or `<configure name="USE_SERVOS_1AND2" value="TRUE"/>`

flixr:
@martinmm could you please clarify what you think would be be the best way to configure this?
E.g. simply do it via `<configure name="RADIO_CONTROL_PPM_PIN" value="PA7"/>` as in flixr@9ba4315eb7aa5a806f840ec0161757e54198d8ba

martinmm:
Sounds good

PR #1479, opened 2015-12-15
Rate controller: sp in degrees/s

EwoudSmeur:
This is a small improvement so you actually know what you fill in the airframe file for the rate setpoint. Changing all the airframe files seems like a bit of a pain though. Adding unit="deg/s" should be simple, but is there a command to divide the rate setpoint by 1.0/180.0*PI*2^12 everywhere?


flixr:
You can probably just set them to the same values for all airframe files, rate mode hasn't been used much anyway...

On a side note: when you improve rate mode, you might as well do it in float right away... not really any significant gain to use fixed point here...

EwoudSmeur:
ok, that makes a lot of sense to me :D

I managed to string replace most of the setpoints in the airframe files, so I will add this later

PR #1482, opened 2015-12-16
Rate integrator

EwoudSmeur:
The integrator was doing very little for rate control as it was scaled to a very small value and it accumulated very fast up to the maximum. With these changes it takes longer to reach the maximum


flixr:
Wouldn't it make more sense to directly convert it to floating point instead of fiddling with the scales?

EwoudSmeur:
In the meantime, can we just merge this? It works now...

gautierhattenberger:
Can you rebase on master in your development branch? It can't be merged directly anymore.

PR #1485, opened 2015-12-18
fixed br optitrack file, fixed my own conf

rmeertens:
Removed the non existing module. Also changed my own control panel.


flixr:
what about the module in the tudelft_ardrone2_oa_clint_roland airframe?

PR #1486, opened 2015-12-18
[stereocam] New module for sending measurements from external stereoc…

knmcguire:
…amera to state, with seperate conf files

This has been tested on the lady lisa s

### DIFF
--- a/conf/modules/stereocam_stereocam2state.xml
+++ b/conf/modules/stereocam_stereocam2state.xml
@@ -1,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="stereocam2state" dir="stereocam/stereocam2state">
+  <doc>
+    <description>This module sends the data retreived from an external stereocamera modules, to the state filter of the drone. This is done so that the guidance modules can use that information for couadcopter</description>
+  </doc>
+  <header>
+    <file name="stereocam2state.h"/>
+  </header>
+  <init fun="stereo_to_state_init()"/>
+  <periodic fun="stereo_to_state_periodic()"  start="stereo_to_state_start()" stop="stereo_to_state_stop()" autorun="TRUE"/>
+  <makefile>
+    <file name="stereocam2state.c"/>
+  </makefile>
+</module>
+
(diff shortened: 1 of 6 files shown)


conf/modules/stereocam_stereocam2state.xml
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="stereocam2state" dir="stereocam/stereocam2state">
+  <doc>
+    <description>This module sends the data retreived from an external stereocamera modules, to the state filter of the drone. This is done so that the guidance modules can use that information for couadcopter</description>
```
flixr: First line of description should be the brief description "name" of the module (see https://wiki.paparazziuav.org/wiki/Modules#In_the_module_file) as it is used in the generated docs at http://docs.paparazziuav.org/latest/onboard_modules.html

conf/modules/stereocam_stereocam2state.xml
```diff
@@ -0,0 +1,16 @@
...
+  </doc>
+  <header>
+    <file name="stereocam2state.h"/>
+  </header>
+  <init fun="stereo_to_state_init()"/>
```
flixr: If you don't need the init and start stop functions, please remove them and only keep the periodic that you actually use...

sw/airborne/modules/stereocam/stereocam2state/stereocam2state.c
```diff
@@ -0,0 +1,139 @@
...
+
+#if STATE_MEASURE_OPTICFLOW
+  // Calculate velocity in body fixed coordinates from opti-track and the state filter
+  struct NedCoor_f coordinates_speed_state;
+  coordinates_speed_state.x = stateGetSpeedNed_f()->x;
+  coordinates_speed_state.y = stateGetSpeedNed_f()->y;
+  coordinates_speed_state.z = stateGetSpeedNed_f()->z;
+
+  struct NedCoor_f opti_state;
+  opti_state.x = (float)(gps.ecef_vel.y) / 100;
```
flixr: On a sidenote: accessing the GPS struct directly is kind of deprecated and won't be possible any more soon(-ish)...
Should be changed to ABI subscription of the GPS message....

PR #1491, opened 2015-12-21
Attempt to clean and remove ifdefs from main_fbw

podhrmic:
Make the fbw code more clean and legible, also logically group *ifdef* directives. Also, moved includes into the header file where they belong.
First attempt, let me know what you think. The next step would be to do something similar with `main_ap` code

Also I applied the correct formatting on both files.

PS: Sorry about the python stuff, it got included after I did `rebase` :( It wont make any problems after merge though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+Paparazzi v5.8.0_stable
+=======================
+
+Stable version release.
+
+- module: nav_catapult: avoid conditional directives
+  [#1454] (https://github.com/paparazzi/paparazzi/pull/1454)
+- module: gps_ubx_ucenter fix autobaud and autoconf for ublox-8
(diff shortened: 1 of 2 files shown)


flixr:
Can you please rebase it again to remove the python commits?
You can use `git rebase -i paparazzi/master` and then simply remove the commits that you don't want if it doesn't already recognize that they are already applied...

flixr:
Like the cleanup :-)
But I disagree that all includes belong into the header, quite the contrary... only the includes needed for the function declarations in the header should be there. All includes that are only needed in the c file should only be added there...
Adding unnecessary includes to the headers makes it harder to re-use, increases compilation time and pollutes the "namespace" of other files that might include that header but never need any of the other includes.

podhrmic:
fixed in #1492 (couldn't get it done in this pull request). Let me know what to do you think.

PR #1496, opened 2015-12-22
Stereocam updates and ins_int.c changes

knmcguire:
- Updates to stereocam2state related conf files
- in ins_int.c 
       -  at  vel_est_cb() the velocity is intergrated to position for the guidance module
       -  at gps_cb() USE_GPS define is used around the update function. I figured out that, even though USE_GPS is false in the airframe file, GPS measurements are still used for calculating the state (which is weird right?). So this will fix that problem for now.

### DIFF
--- a/conf/TUDELFT/tudelft_KM_control_panel.xml
+++ b/conf/TUDELFT/tudelft_KM_control_panel.xml
@@ -68,8 +68,8 @@
         <arg flag="-speech" />
       </program>
       <program name="NatNet">
-        <arg flag="-ac" constant="1"/>
-        <arg flag="12"/>
+        <arg flag="-ac" constant="7"/>
+        <arg flag="114"/>
         <arg flag="-small"/>
         <arg flag="-tf" constant="5"/>
       </program>
--- a/conf/airframes/TUDELFT/tudelft_ladylisa_bluegiga_stereoboard.xml
+++ b/conf/airframes/TUDELFT/tudelft_ladylisa_bluegiga_stereoboard.xml
@@ -187,8 +187,8 @@
  <define name="USE_GPS_HEADING" value="FALSE"/>
 
  <section name="GUIDANCE_H" prefix="GUIDANCE_H_">
-   <define name="PGAIN" value="40"/>
-   <define name="DGAIN" value="10"/>
+   <define name="PGAIN" value="10"/>
+   <define name="DGAIN" value="40"/>
    <define name="IGAIN" value="0"/>
    <define name="REF_MAX_SPEED" value="0.5"/>
    <define name="USE_REF" value="FALSE"/>
--- a/conf/airframes/TUDELFT/tudelft_ladylisa_bluegiga_stereoboard.xml
+++ b/conf/airframes/TUDELFT/tudelft_ladylisa_bluegiga_stereoboard.xml
@@ -181,10 +181,10 @@
     <define name="H_X" value="0.39049610"/>
     <define name="H_Y" value="0.00278894"/>
     <define name="H_Z" value="0.92060036"/>
-    <define name="USE_GPS_HEADING" value="1"/>
+    <define name="USE_GPS_HEADING" value="0"/>
   </section>
-<define name="USE_GPS" value="FALSE"/>
- <define name="USE_GPS_HEADING" value="FALSE"/>
+
+
 
  <section name="GUIDANCE_H" prefix="GUIDANCE_H_">
    <define name="PGAIN" value="10"/>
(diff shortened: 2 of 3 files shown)


flixr:
Using a velocity as if it was a (delta) position doesn't result in anything meaningful, the scale is not correct and it strongly depends on the dt between the updates.

If you want to do a simple integration here, you need to do something like `pos = pos + vel * dt` and hence remember the last time you got a velocity estimate to calc the dt.
I wonder though what the reason is that you don't use the horizontal filter, which already should do that for you...

Also if `USE_GPS` is false, then `ins_int_update_gps` is already an empty function...

knmcguire:
Yes you are right about the intergration, let me do a commit where I implement pos = pos + vel * dt.

About the horizontal filter, right now only hf_float exists, which does not do very well on the lisa-s. So there are plans within the lab to make a better horizontal filter and based on integers.  Untill that one is finished, I will use it on the lisa s but for now I'll be fine with sending it right through to the state. 

About USE_GPS, yes I know that, that is why I found it really strange that it still sends when I put USE_GPS on false. Look at the plot that I have made of my data, it is low resolution but I think that it shows clearly that it is switching between velocity estimate and velocity from gps (in this case optitrack)
optitrack_opticalflow. After my last commit, this does not happen anymore

flixr:
I still don't see how adding `#if use_GPS` in `gps_cb` makes any sense... also note that you use a lower case `use_` here, since that is not defined it is always false.
How did you configure/define `USE_GPS`? To disable it, use either `<define name="USE_GPS" value="0"/>` or `<define name="USE_GPS" value="FALSE"/>`

knmcguire:
I implemented your first comment, which I based on the way of accel_cb()

About the gps measurement,

knmcguire:
My bad, hit close by accident but it was not the intention,

about your comments:
- I'll implement the velocity intergration outside the #if in the next commit
- In line 186 of  conf/airframes/TUDELFT/tudelft_ladylisa_bluegiga_stereoboard.xml (the airframe I use I define USE_GPS as <define name="USE_GPS" value="FALSE"/>.  (I do see that USE_GPS_HEADING is multiple times defined, however that define is not used in ins_int, so that shouldn't be the problem I think.)

flixr:
To debug you could simply add a `#warn gps enabled/disabled` or so where the ins_int_update_gps is implemented to make 100% sure that you get the empty function...
I really see no possible use of `#if use_GPS` in the gps_cb....

knmcguire:
okay I'll try to look more closely where the real problem is then. Hopefully I'll find it soon

knmcguire:
Yes you were right. If I put a #warning in  ELSE of USE_GPS, it shows nothing while compiling. So somewhere it still puts USE_GPS on true. If I define this within the subsystem environment of ins, it does show the warning, together with a lot of other redefine warnings of USE_GPS...

If USE_GPS is used for some many other functions, maybe best to make a new define for this? like USE_GPS_INS ?

flixr:
I don't see the point of adding yet another define... just checked, `USE_GPS` is defined if you add a gps subsystem... so you seem to have added one, but don't plan to use it? Or do you need it for something else? If not, just don't add a GPS...

Also note that you can "simply" disable ABI callbacks by specifying `ABI_DISABLE` in `AbiBindMsgX`, so it would make more sense to add a define `INS_INT_GPS_ID` and use that instead of `ABI_BROADCAST` so you can configure the GPS senders that are accecpted...

knmcguire:
I do want to use the GPS messages but I do not want it to be used to calculate the state. I'm working on a indoor navigation scheme that should work without an optitrack system and that is using the existing horizontal guidance control which uses this state. I could write my own guidance module which uses the velocity estimate directly, but that would not make any sense too. 

 However in order to validate whatever I'm working on, I do need GPS/optitrack data logged at the same system clock of the autopilot. That is why I want to specifically block GPS to have any effect on the state.

Now I made the following lines on the top of ins_int.c. 
#ifndef INS_INT_GPS_ID
#if USE_GPS
#define INS_INT_GPS_ID ABI_BROADCAST
#else
#define INS_INT_GPS_ID ABI_DISABLE
#endif
#endif

 However, I saw in ins_int_init() for sonar that that the binding is prevented by the USE_SONAR define. Do you prefer the first or second solution?

flixr:
IMHO just
#ifndef INS_INT_GPS_ID
#define INS_INT_GPS_ID ABI_BROADCAST
#endif
is enough

knmcguire:
okay, than I'll define the sender ID in the airframe file in that case

flixr:
You can't really define `USE_GPS` to false in your airframe file while using a GPS subsystem anyway...

knmcguire:
I believe I removed <define name="USE_GPS" value="FALSE"/> (line 186) already in the airframe file in the last commit.

The only thing I'm specifying extra in my airframe file now is INS_INT_GPS_ID in the subsystem INS

PR #1480, opened 2015-12-15
[gps] fix ubx ucenter autobaud and autoconf

gautierhattenberger:
should fix #1459 
tested on UBX M8M module with default config at startup

PR #1497, opened 2015-12-22
[ucenter] BUGFIX ucenter change baudrate

dewagter:
baudrate was configured to the value currently in the GPS_PORT, but after an auto-baudrate the baudrate was always the value from the device, so the baudrate was never changed.


flixr:
Seems to me this was already fixed by @gautierhattenberger in #1480 ?

dewagter:
shit, then I lost time for nothing.
In any case, this was the reason of my ublox8 errors.

flixr:
To prevent that I explicitly referenced you in gautiers pull request, seems you missed that...
Closing as duplicate of #1480

dewagter:
thanks.
yes, you can close/delete

PR #1501, opened 2015-12-28
[python] ivy_msg_interface: find message class by name

flixr:
and don't treat sender_name (first string in ivy message) as msg_class

### DIFF
--- a/sw/lib/python/ivy_msg_interface.py
+++ b/sw/lib/python/ivy_msg_interface.py
@@ -72,28 +72,21 @@ class IvyMessagesInterface(object):
             return
 
         # check which message class it is
-        # pass non-telemetry messages with ac_id 0
-        if data[0] in ["sim", "ground_dl", "dl"]:
-            if self.verbose:
-                print("ignoring message " + larg[0])
-                sys.stdout.flush()
+        msg_name = data[1]
+        msg_class, msg_name = messages_xml_map.find_msg_by_name(msg_name)
+        if msg_class is None:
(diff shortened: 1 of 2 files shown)


flixr:
@gautierhattenberger do you know why some datalink messages from server are sent with "sender_id" (first string) `ground` and some with `ground_dl`?
In server there is a comment that it is a "hack": https://github.com/paparazzi/paparazzi/blob/v5.8.0_stable/sw/ground_segment/tmtc/server.ml#L740

gautierhattenberger:
From what I can see, `ground` id is used for messages from server to GCS while `ground_dl` is used for messages from server to link. At least I think it was the original idea, maybe it is not completely true now.

flixr:
So how do you think we should handle it? Also use `ground_dl` for the `ACINFO` datalink message meant for link?
Or rather directly set `server` as sender_name?
I don't see where it would be needed to have "ground" or "ground_dl" or even why "ground_dl" was even added in the first place in 23f981b5e0ea7ab24b8d857811f15ec4dda0bb36

flixr:
At least with this PR, the ivy_msgs_interface should handle the messages correctly and not rely on the sender_id anymore...

gautierhattenberger:
ACINFO is going from server to link, so according to the initial idea, using `ground_dl` id is fine (seems to be the case currently).
But this PR is not about this, and your changes make things much better.

flixr:
Thx, ACINFO is currently sent with `ground`, I'll change it to `ground_dl` to be at least consistent with the other datalink messages...

PR #1502, opened 2015-12-28
[rotorcraft] start adding velocity commands for guided mode

flixr:
move at specified NED velocity in meters/sec with absolute heading

Note that this doesn't add a real dedicated velocity guidance loop, but rather updates the position reference setpoint from a given velocity.

Feedback on API welcome... not particularly happy with it, but can't think of a nicer one right now...


kirkscheper:
This is useful for what I am working on right now. The api looks to be inline with the other modes (not sure how much that says haha). I can flight test this on Wednesday if you like.

flixr:
It works fine in simulation... but real flight tests are always appreciated :-)

Why this isn't merged yet is mostly because I wanted some more feedback on the "higher level API" functions and their naming/arguments: So `autopilot_guided_move_ned` and actually all other already added `autopilot_guided_goto_*` ones as well...

kirkscheper:
first thought would be if go to a position is 'goto' then logically set speed would be 'go' instead of move, but maybe it isn't actually any clearer.

flixr:
Yeah, same here... IMHO 'goto' is good for position, but 'move' or 'go' are not really that clear...
Also thought about something like 'move_at_vel', but it gets long...
Maybe `autopilot_guided_velocity` would be more clear, but then it's a different naming scheme (not an action anymore).
Later we will probably have other combinations like setting horizontal position and vertical velocity...
How would you call that in an "action" naming scheme?

flixr:
@martinmm @dewagter @gautierhattenberger maybe you have some good ideas on a API/naming scheme here? Especially considering that later we might want to omit parameters (e.g. not change anything in height) or mix them in other combinations.

hooperfly:
Hi @flixr,

I'm currently using the guided mode as part of the Flying Robot commander stuff. Something that might be nice is to have some verbiage around what each frame mode/velocity command and how they are used. Given the following code snippet:
```
      switch (flags) {
        case 0x00:
        case 0x02:
          /* local NED position setpoints */
          autopilot_guided_goto_ned(x, y, z, yaw);
          break;
        case 0x01:
          /* local NED offset position setpoints */
          autopilot_guided_goto_ned_relative(x, y, z, yaw);
          break;
        case 0x03:
          /* body NED offset position setpoints */
          autopilot_guided_goto_body_relative(x, y, z, yaw);
          break;
        case 0x70:
          /* local NED with x/y/z as velocity and yaw as absolute angle */
          autopilot_guided_move(x, y, z, yaw);
          break;
        default:
          /* others not handled yet */
          break;
      }
```
It might be nice to have the bitmask values abstracted for each case vs. the raw hex bitmask? Does that seem reasonable? That way the values are encapsulated in a header file and easily altered if the need arises. As well as a nice place to document each bitmask similar to how you have documented the above  switch statement.

flixr:
@hooperfly sure, we can add defines for the common combinations used in the switch statement.
But you probably wouldn't use them anywhere else anyway.
For onboard usage (modules, flightplan) you would directly call the `autopilot_guided_x` functions.
For setting this via datalink the description of the message has to be clear enough... is it?
http://docs.paparazziuav.org/latest/paparazzi_messages.html#GUIDED_SETPOINT_NED

hooperfly:
@flixr maybe some semi-colons in the appropriate spots or breaking the flags out line by line will help with the documentation:

Currently:
```
Set vehicle position or velocity in NED. Frame can be specified with the bits 0-3 in the flags field: 0x0: LOCAL_NED 0x1: LOCAL_OFFSET_NED 0x2: BODY_NED 0x3: BODY_OFFSET_NED
```

Maybe:
```
Set vehicle position or velocity in NED. Frame can be specified with the bits 0-3 in the flags field: 
0x0: LOCAL_NED 
0x1: LOCAL_OFFSET_NED 
0x2: BODY_NED 
0x3: BODY_OFFSET_NED
```
... and based on the current switch statement, LOCAL_NED and BODY_NED are the same.

flixr:
It's not in one line in the [messages.xml](https://github.com/paparazzi/paparazzi/blob/v5.8/conf/messages.xml#L2441-L2448)
The lines get joined by the generator https://github.com/paparazzi/paparazzi/blob/master/sw/tools/doxygen_gen/gen_messages_doc.py#L32
There is probably a nicer way to prevent this (probably explicitly specifying that only spaces and tabs should be stripped)
Fixes for this welcome ;-)

Regarding position LOCAL_NED and BODY_NED:
Absolute position doesn't really make much sense in BODY_NED, so we treat it the same as LOCAL_NED.
Same as in mavlink...

flixr:
@kirkscheper did you get a chance to test this?

flixr:
Unless someone has a better idea on naming, I suggest we merge this now...
It's working and already used anyway...

kirkscheper:
Last wee I got really busy so couldn't. Will try again this week.

PR #1504, opened 2015-12-29
Chibios integration

podhrmic:
Merging current work based on https://github.com/podhrmic/paparazzi/tree/rt_chibios into latest master.

The code is split into appropriate tasks, the tasks have different priorities and allocated certain memory space. As a result the code will never be fully compatible with non-rt paparazzi (some extra defines/guards are needed), but having RTOS to run one big task with the whole autopilot doesn't really make sense.

Tested with Vectornav in HITL, sadly I don't have a board with working IMU ( @flixr @gautierhattenberger maybe you would be willing to test the baro/imu test cases?). 

For GPS subsystems there needs to be defined another thread.

Features:
- working drivers (uart, adc, i2c, spi, gpio)
- working compilation with DEBUG/RELEASE options
- test cases for individual drivers
- working example of fixedwing RT code
- switched to new ChibiOS 3.0
- works on Lisa MX only (for now)

Drawbacks:
- currently fails to compile old Apogee example, because that one expects ChibiOS 2.6 (@gautierhattenberger are you still using the chibios3-libopencm3 hybdrid? )

Benefits:
- faster and more efficient code (less CPU usage)
- potentially safer code (tasks have different priorities)

Drawbacks:
- longer compilation time
- larger executable (although I haven't measured by how much)

The functionality is there, and the basic structure too, let me know your feedback!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,4 +18,4 @@
 	url = https://github.com/paparazzi/mavlink.git
 [submodule "sw/ext/chibios"]
 	path = sw/ext/chibios
-	url = https://github.com:ChibiOS/ChibiOS.git
+	url = https://github.com/ChibiOS/ChibiOS.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,4 +18,4 @@
 	url = https://github.com/paparazzi/mavlink.git
 [submodule "sw/ext/chibios"]
 	path = sw/ext/chibios
-	url = git@github.com:ChibiOS/ChibiOS.git
+	url = https://github.com:ChibiOS/ChibiOS.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "sw/ext/fatfs"]
 	path = sw/ext/fatfs
 	url = https://github.com/enacuavlab/fatfs.git
-[submodule "sw/ext/chibios"]
-	path = sw/ext/chibios
-	url = https://github.com/mabl/ChibiOS.git
 [submodule "sw/ext/libzbar"]
 	path = sw/ext/libzbar
 	url = https://github.com/paparazzi/libzbar
@@ -19,3 +16,6 @@
 [submodule "sw/ext/mavlink"]
 	path = sw/ext/mavlink
 	url = https://github.com/paparazzi/mavlink.git
+[submodule "sw/ext/chibios"]
+	path = sw/ext/chibios
+	url = git@github.com:ChibiOS/ChibiOS.git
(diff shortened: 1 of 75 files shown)


flixr:
The the chibios3-libopencm3 hybrid is used by quite a few people for logging...

How does this compare to the enac/rt_chibios branch?

.gitmodules
```diff
@@ -19,3 +16,6 @@
 [submodule "sw/ext/mavlink"]
 	path = sw/ext/mavlink
 	url = https://github.com/paparazzi/mavlink.git
+[submodule "sw/ext/chibios"]
+	path = sw/ext/chibios
+	url = git@github.com:ChibiOS/ChibiOS.git
```
flixr: Please use https URLs for the submodules, so it works without ssh keys (as on the CI server) and through a proxy

gautierhattenberger:
Yes, I using the logging a lot, my work on the enac/rt_chibios branch is mostly to port the logging to a pure chibios arch.
It seems that my branch is doing a bit more on dynamic settings of pins (I need that for Apogee's AUX pins) and pwm_input driver. You have made some necessary work on state with mutex that we should definitely keep. But for the straight splitting of the code into threads, I think it is a bit early for master. Your solution is efficient indeed but requires a lot of specific code, I'm still hoping to find a cleaner global integration.
I understand that some drivers, especially UART, is not efficient at all in my branch, but I'm confident I can make it better just by pushing several bytes instead of one-by-one.

podhrmic:
For easier access, here is the link to enacuav chibios branch: https://github.com/enacuavlab/paparazzi/commits/rt_chibios

podhrmic:
Just FYI we now have fully functional RT paparazzi for both fixedwing and rotorcraft. I added a better set of debugging settings, and cleaned up the code a bunch. 

However, it is still the same philosophy of multiple threads that are not used in non-rt paparazzi (so there is a bunch of extra code and ifdefs). It is not that much, since the chibios specific code is (besides a few exceptions) in separate files that are not included if you are compiling regular paparazzi. 

So imho the overhead is not that bad. Again, I am more than happy to integrate this code into master, if you are OK that the rt paparazzi isn't 100% identical. We have the code based on v5.8, so it would be easier to integrate it now, before master branch changes too much.

Let me know what you think. If you re certain that this is not the way the rt paparazzi should look like, then it will be probably up to @gautierhattenberger to integrate some future version of rt paparazzi, since we wont be able to spend as much time on sw development in the future as we do now.

sw/airborne/arch/chibios/mcu_periph/uart_arch.h
```diff
@@ -0,0 +1,80 @@
...
+#define B19200   19200
+#define B38400   38400
+#define B57600   57600
+#define B115200  115200
+#define B230400  230400
+#define B921600  921600
+#define B100000  100000
+#define B3000000 3000000
+
+#define ch_uart_receive_downlink(_port, _flag, _callback, _arg) {   \
```
flixr: should probably be a function and follow code style...

gautierhattenberger:
closed in favor of #1699

PR #1506, opened 2016-01-06
[gps_datalink] update small message protocol

kirkscheper:
Removed USE_DATALINK_SMALL definition. Local reference frame now initialized from flight plan.
Updated flight plans that used USE_DATALINK_SMALL.
Added z speed and higher resolution heading to message.
Added gps z speed to ins vertical filer.
Ran fix_code_style on natnet2ivy.c.


flixr:
Looks basically good to me, like that you got rid of USE_DATALINK_SMALL ;-)

Some comments inline in the changes...

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -27,35 +27,19 @@
  * GPS structure to the values received.
  */
 
+#include "messages.h"
+#include "generated/airframe.h"           // AC_ID
```
flixr: AC_ID doesn't seem to be used in this file...

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -27,35 +27,19 @@
  * GPS structure to the values received.
  */
 
+#include "messages.h"
```
flixr: messages.h and downlink.h don't seem to be needed...

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -66,70 +50,67 @@ void gps_impl_init(void)
...
-  enu_speed.z = 0;
-
-  // printf("ENU Speed: %u (%d, %d, %d)\n", speed_xy, enu_speed.x, enu_speed.y, enu_speed.z);
 
-  ecef_of_enu_vect_i(&gps.ecef_vel , &tracking_ltp , &enu_speed);
+  ecef_of_enu_vect_i(&gps.ecef_vel , &ltp_def , &enu_speed);
   SetBit(gps.valid_fields, GPS_VALID_VEL_ECEF_BIT);
 
-  gps.hmsl = tracking_ltp.hmsl + enu_pos.z * 10; // TODO: try to compensate for the loss in accuracy
+  gps.ned_vel.x = enu_speed.x;
```
flixr: I guess this should be `gps.ned_vel.x = enu_speed.y;`?
NED and ENU have x and y switched...

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -374,6 +378,7 @@ void ins_int_update_gps(struct GpsState *gps_s)
 
 #if INS_USE_GPS_ALT
   vff_update_z_conf(((float)gps_pos_cm_ned.z) / 100.0, INS_VFF_R_GPS);
+  vff_update_vz_conf(((float)gps_speed_cm_s_ned.z) / 100.0, INS_VFF_VZ_R_GPS);
```
flixr: Maybe it should be possible to enable/disable the velocity update separately from the alt update?

On a sidenote: it might be nice to change these defines to variables, so you can change this via settings during runtime and just use the defines to set the initial value at init...

gautierhattenberger: that would make sense since since velocity or altitude could come from different sensors or at different speed after all

kirkscheper:
No prob with the separate define for alt and alt_speed but do I really have to add the settings for the noise? Not so interesting for me. Can I just make a comment that it is a todo? I assume you can also do that for the sonar...

flixr:
I didn't mean settings for the noise, but rather make the update of alt and vel not only configurable via a define, but via a runtime variable...
But it look ok to me for now...
@gautierhattenberger if you are fine with it as well, merge it...

PR #1507, opened 2016-01-06
[datalink] update bluegiga protocol and add rssi abi message

kirkscheper:
Added abi message for rssi from bluegiga bluetooth module.
Simplified the bluetooth check and parse to pprz_check and parse.
Updated bluetooth the recent build of firmware (telemetry fully backwards compatible).
Update bluetooth ground code.

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -73,7 +73,7 @@
       <field name="noise" type="float"/>
     </message>
     
-    <message name="BLUEGIGA_RSSI" id="13">
+    <message name="RSSI" id="13">
       <field name="ac_id" type="uint8_t"/>
       <field name="source_strength" type="int8_t" unit="dB"/>
       <field name="rssi" type="int8_t" unit="dB"/>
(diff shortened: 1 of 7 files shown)


flixr:
Hi Kirk,
Some questions:
- What is the BLUEGIGA_RSSI message intended for? Doesn't seem to be used yet...
- Why does it (need to) contain the ac_id?

Just wondering if it wouldn't make sense to make a more generic RSSI message...

gautierhattenberger:
I also think the RSSI message should be generic. It could be used with some xbee modules too

kirkscheper:
@flixr The rssi message should contain the signal strength and broadcast strength (if available) from the message received and an rssi was available. ac_id is useful for drone2drone coms that I and a student are working on. It is currently not used but will be shortly to help estimate the distance between a broadcaster and a receiver. This can of course also be used with the ground station (with ac_id 0).

I have no problem making that just RSSI...

flixr:
Looks ok to me now, @gautierhattenberger if you are also happy now, feel free to merge it...

PR #1508, opened 2016-01-06
[imu] add support of mpu9150 on Apogee

gautierhattenberger:
- driver for ak8975 mag
- only working in passthrough mode
- used with apogee v1 imu when mpu9150 is mounted instead of the mpu6050

#1498 #587

### DIFF
--- a/conf/firmwares/subsystems/shared/imu_apogee_mpu9150.makefile
+++ b/conf/firmwares/subsystems/shared/imu_apogee_mpu9150.makefile
@@ -1,0 +1,16 @@
+# Hey Emacs, this is a -*- makefile -*-
+#
+# Apogee IMU
+#
+
+include $(CFG_SHARED)/imu_apogee.makefile
+
+IMU_APOGEE_MPU9150_CFLAGS = -DAPOGEE_USE_MPU9150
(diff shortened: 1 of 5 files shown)


gautierhattenberger:
@philipan You can test this on your apogee. At least it is working on mine.

philipan:
@gautierhattenberger Thank you so much!  This is really great! It works. Interestingly also if I keep ```<subsystem name="imu"     type="apogee"/>``` I attach my airframe file ...

[AULA_FUNJET_I_01.xml.txt](https://github.com/paparazzi/paparazzi/files/80248/AULA_FUNJET_I_01.xml.txt)

gautierhattenberger:
hu... that's hardly possible since the driver is not compiled then... are you sure you completely rebuild and re-flash the code?

gautierhattenberger:
merged with 26ebe9daef3055c8680bc4a6de9d40c5fc0fbeda

philipan:
got it: it is only possible if I first compiled with ```apogee_mpu9150``` and then with ```apogee``` for the same ```A/C``` (so that make keeps the driver even if I said make clean at the PPRZ-Centre). If I create a new ```A/C```, ```<subsystem name="imu" type="apogee"/>``` gives me zero for all air_data as expected. However, it works now! Thx!

PR #1509, opened 2016-01-06
[rotorcraft] add traffic_info to rotorcraft

kirkscheper:
Added drone to drone messaging to datalink.
Added ac traffic logging capability (TRAFFIC_INFO should be defined in airframe file).
Works with bluegiga (although I use a smaller message which isn't here yet)

### DIFF
--- a/sw/airborne/subsystems/navigation/traffic_info.c
+++ b/sw/airborne/subsystems/navigation/traffic_info.c
@@ -26,9 +26,12 @@
  *
  */
 
-#include <inttypes.h>
 #include "subsystems/navigation/traffic_info.h"
-#include "generated/airframe.h"
+
+//#include "subsystems/navigation/common_nav.h"
+#include "generated/airframe.h"     // AC_ID
+#include "math/pprz_geodetic_int.h"
+#include "math/pprz_geodetic_float.h"
 
 uint8_t acs_idx;
 uint8_t the_acs_id[NB_ACS_ID];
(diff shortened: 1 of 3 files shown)


flixr:
Enabling generation of get_macros for telemetry again will break support for multiple variable length arrays again, see #1385

sw/airborne/subsystems/navigation/traffic_info.c
```diff
@@ -46,3 +49,45 @@ struct ac_info_ *get_ac_info(uint8_t id)
...
+  if (acs_idx < NB_ACS) {
+    if (id > 0 && the_acs_id[id] == 0) {
+      the_acs_id[id] = acs_idx++;
+      the_acs[the_acs_id[id]].ac_id = id;
+    }
+
+    struct EcefCoor_i ecef_pos;
+    struct LlaCoor_i lla = {.lat = lat, .lon = lon, .alt = alt};
+    ecef_of_lla_i(&ecef_pos, &lla);
+    the_acs[the_acs_id[id]].east = (float)ecef_pos.x * 100;
```
gautierhattenberger: `ecef.x` is clearly not an `east` value, same for other parameters. This could be confusing in a future use. The structure should be probably extended to use proper position and speed formats.

kirkscheper:
How about this @flixr, only for printing macros, we just skip the fields after the variable length array. In the (far) future when it becomes necessary to read multiple arrays from a telemetry message onboard we can revisit this. How does that sound :)

gautierhattenberger:
Another option: if the message is aligned without variable length array, use old macro style (maybe they even can be improved), if not, use a new type of macro to fill the data byte by byte.
So it could work for any message of any class without loosing efficiency.
Also consider that their is an ongoing work to split the messages tools (libs, generator,...) into a separated sub project https://github.com/paparazzi/pprzlink (with Python generator) and the dev branch here: https://github.com/enacuavlab/paparazzi/tree/pprzlink
I'll make pull request when I'm done with the cleaning.

flixr:
@gautierhattenberger would be great if you can resolve this together with @kirkscheper, I don't really have much time right now...

kirkscheper:
@gautierhattenberger let me know what you would like me to do with this in the interim till the pprzlink is complete. I would like to get this included in master soon so we can get the drone2drone in there. I need to solve the merge conflict with datalink.c first btw...

I would also like to start merging the rotorcraft and fixed wing datalink files btw. I was thinking of adding a subsystems/datalink/datalink.c file where all common datalink handling goes which then calls the firmware specific handling then the module datalink handling. This will reduce the maintenance and help get us to integrated firmware. I will do this in a different pull request though.

gautierhattenberger:
@kirkscheper this is a tough question. PPRZLINK is almost ready, but since it is a big change, it would be nice to have a bit more feedback before merging. I still would like to do this soon to avoid blocking of developments like yours. There is no point in doing the job twice...
So my question is, assuming we have switched to PPRZLINK, what it really missing to have this working ? Only the alignment issue for telemetry messages ?
The other point that is annoying me as mentioned earlier is the traffic info structure itself where utm fields are used for rotorcraft. Since you didn't provide any code where this is used for rotorcraft (unless you're re-using the fixedwing modules straight ?), it is difficult to tell if it is wise or not to keep this format. e.g. should we have something more complex like the state interface where several formats can be stored and conversion are done on the fly depending of the actual needs.

kirkscheper:
So if you use the byte by byte approach as currently in the gen_messages then you don't need to align the telemetry messages but then you can't use multiple variable length arrays in the message.

In the original traffic_info, the position was saved as the utm_pos - nav_pos0. The the modules that use this position just compare that difference with the ecef position from the state. Of course this isn't very accurate as you move far away from the initial reference frame and as the initial position of the utm frame isn't loaded into the rotorcraft firmware I can't do that. The benefit is that those modules can actually be used in rotorcraft if the utm frame position is loaded in (as far as I can see).

ATM I am just using the full utm position of the intruder and computing the rotorcraft utm position using one of the api help functions. Alternatively, I can do the same hack as previously used by including the utm frame position or implement what you mentioned above. For now I am just trying to get it all working and along the way will try to find out what the most useful method is to receive and store all the required data.

kirkscheper:
Hey guys, just pulled the pprzlink changes but the dl macros are not generated for the telemetry messages for decoding messages onboard. Is there any plan to change this?

gautierhattenberger:
Yes, I want to improve the reading macros to work all the time with aligned data or not

kirkscheper:
Alrighty, well I will change this pull request to only update the tracffic code to include the lla messages and to add it to rotorcraft.

gautierhattenberger:
We need to improve this in the future, but for now I guess it is fine like this. @flixr do you agree ?

PR #1511, opened 2016-01-06
Multigps Integrated

masierra:
Added some changes from #1455.
Gone is the gps_multi subsystem. Still set with subsystem="gps" type="multi". Now integrated with gps.c/gps.h. Currently, the only supported units are one piksi with one ublox but its still compatible with other implementations of (single) gps's.
To add multigps support to a gps unit just change GPS_LINK, add a GpsState and GpsTimeSync to that device and add the proper callback registrations.
It's important to note that when using multigps or ublox or piksi, GPS_TYPE_H is undefined and in its place is PRIMARY_GPS_TYPE_H. This is important for compatibility with the other GPS's not addressed in this PR.


flixr:
Thx, that's already close to a working solution :-)

Could you please rebase your branch on the paparazzi master and (force) push again?

We still need to make it work with all the single GPSs implementations again...

masierra:
@flixr rebased!

flixr:
Thx, but I actually meant to rebase it on the upstream paparazzi master and not your own master (which doesn't even differ from upstream master apart from the merge requests).
You could just reset your master to the upstream master and then rebase...
That is not super important, but makes things a bit easier...

Regarding the implementation:
I think it would be better to send ABI messages from every single implementation (like before) and subscribe to all of them in the multi GPS case (and send another ABI message with the currently used one with GPS_MULTI_ID like you already do).
Then you can "simply" set the ABI ID to GPS_MULTI_ID in the filters and other consumers if you have multiple GPS...
IMHO this support for multi GPS is tightly tied in with getting rid of the globally used `gps` struct and giving each GPS implementation their own struct (like you already started for ublox and piksi), otherwise we break a lot of single gps usage or needs loads of ifdefs.

Not completely sure what the best way is to proceed... maybe we can pull this into a branch in the paparazzi repo until we have resolved the issues there?
Unfortunately I don't have much time to spend on this right now...

flixr:
closing in favor of the updated version in #1532

PR #1513, opened 2016-01-08
Pprzlink

gautierhattenberger:
New external message library and tools. The main goal is to allow third party applications to communicate more easily with Paparazzi.
It currently provides:
- messages definition (custom messages file can still be used)
- ocaml and python libraries for ground or embedded CPU usage
- C code generator (written in python, based on mavlink generators) to generate C embedded code

Regarding the integration to the main project:
- all airborne code is converted
- all ocaml programs are converted
- audio and aerocomm telemetries have been removed (obsolete)

This pull request is currently at initial review stage and might need some more cleaning


gautierhattenberger:
Pprzlink should be mostly functional now. Their is one remaining question: should we use the message definition from paparazzi (good old conf/messages.xml) or use the one provided by pprzlink (message_definitions/v1.0/messages.xml in the pprzlink tree). Both are possible
Option one might avoid people from loosing their custom messages, option two limits duplicated work, since the two file should be synced. Also custom modifications from pprzlink file might be troublesome with the perspective of binary install.
Option three: use messages.xml from paparazzi if it exists, fallback to pprzlink otherwise

All comments are welcome, since this stuff is a bit blocking for other PR like #1509

dewagter:
Custom messages are a pain even with conf/messages.XML

What about creating an option to include personal messages in pprzlink and
then using only that. E.g. reserve last 10 message id's for an included
personal.XML

gautierhattenberger:
this will be part of the next version of the protocol, I want to avoid this kind of modification for v1.0

dewagter:
IMHO: the kindest approach would be:

step 1: be able to select which messages.xml one you want to use: default
new pprzlink
(so people with custom messages are not screwed)
step 2: a.s.a.p. add custom messages to pprzlink v2.0
step 3: remove ./conf/messages.xml

(person

-Christophe

gautierhattenberger:
So how do you see step 1 without step 3 ? removing messages.xml is harmless if you don't have modifications and will keep it (with some complains) if you have some. But if you have both, how to you select ? Editing makefile is a pain, exporting env variables or using different makefile rules is error prone.

dewagter:
Symlink?

Maybe make a poll to see if all steps are wanted by many. Personally I
would do the hard switch right away.

gautierhattenberger:
Then I vote for removing old messages.xml and use the new one unless a custom one exist.
Next step (for an other PR) would be to update pprzlink version (protocol, messages, ...)

flixr:
How do you want to check if a custom one exists?
I would be happy with any solution where we don't have multiple messages.xml that can get out of sync...

gautierhattenberger:
just checking if a file `$(PAPARAZZI_HOME)/conf/messages.xml` exists from the makefile and if yes pass it to pprzlink to overwrite default value.

flixr:
Ok,  I think that should work... as long as nothing directly uses conf/messages.xml anymore...

flixr:
@gautierhattenberger there are still some warnings that XML validation is not performed.
Seems this is because the respective libs are missing in pprzlink...
Was that intentional?

gautierhattenberger:
kind of
it requires some extra python lib not in the dependency of paparazzi-dev yet. Currently I'm using a parser that seems to come with standard Python stuff and that is used by mavlink. But it is maybe not the best choice and maybe we should switch to lxml that I think we already use. Also lxml has native DTD check support while the library used for mavlink is using xsd schema.

flixr:
AFAIK xsd is more powerful/descriptive than dtd and it seems that mavlink includes the `minixsv` lib in the tree...

I quickly tried to add the libs and then it will try to validate (takes a while though), but it fails since the xsd doesn't completely match our allowed values in the messages.xml...

In any case getting those warnings is not so nice if it is not even supported by us...

flixr:
See https://github.com/paparazzi/pprzlink/pull/1 for validation...

gautierhattenberger:
What else is missing ? Does anyone has some time to try/comment this ? @dewagter @fvantienen @esden  ?

flixr:
Looks good to merge to me...

PR #1514, opened 2016-01-11
[module] adc to uart expansion board for lisa/S

bartslinger:
I made a small device which plugs into the Lisa/S and provides additional ADC ports. The device sends PPRZ messages with three unsigned 16-bit values. This allows me to read servo positions for a new helicopter INDI controller which is in production.

photo_2016-01-11_10-00-58

### DIFF
--- a/conf/modules/adc_expansion_uart.xml
+++ b/conf/modules/adc_expansion_uart.xml
@@ -1,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="adc_expansion_uart">
+  <doc>
+    <description>Get analog signals from expansion board to main autopilot over uart</description>
+  </doc>
+  <header>
+    <file name="adc_expansion_uart.h"/>
+  </header>
+  <init fun="adc_expansion_uart_init()"/>
+  <datalink message="ADC_DATA" fun="adc_expansion_uart_process_msg()"/>
+  <makefile>
+    <file name="adc_expansion_uart.c"/>
+  </makefile>
+</module>
+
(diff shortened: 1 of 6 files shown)


conf/modules/adc_expansion_uart.xml
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="adc_expansion_uart">
+  <doc>
+    <description>Get analog signals from expansion board to main autopilot over uart</description>
```
flixr: First line should be the name that is displayed in the generated list: http://docs.paparazziuav.org/latest/onboard_modules.html

flixr:
Regarding the added messges: I'm wondering if we can't find something more generic...
Maybe @gautierhattenberger has some ideas...

conf/modules/adc_expansion_uart.xml
bartslinger: I used the module generator script here, which is then also wrong.

flixr: Well, gen_modules could be improved, true...
But "simply" using something like
```
ADC expansion via UART. Get analog signals from expansion board to main autopilot over uart.
```
as description would have also done the job...

flixr: and some more details on how it works and how it's supposed to be used also wouldn't hurt ;-)
I.e. that you should send the ADC_DATA message via datalink and how this modules then uses that info...

bartslinger:
With regard to the messages, I think the message system itself is the problem. It has no modularity or whatsoever. It would be better to make the messages configurable with airframe or module.

flixr:
You will need one globally valid messages.xml in any case, regardless whether you split it in several files or so...
You e.g. can't do something like: use msg id 42 for message FOO in one module/airframe and for message BAR in another...

bartslinger:
Well, that's exactly why the message system needs a make-over. The only reason I want this in master is because then I have this message-id claimed and it's easier to keep up-to-date with master. The module itself is of no use to anyone except myself at this moment.

flixr:
Yes, I know... but even with something where you have "user message classes" you only shift the problem to another part (or delay it since you have more IDs): within one system the IDs still need to be consistent... no way around that.

In any case that is a discussion for another time/topic..

PR #1515, opened 2016-01-12
Stereocamera resolution sendarray

knmcguire:
- Added bitoperations to retrieve int16 from uint8 message from stereocam
- Added add int16 velocity messages to SEND_STEREO_IMG for high resolutions plots

Let me know if there is anything I should improve to this pull request before merging

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_KM_conf.xml
+++ b/conf/airframes/TUDELFT/tudelft_KM_conf.xml
@@ -6,7 +6,7 @@
    radio="radios/cockpitSX.xml"
    telemetry="telemetry/default_rotorcraft_slow.xml"
    flight_plan="flight_plans/rotorcraft_optitrack.xml"
-   settings="settings/rotorcraft_basic.xml settings/control/stabilization_att_int.xml settings/control/rotorcraft_guidance.xml settings/control/stabilization_rate.xml"
+   settings="settings/rotorcraft_basic.xml settings/control/stabilization_att_int.xml settings/control/rotorcraft_guidance.xml settings/control/stabilization_rate.xml settings/estimation/body_to_imu.xml"
    settings_modules=""
    gui_color="blue"
   />
--- a/conf/airframes/TUDELFT/tudelft_KM_control_panel.xml
+++ b/conf/airframes/TUDELFT/tudelft_KM_control_panel.xml
@@ -80,7 +80,7 @@
       </program>
     <program name="BluegigaUsbDongle" command="sw/tools/bluegiga_usb_dongle/bluegiga_usb_driver">
       <arg flag="/dev/ttyACM2"/>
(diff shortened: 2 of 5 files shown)


gautierhattenberger:
it looks fine for me

flixr:
I find it a bit weird to add velocity information to a STEREO_IMG message.... since from that message name you would actually expect it to contain - well - an image... and the vel fields are not used in stereocam.c anyway.
So what information is actually in the image_buffer from your stereo module?

If it is not necessary to have this velocity in the same message as the stereo data it would IMHO make sense to add a separate message to only send the velocity information.
Best would be to find a more generic message that can be used by multiple modules providing this information. Maybe we could even directly rename the [PX4FLOW](docs.paparazziuav.org/v5.8/paparazzi_messages.html#PX4FLOW) message and use that?

If it makes more sense for you to add velocity to STEREO_IMG, maybe we should rename it and add some description to the message, otherwise someone else writing stereo code might be tempted to use this apparently generic message...

knmcguire:
That is the thing, what comes out of stereocam is not necessary an image buffer. The software on that stereocam is now in high development and we have now different modes on it. Usually it feeds a video stream, but in my case I just send the already interpreted data through in that int8_t array, like velocity and quality measures as such. 

The extra velocity message is the decoded value from that int8_t array, that I need to log in higher resolution. I think it should stay within that message since it is specific for values decoded from the stereocam. However I could rename it to decoded_value_1 ...2 ..3..? So it could also be used for different decoded values?

flixr:
A rename of at least the fields would make sense, so you can use it for other modes as well...

For me this still looks like it would make more sense in a separate message, but if you guys (cc @dewagter ) don't mind sending useless zeros from stereo.c, I'm fine with merging it...

knmcguire:
It would make more sense to make a separate message, since this is specific for just stereocam2state. It was my initial idea actually but somebody advised against it since it will increase the clutter of messages. 

Let me make more changes to make this cleaner and I'll push a new commit today for this.

flixr:
Would the contents of the px4flow message be ok for you? Then we could rename that and use it as a general message... otherwise I agree with the advice that it's better to add it to the STEREO_IMG message instead of adding another debug message just for your case...

knmcguire:
It seems that this commit now has merge conflicts.. is it still mergeable for you  or do you want me to make another branch based on the most current master?

gautierhattenberger:
you can merge with master or rebase everything on top of master and then update this pull request by pushing to the same branch

knmcguire:
Okay, I was thinking about merging it with master indeed but I'll do rebase for now, however I did had to force push it to the branch. I see that there are no build errors after the rebase

knmcguire:
I'm closing this pull request since it already diverted again from the master. I will make a new one if the development of the stereoboard has converged to a more permanent solution

Issue #1503, opened 2015-12-29
[conf] Group together conf files


podhrmic:
Would be possible to ground the conf files into one folder?

Right now in `conf` there are various folders that are not related to the actual paparazzi code, such as `AGGIEIAR`, `CDW` etc.

The conf files used to be together with the airframe files, but since that was not desirable for some reason, can the folders be grouped into one? For example in a new folder `aiframe_config`:
```
conf/aiframe_config/AGGIAIR
conf/aiframe_config/CDW
conf/aiframe_config/...
```

Right now it is really cluttering the `conf` directory, so it is harder to find the folder you are looking for, and it is making comparisons between branches more difficult. I could image that it is also confusing for users.

Thanks for consideration.

flixr:
I'm not quite following... what exactly do you want to put in the `airframe_config` folder and what keep in `airframes`?

podhrmic:
`airframe_config` ->The uppercase folders that now contain the conf.xml files that are selected with the start.py script. (for example AGGIEAIR, CDW...)

`airframes` -> keep it as is

That way there will be only two folders in `conf` that are related to user airframes, rather than 10 (and possibly growing).

Does it make sense now?

flixr:
hm... to me it does not make more sense to have some airframe files in conf/airframes (like the examples) and some in conf/airframe_conf ("user specific"), especially since quite some conf.xml have airframe files from examples and from user specific dirs...

If people agree that changing this helps, OK... personally I find it even more confusing to look for airframe files in conf/airframes and conf/aircframe_conf...
If the uppercase USER_FILES vs. examples is not distinction enough, I would rather have a subdir under conf/airframes where all user airframes go (e.g. conf/airframes/USERS/AGGIEAIR ) or so...

Also the user files are often more "related to actual paparazzi code" than the examples, because they are test flown more...

podhrmic:
I see - I guess what I wanted to say was that the uppercase directories in `conf` folder are distracting. If they are in lets say `conf/airframes` instead that is totally cool, because then you don't have to see them unless you want to see some specific user files.

For me as a developer it is confusing because now instead of having clearly named directories in `conf` there are now mixed other unrelated folders with user files.

So I would advocate for having as few folders in `conf` as possible, like it used to be before (for example https://github.com/paparazzi/paparazzi/tree/v5.6/conf )

flixr:
[...]

PR #1516, opened 2016-01-13
Config files cleanup

podhrmic:
Config files moved from the `conf` folder back to the airframe folder, so the user directories wont clutter the `conf` folder any more.


podhrmic:
Related to #1503

flixr:
:+1: from me, @OpenUAS as you originally moved them you might want to comment/object now ;-)

dewagter:
To have all conf in between the makefiles and messages.xml is not very nice, so for me this change is OK.

If having conf mixed with airframes is not desired I could also image that config files would reside in a special folder like ./conf/personal or ./conf/conf or ./conf/config/

podhrmic:
So far 3 agree.
@gautierhattenberger @OpenUAS any objections to the change?

gautierhattenberger:
Having personal conf in airframes directory is probably not ideal but still okay with the current configuration structure. So I'm fine with this.

flixr:
We could also put the personal confs in a folder like @podhrmic and @dewagter already suggested.
Either way I'm fine with it....

If @OpenUAS doesn't object until tonight, I'll merge it....

podhrmic:
@flixr merge then?

PR #1522, opened 2016-01-30
[refactor] add shared datalink for all firmware

kirkscheper:
Trying to help the joining of fixedwing and rotorcraft. This way all shared datalink code and updates only have to be done once. Firmware specific code not yet shared can be kept separate in the original files.

Pull request also contains a start for drone to drone messaging but reading the messages is waiting for the dl macros in telemetry.


flixr:
Nice :-)

Apart from the missing declarations it would probably make sense to rename the firmware specific `datalink.c` to `rotorcraft_datalink.c` and `fixedwing_datalink.c`...

kirkscheper:
ah yea, forgot to include the datalink.h.

Isn't it enough that the files are in their respective firmware folders? If not let me know and I will rename them.

flixr:
IMHO it is cleaner if files that are meant to be compiled together have unique names...

kirkscheper:
done

flixr:
Thx!

PR #1523, opened 2016-02-01
Speech recognition to jump to a block

rmeertens:
Maybe not the best pull request ever, but I played around with speech recognition last weekend. 
I added one simple python example for jumping to a new block, and one example on how to do this with speech recognition (although a certain python library is needed). 

If anybody likes the idea of a speech controlled airplane, let me know!

### DIFF
--- a/sw/ground_segment/python/jump_to_block_example.py
+++ b/sw/ground_segment/python/jump_to_block_example.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import sys
+from os import path, getenv
+from time import sleep
+
+# if PAPARAZZI_SRC not set, then assume the tree containing this
+# file is a reasonable substitute
+PPRZ_SRC = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../')))
+sys.path.append(PPRZ_SRC + "/sw/ext/pprzlink/lib/v1.0/python")
+
+from ivy_msg_interface import IvyMessagesInterface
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
At least it would be nice to factorize the jump_to_block code and provide some information on which library to install to make this running

rmeertens:
Can and will do!

flixr:
Should also use 4 spaces for indentation of the python code and be moved to pprzlink...

And I don't know about any python interpreter for the flight plan... what exactly were you thinking about?

rmeertens:
Will do. 
As for the python interpreter: currently I have commands that say "Go to block 4"... which works. But I would like to say things like: "Go to waypoint 'river'". Or: "execute block 'land here'".

flixr:
AFAIK there is nothing like that yet...

OpenUAS:
A fun thing the idea of a speech controlled airplane, although for my no telemetry flights I need a mike in my plane and a megaphone ;), Anyhow... adding something like select AC "ac name" or ID for multi UAS flights. Also a conformation feedback [Really do XYZ? Yes, No?] makes it a bit less adventurous :) Also with multiple people having a say does not always end well ... http://vignette3.wikia.nocookie.net/smurfs/images/4/4c/Weather_Smurfing_Machine_Comic.jpg

flixr:
@rmeertens should we keep this open? If it only depends on pprzlink, it should go in a tools/examples dir there...

rmeertens:
Close it for now, don't know if anybody is interested at all in having a speech controlled autopilot ;)

PR #1524, opened 2016-02-01
direct NED to ENU for N and E

EwoudSmeur:
I noticed that waypoints I enter in NED do not exactly result in exactly those coordinates in the generated flight plan. In the generated flight plan, my NED waypoint {-2.0, 0.0, 1.6} goes to {-2.00, 0.04, 1.60}, / ENU in meters / . The ENU waypoint is what is used in the navigation.

This seems kind of silly to me, as it should be really simple to go from NED to ENU. The transformation happens in the flight plan generator in the function print_waypoint_enu. In this function the waypoints are first converted to ECEF and then to NED. This is where I think the error is introduced. I suppose this is to get the correct altitude if you use alt or height. But there is no reason to use the east and north values from NED->ECEF->ENU.

### DIFF
--- a/sw/tools/generators/gen_flight_plan.ml
+++ b/sw/tools/generators/gen_flight_plan.ml
@@ -136,7 +136,7 @@ let print_waypoint_enu = fun utm0 default_alt waypoint ->
   let ecef0 = Latlong.ecef_of_geo Latlong.WGS84 (Latlong.of_utm Latlong.WGS84 utm0) !ground_alt in
   let ecef = Latlong.ecef_of_geo Latlong.WGS84 (Latlong.of_utm Latlong.WGS84 (Latlong.utm_add utm0 (x, y))) (float_of_string alt) in
   let ned = Latlong.array_of_ned (Latlong.ned_of_ecef ecef0 ecef) in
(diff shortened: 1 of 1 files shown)


flixr:
If I'm not mistaken then all waypoint `x`, `y` coordinates are currently handled as UTM coordinates and not NED.
So this is why you see a discrepancy if they are converted to ENU.

We currently don't have a way to specify that the x,y coordinates are supposed to be NED coordinates as opposed to UTM.

@gautierhattenberger is that correct?

gautierhattenberger:
Correct, and for longer distance, it really makes a big difference (like missing the roof where you're suppose to land during an IMAV competition...)

EwoudSmeur:
Ah ok, I see. I think it would be a nice feature to be able to set waypoints in ENU of NED. For my experiments, I want the drone to fly exactly two meters east, which is hard to achieve with the current setup.

flixr:
@EwoudSmeur so can we close this then?
Feel free to open an issue for a feature request to be able to specify the coordinate system.

EwoudSmeur:
Yes, thanks for the info. I will open an issue.

PR #1527, opened 2016-02-04
[pprzlink] add read macros for all messages classes

gautierhattenberger:
- support aligned or non aligned data depending of the arch
- will help air-to-air communications (#1509)

### DIFF
--- a/conf/Makefile.arm-linux
+++ b/conf/Makefile.arm-linux
@@ -41,6 +41,9 @@ endif
 
 ARCH_CFLAGS ?= -mtune=cortex-a8 -march=armv7-a
 
+# with armv7, unaligned data can still be read
+CFLAGS += -DPPRZLINK_CAST_ALIGN=0
+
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
@kirkscheper If this work, you should be able to parse incoming telemetry messages from the airborne side.
Just be careful with arrays, as it is not possible to read them on all architecture, reading unaligned data must be supported for that. If it is not the case, the array size should be forced to 0 to tell the user it should not be read.

flixr:
I haven't found any good document on how this is handled on ARM... did you find a resource with documentation on that?
Normally gcc knows the ABI of the target system and only warns if unaligned access is problematic (e.g. shouldn't warn on x86/64).

Also if it does warn, can we plz only disable the `-Wcast-align` warnings for that particular set of macros instead of for the whole code base?

gautierhattenberger:
closed with #1528

PR #1530, opened 2016-02-04
[flight_plan] allow variable declaration from flight plan

gautierhattenberger:
- can be used localy in the flight plan to avoid adding new variables in
  the base code when it can be avoided
- with appropriate attributes, will generate a settings accessible from
  the GCS


flixr:
Do you have an example or docs on how this should be used?
What is the keyword?

gautierhattenberger:
Will add an example somewhere.
Basically:
```
<variables>
  <variabla var="boo"/>
  ...
</variables>
```
creates a float initialized to 0.
`type` and `init` can control this.
and if at least `min`, `max` and `step` are added, it generates automatic settings.

flixr:
Nice addition!
Looks good to me, you want do add anything else before merging?

PR #1531, opened 2016-02-07
added flask server 

rmeertens:
Inspired by hooperfly who controls his drones using a flask server.... I made a flask server which can process messages. 

If you want to put stuff online, simple call this server.

### DIFF
--- a/conf/control_panel_example.xml
+++ b/conf/control_panel_example.xml
@@ -14,6 +14,7 @@
     </program>
     <program name="Messages" command="sw/ground_segment/tmtc/messages"/>
     <program name="Messages (Python)" command="sw/ground_segment/python/messages_app/messagesapp.py"/>
+    <program name="Flaskserver" command="sw/ground_segment/python/flaskserver/flaskserver.py"/>
     <program name="Settings" command="sw/ground_segment/tmtc/settings">
       <arg flag="-ac" constant="@AIRCRAFT"/>
     </program>
--- a/sw/ground_segment/python/flaskserver/callbackexample.html
+++ b/sw/ground_segment/python/flaskserver/callbackexample.html
@@ -1,0 +1,33 @@
+<html>
+	<head>
+		<title>Subscriber example</title>
+		<style>
+			body { margin: 0; }
+			canvas { width: 100%; height: 100% }
+		</style>
+	</head>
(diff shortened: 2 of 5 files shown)


flixr:
please use 4 spaces for indentation of python files...

Also wondering if we should put this in pprzlink (as that is all that it depends on)

hooperfly:
@flixr yes, we might want to come up with a plan for various new tools/services that are created using `PPRZLink`. I will be including parts @rmeertens work here into my `frc.py` handler as well. I think `PPRZLink` will be a popular interface for exploring/developing PPRZ feature sets.

gautierhattenberger:
I agree it could be part of the tools of pprzlink

rmeertens:
So. What to do with the pprzlink idea? Do we want to merge it here for now? Or already merge this in pprzlink...?

gautierhattenberger:
I think can make a PR directly in pprzlink for the server itself and maybe the examples. Then you can update this pull request to update pprzlink in main paparazzi and add the tool to the control panel if you want.

flixr:
See also paparazzi/pprzlink#2

hooperfly:
@rmeertens I took some time this weekend to run and test your code. I like where it's going. Apropos @flixr feedback in [paparazzi/pprzlink#2](https://github.com/paparazzi/pprzlink/pull/2), it might be nice to build a streamlined example of a `flask` server that leverages various aspects of `PPRZLink` that includes a nice set of `PPRZLinkUtils` to access both static(compile-time) and dynamic(run-time) flight information.

hooperfly:
@rmeertens I'm integrating your `flaskserver.py` concepts into the `Flying Robot Commander` server. I refactored it a bit for my use, but it's working just fine. I'm thinking about the static vs. dynamic aspects of the data model and how to gracefully mange non-runtime(pre/post flight/simulation) vs. runtime(flight/simulation). Things like the `aircraft id` have a static set view, so to speak, as well as a runtime set; where {`ac_id` runtime} <= {`ac_id` static}. So the server would only subscribe to the message bus, if there is a runtime dynamic needed.

flixr:
closing this as it would go in pprzlink if cleaned up, see https://github.com/paparazzi/pprzlink/pull/2

PR #1532, opened 2016-02-07
support for multiple GPS

flixr:
This is a cleaned up and finished version of #1511 

To e.g. use piksi as a secondary GPS:
```
<subsystem name="gps" type="ubx">
  <configure name="UBX_GPS_PORT" value="UART2"/>
</subsystem>
<subsystem name="gps" type="piksi">
  <configure name="PIKSI_GPS_PORT" value="UART3"/>
  <configure name="SECONDARY_GPS" value="piksi"/>
</subsystem>
 ```

Each GPS implementation has their own `GpsState` struct now.
We still have the global `gps` struct for now (which gets set to the currently best GPS), but the goal is to get rid of that global struct later as well...


flixr:
Ok, the LisaMX example using UART5 for piksi is kind of bad, since TX of that uart isn't exposed...

podhrmic:
You can use the pin labelled as PC12 for UART5 TX and then change accordingly in the board file.

masierra:
Yeah I had to the mod to get UART1 on my LisaMX

gautierhattenberger:
really nice feature, especially because it starts to get rid of the global gps struct

gautierhattenberger:
@flixr do you want to add something ? or are we just waiting for more tests ?

flixr:
I started refactoring the xsens INS, but that shouldn't be part of this PR.
So IMHO it's good to merge as soon as it's tested...

gautierhattenberger:
Tested and working with a UBX+PIKSI on Apogee. Manual and automatic switch seems to work as expected.
I just changed a protection flag name to be consistent, see multigps_ms branch in paparazzi.
@flixr I think you can update and merge.

PR #1533, opened 2016-02-10
[generator] Qt project generator

fvantienen:
Add a Qt Creator project Generator for a specific AC. This can either be done trough commandline by using:

```
make -f Makefile.ac AIRCRAFT=MyAircraftName ap.qt
```

Or automatically on each build by setting the following environment variable:

```
export PAPARAZZI_QT_GEN=1
```

After the creation of a project you can open Qt Creator and then Select `File->Open Project`. Then browse to the paparazzi folder and open the `ap.creator` file(Or `nps.creator`, `fbw.creator` if you want a different target).
You don't need to reload/reopen Qt Creator if you use the Auto switch on each build as your project will automatically update.

Some feature additions would be:
- [ ] Add the modules xml files from the airframe file
- [ ] Add compile/upload configurations automatically to the creator file
- [ ] Add code style to the creator file
- [ ] Add a button to Paparazzi center to execute a switch to a new AC

### DIFF
--- a/sw/airborne/Makefile
+++ b/sw/airborne/Makefile
@@ -38,16 +38,16 @@ VPATH = .
 ifneq ($(MAKECMDGOALS),clean)
   include $(AIRCRAFT_BUILD_DIR)/Makefile.ac
   $(TARGET).srcs += $($(TARGET).EXTRA_SRCS)
-  include ../../conf/Makefile.local
+  include $(PAPARAZZI_HOME)/conf/Makefile.local
 
 # check if ARCHDIR is set
   ifeq ($($(TARGET).ARCHDIR), )
     $(error Architecture not set, maybe you forgot to add the target? e.g. <target name="tunnel" board="twog_1.0"/>)
   else
     ifdef $(TARGET).MAKEFILE
-      include ../../conf/Makefile.$($(TARGET).MAKEFILE)
+      include $(PAPARAZZI_SRC)/conf/Makefile.$($(TARGET).MAKEFILE)
     else
-      include ../../conf/Makefile.$($(TARGET).ARCHDIR)
+      include $(PAPARAZZI_SRC)/conf/Makefile.$($(TARGET).ARCHDIR)
     endif
   endif
 
--- a/sw/tools/generators/gen_aircraft.ml
+++ b/sw/tools/generators/gen_aircraft.ml
@@ -397,9 +397,9 @@ let () =
 
     (* Get TARGET env, needed to build modules.h according to the target *)
     let t = try Printf.sprintf "TARGET=%s" (Sys.getenv "TARGET") with _ -> "" in
+    make_opt "radio_ac_h" "RADIO" "radio";
     make_opt "flight_plan_ac_h" "FLIGHT_PLAN" "flight_plan";
-    make "all_ac_h" t;
-    make_opt "radio_ac_h" "RADIO" "radio"
+    make "all_ac_h" t
   with Failure f ->
     prerr_endline f;
     exit 1
--- a/sw/tools/qt_project.py
+++ b/sw/tools/qt_project.py
@@ -1,0 +1,102 @@
+#!/usr/bin/python
+import sys
+import re
+from os import path, getenv
+from lxml import etree
+
+# Get the paparazzi home dir (if not set assume this folder is correct)
+home_dir = getenv("PAPARAZZI_HOME", path.normpath(path.join(
+        path.dirname(path.abspath(__file__)), '../../')))
+
+
+# Check if we have enough arguments
+if len(sys.argv) != 4:
+    print("Usage: qt_project.py AIRCRAFT CONF_XML AP_SRCR_LIST_FILE")
+    print("Not enough arguments!\n")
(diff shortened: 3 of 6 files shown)


flixr:
Nice, that should be really useful for anyone wishing to use Qt creator!

Still gives quite a few errors when building sim/nps though, e.g.
```
cc: error: nps/nps_atmosphere.c: No such file or directory
```
guess you need to pass the include dirs to `gcc -MM`

Also please use 4 spaces for indentation of python files, see http://docs.paparazziuav.org/latest/stylepython.html

sw/tools/generators/gen_aircraft.ml
```diff
@@ -397,9 +397,9 @@ let () =
 
     (* Get TARGET env, needed to build modules.h according to the target *)
     let t = try Printf.sprintf "TARGET=%s" (Sys.getenv "TARGET") with _ -> "" in
+    make_opt "radio_ac_h" "RADIO" "radio";
```
flixr: Wondering what's up with that change?
Did that result in wrong deps in the old order?

sw/airborne/Makefile
```diff
@@ -38,16 +38,16 @@ VPATH = .
 ifneq ($(MAKECMDGOALS),clean)
   include $(AIRCRAFT_BUILD_DIR)/Makefile.ac
   $(TARGET).srcs += $($(TARGET).EXTRA_SRCS)
-  include ../../conf/Makefile.local
+  include $(PAPARAZZI_HOME)/conf/Makefile.local
```
flixr: Hm.. not quite sure if these makefiles should really be in `PAPARAZZI_HOME` (instead of the conf in `PAPARAZZI_SRC`).
Probably doesn't matter that much right now as the separation is not really much used and probably rather untested, but IMHO only the conf xml files should be in HOME (and var).

sw/tools/qt_project.py
```diff
@@ -0,0 +1,102 @@
...
+from os import path, getenv
+from lxml import etree
+
+# Get the paparazzi home dir (if not set assume this folder is correct)
+home_dir = getenv("PAPARAZZI_HOME", path.normpath(path.join(
+    path.dirname(path.abspath(__file__)), '../../')))
+
+
+# Check if we have enough arguments
+if len(sys.argv) != 4:
```
flixr: you probably rather want to use argparse...

sw/airborne/Makefile
gautierhattenberger: Most of this file is actually useless, but you're probably right, it could be somewhere else in SRC part. Same for most of the makefiles in `conf` I guess.

flixr: yes, it was more about all the makefiles in conf, like the ones below.
with the relative path (as it was before) will "automatically" look for them in SRC

sw/tools/generators/gen_aircraft.ml
fvantienen: Since the project is made at the "all_ac_h" and it uses `gcc -MM` to get the files it also needs the radio header file.

sw/airborne/Makefile
fvantienen: Ok will change it to SRC

flixr:
@fvantienen what is the status here?

fvantienen:
Sorry was on holiday, will try to look into it again. It's harder than expected.
My idea now was to use the `.d` files to generated the header dependencies, but then a different compile step for those would be prefered.

flixr:
If the VPATH is the problem, you can probably also just set it for all possible cases (right now only simulator and pprzlink) in the command generating the files..

kevindehecker:
I would like to voice my support over this PR. I use it and get quite some questions about it from other interested people.

flixr:
I would gladly merge this if the conflicts and the VPATH problems are resolved...

flixr:
@fvantienen any updates here?

podhrmic:
FYI you can use Eclipse IDE to generate/compile a specific AC. If there was interest I could post some howto on wiki.

flixr:
@podhrmic I think quite a few people would appreciate that very much...

podhrmic:
Here is a guide for settings up Eclipse to compile flight code: https://wiki.paparazziuav.org/wiki/RT_Paparazzi#Compiling_Paparazzi_code_in_Eclipse
@flixr do you think I should send it to the mailing list too or is that not necessary?

flixr:
@podhrmic you already sent a message ;-)
But maybe it would make more sense to put the Eclipse information on https://wiki.paparazziuav.org/wiki/DevGuide/CodeEditors instead of the RT_Paparazzi page, as it applies to the other targets as well?

podhrmic:
Wiki updated. I am not sure about the Eclipse debugging because you can surely use it to debug non-RT paparazzi too, I've just never done it so I can't guarantee that it works the same way - so I would leave that part on the RT page for the time being.

dewagter:
revive++

podhrmic:
Is this still relevant after merging #1940 ?

kevindehecker:
No, this one can be closed now.

PR #1534, opened 2016-02-10
Modules makefile improvements

gautierhattenberger:
Two improvements for the makefile section of modules in order to efficiently replace subsystems:
* case modifier for upper/lower version makefile variables: much cleaner xml code
* add firmware option to allow a single module to handle specific firmware options

### DIFF
--- a/conf/modules/airspeed_amsys.xml
+++ b/conf/modules/airspeed_amsys.xml
@@ -34,11 +34,7 @@
   <event fun="AirspeedAmsysEvent()"/>
 
   <makefile>
-    <raw>
-      AIRSPEED_AMSYS_I2C_DEV ?= i2c0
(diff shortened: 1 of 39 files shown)


flixr:
Very nice!

Just wondering if we maybe should make `default="TRUE"` the default, since that is probably used with nearly every configure option that sets a value.
E.g. replace it by something like `set="TRUE"` for the case where you don't want to specify a default value with `?=`

gautierhattenberger:
I guess we can change this. The only point I see is that `configure` will have different behavior if they come from airframe file (regular variable) or from the module (optional variable). If you're fine with this, I can change it.

flixr:
Good point.
I think even nicer to use would be
`<configure name="FOO_I2C_DEV" default="i2c0" case="upper|lower"/>`
and it wouldn't collide with `<configure name="BAR" value="BAZ"/>` as  the airframe file.

gautierhattenberger:
Can I merge like this ?

flixr:
:+1:

Issue #101, opened 2011-12-11
General vehicle state interface


flixr:
A general state interface that holds all the most important vehicle states like position, speed, attitude, etc. for fixedwings and rotorcrafts alike would be nice. So any subsystem/module that needs to know any of these states should not get it directly from the ahrs, gps, airspeed sensors or whatever and just read it from the state interface. It should also do the appropriate coordinate and fixed/floating point conversions automatically and only when needed. This should decouple e.g. estimation and control more and also the "consumers" of vehicle states don't need to worry about where the data is actually coming from and in what format it was estimated or measured.

I started on this quite a while ago already, see [state.h](https://github.com/paparazzi/paparazzi/blob/state_interface/sw/airborne/state.h) and [state.c](https://github.com/paparazzi/paparazzi/blob/state_interface/sw/airborne/state.c) in the [state_interface](https://github.com/paparazzi/paparazzi/tree/state_interface) branch.

Current status:
- Most of the state interface and needed conversions are done
- You can set e.g. the position in any coordinate system you wish: `stateSetPositionNed_i(pos_ned_i)` to set the position in fixed-point NED coordinates
- If you need to read the position somewhere else in a different representation, simply call: `stateGetPositionLla_f()` and only then the LLA float position representation is calculated and returned on the fly. It's also only calculated once, until a new position is set.

There are however some open questions that need addressing before we go further:
-  How to deal with the problem that a given representation can't be calculated or is not valid, e.g.
  - something needed for conversion not initialized (e.g. LTP origin not initialized)
  - simply no measurement of the state available (e.g. no airspeed sensor)
  - state not valid (e.g. no 3d GPS fix, hence pos/speed not valid)
- Add extra functions to set multiple representations of the same state at once, or allow to set a state without marking the other representations as uncalculated?
  - E.g. when the GPS already does the work and calculates several representations, why not use that instead of only setting one and doing the conversions again ourselves...
- We could get rid of the AHRS and INS interfaces and only have implementation specific structs that hold only the necessary data (e.g. fixed point in imu frame) and then just use the `stateSetFoo_i(bar)` functions to copy to the state interface where the other representations (e.g. float in body frame) will be calculated automatically on the fly when needed.
- more that I don't recall right now...

PR #1535, opened 2016-02-12
[refactoring] changed rotorcraft_radio_control message to just radio_control

kirkscheper:
requires change to pprzlink msg 160 from 
<message name="ROTORCRAFT_RADIO_CONTROL" id="160">

to 
<message name="RADIO_CONTROL" id="160">

### DIFF
--- a/conf/telemetry/default_fixedwing.xml
+++ b/conf/telemetry/default_fixedwing.xml
@@ -33,6 +33,7 @@
       <message name="FBW_STATUS"          period="2"/>
       <message name="AIR_DATA"            period="1.3"/>
       <message name="VECTORNAV_INFO"      period="0.5"/>
+      <message name="RADIO_CONTROL"       period="0.51"/>      
     </mode>
     <mode name="minimal">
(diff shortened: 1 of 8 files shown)


kirkscheper:
Didn't realize there already is already a generic rc message (#101). why is there a specific rotorcraft_rc message? Will close this and just add the periodic to the fixedwing/autopilot.c, it is currently only in the fbw firmware.

gautierhattenberger:
@kirkscheper the ROTORCRAFT_RC message dates back to the very beginning of the rotorcraft support in Paparazzi and I guess it was easier to have named RC channels since the order of the generic RC message depends on you radio config file (ppm) or radio type (spektrum, futaba,...).
Also you can't add the RC message to the `ap` part of fixedwing since it would break the logical separation between `ap` and `fbw` (http://wiki.paparazziuav.org/wiki/DevGuide/DesignOverview#Airborne_Functional_Diagram). And the `RC` message is already registered for the rotorcraft ap, so you just need to add it to your telemetry file (or make a PR to have it in the standard files)

PR #1537, opened 2016-02-17
+logging function for natnet2ivy

rmeertens:
added a logging function to see what was sent to the drone.


kirkscheper:
This removes lots of previous work, can you merge instead of overwriting?

rmeertens:
Mmm, looks like I messed up again. Will do!

kirkscheper:
also this only logs if small messages are not set, can you move the log outside the brackets to log regardless.

kirkscheper:
maybe also give a unique time to each log and have them stored in the var folder and not the paparazzi home folder?

PR #1538, opened 2016-02-17
logging natnet messages

rmeertens:
added logging of natnet2ivy messages.


rmeertens:
fixed Kirks first remark. @kirkscheper , for me the time is not really necessary... do you really want it???

kirkscheper:
I think it is, else you have the chance will overwrite logs.

PR #1539, opened 2016-02-17
mavlab course to master

rmeertens:
Hey everybody, 

In the MAVLab we are giving a course on autonomous drones that uses paparazzi. 
As a part of this course I made a nice example of a drone autonomously avoiding orange obstacles. Let's put it in master to make sure it will not be lost next year.

### DIFF
--- a/conf/modules/orange_avoider.xml
+++ b/conf/modules/orange_avoider.xml
@@ -1,0 +1,21 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="orange_avoider">
+  <doc>
+    <description>
+	Avoid all objects that are orange!
+	This example module shows how you can use the camera stream and colorfilter to detect orange objects. 
+	By adding this module to your flightplan and flying in the cyberzoo with the flightplan TUDELFT/tudelft_course2016_avoid_orange_cyberzoo.xml you will avoid every obstacle that is orange.
+    </description>
+  </doc>
+  <depends>cv_colorfilter.xml</depends>
+  <header>
+    <file name="orange_avoider.h"/>
+  </header>
+  <init fun="orange_avoider_init()"/>
+  <periodic fun="orange_avoider_periodic()" freq="4"/>
+  <makefile>
+    <file name="orange_avoider.c"/>
+  </makefile>
+</module>
+
--- a/sw/airborne/modules/orange_avoider/orange_avoider.c
+++ b/sw/airborne/modules/orange_avoider/orange_avoider.c
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) Roland Meertens
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/orange_avoider/orange_avoider.c"
+ * @author Roland Meertens
+ * Example on how to use the colours detected to avoid orange pole in the cyberzoo
+ */
+
+#include "modules/orange_avoider/orange_avoider.h"
+#include "modules/computer_vision/colorfilter.h"
+#include "firmwares/rotorcraft/navigation.h"
+#include "state.h"
+#include <time.h>
+#include <stdlib.h>
+
+uint8_t safeToGoForwards = FALSE;
+int tresholdColorCount = 200;
+int32_t incrementForAvoidance;
+
+void orange_avoider_init()
+{
+  // Initialise the variables of the colorfilter to accept orange
+  color_lum_min = 0;
+  color_lum_max = 131;
+  color_cb_min = 93;
+  color_cb_max = 255;
+  color_cr_min = 134;
+  color_cr_max = 255;
+  // Initialise random values
+  srand(time(NULL));
+  chooseRandomIncrementAvoidance();
+}
+void orange_avoider_periodic()
+{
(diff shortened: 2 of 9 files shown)


sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -110,7 +110,7 @@ static inline void nav_set_altitude(void);
 
 /** minimum horizontal distance to waypoint to mark as arrived */
 #ifndef ARRIVED_AT_WAYPOINT
-#define ARRIVED_AT_WAYPOINT 3.0
+#define ARRIVED_AT_WAYPOINT 0.2
```
gautierhattenberger: this should be changed in airframe file since it is way to low for a regular GPS

sw/airborne/modules/orange_avoider/orange_avoider.c
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+#include "modules/orange_avoider/orange_avoider.h"
+#include "modules/computer_vision/colorfilter.h"
+#include "firmwares/rotorcraft/navigation.h"
+#include "state.h"
+#include <time.h>
+#include <stdlib.h>
+
+uint8_t safeToGoForwards=FALSE;
```
gautierhattenberger: please follow code style and/or use the `fix_code_style.sh` script

sw/ground_segment/misc/natnet2ivy.c
```diff
@@ -19,14 +19,14 @@
...
-* navigate inside the Optitrack system.
-*/
+ /** \file natnet2ivy.c
+ *  \brief NatNet (GPS) to ivy forwarder
+ *
+ *   This receives aircraft position information through the Optitrack system
+ * NatNet UDP stream and forwards it to the ivy bus. An aircraft with the gps
+ * subsystem "datalink" is then able to parse the GPS position and use it to
+ * navigate inside the Optitrack system.
+ */
```
gautierhattenberger: What are exactly the changes in this file beside the logging and breaking a part of the code style ? Is this a duplicate of #1538 ?

rmeertens:
Welp
Thought I left the files with your remarks out. Somehow I screwed up something with git. Will remove the two files with your remarks.

rmeertens:
Think I now fixed everything. Note that I took the same commit for the pprzlink as is currently in the master (that is, not the most recent one).

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -48,13 +48,13 @@
 
 // Downsize factor for video stream
 #ifndef VIEWVIDEO_DOWNSIZE_FACTOR
-#define VIEWVIDEO_DOWNSIZE_FACTOR 4
+#define VIEWVIDEO_DOWNSIZE_FACTOR 1
```
flixr: Any specific reason to change the defaults here?
How will that affect current configurations for e.g. ARDrone, Bebop?
Also if changing defaults, please update the module xml description accordingly.

conf/modules/orange_avoider.xml
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="orange_avoider">
+  <doc>
+    <description>Example on how to use the colours detected to avoid orange pole in the cyberzoo</description>
```
flixr: first line of description should be the title/name of the module, extended description after that.

flixr:
Missing either a dependency on another module or `computer_vision/colorfilter.c`?

Please also squash and rebase on master...

flixr:
You also might want to use the new functionality to load the module from the flight plan (see #1540)

sw/airborne/modules/computer_vision/viewvideo.c
rmeertens: Makes everything more flyable... when using this module as it is in master on the ARDrone it hardly receives telemetry updates (such as from the joystick) as the whole wifi fills with video. IMO this downsize factor makes it way better.

flixr: I don't get it, you increase the size of the images/video as you don't downsize it anymore... you also increased the quality...
In any case, this change needs to be reflected in the module xml

rmeertens: Welp, think you are right. Will revert.

rmeertens:
Bam. Squashed and added dependency.

flixr:
Thx, just one question: do you actually also use the conf/airframes/examples/course_bebop_colorfilter.xml airframe? It's not in any conf, so won't be test compiled...

Also the note about module order points to a problem we should consider solving using the dependencies:
```
Note that for vision the order in which modules are loaded is important
If you place cv_colorfilter before your own module your own module will only get the filtered image
```
This dependence on the module init order is kind of brittle...

rmeertens:
It is included in the conf/TUDELFT/tudelft_course2016_conf.xml configuration. 

And good point about that relation. For now this works, but I will probably update this part of the course next year to reflect what is working at that time. 

Merge it now?

flixr:
Sorry missed that... merging.

Issue #924, opened 2014-11-04
XSens MTi-G 700 driver coordinate system cleanup and improvements


flixr:
It seems like the xsens 700 must be first manually set up to output data in correct frame (NED or NWU?), otherwise the driver will output the data in the wrong coordinate system.
Would be good if the  driver would always configure the xsens imu to that output format.

@dewagter you wrote on the mailing list that you setup the xsens to output in NED, but from the code it rather looks like you need to set up to NWU (NorthWestUp).
It also seems like this is taken into acount for the euler attitude and rates, but e.g. not acceleration...

Also it would make a lot of sense to not use the euler but the quaternion output and properly implement body_to_imu and get rid of the ins_neutrals hack (that is only meant for minor adjustments on attitude and doesn't correct the other output data like acceleration, rates...
##

gautierhattenberger:
is it still relevant ?

dewagter:
Actually, I tried recently to flash our TWOG, but the LED does not even
blink anymore. Do you know anyone who can confirm that the LPC21 is still
actually working in master?

-Christophe

gautierhattenberger:
There is at least an issue when using the Xbee in API mode, but it should work in transparent mode. What is your current conf ?

podhrmic:
@dewagter I ll test Umarim and let you know

dewagter:
It was the xbee-api problem...

podhrmic:
@dewagter confirmed - Umarim works with transparent telemetry (LED blinks, that is the extent I tested it), but with xbee-api there is no LED at all.

gautierhattenberger:
just for the fun: start the board with Xbee mode and wait long enough

hint: the active wait in xbee init is doing some mess with the sys_time module apparently, so if you wait long enough until the internal timer roll over, it will continue the init

advice: remove the propellers, it was a bit scary to have the ESC doing the init several minutes after startup while I was concentrating on the code looking for the issue.

PR #1541, opened 2016-02-17
Refactor xsens

flixr:
Might still be missing some stuff and is untested...


gautierhattenberger:
I don't have working XSens either, but it looks much better like this. Can someone test it ?
Otherwise I still suggest to merge, it makes the code cleaner.

dewagter:
For Delft you can merge the new xsens code.

-Christophe

podhrmic:
If nobody has means to test it, shouldn't we just remove the code or at least mark is as obsolete?

flixr:
I thought that @dewagter and co. have an xsens in one of their planes...

dewagter:
We do have 2 xsens planes. For now they fly with older code. We need to
upgrade them with new paparazzi. We do have the means to test it, just not
this week as I'm not in Delft.

-Christophe

flixr:
I would wait with merging until it has been tested as the handling of GPS position/vel is different now and I'm not sure exactly which information is meant to be used for the INS and which for the "raw" GPS part...

Also still TODO:
* fix ins_xsens modules or directly convert the subsystems to modules
* multi GPS support?

flixr:
The different signs (IMU/sensor orientation) between the different xsens types is also weird and unclear.
See also related #924

flixr:
@dewagter any chance someone in your Lab can try this soon-ish?

flixr:
Also converted the xsens subsystems to modules...
I would like to merge this ASAP so we can get on with things.

This still really needs to be tested with a real XSens though... also the `XSENS_OUTPUT_MODE` might not be correct for all cases...

PR #1542, opened 2016-02-19
Fixed and tested magnetometer

OpenUAS:
Fixed and tested magnetometer. Tested on four different Bebop's from different batches. All is now fine, note that this is a fix for regular Bebop  and has nothing to do with Bebop2, Bebop2 is another **product,** not another version. Raw values monitored, scaled compared etc. etc.

Yes, it is only one - (minus) sign difference that make all the difference from non- autonomous flight to OK autonomous flight.

### DIFF
--- a/sw/airborne/subsystems/imu/imu_bebop.c
+++ b/sw/airborne/subsystems/imu/imu_bebop.c
@@ -134,8 +134,8 @@ void imu_bebop_event(void)
 #if BEBOP_VERSION2
     // In the second bebop version the magneto is turned 90 degrees
     VECT3_ASSIGN(imu.mag_unscaled, -imu_bebop.ak.data.vect.x, -imu_bebop.ak.data.vect.y, imu_bebop.ak.data.vect.z);
-#else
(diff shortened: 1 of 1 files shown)


OpenUAS:
Note that fix should also be backported to v5.8

fvantienen:
Why didnt we have any problems before?

flixr:
This is weird, it would mean that it has been wrong all along, but the bebop was flying nicely regardless...
And I doubt that has anything to do with the firmware version, right?

OpenUAS:
Right..., @Freek, a look at the Blame and your mystery will be revealed ;) anyone flew 100% autonomous in January? Anyhow one can look oneself at RAW magneto, have a - and without and also leave on desk for 1 minute and look where your magneto goes :)

flixr:
@OpenUAS thanks!
Seems this is a regression from commit 25fddec9d0d89a7506e1ab54bcb38a5092b9dfcc

flixr:
Fix backported to v5.8 with 0a2c64a9a7cb6be6fbb6cdc32e2325e5a97f5d5e

PR #1547, opened 2016-02-26
[paparazzi center] make console non editable

flixr:
You could type in the console, but it is actually only used for output... modifying text in it makes no sense.
So make it non editable for the user...


hooperfly:
@flixr so maybe it's a status window vs. a console? Just semantics but maybe a more accurate term?

flixr:
Sure, but I don't care about the term used in the glade file...

gautierhattenberger:
I'm fine with both this one and #1546, maybe you want to integrate them without making merge points?

kevindehecker:
Uhm, I do not fancy to edit, but I very much did like to be able to clear the console. (which I did with select all + delete).... Should I make an issue out of this?

PR #1550, opened 2016-03-02
Pixhawk flies with intermcu on Iris

kevindehecker:
First PR for getting Pixhawk to play nice with PPRZ. 

The Pixhawk can be flashed without external tools (i.e. no JTAG interface is needed). Instead, the PX4 bootloader is kept in place and used by pprz to flash both the AP (stm32f4) and FBW (stm32f1) devices. One important remark is that for the FBW device, currently the computer needs to be connected to the Telem2 port using a FTDI cable. This is because pprz does not support serial communication over USB, or so I've heard.

In any case, the Iris+ drone factory equipped with a Pixhawk flies both on INDI and PID in ATT mode. However, PPRZ currently uses the MPU6000 IMU, which was broken in my Iris+ (and there are indications that maybe in others Iris' and X8's as well). I had to replace it with a fresh Pixhawk.

Also tested with a Pixhawk clone: seems to work just as well.

There are some things on the todo list:
1. Enable USB2serial communication :)
2. Led 3 (which is a led embedded in a separate push button to be mounted somewhere on the air-frame) is used to signal the FBW status. Should be made generic (either by a module, or by a LED def or something)
3. The 1500000 baud rate is necessary to remain compatible with flashing back to original PX4 firmware, however the communication has shown some stability issues. Strange, because original PX4 also uses this baud... The 24MHz external crystal used with the f1 is however outside the specs, so it may have something to do with that.
4. Drivers are missing for:
     The spi L3GD20H gyro 
     The multi color led
5. Only ATT mode has been tested. Autonomous flight using  GPS  baro and magneto has not been done or tested.

Specific to the Iris:
6. Enable the Gimbal controll through the RC


flixr:
Cool!
It would be really great if you could split this up in logical commit units and comment some of the important changes you made.

flixr:
Also please follow the style: 2 spaces for indentation of C files, no trailing whitespaces...

sw/airborne/subsystems/radio_control.c
```diff
@@ -64,7 +64,11 @@ void radio_control_periodic_task(void)
     radio_control.time_since_last_frame++;
   }
 
-#if defined RADIO_CONTROL_LED
+#ifndef RADIO_CONTROL_LED
```
flixr: What is this about?
`RADIO_CONTROL_LED` should simply be not defined if you don't have one...

flixr:
Btw, sw/tools/px4/ seems to be missing...

kevindehecker:
Yeah, I just noticed... I am preparing another push...!

flixr:
btw, in cases like this, there is no need to close the PR and create a new one.
You can simply update it by pushing additional commits to that branch (or even replace them by force pushing)

kevindehecker:
Sorry, it's what I did. must have clicked the wrong button... However, now it does not let me reopen it !?!?

flixr:
Hm... seems that you can't reopen the PR if you already force pushed to the branch after closing it...
Thats new ;-)
So I guess this time you will really have to make a new one...

kevindehecker:
Yeah. very weird, I'm even thinking I was not allowed to push -f / replace the first commit or something, because I really do not remember closing this.  Anyway, I'll open another one.

PR #1551, opened 2016-03-02
Pixhawk flies with intermcu on Iris (reopened)

kevindehecker:
First PR for getting Pixhawk to play nice with PPRZ.

The Pixhawk can be flashed without external tools (i.e. no JTAG interface is needed). Instead, the PX4 bootloader is kept in place and used by pprz to flash both the AP (stm32f4) and FBW (stm32f1) devices. One important remark is that for the FBW device, currently the computer needs to be connected to the Telem2 port using a FTDI cable. This is because pprz does not support serial communication over USB, or so I've heard.

In any case, the Iris+ drone factory equipped with a Pixhawk flies both on INDI and PID in ATT mode. However, PPRZ currently uses the MPU6000 IMU, which was broken in my Iris+ (and there are indications that maybe in others Iris' and X8's as well). I had to replace it with a fresh Pixhawk.

Also tested with a Pixhawk clone: seems to work just as well.

There are some things on the todo list:
1. Enable USB2serial communication :)
~~2. Led 3 (which is a led embedded in a separate push button to be mounted somewhere on the air-frame) is used to signal the FBW status. Should be made generic (either by a module, or by a LED def or something)~~
3. The 1500000 baud rate is necessary to remain compatible with flashing back to original PX4 firmware, however the communication has shown some stability issues. Strange, because original PX4 also uses this baud... The 24MHz external crystal used with the f1 is however outside the specs, so it may have something to do with that.
4. Drivers are missing for:
The spi LSM303D accelero+magneto 
The multi color led
5. Only ATT mode has been tested. Autonomous flight using GPS baro and magneto has not been done or tested.

Specific to the Iris:
6. Enable the Gimbal controll through the RC


kevindehecker:
Ran fix_code_style on all the .c and .h files, and added all the files (all new files got unstaged because I did a "manual squash").

Splitting in logical commits is going to be harder, because my original commit list is very messy because of countless merges and debug attempts and what not... Any suggestions here?

sw/airborne/arch/stm32/px4fmu_2.4.ld
```diff
@@ -19,15 +19,14 @@
...
 
 /* Define memory regions. */
 MEMORY
 {
-	/* only 192K (SRAM1 and SRAM2) are accessible by all AHB masters. */
-	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 192K
+    ram (rwx) : ORIGIN = 0x20000000, LENGTH = 192K
     /* Reserving 128kb flash for persistent settings. */
-    rom (rx) : ORIGIN = 0x08000000, LENGTH = 1920K
+    rom (rx) : ORIGIN = 0x08004000, LENGTH = 1008K
```
flixr: What's up with these changes?
This STMF427 should have 2M flash...

IMHO changing only this line to
`rom (rx) : ORIGIN = 0x08004000, LENGTH = 1904K`
would be correct to account for the change of the origin (where the px4 bootloader is)

conf/Makefile.px4
```diff
@@ -0,0 +1,36 @@
...
+# along with paparazzi; see the file COPYING.  If not, see
+# <http://www.gnu.org/licenses/>.
+#
+
+#
+# This Makefile uses the generic Makefile.arm-linux and adds upload rules for the PixHawk
+#
+include $(PAPARAZZI_SRC)/conf/Makefile.stm32-no_upload
+
+ifeq ($(FLASH_MODE),PX4_BOOTLOADER)
```
flixr: IMHO it would make more sense to add this to `conf/Makefile.stm32-upload` and remove `Makefile.px4` and `Makefile.stm32-no_upload`
That way there is no duplication and you also open the option to use other flash modes for px4 again...

kevindehecker: Ah, so that's how that is supposed to work. Will try to make that happen Friday.

kevindehecker:
I have neither defined or configured RADIO_CONTROL_LED in my [airframe file](https://github.com/kevindehecker/paparazzi/blob/PX4/conf/airframes/TUDELFT/tudelft_iris_indi.xml). It seems to be "none" by default, but only for the ap side. No idea why, it is kind of hard to track this kind of stuff through pprz.

I guess I could add a configure to fix it, but I don't really like that kind of pollution either... Will see if I can find out why it is none next Friday.

sw/airborne/arch/stm32/px4fmu_2.4.ld
kevindehecker: I got that from the PX4 repo [ld script](https://github.com/PX4/Firmware/blob/afb68cb0abee4034582f93d7f383c287b9c9a109/nuttx-configs/px4fmu-v2/scripts/ld.script). See the comments over there. Particularly note the "Disabled due to silicon errata flash (rx)". They have had a badge of Pixhawks with bad flash...

conf/firmwares/subsystems/rotorcraft/intermcu_uart.makefile
```diff
@@ -4,16 +4,18 @@
...
   fbw.srcs += subsystems/intermcu/intermcu_fbw.c
 else
   INTERMCU_PORT ?= UART3
+  INTERMCU_BAUD ?= B230400
   INTERMCU_PORT_LOWER = $(shell echo $(INTERMCU_PORT) | tr A-Z a-z)
   ap.CFLAGS += -DINTER_MCU_AP -DINTERMCU_LINK=$(INTERMCU_PORT_LOWER)
-  ap.CFLAGS += -DUSE_$(INTERMCU_PORT) -D$(INTERMCU_PORT)_BAUD=B230400
+  ap.CFLAGS += -DUSE_$(INTERMCU_PORT) -D$(INTERMCU_PORT)_BAUD=$(INTERMCU_BAUD)
   $(TARGET).CFLAGS += -DRADIO_CONTROL_TYPE_H=\"subsystems/intermcu/intermcu_ap.h\" -DRADIO_CONTROL
   $(TARGET).CFLAGS += -DRADIO_CONTROL_LED=$(RADIO_CONTROL_LED)
```
flixr: This should be
```
RADIO_CONTROL_LED ?= none
ifneq ($(RADIO_CONTROL_LED),none)
	$(TARGET).CFLAGS += -DRADIO_CONTROL_LED=$(RADIO_CONTROL_LED)
endif
```

flixr:
Regarding USB serial: if you want telemetry via USB, simply choose `<subsystem name="telemetry" type="transparent_usb"/>`....

conf/modules/px4io_flash.xml
```diff
@@ -0,0 +1,33 @@
...
+    <description>Flashes the px4io f1 through the px4 bootloader.</description>
+  </doc>
+  <header>
+    <file name="px4io_flash.h"/>
+  </header>
+  <init fun="px4ioflash_init()"/>
+  <event fun="px4ioflash_event()"/>
+  <makefile target="ap">
+    <raw>
+        PX4IO_UART ?= UART6
```
flixr: Since #1534 this can and should be replaced by
```
<configure name="PX4IO_UART" default="uart0" case="upper|lower"/>
<configure name="PX4IO_BAUD" default="1500000"/>

<configure name="TELEM2_UART" default="uart3" case="upper|lower"/>
<configure name="TELEM2_BAUD" default="B115200"/>
```

conf/modules/px4io_flash.xml
```diff
@@ -0,0 +1,33 @@
...
+        PX4IO_BAUD ?= 1500000
+        PX4IO_UART_LOWER=$(shell echo $(PX4IO_UART) | tr A-Z a-z)
+
+        TELEM2_UART ?= UART3
+        TELEM2_BAUD ?= B115200
+        TELEM2_UART_LOWER=$(shell echo $(TELEM2_UART) | tr A-Z a-z)
+    </raw>
+    <file name="px4io_flash.c"/>
+    <define name="USE_$(PX4IO_UART)"/>
```
flixr: should be `<define name="USE_$(PX4IO_UART_UPPER)"/>`

sw/airborne/arch/stm32/px4io_2.4.ld
```diff
@@ -0,0 +1,36 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/* Linker script for Lisa-M (STM32F105RCT6, 256K flash, 64K RAM). */
```
flixr: Please fix the comments.
Also doesn't seem that it has space reserved for persistent settings...

sw/airborne/firmwares/rotorcraft/main_fbw.c
```diff
@@ -175,9 +178,17 @@ STATIC_INLINE void main_periodic(void)
     }
   }
 
+
+  static uint16_t dv = 0;
+  // TODO make module out of led blink?
   /* set failsafe commands     */
   if (fbw_mode == FBW_MODE_FAILSAFE) {
     SetCommands(commands_failsafe);
+    if (!(dv++ % (PERIODIC_FREQUENCY / 20))) { LED_TOGGLE(3);}
```
flixr: Please no hardcoded LED numbers!

sw/airborne/modules/spektrum_soft_bind/spektrum_soft_bind.c
```diff
@@ -0,0 +1,30 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/spektrum_soft_bind/spektrum_soft_bind_fbw.c"
```
flixr: plz fix file header

sw/airborne/modules/spektrum_soft_bind/spektrum_soft_bind_ap.c
```diff
@@ -0,0 +1,39 @@
...
+#include "modules/spektrum_soft_bind/spektrum_soft_bind_ap.h"
+#include "subsystems/intermcu/intermcu_ap.h"
+#include "led.h"
+
+void spektrum_soft_bind_init(void) {}
+
+uint8_t bind_soft_value;
+void spektrum_soft_bind_click(uint8_t val __attribute__((unused)))
+{
+  LED_OFF(1);
```
flixr: Please no hardcoded LEDs!
Or if this is supposed to be a gpio, don't use the LED_OFF, but `gpio_set(port, gpio);`

sw/airborne/modules/spektrum_soft_bind/spektrum_soft_bind_fbw.c
```diff
@@ -0,0 +1,56 @@
...
+{
+  gpio_setup_output(SPEKTRUM_POWER_PIN_PORT, SPEKTRUM_POWER_PIN);
+  gpio_set(SPEKTRUM_POWER_PIN_PORT, SPEKTRUM_POWER_PIN);
+}
+
+void received_spektrum_soft_bind(void)
+{
+
+  //power cycle the spektrum
+  LED_OFF(1);
```
flixr: Please no hardcoded LEDs

conf/modules/px4io_flash.xml
flixr: Also Baud rates should be always prefixed with `B`, so `B1500000` instead of `1500000`
That define then needs to be added in `sw/airborne/arch/stm32/mcu_periph/uart_arch.h`

conf/modules/px4io_flash.xml
kevindehecker: Hmm, I got the PX4IO part to work but if I try to do the same for the TELEM2 port, it gives me 
modules/px4io_flash/px4io_flash.c: In function 'px4ioflash_event':
<command-line>:0:13: error: 'uart3' undeclared (first use in this function)

sw/airborne/firmwares/rotorcraft/main_fbw.c
kevindehecker: Pixhawk has this pushbutton with integrated led, that can be mounted somewhere on the airframe. So, I've added `PUSH_BUTTON_LED    ?= 3` to the px4io_2.4.makefile, but that does not seem to do it:

"/home/houjebek/paparazzi/sw/airborne/arch/stm32/led_hw.h:43:35: error: 'LED_PUSH_BUTTON_LED_GPIO_PIN' undeclared (first use in this function)"

How to do this?

flixr: Just creating a makefile variable doesn't make a C define magically appear.
You need to add in the intermcu makefile something like for the RADIO_CONTROL_LED.

And it would probably be better to name this led something like FBW_MODE_LED

sw/airborne/modules/px4io_flash/px4io_flash.c
```diff
@@ -0,0 +1,276 @@
...
+{
+  // setToBootloaderMode=true;
+  if (PX4IO_PORT->char_available(PX4IO_PORT->periph)) {
+    if (!setToBootloaderMode) {
+      //ignore anything coming from IO if not in bootloader mode (which should be nothing)
+    } else {
+      //relay everything from IO to the laptop
+      unsigned char b = PX4IO_PORT->get_byte(PX4IO_PORT->periph);
+      TELEM2_PORT->put_byte(TELEM2_PORT->periph, b);
+      TELEM2_PORT->send_message(TELEM2_PORT->periph);
```
kevindehecker: `TELEM2_PORT->put_byte(TELEM2_PORT->periph, b);
TELEM2_PORT->send_message(TELEM2_PORT->periph);`

When using usb serial, I have to add this send_message line. Otherwise, bytes will only be send when the fifo is full. With normal uart this is not the case.

Secondly, after switching to usb serial in commit 6aa775c44c76fc774fd229854d7818c8723ee130, the uploading of firmware resulted in a corrupted flash (does not boot anymore). Not 100% sure if this is the usb driver libopencm3 problem @fvantienen spoke to me about... but in any case this does not work yet. Any ideas?

flixr: `send_message` should be called each time a message if fully assembled to send the completed buffer:
https://github.com/paparazzi/pprzlink/blob/2b246f6480ac06bb6fc58b2a7947a59cec360a04/tools/generator/C/include_v1.0/pprzlink_device.h#L57

While this is not needed for pure character oriented devices like UART, it is the case for UDP and USB...
Calling it for after every byte is not how it's meant to be used...

kevindehecker: I managed to get FBW flashing over USB working... but currently I need a hack to do so: 4f220522862c0b830b5267d81c20378115afe7a6
In any case I've added automatic sending of the tx fifo if anything is in there from the event function, I also tried to make sure that manually calling the send_message function is not impacted by this change. However, chances are this fails due to the same reason I need the timeout hack. 

For some reason the usbd_ep_stall_get function does not seem to function properly always... Internally it checks the USB_EP_TX_STAT bit, which seems about right. No clue on how to further debug this. Any ideas?

flixr:
@podhrmic wrote the Paparazzi usb serial driver, hopefully he has some insight into this...

flixr:
Maybe you could also test with the latest libopencm3 version, as there have been some USB related fixes.
See #1557

kevindehecker:
@podhrmic , I think you may have gotten that reversed. The timeout hack is something I put there to get usb comm work flawlessly. In the last commit I only removed the timeout hack.

flixr:
@kevindehecker do you maybe want to merge the additions up to the uart flashing (which worked) and address the usb serial in a separate pull request or would you rather do it all at once?

kevindehecker:
@flixr, merge up to usb is fine by me! However, for other people to be able to use this easily the usb part is vital. So I hope to get that in one way or the other soon as well.

flixr:
Ok, merging the first part then.
Please make a new pull request with the usb serial stuff.

flixr:
@kevindehecker maybe you guys can also look into how to best handle some other split ap/fbw airframes (like TUDELFT/tudelft_logo600.xml) that are now failing?

PR #1554, opened 2016-03-07
simulate airspeed sensor in NPS

flixr:
Get airspeed, pressure and temperature from FDM and add an airspeed sensor that directly pushes equivalent airspeed to the state interface.

Also fix some other NPS related things:
- disable airspeed_adc module for NPS
- fix NPS_BYPASS_NPS for fixedwings

### DIFF
--- a/conf/airframes/examples/easystar_ets.xml
+++ b/conf/airframes/examples/easystar_ets.xml
@@ -2,49 +2,54 @@
 
 <!--
     Multiplex EasyStar, using rudder
-    TWOG v1 board
-    Tilted infrared sensor
+    Lisa/MX board
     XBee 2.4GHz modem in transparent mode
 -->
 
-<airframe name="EasyStar2 - TWOG v1">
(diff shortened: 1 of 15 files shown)


flixr:
@gautierhattenberger anything to add here?
While it might be better to get airspeed directly from the FDM (if possible), this seems to work ok for me..

gautierhattenberger:
It looks fine for me. At some point I think only air_data should update state interface and sensor driver only publish ABI messages. But it is for a later PR. Also, here it is TAS that is computed. Probably good enough for simulation, but actual aircraft are using EAS to fly.

flixr:
Added airspeed, pressure and temperature to the FDM interface (and get it in JSBSim case).
Airspeed sensor now uses equivalent airspeed from FDM.

As you said it would make more sense to only update the state interface from air_data, but currently most airspeed sensor modules already compute airspeed from dynamic pressure and set the state interface directly.

PR #1557, opened 2016-03-09
[ext] update to latest libopencm3

flixr:
and update clock setup accordingly

### DIFF
--- a/sw/airborne/arch/stm32/mcu_arch.c
+++ b/sw/airborne/arch/stm32/mcu_arch.c
@@ -41,7 +41,7 @@
 
 #if defined(STM32F4)
 /* untested, should go into libopencm3 */
-const clock_scale_t hse_24mhz_3v3[CLOCK_3V3_END] = {
+const struct rcc_clock_scale rcc_hse_24mhz_3v3[RCC_CLOCK_3V3_END] = {
   { /* 48MHz */
     .pllm = 24,
     .plln = 96,
@@ -51,10 +51,24 @@ const clock_scale_t hse_24mhz_3v3[CLOCK_3V3_END] = {
     .ppre1 = RCC_CFGR_PPRE_DIV_4,
     .ppre2 = RCC_CFGR_PPRE_DIV_2,
     .power_save = 1,
-    .flash_config = FLASH_ACR_ICE | FLASH_ACR_DCE |
-    FLASH_ACR_LATENCY_3WS,
+    .flash_config = FLASH_ACR_ICE | FLASH_ACR_DCE | FLASH_ACR_LATENCY_3WS,
+    .ahb_frequency  = 48000000,
     .apb1_frequency = 12000000,
     .apb2_frequency = 24000000,
+  },
+  { /* 84MHz */
+    .pllm = 24,
+    .plln = 336,
+    .pllp = 4,
+    .pllq = 7,
+    .hpre = RCC_CFGR_HPRE_DIV_NONE,
+    .ppre1 = RCC_CFGR_PPRE_DIV_2,
+    .ppre2 = RCC_CFGR_PPRE_DIV_NONE,
+    .power_save = 1,
+    .flash_config = FLASH_ACR_ICE | FLASH_ACR_DCE | FLASH_ACR_LATENCY_2WS,
+    .ahb_frequency  = 84000000,
+    .apb1_frequency = 42000000,
+    .apb2_frequency = 84000000,
   },
   { /* 120MHz */
     .pllm = 24,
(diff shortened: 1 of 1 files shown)


kevindehecker:
This does not compile for me... 

arch/stm32/mcu_arch.c:44:48: error: 'RCC_CLOCK_3V3_END' undeclared here (not in a function)
 const struct rcc_clock_scale rcc_hse_24mhz_3v3[RCC_CLOCK_3V3_END] = {
                                                ^
arch/stm32/mcu_arch.c:46:5: error: field name not in record or union initializer
     .pllm = 24,
     ^
arch/stm32/mcu_arch.c:46:5: error: (near initialization for 'rcc_hse_24mhz_3v3')
arch/stm32/mcu_arch.c:47:5: error: field name not in record or union initializer
     .plln = 96,
     ^
arch/stm32/mcu_arch.c:47:5: error: (near initialization for 'rcc_hse_24mhz_3v3')
arch/stm32/mcu_arch.c:48:5: error: field name not in record or union initializer

......

dewagter:
@kevin: was this not our clock setup code for the 24MHz crystal? Probably
it must be re-copy pasted and changed with the new opencm define names.

kevindehecker:
I tried it with a clean master on the Logo600 airframe. Should not have anything to do with PX4 config.

flixr:
@kevindehecker then you didn't call toplevel make first, it needs to fetch the updated submodule and recompile the libopencm3 libs...

kevindehecker:
Nop, did that to. However, I first commited after the cherry-pick because of that submodule  nonsense ( it reported uncommited changes for the submodule). Maybe that was something I was not supposed to do...

flixr:
you probably changed the submodule back to the previous version or so...

kevindehecker:
Apparently I did...  Managed to compile the AP for my Iris. 

Next problem; after uploading the AP, ttyACM0 does not show up anymore (or any other new tty device for that matter)... Sys time led is going strong though, so the AP seems to run fine.

flixr:
hm... so there must have been some changes in libopencm3 that make this silently not work anymore.
I can test this on my Lisa/MX on the weekend.

flixr:
So transparent_usb telemetry still works fine on my LisaMX.

Other things are working fine though? Are you sure the clock setup is correct in your case?
Also is that on the F4 or the F1?

kevindehecker:
It's on the F4, so the clock config should be fairly standard. I'm however not using telemetry over usb though (and as such did not set transparant_usb), but using usb serial directly from a module. Which works with the current libopencm3, but apparently not with this newest version.

flixr:
transparent_usb uses the same usb serial code, so it shouldn't really make a difference... (except maybe where/when you call send_message in your module?).
Also note that I tried it with the usb_ser_hw from this PR, so without the changes you made for the px4io flashing...
Maybe you could first confirm that transparent_usb is working for you, then hunt for the problem?

kevindehecker:
OK, also transparant_usb is not working. No /dev/ttyACM device appears.

kevindehecker:
Think I may have found something, the new libopencm contains another clock config (84MHz), which is not implemented in the pprz version of the 24 ext crystal function...

sw/airborne/arch/stm32/mcu_arch.c
```diff
@@ -41,48 +41,51 @@
...
+    .hpre = RCC_CFGR_HPRE_DIV_NONE,
+    .ppre1 = RCC_CFGR_PPRE_DIV_4,
+    .ppre2 = RCC_CFGR_PPRE_DIV_2,
+    .power_save = 1,
+    .flash_config = FLASH_ACR_ICE | FLASH_ACR_DCE |
+    FLASH_ACR_LATENCY_3WS,
+    .ahb_frequency  = 48000000,
+    .apb1_frequency = 12000000,
+    .apb2_frequency = 24000000,
+  },
```
kevindehecker: Yep, there should be another 84MHz config here! That solves my problem.

flixr:
Rebased, added missing 84MHz entry and updated F1 24MHz setup accordingly as well...

flixr:
@kevindehecker just to clarify: what is working for you now? You get a device again, but is the px4io flashing working now as well, or same problem as before the libopencm3 update?

kevindehecker:
Yes, the device works, but unfortunately the new libopencm3 does not solve my PX4 usb flashing problem :(.

flixr:
Bummer... but I think it would be good to merge this anyway.
Maybe @podhrmic @gautierhattenberger can confirm stuff is still working on Apogee, Lia, etc..

kevindehecker:
Yes please merge. I will work on an usb issue Monday.

flixr:
@gautierhattenberger any objections in merging this?
Maybe you could test on apogee as well first?

gautierhattenberger:
Seems to work on Apogee with and without ChibiOS mix.

PR #1558, opened 2016-03-09
edgeflow: improved optic flow calculation

knmcguire:
Hi,

I developed a new method to calculate the global optical flow in a efficient way. It should run twice as fast as the current implementation.

I tried to rewrite the opticflow_calculator.c a bit so that more methods can be added more easily, since before it was specifically written for fast9/LK implementation. On top of that I tried to use the existing structures as much as possible, so that I can switch between different methods on the fly. A couple of new files are added to the lib folder (edge_flow.c & .h)

Any comments/ suggestions for changes are welcome. @dewagter, can you also take a look at this? 

Cheers, K


flixr:
Looks good to me, and nicely documented parameters, etc :+1:

although I have no means of testing it...

flixr:
Just as a reminder for all involved, not really for this pull request:
The capturing should not be done by the opticflow module...
@dewagter already changed that once, see #1370

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
```diff
@@ -308,6 +328,167 @@ void opticflow_calc_frame(struct opticflow_t *opticflow, struct opticflow_state_
...
+void opticflow_calc_frame(struct opticflow_t *opticflow, struct opticflow_state_t *state, struct image_t *img,
+                          struct opticflow_result_t *result)
+{
+  if (opticflow->method == 0) {
+    calc_fast9_lukas_kanade(opticflow, state, img, result);
+  } else {
+    if (opticflow->method == 1) {
+      calc_edgeflow_tot(opticflow, state, img, result);
+    } else {
+      PRINT_CONFIG_MSG("Both edgeflow and Lukas kanade is not turned on. Define either USE_LK or use_EDGEFLOW on TRUE!");
```
flixr: This does not work as you probably intended...
`PRINT_CONFIG_MSG` will print the message during compilation if the "print config" checkbox is enabled.
Since this already a preprocessor macro, the else condition does nothing here and it will always be printed.

knmcguire:
Hi Felix thanks for your comments, I'll try to fix these this afternoon.

Also it seems that the opticflow airframe in TUDELFT folder was not updated to include
+    <load name="video_thread.xml">
 +      <define name="VIDEO_THREAD_FPS" value="30"/>
 +      <define name="VIDEO_THREAD_CAMERA" value="bottom_camera"/>
 +    </load>

I used that airframe to test it...However... it still captured the bottom camera.

 I see that in the opticflow module, the video device is also initialized, so with this extra module it will be doing this double, which it is now doing in the example opticalflow airframe. I think I should remove this completely from the opticflow_module.c to make it more up to date to the current changes right?

knmcguire:
The dots mean  <load name="video_thread.xml"> etc.. btw

knmcguire:
with the dots I mean the video_thread module btw

flixr:
Well, while the capturing should indeed be done by the video_thread module instead of directly in the opticflow_module (which btw is a bit of a misnomer), this will only be a real option as soon as the video_thread module supports capturing multiple video sources... otherwise you'd have no option to use the front camera at the same time...

knmcguire:
Ah okay, I also see now that the example optical flow doesn't have the thread module yet indeed, so I'll leave it at this for now. I did change the config msg to a error message on the top at the defines, so it will only show if define METHOD goes over 1. 

Also I added a derotation setting to turn it off if the gyroscopes measurements is too noisy or such. It was not the case with my ardrone but I experienced cases where the quality is better when it is turned off. 

Anyway, I hope the pull request is okay now?

flixr:
Since you also requested a review from @dewagter, I'll let him review and merge it...

knmcguire:
Oke no problem. @fvantienen is a good candidate as well since he made the major change to the code the last time

dewagter:
@flixr Just read through the code. Looks great. I see no reason not to merge as is.

Will keep in mind to plug opticflow in videothread as soon as we can start multiple threads.

PR #1561, opened 2016-03-09
[stm32] also compile C++ files

flixr:
only tested with some dummy C++ files, but it should work fine for all.


podhrmic:
@flixr why do we want to be able to compile C++ files for STM32 autopilots?

flixr:
Why not?
Somebody might want to use a C++ lib (e.g. Eigen) or port something to C++, or ...

podhrmic:
OK, I guess it is more like "useful in the future" feature then. It would be interesting what including C++ files does with code size, but that is a different story...

flixr:
That would strongly depend what C++ features you use...

conf/Makefile.stm32
```diff
@@ -199,8 +222,8 @@ sym: $(OBJDIR)/$(TARGET).sym
 
 # Link: create ELF output file from object files.
 .SECONDARY : $(OBJDIR)/$(TARGET).elf
-.PRECIOUS : $(COBJ) $(AOBJ)
-%.elf: $(COBJ) $(AOBJ) | $(OBJDIR)
+.PRECIOUS : $(OBJS) $(AOBJ)
+%.elf: $(OBJS) $(AOBJ) | $(OBJDIR)
 	@echo LD $@
 	$(Q)$(LD) $(LDFLAGS) $($(TARGET).LDFLAGS) -o $@ $(COBJ) $(AOBJ) $(LDLIBS)
```
dewagter: Whenever a cpp file is compiled, you need to link with g++ instead of gcc

flixr: Any reason not to always use g++ for linking?

podhrmic:
@flixr should we merge this? Seems to be minor enough.

flixr:
With this g++ is always used for linking...
I thought that the only difference linking with g++ is that -lstdc++ added and if nothing from libstdc++ is used, it *should* make no difference.

That is *nearly* the case, the lib order also changed:
* gcc: `-lc -lm -lgcc `
* g++: `-lgcc -lstdc++ -lm -lc`

Linking e.g. Quad_LisaMX with g++ actually results in a slightly smaller elf than linking with gcc, probably because the link order changed there are some symbols taken from libcc instead of libc.

podhrmic:
@dewagter @gautierhattenberger should I merge this as is?

gautierhattenberger:
On the principle I'm fine, but I would prefer to test it before merging (okay, could have done that earlier... )

podhrmic:
Merging since it doesn't seem to break anything and having C++ support is useful.

PR #1562, opened 2016-03-10
Bebop camera luminosity increase.

Pax1601:
Using the datasheet of the Bebop front camera, I've edited the video.c source code to increase the gains and obtain a better image.


flixr:
Sounds good to me :-)
Is the gain always fixed? Only shutter time is adjusted dynamically?
@gautierhattenberger @fvantienen & co.
Could anyone else with a Bebop please test this as well?

Pax1601:
It is my understanding that both the shutter time and the gains are completely fixed (right now). I've modified the values of the gains so that it can work indoors. A better idea would be to allow the user to define the gains, maybe with different presets, for indoors and outdoors usage. These values are a bit too high for outdoors use, but I believe they are better overall.

dewagter:
Shutter time and white balance are not automatic on the bebop front camera
but must be adjusted by user code.
Another commit is expected which will process the image in the ISP which
also returns histograms to adjust both from user space.

In the mean time it's OK for me to adjust the shutter time to indoor
conditions. Even better would be to create a setting that can be adjusted
with a slider from the GCS and saved:

Settings:

LUM: auto (not supported yet) - manual
Shutter - slider

AWB: auto (not supported yet) - manual
Red - slider
Green - slider
Blue - slider

-Christophe

OpenUAS:
A merge as interim solution maybe... @Pax1601 thx for you great work done, the frontcam datasheet is quite a hefty amount of pages...

PR #1564, opened 2016-03-10
fixed course conf

rmeertens:
removed stabilisation rate, now the airframes compile!


flixr:
Can you please also move the files in `conf/TUDELFT` to `conf/aircrafts/TUDELFT` so it's consistent with the rest?

PR #1565, opened 2016-03-11
Nps wind

gautierhattenberger:
some more wind support for NPS
- direct input from flight gear socket
- request for WORLD_ENV message

### DIFF
--- a/sw/simulator/fg_environment.xml
+++ b/sw/simulator/fg_environment.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0"?>
+
+<PropertyList>
+  <generic>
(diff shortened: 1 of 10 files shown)


flixr:
Nice!
Wonder what happens if you have FlightGear and gaia (environment sim) running?
Won't this cause problems and the wind will always switch between the one set in FG vs. gaia?

gautierhattenberger:
I guess you should avoid to use both at the same time for now. At least you can disable FG input by setting the port_in to 0.

flixr:
Ok, I guess we should note that on the wiki then...

PR #1566, opened 2016-03-14
Pixhawk flashing via USB

kevindehecker:
Add flash through usb functionality PX4.

Automatically detects which target is being flashed, and redirects accordingly.


sw/airborne/mcu_periph/uart.h
```diff
@@ -36,8 +36,13 @@
 #define UART_RX_BUFFER_SIZE 128
 #endif
 #ifndef UART_TX_BUFFER_SIZE
+#ifdef STM32F4  //the F4 has enough memory, and the PX4 bootloader needs more then 128
```
flixr: ? why change UART TX buffer when changing to USB?

kevindehecker: In order to flash the fbw f1 through the same USB cable (which is only connected to the AP f4), the f4 is set in a relay mode (pc <-> usb <-> f4 <-> uart <-> f1) . The PX4 upload script uses blocks of 252 bytes, which overan the internal UART buffer of the f4.

podhrmic: Maybe it would be a good idea to increase the default buffer size? See #1453

flixr: I'm really not in favour of increasing default buffer size, it's not needed for 99% of use cases and on some of the low end LPCs/F1s not even possible to to ram size.

Defining a bigger buffer for e.g. linux or certain STMs could be done in their respective `uart_arch.h` files.

flixr:
Please follow pep8 (4 spaces for indentation) for python files as per our style guide...

sw/airborne/mcu_periph/uart.h
kevindehecker: Which is why I took the compromise road here and changed it only for the f4 :)

So, what would you like me to do here? Move the define to the uart_arch.h?

flixr: As you want.. if you move it to uart_arch.h, you might as well rais the rx buffer size to the same value for F4.

flixr:
So is it working reliably now?
Also it's not like longer multi-line descriptions aren't appreciated, just have one very brief one in the first line... ;-)

sw/airborne/mcu_periph/uart.h
kevindehecker: In that case, I prefer leaving it as is. Moving it to uart_arch.h still requires a ifndef and the ifdef f4. Moreover it only makes it less clear to have this definition redefined at another place.

kevindehecker:
Yes, this works reliably as far as I know.

flixr:
Ok, cool!
The usb serial will need to be tested with transparent_usb on other boards as well.

It would really be great if you could add some comments on the timeout and (potientially infinite??) while usbd stall loop.
Also it really needs to be guaranteed that this will never turn into an infinite loop.

flixr:
@podhrmic maybe you can test this as well?

kevindehecker:
I checked the transparant_usb on the Pixhawk (temporarily disabled the flash module) yesterday, it seems to work as before.

Added some extra comments on the timeout stuff. About the potential infinite while loop, not sure what you want me to say here. Internally it checks the USB_EP_TX_STAT bit, which seems about right.... but that is about all I know on this. That, and the fact it has never gone infinite on me up till now :)

flixr:
@gautierhattenberger @podhrmic any feedback here?

podhrmic:
I ll test it on Lisa MX today.

podhrmic:
`transparent_usb telemetry` works on Lisa MX

`usb_serial_stm32_example1` and `usb_serial_stm32_example2` modules don't compile (`undefined reference to `UsbSChAvailable` ... )

OpenUAS:
@podhrmic Thx for testing!on Lisa MX

podhrmic:
You're welcome:) FYI with the latest master I am getting the same results (usb_telemetry compiles, but the usb module examples don't).

flixr:
@podhrmic the usb module examples should be fixed with 0d4c628b9115b0dc3a34d00f60ae34f6e895fbd6

Issue #1545, opened 2016-02-23
Back up message box very unclear


kevindehecker:
![image](https://cloud.githubusercontent.com/assets/5213220/13252710/7b775874-da38-11e5-971d-0869e7d81e05.png)

At seemingly random moments, I'm confronted with this backup dialog during startup of pprz Center, which I do not want to see nor read because:
1. It is totally irrelevant at that moment. (trying to fly here!)
2. it's unclear what I'm supposed to do... even if I read it. 
3. The most persuasive and save looking button (restore auto backup), actually destroys last made settings.

Suggest to remove that dialog, and handle this without user attention. Modern GUI auto-saves without requiring user attention, but automatically creates a back-up of the previous state in case the user did not want to save.

gautierhattenberger:
Hum, I almost sure this discussion already happen somewhere...
1. I agree
2. I don't totally agree
3. Actually reading all the lines tells you that by restoring the backup you will lose the latest manual changes

Yet, I agree this is not the ideal way to do this.

But:
- unless someone wants to do some ocaml, I won't do it because I don't have the time right now
- we have a student that just started working on replacement of paparazzi center written in Python (and with a different way to save sessions)

So in the meantime, I suggest to save your session before leaving the paparazzi center or choose "keep current". Sorry.

kevindehecker:
Oooh, ah new Python Center. I'll happily wait for that :)

dewagter:
I would suggest to add a line to say:
If you recently checked out a different branch, this will have changed your
conf. To keep this new conf and forget the previous conf from the other
branch press ...

kevindehecker:
If we are to keep the msgbox for now, I'm thinking more along the line of:

"Configuration changes to conf/conf.xml were either not saved during the last session, or external changes were applied (e.g. from a git branch).\n In the first case, please choose [Keep current] to save the new settings permanently. In the second case, please choose [Discard local changes] to reset to the new settings of that branch."

[Keep current] [Discard local changes] [View changes]

Also, I'd change the default button to the safer [Keep current] option.

I can create a PR if this is ok

gautierhattenberger:
The first line is fine, but in the first case for instance, I prefer sometimes to discard local changes to go back to the previous version. The rest of the text implies that can't or shouldn't do that even if it is perfectly fine. So I suggest to at least add "or choose [Discard local changes] to restore previous settings.
[...]

PR #1567, opened 2016-03-14
Clarified backup msgbox

kevindehecker:
See #1545.

We've kind of left it there mid-hanging. So, this may not be perfect, but at least clearer than before. And since there is a new pprz center coming up anyway...

### DIFF
--- a/sw/supervision/paparazzicenter.ml
+++ b/sw/supervision/paparazzicenter.ml
@@ -182,8 +182,8 @@ let () =
 
   if Sys.file_exists Utils.backup_xml_file then begin
     let rec question_box = fun () ->
-      let message = "Configuration changes to conf/conf.xml were not saved during the last session.\nYou can either keep the current version or restore the auto-saved backup from the last session to discard the changes.\nIf you made any manual changes to conf/conf.xml and choose [Restore auto-backup] you will lose these." in
-      match GToolbox.question_box ~title:"Backup" ~buttons:["Keep current"; "Restore auto-backup"; "View changes"] ~default:2 message with
+      let message = "Your conf/conf.xml is different than the last automatic backup.\n(e.g. forgot to press save, manually edited conf.xml, or changed git branch).\n" in
+      match GToolbox.question_box ~title:"Warning: unsaved conf changes detected" ~buttons:["Use conf.xml (as is)"; "Discard local changes"; "View changes"] ~default:1 message with
       | 2 -> Sys.rename Utils.backup_xml_file Utils.conf_xml_file
       | 3 -> ignore (Sys.command (sprintf "meld %s %s" Utils.backup_xml_file Utils.conf_xml_file)); question_box ()
       | _ -> Sys.remove Utils.backup_xml_file in


gautierhattenberger:
It seems to the best consensus from previous discussion. I don't have strong opinion thought, but I think it is fine like this.

flixr:
IMHO "discard local changes" is still misleading... to me that implies that there are "local changes" that could be "discarded", but that doesn't have to be the case:
E.g. if you switch branches or configs it doesn't mean you have local changes, but the backup from the last session can be from a different conf and hence not match....

kevindehecker:
I'm still a bit fuzzy on the logic here, but I think that exactly does mean you have local changes. Where the changes originated from is of less concern.

In your example, the backup contains the original conf from branch A, there should be no changes. After you switch to branch B, the backup still contains the conf from branch A ... which are now local changes since we are now in fact on branch B....

flixr:
What I mean is that (at least IMHO) the term "local changes" is confusing, as that usually refers to local changes in your git repository (i.e. changes that are not committed).

E.g. following the setup you nicely described above:
- you have a clean working dir after switching branches
- start paparazzi and click on "discard local changes"
- now you actually end up with changes in your working directory... but wait I thought I discarded all local changes???

Maybe it's just me, but naming that button "discard local changes" doesn't make sense to me.
Personally I preferred something like "restore/use/load auto-backup conf from previous session"

martinmm:
Maybe I am not really understanding the OCAML here, but it seems that the logic functionality is much simpler than what it seems. The center just looks for a conf/conf.xml~ and pops up a message if there is one. It could also be a backup of a backup (if you saved twice externally) or just any file, even an empty one. The many words in the message box give you a (wrong) impression of a lot of checks and logic in the background.

So the real message is more like "Found an older copy of conf.xml, should I restore it?"

(I always pressed "Keep current" but even after all these years it still confuses me and it takes 5 seconds whenever it pops up)

flixr:
@martinmm yes, it's actually that simple...

Maybe we are looking at this the wrong way anyway?
I wouldn't mind if we got rid of that "auto-backup" and hence also the need for this dialog.
That could mean that instead of saving to a backup, we always directly save to the conf.xml or require the user to explicitly save like it's the case so far.
IMHO, ever since you could switch between different conf.xmls, this simple backup didn't make that much sense any more...

podhrmic:
@martinmm nice summary - I actually like your "message" the most so far, because it is easy to understand:)

kevindehecker:
OK, so "Found an older copy of conf.xml, should I restore it?" is basically what the msgbox in this PR says: "Your conf/conf.xml is different than the last automatic backup". And while I'd have no problem changing the text to the former, then we will get another discussion about some tiny things that are not 100% correct. ('older'?).  So a consensus here will be unlikely (proven by the 30 or so comments up to now).

@flixr your git workflow is a valid point. However, the git use case should be of lower priority than actual pprz use cases. 

I do not believe there is a perfect solution for this using the msgbox. This has been my starting point all along in #1567... However, changing the logic requires OCAML business, which I won't do for all sorts of reasons (an important one being the upcoming new python pprz center). 

<sarcasm alert> Currently the msgbox is real bad... my favorite go-to example on how to not do GUI design kind of bad. I've used it in no less then 2 powerpoints :). Even this would be a better implementation:
image
At least here it will not have you loose changes on purpose :)

So, can we please merge this as is? And then make another PR for improved logic whoever wants to?

flixr:
Personally I'm not going to merge it as is with "discard local changes", since as I think it's more confusing than before.
And what is "actual pprz use case" supposed to mean?
If everyone else likes this better, someone else can still merge this...

EwoudSmeur:
haha, nice discussion. I am pro deleting the message altogether, I only pressed the keep current button anyway (but only after I learned the use from reinforcement learning)

gautierhattenberger:
Let's add a bit noise here: adding a confirmation box to save the conf on window closing was actually pretty easy, it was already the case when closing the window from the menu.
See https://github.com/paparazzi/paparazzi/commit/f6b22dd92f6ebb8b680581bfc8b062cadd19b5da
But actually it is not solving all the problems since you can change the conf by hand or from git checkout.

flixr:
So shall we close this no or still re-word the msgbox text?

dewagter:
@gautierhattenberger: confirmation on exit is great

@flixr: for me the phrasing "discard local changes" is confusing. I'd
rather have "use conf/conf.xml" & "use var/temp/autosave/conf.xml" or
something more technical than some vague statement about losing changes:
you loose with either button and both are local

@all: about the "load" dialog:

IMHO:

 - previously conf files were not committed, so it was crucial not to
overwrite them and make an auto-backup in a different folder
 - now that they are on git, so for me we can simply auto-save to the real
conf (no more temp conf)
 - since there is an option to skip autosave on exit, this means you only
overwrite conf/conf.xml if you chose to
 - this would remove the need for the message box completely

-Christophe

flixr:
With the danger of repeating myself:
- `conf.xml` was and is always saved, there is and was no "temp conf"
- the "backup" `conf.xml~` is created as soon as you change the conf
- upon clicking "save" the "backup" `conf.xml~` was/is simply deleted
- you normally shouldn't see the msgbox at startup anymore because `conf.xml~` won't exist if you exited normally (either chose "keep changes" or "restore old version")

So I guess the wording doesn't matter much any more since the msgbox will only appear if you killed pprz center (instead of normal quit) after making changes

dewagter:
@felixr

I was just wondering about the need for this "step 2". Why not replace it
with: (if not-saved-closing-dialog-checkbox "always safe" is checked: then
overwrite conf.xml right away, and otherwise do nothing [until you close
paparazzi where you will get a question])

-Christophe

flixr:
I think you still misunderstand: conf.xml has always been directly modified (and still is), since all tools operate on that

dewagter:
Yes, my message is indeed the wrong way around, sorry.

MNHPS: _(My New Humble Personal Suggestion)_

 - if you checked `"always save"` in the exit saving dialog, then never make a `conf.xml~`
 - if you have a `conf.xml~` (e.g. killed paparazzi center), use @martinmm formulation: `"Found an older copy of conf.xml, should I restore it?"`
 - when you close paparazzi center without saving via the menu: change the `"your conf was not saved"` (which is not correct) into something like: `"All you changes to conf.xml are saved instantaneously. If you want to undo the changes you made since you started paparazzi center, click **REVERT**"`

podhrmic:
Btw this is probably the most discussed pull request in a while - although almost certainly not the most important one...

flixr:
> if you checked `"always save"` in the exit saving dialog, then never make a `conf.xml~`

Sure we can do that, but that precludes your next point (in case you selected "always save") and also the option that @kirkscheper proposed (a menu entry that allows you to "restore old version" at any time).
Either way it's fine with me...

The proposed rewording sounds good, tried to be a bit more brief though... see #1640

flixr:
@podhrmic yeah... very true...

flixr:
Closing this PR as the wording is hopefully ok now.
If someone wants different behavior of the session backup or a possibility to revert from the menu, please open a new issue/PR.

PR #1568, opened 2016-03-22
Navigation oval pullreq

tcunis:
So far, oval navigation as described in the [wiki](http://wiki.paparazziuav.org/wiki/Flight_Plans#Oval) has been implemented for fixed-wing only. I have adapted the fixed-wing implementation and ported to rotorcraft as well. Unlike fixed-wing, the rotorcraft implementation uses binary fixed-point representation rather than floats.


flixr:
Nice!
I think eventually we should convert all these navigation routines to modules, but it's ok like this for now I guess...

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -148,11 +148,16 @@ static void send_wp_moved(struct transport_tx *trans, struct link_device *dev)
   static uint8_t i;
   i++;
   if (i >= nb_waypoint) { i = 0; }
+  struct EnuCoor_i wp_i;
```
flixr: What's up with those changes?
They don't seem to change anything functionally, just add some extra copying...

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -547,3 +552,107 @@ bool_t nav_set_heading_current(void)
...
+  }
+  int32_t qdr_anticipation = ANGLE_BFP_OF_REAL(radius > 0 ? -15 : 15);
+
+  switch (oval_status) {
+    case OC1 :
+      nav_circle(&p1_center, POS_BFP_OF_REAL(-radius));
+      if (NavQdrCloseTo(INT32_DEG_OF_RAD(qdr_out_1) - qdr_anticipation)) {
+        oval_status = OR12;
+        InitStage();
+//        LINE_START_FUNCTION;
```
flixr: I think it would be worth keeping those (for the cam modules).
You can let them default to something empty like in fixedwing/navigation.c

sw/airborne/firmwares/rotorcraft/navigation.h
```diff
@@ -111,6 +111,8 @@ extern bool_t nav_set_heading_current(void);
 #define NavSetWaypointHere(_wp) ({ waypoint_set_here_2d(_wp); FALSE; })
 #define NavCopyWaypoint(_wp1, _wp2) ({ waypoint_copy(_wp1, _wp2); FALSE; })
 #define NavCopyWaypointPositionOnly(_wp1, _wp2) ({ waypoint_position_copy(_wp1, _wp2); FALSE; })
+#define NavSetWaypointPos(_wp, _enu_x, _enu_y) ({ waypoint_set_xy(_wp, _enu_x, _enu_y); FALSE; })
```
flixr: What's the reason for adding these?
I'm not in favour of adding more macros unless needed...

sw/airborne/firmwares/rotorcraft/navigation.c
tcunis: That's an artefact of the project I am actually working on, wherefore I needed to increase INT32_POS_FRAC. Then, this was to ensure the telemetry precision was still 8 bits as defined in messages.xml.

sw/airborne/firmwares/rotorcraft/navigation.c
tcunis: I agree. To be honest, I wasn't sure where these macros were for.

sw/airborne/firmwares/rotorcraft/navigation.h
tcunis: Also not related I am afraid. However, I prefer calling macros rather than functions within the flight plan.

OpenUAS:
@tcunis Nice, if a bit more discussed what would be the best way to add with PPRZ team, and in Lab... so almost there to merge :+1:

tcunis:
I have included the changes as discussed so far (line-start / -stop, removed un-related macros etc.).

flixr:
@gautierhattenberger you can merge it if it looks good to you...

sw/airborne/firmwares/rotorcraft/navigation.c
gautierhattenberger: I understand why you have that, but I don't think it should appear in master. Would be nice if you remove it for now, and eventually add it back if you decide to make a pull request for the code that is actually using this.

gautierhattenberger:
merged with ab1249b7092b0709516581b61c386ab92c600a57 (and c4d7283b8deb77336d0bdb5f3550598c2566a569 to fix my own mistake, sorry)

PR #1569, opened 2016-03-22
Px4 imu

kevindehecker:
1. Wrote a spi driver for the lsm303dlhc. (loosely based on the I2C driver, of which I'm 99.9% sure that it didn't work). May still need some streamlining, but because the I2C driver is such a mess that's not trivial.
2. Created a PX4 imu, which consists out of 
   a) lsm303dlhc (acc) + l3g (gyro) IMU, both on spi
   b) the MPU6050 (acc + gyro) IMU, on spi.
   c) 2 magneto's: one in the lsm303dlhc and an external i2c hmc5811 in the 3dr gps
3. Changed the ms5611 baro to spi

(this PR builds on #1566, which was not merged yet)


flixr:
Nice!
Will be great to have a SPI driver for the lsm303d :-)
But as you said I think this still needs some cleanup and refactoring (i.e. pulling out the common parts).

Also I don't understand why you would add the external hmc5811 mag to the imu driver?
You should use a module if you want to use the external mag...

Please also remove the qt_project.py as it's unrelated to this PR.

kevindehecker:
About the clean up, yeah I was considering removing the I2C stuff, because I don't have the hardware and I'm sure it did not work before anyway... (so apparently nobody is using it). 

About the hmc, uhm I was following the imu_mpu6000_hmc5883 code as an example here... will figure out the module thing.

flixr:
The imu_mpu6000_hmc5883 driver is for hardware modules that have the hmc included, on the pixhawk this is an optional and external module..

flixr:
It would be nice if we could keep the lsm303 i2c driver, it was cherry picked from https://github.com/paparazzi/paparazzi/pull/625

flixr:
Please also note that updating the global `imu` struct from multiple sources is not properly supported, i.e. you can't tell from which sensor it was last updated...
Also if you have multiple ABI accel/gyro/mag publishers, one should configure the consumers/subscribers to use only one specific IMU (filter on sender ID)...

flixr:
Also right now you can have only one set of IMU sens/neutral defines, here it would apply the same to both IMUs...
I don't think enabling both sensors (MPU and LSM303D) makes much sense at this point...

kevindehecker:
> Please also note that updating the global imu struct from multiple sources is not properly supported, i.e. you can't tell from which sensor it was last updated...

Hmm, that might explain some troubles I'm having calibrating the Iris...

>  if you have multiple ABI accel/gyro/mag publishers, one should configure the consumers/subscribers to use only one specific IMU (filter on sender ID).

Well. I do just that in the Iris xml right?
`+    <subsystem name="ahrs" type="int_cmpl_quat" >
 +      <define name="AHRS_ICQ_IMU_ID" value="IMU_PX4_ID" />         
 +      <define name="AHRS_ICQ_MAG_ID" value="IMU_MPU6000_HMC_ID" /> 
 +    </subsystem>`

Anyway, to bad that does cover all the cases. I guess I'll  remove the mpu6000 from the PX4 imu for now then.

flixr:
The AHRS is not the only subscriber to IMU messages... e.g. the aligner and INS also subscribe to IMU

kevindehecker:
I think I handled all concerns.

conf/firmwares/subsystems/shared/imu_px4fmu_v2.4.makefile
```diff
@@ -0,0 +1,52 @@
+# Hey Emacs, this is a -*- makefile -*-
+#
+# PX4 Pixhawk IMUconsists of two (internal) IMU's and one internal magneto. Also an optional external magneto
+#
+# L3GD20 +  LSM303D
+
+include $(CFG_SHARED)/spi_master.makefile
+
+IMU_PX4FMU_CFLAGS += -DIMU_TYPE_H=\"imu/imu_px4fmu_v2.4.h\"
+IMU_CFLAGS  = -DIMU_TYPE_H=\"imu/imu_px4fmu_v2.4.h\"
```
flixr: preferably all these variables have the imu name in them, so like `IMU_PX4FMU_CFLAGS` above...
Or at least the currently unused `IMU_PX4FMU_CFLAGS` above...

sw/airborne/subsystems/imu/imu_px4fmu_v2.4.c
```diff
@@ -0,0 +1,106 @@
...
+void imu_impl_init(void)
+{
+
+  /* L3GD20 gyro init */
+  /* initialize gyro and set default options */
+  l3gd20_spi_init(&imu_px4.l3g, &IMU_SPI_DEV, IMU_L3G_SPI_SLAVE_IDX);
+
+  /* LSM303dlhc acc + magneto init */
+  lsm303dlhc_spi_init(&imu_px4.lsm_acc, &IMU_SPI_DEV, IMU_LSM_SPI_SLAVE_IDX, LSM_TARGET_ACC);
+#if !MODULE_HMC58XX_UPDATE_AHRS
```
flixr: IMHO there should be no usage of other module flags in here... e.g. what if you use a different external mag?

kevindehecker: I'm guessing the Pixhawk will be used fairly often together with the 3dr gps (with external magneto embedded). Of course, no guarantees...

What other way would you suggest to check if any other external magneto is used? I like to not have additional needed configuration when the external magneto module is removed.

flixr: We clearly need better/proper support for multiple IMUs and get rid of the global `imu` struct to really solve this problem...

While I can understand the desire to keep the amount of configuration low, I personally prefer something more generic so that you can also disable the px4 mag if you use a different external GPS.

One way to achieve that could be to add a define/flag `IMU_PX4_DISABLE_MAG` and check for that.
Then if it is not defined, but `MODULE_HMC58XX_UPDATE_AHRS` is defined to `TRUE`, you can set `IMU_PX4_DISABLE_MAG` to true as well...
That way it would still "automatically" work with the hmc module, but give the option to explicitly disable the px4 mag even if you use a different external one...

sw/airborne/modules/sensors/mag_hmc58xx.c
```diff
@@ -30,6 +30,8 @@
 #include "pprzlink/messages.h"
 #include "subsystems/datalink/downlink.h"
 
+#include BOARD_CONFIG
```
flixr: Why include the board config? It's not used...

sw/airborne/modules/sensors/mag_hmc58xx.c
```diff
@@ -39,6 +41,15 @@
 #ifndef HMC58XX_CHAN_Z
 #define HMC58XX_CHAN_Z 2
 #endif
+#ifndef HMC58XX_CHAN_X_SIGN
```
flixr: These new options should be documented accordingly in the module xml file...

sw/airborne/subsystems/imu/imu_px4fmu_v2.4.c
kevindehecker: Cool, ok done!

sw/airborne/modules/sensors/mag_hmc58xx.c
kevindehecker: Done + I also added the HMC58XX_CHAN_... to the doc section.

conf/modules/mag_hmc58xx.xml
```diff
@@ -9,6 +9,12 @@
     <configure name="MAG_HMC58XX_I2C_DEV" value="i2c1" description="I2C device to use (e.g. i2c1)"/>
     <define name="MODULE_HMC58XX_SYNC_SEND" value="TRUE|FALSE" description="Send IMU_RAW message with each new measurement (default: FALSE)"/>
     <define name="MODULE_HMC58XX_UPDATE_AHRS" value="TRUE|FALSE" description="Copy measurements to imu and send as ABI message (default: FALSE)"/>
+    <define name="HMC58XX_CHAN_X_SIGN" value="+|-" description="Reverse polarity of x axis (default: +)"/>
+    <define name="HMC58XX_CHAN_Y_SIGN" value="+|-" description="Reverse polarity of y axis (default: +)"/>
+    <define name="HMC58XX_CHAN_Z_SIGN" value="+|-" description="Reverse polarity of z axis (default: +)"/>
+    <define name="HMC58XX_CHAN_X" value="0|1|2" description="Channel id of x axis (default: 0)"/>
+    <define name="HMC58XX_CHAN_Y" value="0|1|2" description="Channel id of y axis (default: 0)"/>
```
flixr: default is `1` for y, `2` for z...

kevindehecker: Oops :)

Fixed

flixr:
Since #1577 is now merged, could you please rebase it on master again and fix the conflicts?

kevindehecker:
Done

sw/airborne/subsystems/imu/imu_px4fmu_v2.4.h
```diff
@@ -0,0 +1,59 @@
...
+#include "std.h"
+#include "generated/airframe.h"
+#include "subsystems/imu.h"
+
+#include "subsystems/imu/imu_px4_defaults.h"
+#include "peripherals/l3gd20_spi.h"
+#include "peripherals/lsm303dlhc_spi.h"
+
+#ifndef IMU_PX4_DISABLE_MAG
+#ifdef MODULE_HMC58XX_UPDATE_AHRS
```
flixr: You shouldn't check if it's defined, but if it's TRUE, so only `#if MODULE_HMC58XX_UPDATE_AHRS`

kevindehecker: You are very right ... tnx :)

kevindehecker:
Ow fyi, I have not tested the I2C lsm303dlhc driver, but I'm pretty sure it doesn't work perfectly. Maybe add a comment or something? I spend quite a lot of time on the I2C driver until I discovered it was never used (and probably non working) in the first place...

flixr:
I don't think the driver was never used, apparently it was working fine for the guy who made the PR...
But printing a warning if that code is compiled is a good idea...

flixr:
Actually I just realized that you used (and rewrote) the regs and driver for the LSM303DLHC, but you are actually using a LSM303D (those are not the same).
That's probably also why you think that the i2c driver was never working, since it was written for a slightly different sensor (which is also deprecated now).
Also there is https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/peripherals/lsm303d_regs.h which should have had the correct register defs and useful enums already...

kevindehecker:
Hmmmm :)

PR #1570, opened 2016-03-22
[ocaml] print more verbose info on xml loading errors

flixr:
Try to use our own `ExtXml.parse_file` instead of plain `Xml.parse_file` in most cases to get a better error report if the xml file is invalid.
e.g.
```
/home/flixr/code/paparazzi/conf/airframes/examples/bebop.xml: End of tag expected : 'load' line 43 characters 2-11
make: *** [ap.ac_h] Error 1
```
instead of only
```
Fatal error: exception Xml.Error(_)
make: * [ap.ac_h] Error 2
```


OpenUAS:
Very welcomed, I know a 100(student) reasons why ;) 
Ideas:
For users encountering issues, the "make: *** [ap.ac_h] Error 1" likely is not all that informative an IMHO can be chopped of the message. 

And we could add if XMLerror encountered start with something descriptive to make sure it is a user error that can be fixed.

"There is an Error in your XYZ.shortname.XML you can fix. Your problem to fix is details are:"

flixr:
I don't quite follow... is the message printed with this update not descriptive enough?

sw/ground_segment/cockpit/sectors.ml
```diff
@@ -32,7 +32,7 @@ let load = fun geomap () ->
       None -> ()
     | Some f ->
       try
-        let xml = Xml.parse_file f in
+        let xml = ExtXml.parse_file f in
```
gautierhattenberger: not sure this one should be changed, it breaks the `try ... catch` and the popup message

sw/ground_segment/cockpit/sectors.ml
```diff
@@ -44,7 +44,7 @@ let load_kml = fun geomap () ->
...
```
gautierhattenberger: same here

sw/tools/generators/gen_autopilot.ml
```diff
@@ -423,7 +423,7 @@ let write_settings = fun xml_file out_set ap ->
   *)
 let gen_autopilot main_freq xml_file h_dir out_set =
   try
-    let ap_xml = Xml.parse_file xml_file in
+    let ap_xml = ExtXml.parse_file xml_file in
```
gautierhattenberger: DTD errors were already caught here

sw/tools/generators/gen_mtk.ml
```diff
@@ -125,7 +125,7 @@ let _ =
   end;
   let xml_file = Sys.argv.(1) in
   try
-    let xml = Xml.parse_file xml_file in
+    let xml = ExtXml.parse_file xml_file in
```
gautierhattenberger: DTD errors already caught

sw/tools/generators/gen_ubx.ml
```diff
@@ -155,7 +155,7 @@ let _ =
   end;
   let xml_file = Sys.argv.(1) in
   try
-    let xml = Xml.parse_file xml_file in
+    let xml = ExtXml.parse_file xml_file in
```
gautierhattenberger: DTD errors already caught here

sw/tools/generators/gen_xsens.ml
```diff
@@ -224,7 +224,7 @@ let _ =
   end;
   let xml_file = Sys.argv.(1) in
   try
-    let xml = Xml.parse_file xml_file in
+    let xml = ExtXml.parse_file xml_file in
```
gautierhattenberger: DTD errors caught here

sw/tools/generators/gen_autopilot.ml
flixr: does it make sense to keep duplicating the error handling here instead of using ExtXml.parse_file?

flixr:
@gautierhattenberger thanks for the corrections, updated...

PR #1572, opened 2016-03-22
statefilter-in-a-module pull request

tcunis:
If someone is implementing a new, non-standard statefilter, e.g. in order to test a new filter approach or just for research, it would be both handy and favourable to do so within a module rather than directly in `subsystems/ins.` In this manner different approaches can be realized and tested in a modular way.

Thus, this allows to implement an individual state filter within a paparazzi module: If the define `INS_USE_MODULE` is set in the airframe file, the corresponding init, reset, propagation, and update functions need to be implemented in the module and will be called by the state filter subsystem.

Similar to `guidance_module.h,` the header file `ins_module_int.h` includes the module headers and describing the interface as well.

### DIFF
--- a/sw/airborne/subsystems/ins/ins_int.c
+++ b/sw/airborne/subsystems/ins/ins_int.c
@@ -57,9 +57,15 @@
 #include "generated/flight_plan.h"
 
 
+#ifndef INS_USE_MODULE
+#define INS_USE_MODULE FALSE
+#elif INS_USE_MODULE
+#include "ins_module_int.h"
+#endif
+
 #if USE_SONAR
-#if !USE_VFF_EXTENDED
-#error USE_SONAR needs USE_VFF_EXTENDED
+#if !USE_VFF_EXTENDED && !INS_USE_MODULE
+#error USE_SONAR needs either USE_VFF_EXTENDED or INS_USE_MODULE
 #endif
 
 /** default sonar to use in INS */
@@ -226,6 +232,10 @@ void ins_int_init(void)
   INT32_VECT3_ZERO(ins_int.ltp_speed);
   INT32_VECT3_ZERO(ins_int.ltp_accel);
 
+#if INS_USE_MODULE
+  ins_module_int_init();
+#endif
+
 #if PERIODIC_TELEMETRY
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_INS, send_ins);
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_INS_Z, send_ins_z);
@@ -247,6 +257,10 @@ void ins_reset_local_origin(void)
   }
 #else
   ins_int.ltp_initialized = FALSE;
+#endif
+
+#if INS_USE_MODULE
+  ins_module_int_reset_local_origin();
 #endif
 
 #if USE_HFF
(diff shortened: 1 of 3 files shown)


flixr:
Implementing INS algorithms in a module is a good idea!
We even plan on converting all (INS) subsystems to modules...

But IMHO there is no reason for those to use ins_int (which is an implementation).
It's already perfectly possible to implement a new INS subsystem without ins_int (e.g. ins_invariant).
So rather than adding even more stuff and ifdefs to ins_int, I would be in favour of simply making a completely new INS (as a proper module).

OpenUAS:
@tcunis  First of all, thank for your seriously good and useful extention and work on PPRZ Will discuss this one then one on one tomorrow if you have some time.

tcunis:
I do not fully agree: Yes, the solution with `#if` directives is definitely not optimal. However, if each new INS module is implemented from scratched, there is a lot of unnecessary code duplication -- since most state-filter implementation uses similar functions: _init_, _align_, _propagate_ on a acceleration signal, _update_ on GPS/baro/SONAR signal, etc. 
In order to re-use code, one would like to have an abstract super module to implement the default behaviour. Using `#if` is one solution; on the other hand, this could also be solved using _weak_ functions:

    void WEAK ins_module_int_propagate(struct Int32Vect3* meas_accel_ltp, float dt) {
      // ins_int default implementation
    }
    
    void ins_int_propagate(struct Int32Vect3* accel, float dt)
    {
      // pre-math (body-to-ned rotation etc.)
     
      // call 'abstract' function
      ins_module_int_propagate(&meas_accel_ltp, dt);
      
      // after-math
      ins_ned_to_state();
    }

Here, the weak function `ins_module_int_propagate` constitutes a kind of _statically virtual_ method in C; thus, also the different modes of ins_int (hff, vff, vff_extended, ...) can be realised as subsystem-types or modules respectively implementing the weak functions rather than using `#if`.

This proposition clearly requires a lot of re-structuring of the current INS. If already agree about the identifiers, the `#if`-solution I initially proposed can be an intermediate.

flixr:
I don't mind a sort of skeleton INS that already provides the callbacks from the ABI messages, etc...
However I don't think it's a good idea to do this in ins_int, IMHO it would be better to make a new skeleton/template INS.
Also a good design would be to separate the INS implementation functions from the ones that update the state interface, handle ABI and telemetry... (similar to AHRS with separate wrappers).

tcunis:
Well, I guess we will need similar structure both for float and integer...

A separation of implementation and back-end would/could follow from the proposed solution, wouldn't it?

flixr:
> Well, I guess we will need similar structure both for float and integer...

IMHO it's not really about float vs. integer, but rather about what the specific INS implementation estimates and what sensors it uses.
E.g. we have
- simple "passthrough" INS that basically just forward GPS data
- GPS for horizontal position/velocity + accel and baro for vertical
- using accel to propagate velocity->position in 3D
  - combined with updates from gps/baro/sonar
  - some estimate baro offset as well as accel bias
  - without GPS but visual odometry/opticflow
- full INS that estimate biases, attitude, velocity, position and hence also subscribe to gyro,mag, etc
- INS that get all of their data from an external sensor

> A separation of implementation and back-end would/could follow from the proposed solution, wouldn't it?

Sure it could... I just don't think that it makes sense to try to provide one skeleton/template/wrapper that fits all...
Also that precludes the possibility to run multiple different INS at the same time.

flixr:
@tcunis how about something like #1575

flixr:
Closing this as we have a cleaner solution with #1575

PR #1574, opened 2016-03-23
Edited image.c for YUV422 images.

Pax1601:
I've edited the image_draw_line function to take into account the fact that in yuv422 images pixels can only be edited in pairs in the x direction, so the increment in the x direction should be equal to 2. I hope not to be mistaken, but this way it is actually possible to draw line of the specified color.

### DIFF
--- a/sw/airborne/modules/computer_vision/lib/vision/image.c
+++ b/sw/airborne/modules/computer_vision/lib/vision/image.c
@@ -493,9 +493,9 @@ void image_draw_line(struct image_t *img, struct point_t *from, struct point_t *
      line.
   */
   int8_t incx, incy;
-  if (delta_x > 0) { incx = 1; }
+  if (delta_x > 0) { incx = pixel_width; }
   else if (delta_x == 0) { incx = 0; }
-  else { incx = -1; }
+  else { incx = -pixel_width; }
 
   if (delta_y > 0) { incy = 1; }
   else if (delta_y == 0) { incy = 0; }


OpenUAS:
@Pax1601 , sadly hope is not enough to honour a pull request ;) A retest and fix and after you've retested a pull request redux welcomed. TIA

Pax1601:
Actually I don't understand the error. I've made such a simple edit, plus it is build with no errors on my machine.

flixr:
The errors seems to come from an Ubuntu packaging problem:
```
Unpacking libstdc++6-armhf-cross (4.8.4-2ubuntu1~14.04.1cross0.11.1) ...
dpkg: error processing archive /var/cache/apt/archives/libstdc++6-armhf-cross_4.8.4-2ubuntu1~14.04.1cross0.11.1_all.deb (--unpack):
 trying to overwrite '/usr/share/gcc-4.8/python/libstdcxx/__init__.py', which is also in package libstdc++6:amd64 4.8.4-2ubuntu1~14.04
```
See also https://bugs.launchpad.net/ubuntu/+source/gcc-4.8-armhf-cross/+bug/1557205

flixr:
> I hope not to be mistaken, but this way it is actually possible to draw line of the specified color. 

@Pax1601 this sounds good to me :-)
@dewagter or @OpenUAS if you guys are happy with it, feel free to merge it...

fvantienen:
This seems fine for me, but maybe you should also add the second Y then to the coloring?
Because now only the first 3 will be set: UYV and not the second Y. This way you can generate colors.

Als you then should take into account if you start at a x position which is odd you should start with setting V so the ordering will be VYUY.

fvantienen:
While taking a better look it maybe seems that the implementation now becomes incorrect and the line isn't placed correctly. Because delta_x should be divided by the pixel_width now.

Pax1601:
Why do you think so? Which line are you referring to?

flixr:
@fvantienen what is the status here?

flixr:
@Pax1601 @OpenUAS @fvantienen So, do we merge this or not?

flixr:
@Pax1601 @fvantienen ping again....

fvantienen:
We can merge this ;)

PR #1575, opened 2016-03-23
[modules] add ins_skeleton

flixr:
INS module skeleton.
This module is meant to be used with an additional module which implements any of the following
  - ins_module_init()
  - ins_module_propagate(struct Int32Vect3 *accel, float dt)
  - ins_module_update_gps(struct GpsState *gps_s, float dt)
  - ins_module_update_baro(float pressure)
  - ins_module_reset_local_origin(void)

In each of those functions the updated estimate has to be written to the ins_module struct,
which is then used to update the state interface.

alternative proposal to #1572

### DIFF
--- a/conf/modules/ins_skeleton.xml
+++ b/conf/modules/ins_skeleton.xml
@@ -1,0 +1,32 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="ins_skeleton" dir="ins">
+  <doc>
+    <description>
+      INS module skeleton.
+      This module is meant to be used with an additional module which implements any of the following
+        - ins_module_init()
+        - ins_module_propagate(struct Int32Vect3 *accel, float dt)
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
This solution seems much easier to maintain than #1572 and with more capabilities at the end.

flixr:
@tcunis would this satisfy your needs?

flixr:
ping @tcunis

tcunis:
Hello Felix,

sorry, I have been busy with my thesis recently.

To be honest, from computer science / software architecture point of view I don't really like this solution; on the other hand and for now they are satisfying my needs and I can work with this.

flixr:
I don't quite understand why you don't like this (IMHO much cleaner and more maintainable) solution... Also this way you can more easily run multiple INS at the same time (we are not quite there yet).

PR #1577, opened 2016-03-24
replace bool_t with C99 bool from stdbool.h

flixr:
Commonly people use `bool` from `stdbool.h` and this is usually the recommended way.
This requires C99, but we already use that anyway...

This should make integration with some other things easier, e.g.
- chibios, which has it's own `bool_t` typedef
- building python wrappers that map it to a python bool


gautierhattenberger:
`sed` is your friend

I think it is fine for merging.

flixr:
@dewagter @martinmm @fvantienen @podhrmic @esden 
Do you see any reasons not to merge this and keep the `typedef uint8_t bool_t`?

gautierhattenberger:
Guys, does anyone have comments ? I have some stuff I'd like to rebase on top of this if it is integrated, it would save some conflict merging issues...

martinmm:
Don't see any issue with changing it

OpenUAS:
Really would like to see this indeed

PR #1578, opened 2016-03-27
Added pthread cancelling when calling opticflow_module_stop()

npotdar:
Added pthread cancelling of opticflow thread and prints error on failure.


flixr:
Thx!

PR #1580, opened 2016-03-29
[gps_i2c] gps_i2c module rewritten, added ubx ucenter compatibility

masierra:
unbroke gps_i2c module and added ucenter compatibility

```
<subsystem name="gps"           type="ublox">
  <configure name="UBX_GPS_PORT" value="gps_i2c"/>
</subsystem>
```


flixr:
Nice! Will be great to finally have a functioning ublox i2c driver again!

I wonder though if there isn't a better way to set up and use a I2C link device...

sw/airborne/modules/gps/gps_i2c.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2009  ENAC, Pascal Brisset, Michel Gorraz,Gautier Hattenberger
+ * Copyright (C) 2016 Michael Sierra <sierramichael.a@gmail.com>
```
gautierhattenberger: It is usually better to keep the previous authors, even when the original code sucks...

gautierhattenberger:
The way UBX gps are using i2c is really stupid. They more or less do stream over i2c.
I think the complete gps_ubx driver should be reorganized to have the generic part independent of the uart/i2c part.
Also their is a lot of `ifdef` in ucenter module. Not sure if it can be improved.

I would say that have this thing working is already a good thing and we could merge it as his.
Future improvements should cover:
- reorganizing gps_ubx driver
- moving gps_i2c files to `modules/gps` folder (or we do that now)
- implementing the generic pprzlink_device for i2c in the actual i2c driver

sw/airborne/modules/gps/gps_i2c.c
masierra: fixed

flixr:
Also since this module works only for Ublox I2C GPSs, we should probably rename the module to `gps_ublox_i2c` and make it depend on `gps_ublox`.

Even better would if we could add the relevant source files in the normal gps_ublox module if i2c is given as port... that part would actually already be possible, but adding the extra init/periodic/event functions not... so maybe an improvement for later...

flixr:
@masierra Since subsystem will go away (be replaced by modules), you should replace `subsystem` by `module`, in this case:
```
<module name="gps"           type="ublox">
  <configure name="UBX_GPS_PORT" value="gps_i2c"/>
</module>
```

flixr:
Can you please also add doxygen blocks to the files that are missing them?
See http://docs.paparazziuav.org/latest/styledoxygen.html

Issue #1576, opened 2016-03-23
module target inversion with ! doesn't work for all targets


flixr:
Prefixing module targets with `!` is supposed to invert the list.
However that doesn't seem to work properly, e.g.
- `<makefile target="!nps">` is supposed to add it to all targets except for `nps`, however the sources are not added for e.g. `ap` target (but it's added in `generated/modules.h`)
- `<makefile target="ap|!nps">` however adds it for `ap` target

PR #1583, opened 2016-03-30
[generators] fix targets testing and inversion for modules

gautierhattenberger:
see #1576 

This could be made much better, but at least it as the desired behavior


flixr:
:+1: Seems to work as expected here...

PR #1584, opened 2016-03-30
bounding of the spektrum input

EwoudSmeur:
My RC was outputting slightly negative throttle values. This resulted in an overflow of the throttle curve. The radio values are expected to always be within pprz bounds, so there should be a bound on the radio values.


OpenUAS:
Nice find :+1:

PR #1587, opened 2016-03-31
PX4 motor random spin up issues

kevindehecker:
Turned out that having 1500000 baud between the f1 and f4 on the Pixhawk board was rather heisenbuggy. After some time (ca half an hour stationary, but with power from battery) packages started getting lost. 

To remain backwards compatible with PX4 firmware, the 150000 baud is still required. However, during flight not. So the baud rate is set to 230400 after 10 seconds.

This PR
-fixes the above baud issue, 
-adds some signalling information to the FBW LED,
-fixes some issues related to FBW_MANUAL_MODE , which caused the motors to spin up shortly during intermcu loss.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/main_fbw.c
+++ b/sw/airborne/firmwares/rotorcraft/main_fbw.c
@@ -37,18 +37,16 @@
 #include "subsystems/actuators/motor_mixing.h"
 #endif
 
-
 #include "subsystems/electrical.h"
-
 #include "subsystems/radio_control.h"
-
 #include "subsystems/intermcu/intermcu_fbw.h"
-
 #include "firmwares/rotorcraft/main_fbw.h"
 #include "firmwares/rotorcraft/autopilot_rc_helpers.h"
 
 #define MODULES_C
 #include "generated/modules.h"
+
+
 
 /** Fly by wire modes */
 typedef enum {FBW_MODE_MANUAL = 0, FBW_MODE_AUTO = 1, FBW_MODE_FAILSAFE = 2} fbw_mode_enum;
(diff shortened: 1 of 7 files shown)


sw/airborne/firmwares/rotorcraft/main_fbw.c
```diff
@@ -178,18 +176,29 @@ STATIC_INLINE void main_periodic(void)
...
   }
 
+#ifdef BOARD_PX4IO
+  //due to a baud rate issue on PX4, for a few seconds the baud is 1500000 however this may result in package loss, causing the motors to spin at random
+  //to prevent this situation:
+  if (inter_mcu.stable_px4_baud != PPRZ_BAUD) {
+    fbw_mode = FBW_MODE_FAILSAFE;
+    autopilot_motors_on = false;
+    //signal to user whether fbw can be flashed:
+    LED_OFF(FBW_MODE_LED); // causes really fast blinking
```
flixr: please protect this with `#ifdef FBW_MODE_LED`

sw/airborne/subsystems/intermcu/intermcu_fbw.c
```diff
@@ -119,6 +119,19 @@ void intermcu_send_status(uint8_t mode)
   //FIXME
 }
 
+
+void intermcu_blink_fbw_led(uint16_t dv)
+{
+  static uint16_t idv = 0;
+  if (!autopilot_motors_on) {
```
flixr: Please protect this with `#ifdef FBW_MODE_LED`

flixr:
Do I see that correctly that this is currently not usable for the fixedwing firmware?

kevindehecker:
I made and tested it for rotorcraft (for which the manual mode mostly useless, unless an additional IMU is added. This will happen in the near future for some of our airframes). That's why the changes are only in rotorcraft/main_fbw right?

flixr:
We can merge this, just wondering if you also plan to add the stuff required for px4 flashing to the fixedwing intermcu? Since the pixhawk will probably also be used for fixedwings...

kevindehecker:
Ah, I see your point. I guess I can add the `#ifdef BOARD_PX4IO` part from rotorcraft/main_fbw.c to fixedwing, but I wouldn't be able to test it and frankly the fixedwing intermcu code looks quite different (outdated?). I 'm not very enthusiastic about that idea yet. Shall I add a FIXME in the code fixedwing main for now?

To be clear, the flashing part should work just fine. The only problem is that it may switch to FBW_MODE_MANUAL during this first 10s of 150000 baud after powerup. In case of rotorcraft this is a big issue, the motors spin up to minimum rotating thrust. This is I think mostly not the case for fixed wings.

agressiva:
Kevin, i can test the FW firmware.

btw, i tried compile to FW without success.

kevindehecker:
Thx agressiva, but I think there is a long way to go before flight tests :|

It seems intermcu status is not checked at all in fixed wing... There is probably more worked involved offtopic for this PR. I say merge this now.

PR #1589, opened 2016-03-31
[pprzlink] update for new API of pprzlink using fd parameter

gautierhattenberger:
See https://github.com/paparazzi/pprzlink/pull/5

This update in the pprzlink API allow to pass a parameter (long int) to the transport and device function. It can be an address, a file descriptor, an index,... This will allow a more efficient sending of messages especially when using transaction queues or malloc and for the ongoing chibios implementation.

### DIFF
--- a/sw/airborne/subsystems/actuators/actuators_spektrum.c
+++ b/sw/airborne/subsystems/actuators/actuators_spektrum.c
@@ -83,12 +83,12 @@ void actuators_spektrum_set(void)
 static inline void actuators_spektrum_send(struct link_device *dev)
 {
   uint8_t i = 0;
-  dev->put_byte(dev->periph, 0x00); // number missed frames
-  dev->put_byte(dev->periph, 0x12); // 7 channels, 11 bit, 11ms
+  dev->put_byte(dev->periph, 0, 0x00); // number missed frames
+  dev->put_byte(dev->periph, 0x12); // 7 channels, 11 bit, 0, 11ms
 
   /* Transmit all channels */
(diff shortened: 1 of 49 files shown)


kevindehecker:
I think another argument is needed here:
`dev->put_byte(dev->periph, 0x12); // 7 channels, 11 bit, 0, 11ms`

sw/airborne/subsystems/actuators/actuators_spektrum.c
```diff
@@ -83,12 +83,12 @@ void actuators_spektrum_set(void)
...
+  dev->put_byte(dev->periph, 0x12); // 7 channels, 11 bit, 0, 11ms
 
   /* Transmit all channels */
   for (i = 0; i < ACTUATORS_SPEKTRUM_MAX_NB; i++) {
-    dev->put_byte(dev->periph, i << 3 | actuators_spektrum.cmds[i] >> 8);
-    dev->put_byte(dev->periph, actuators_spektrum.cmds[i] & 0xFF);
+    dev->put_byte(dev->periph, 0, i << 3 | actuators_spektrum.cmds[i] >> 8);
+    dev->put_byte(dev->periph, 0, actuators_spektrum.cmds[i] & 0xFF);
   }
 }
```
kevindehecker: I think another argument is needed here:
`dev->put_byte(dev->periph, 0x12); // 7 channels, 11 bit, 0, 11ms`

flixr: thx, looks like a sed command didn't expect a comment with `,` after put_byte ;-)

PR #1592, opened 2016-04-01
Added gimbal control for PX4

kevindehecker:
Module that allows the RC to control the camera pitch angle of the (Tarrot) gimbal through the intermcu Pixhawk board.

Not completely trivial since the gimbal is connected to the pwm of the ap. Which is why I put this in a module instead of just add another actuator in the airframe xml.

### DIFF
--- a/sw/airborne/modules/px4_gimbal/px4_gimbal.c
+++ b/sw/airborne/modules/px4_gimbal/px4_gimbal.c
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) Kevin van Hecke
+ *
+ * This file is part of paparazzi
+ *
+ */
+/**
+ * @file "modules/px4_gimbal/px4_gimbal.c"
+ * @author Kevin van Hecke
+ * Control gimbal camera axis through px4 from rc
+ */
+
+#include "modules/px4_gimbal/px4_gimbal.h"
+#include "subsystems/radio_control.h"
+
+#include "generated/airframe.h" // AC_ID is required
+#include "subsystems/actuators.h"
+
+#ifndef PX4_GIMBAL_PWM_CHAN
+#define PX4_GIMBAL_PWM_CHAN 0
+#endif
+#ifndef PX4_GIMBAL_RC_CHAN
+#define PX4_GIMBAL_RC_CHAN RADIO_AUX2
+#endif
+
+#ifndef PX4_GIMBAL_PWM_MIN
+#define PX4_GIMBAL_PWM_MIN 900
+#endif
+
+#ifndef PX4_GIMBAL_PWM_MAX
+#define PX4_GIMBAL_PWM_MAX 1500
+#endif
+
+
+void px4_gimbal_init() {
+#ifdef INTER_MCU_AP
+   actuators_init();
+#endif
+}
+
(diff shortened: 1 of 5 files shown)


sw/airborne/modules/px4_gimbal/px4_gimbal.c
```diff
@@ -0,0 +1,69 @@
...
+   actuators_init();
+#endif
+}
+
+ void px4_set_gimbal_angle_event() {
+   //reroute gimbal rc input from fbw to pwm output on ap
+
+   float value =radio_control.values[PX4_GIMBAL_RC_CHAN];
+
+   float min = 900;
```
flixr: I guess you want to use `PX4_GIMBAL_PWM_MIN` here... also doesn't seem to really be a float...

sw/airborne/modules/px4_gimbal/px4_gimbal.c
```diff
@@ -0,0 +1,69 @@
...
+   float max = 1500;
+   float range = max - min;
+
+   value = ((9600-value) / 9600.0) * range + min;
```
flixr: Better use MAX_PPRZ for 9600

sw/airborne/modules/px4_gimbal/px4_gimbal.c
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @file "modules/px4_gimbal/px4_gimbal.c"
+ * @author Kevin van Hecke
+ * Control gimbal camera axis through px4 from rc
+ */
+
+#include "modules/px4_gimbal/px4_gimbal.h"
+#include "subsystems/radio_control.h"
+
+#include "subsystems/datalink/downlink.h"
```
flixr: the datalink and airframe includes don't seem to be needed...

sw/airborne/modules/px4_gimbal/px4_gimbal.c
```diff
@@ -0,0 +1,69 @@
...
+#include "generated/airframe.h" // AC_ID is required
+#include "subsystems/datalink/datalink.h"
+#include "subsystems/datalink/telemetry.h"
+#include "subsystems/actuators.h"
+
+#include BOARD_CONFIG
```
flixr: IMHO better to only include generated/airframe.h for the actuators (not AC_ID)

sw/airborne/modules/px4_gimbal/px4_gimbal.c
```diff
@@ -0,0 +1,69 @@
...
+#endif
+
+
+void px4_gimbal_init() {
+#ifdef INTER_MCU_AP
+   actuators_init();
+#endif
+}
+
+ void px4_set_gimbal_angle_event() {
```
flixr: probably also better to do this in a periodic instead of as fast as possible in an event loop....

kevindehecker:
Any more problems merging this?

flixr:
You still have the includes for BOARD_CONFIG and DEFAULT_ACTUATORS, which shouldn't be needed... if they are needed, why?

kevindehecker:
Ah, yes. I had my reasons, but not important for the PR. They are gone now.

PR #1593, opened 2016-04-02
[pprzlink] update to C API with file descriptor param

flixr:
This was already once proposed and mergd with #1589 (and reverted again) since with these changes the simulators don't work anymore.
So this needs some more work and checking before it can be integrated again.

### DIFF
--- a/sw/airborne/arch/linux/mcu_periph/uart_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/uart_arch.c
@@ -240,7 +240,7 @@ void uart_periph_set_baudrate(struct uart_periph *periph, uint32_t baud)
   }
 }
 
-void uart_put_byte(struct uart_periph *periph, uint8_t data)
+void uart_put_byte(struct uart_periph *periph, long fd __attribute__((unused)), uint8_t data)
 {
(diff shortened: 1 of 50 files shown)


flixr:
Ok, I think I found and fixed the remaining bugs, OCaml sim and NPS are working again.
But please double check...

gautierhattenberger:
Tested and working with:
- ocaml sim
- nps rotorcraft
- nps fixedwing
- elle0 with pprz_transport
- ardrone2 (udp, pprz)
- apogee with xbee_transport (with bug fix)
- SD logging on apogee (with bug fix)

flixr:
Thx.
So I guess still untested are the following datalinks:
- bluegiga
- w5100
- spektrum

@kirkscheper could you test bluegiga?
@gtoonstra could you test w5100?

gautierhattenberger:
@flixr since all transports have have been tested (and fixed) and most common devices as well, is it possible to merge this again ?
It seems that Delft team is really busy with outback challenge so they probably won't have time to test bluegiga and superbitrf (@esden ?).
Also modifications on devices where lighter than transports and I checked them again.
It is not like it is urgent for paparazzi/master, but it will make pprzlink pull requests and update easier since `fd` parameter is already in master branch there (okay this was too early as well).

PR #1597, opened 2016-04-07
[energy ctrl] fix accel conversion for correct vdot

flixr:
This is a regression and seems to be wrong since 4bfab9d05000e91c6a6a5a8d45f55e5ea6269543


TobiasMue:
merge it ;-)

flixr:
@TobiasMue Were you able to test this and it was working as expected again?

TobiasMue:
not yet- but it can't make thinks worse, since it is not working at all! I ll try to test on Saturday

flixr:
Ok, please keep us posted here...

flixr:
Also backported fix to v5.8 with 5296cb902eee289546f8a52344e0eda01786ac6d

TobiasMue:
I tested it in flight and every things seems to work as before, nothing unusual so far :-)

PR #1598, opened 2016-04-12
[pprzlink][python] update pprzlink imports

flixr:
serial/ivy_msg_interface are now in pprzlink module

this means anybody using ivy_msg_interface (or serial_msg_interface) needs to change
`from ivy_msg_interface import IvyMessagesInterface`
to
~~`from pprzlink.ivy_msg_interface import IvyMessagesInterface`~~
`from pprzlink.ivy import IvyMessagesInterface`

### DIFF
--- a/sw/ground_segment/python/dashboard/radiowatchframe.py
+++ b/sw/ground_segment/python/dashboard/radiowatchframe.py
@@ -10,7 +10,7 @@ import pygame.mixer
 
 sys.path.append(os.getenv("PAPARAZZI_HOME") + "/sw/ext/pprzlink/lib/v1.0/python")
 
-from ivy_msg_interface import IvyMessagesInterface
+from pprzlink.ivy import IvyMessagesInterface
 
 WIDTH = 150
 HEIGHT = 40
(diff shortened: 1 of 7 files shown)


flixr:
While this already makes things more consistent and importing pprzlink easier, maybe we should rename the files/modules to get something shorter like`from pprzlink.ivy import IvyMessagesInterface` in the end instead of `from pprzlink.ivy_msg_interface import IvyMessagesInterface`?

hooperfly:
@flixr as long as the naming is descriptive and consistent, I like the idea of being more concise. To follow your logic, the serial interface would use `from pprzlink.serial import SerialMessagesInterface`?

flixr:
yes, that would be the idea

hooperfly:
+1

gautierhattenberger:
+1

flixr:
Hm... one should be careful to not do something like `from pprzlink import *` then though, as it would shadow the normal `ivy` and `serial` imports...

@fvantienen @rmeertens @kevindehecker @kirkscheper @bartslinger what do you guys think?

hooperfly:
When this is resolved/merged, I will update the PPRZLINK README.md to reflect the interface updates, Of course, feel free to do so if you have the time. 😉

fvantienen:
For me this looks good ;)

fvantienen:
Can we merge this so that we are back to the pprzlink master?
Because I want to merge a lot of other stuff with the latest master and also an update of pprzlink is needed for that.

PR #1600, opened 2016-04-15
changed default gcs layout

EwoudSmeur:
Working with the GCS has always been a bit painful for me, particularly because of 1 reason: the layout. The settings panel is always too small, and the map is always too wide (unless you really like flying east-west). Dragging the panels around is annoying and time consuming.

But I learned something recently, apparently there are other layouts! The large_left_col has the same panels, but then with a more square map and a wide settings panel. I think this is great for everyone, so I suggest making it the default


EwoudSmeur:
As I mentioned, I think this is something everyone should benefit from, because of the reasons I mention above. If there are reasons why the horizontal layout is preferred by a majority of paparazzi users, I will happily change it in my own control panel.

flixr:
I have no strong feelings about the default...

podhrmic:
FYI We are using this layout:
screenshot from 2016-04-18 12 29 45

My guess is that everyone likes something different so no strong opinion about the default option either.

gautierhattenberger:
I'm also using a custom version of the default layout, and different from the one above. One remark against changing it is that the documentation and wiki examples are build upon the historical layout.

dewagter:
Is there a way to set the layout when starting replay/sim. The lines below
do not work for me:

    <program name="GCS" command="sw/ground_segment/cockpit/gcs">
      <arg flag="-layout" constant="bottom_settings.xml" />
    </program>


-Christophe

EwoudSmeur:
I am thinking about new users here. Of course you can modify and fiddle with all the settings, but I think you can not expect a new user to do this when he is starting out.

So I think we have the possibility to make something that already works nice out of the box, and can still be modified to your own needs. That is why I care about the defaults. 

Indeed I can change the settings for myself, and I actually just did. So for myself, I don't care if you close this pull request. But instead of saying: why don't you just do what you want and I do what I want; maybe we should be saying, what do other (less experienced) people want?
Cheers

flixr:
I actually also prefer layouts with strip on the side and settings at the bottom (especially since scrolling in settings is annoying).
In any case I'm fine with either default (pro keeping the one we have is IMHO mostly(only?) the already existing docs/images).

EwoudSmeur:
So far I only heard people say something like 'why change the default, by the way I am using something other than the default'. I think this is a good reason to change it.

Regarding the wiki issue, Since the same panels are there, just at a different location, I think there will be very little confusion...

gautierhattenberger:
The point was: we all use different defaults...
There is probably a way to store an alternate default for the GCS (in the %gconf.xml file that stores latest user parameters)

EwoudSmeur:
No, the point is: none of you is using the default, you all use custom settings.

I am not sure how I can make my point clearer... This is not about me, I know how to change it. It is about other users then me, the ones that will use the default when they start out. 

And if you use something else, you are NOT using the default.

gautierhattenberger:
I perfectly understand what your saying, but it is hard to find the "best" default since everyone is using a different custom layout. So changing it will not change anything for you, for me or most of the other dev members.
So, if their is something to improve, I would say it is make it easier for new users to change the layout and eventually set one by default, while keeping the hard-coded fallback because we need one at some point.

kevindehecker:
1. I would not say the goal here is to find the best layout... but just a better one. In principle, having a squared map is better. Also, the settings are very often used, and currently by default you need to scroll always making it quite painfull. So, here the proposed layout is also better. (meaning in general, etc etc)
2. The argument of not improving something because ...the documentation may need some work, is quite pitiful.
3. We are having a lot of new users (students), and we are very often helping them with problems. They all use the unhandy default. Eventually this makes our lifes difficult as well, even though *we* have changed the default layout ages ago...

For these reasons, but mostly 3), I am in favor of improving the default layout to something better.

flixr:
@dewagter adding the layout in control_panel.xml as you wrote works fine for me...

Also I'm fine with changing the default, but ideally we would make it easier to save the preferred layout in %gconf.xml like @gautierhattenberger already said.

flixr:
@gautierhattenberger do you want to implement something to make changing/remembering the layout easier?
Or do we use #1671?
We could also just change it in the control_panel_example.xml file, so that new users will get the layout argument appended instead of changing the hardcoded default...

Either way is fine with me, but I like this layout better (although I normally use a modified version of left_col) than the current default.

gautierhattenberger:
I actually like the solution of #1671, it both allow to remember your setting (even after `make clean`) and can be shared with others when providing a control panel file (like giving pre-configured control panel to student).
The only thing I would like to have someday is the possibility to dynamically change the layout, be that would be for an other day...

flixr:
So do we change the hardcoded default or only set `<arg flag="-layout"  constant="large_left_col.xml" />` in control_panel_example.xml?

EwoudSmeur:
I am very much fine with that. Though then we might as well change the default too

flixr:
@gautierhattenberger @martinmm what do you prefer?

1. change hardcoded default
2.  add default layout arg in control_panel_example.xml
3.  both

flixr:
@gautierhattenberger So, should we also change the default that is hardcoded in gcs?
Or do we close this?
Fine with me either way...

gautierhattenberger:
After think again about this, I think I mostly liked the current because I'm used to it and it was well balanced when it was made (with less buttons and settings). Maybe it is not the case anymore.
So let's put an end to this, because it really doesn't matter after all.

PR #1601, opened 2016-04-15
remove scaling from rate feedback

EwoudSmeur:
Currently gains from rate control do not carry over to attitude stabilization, because there is a factor 2 in the rate control and prescaler values (factor 3 for rate) in the attitude control. I searched through the airframes, and there are not many that use rate control. I doubled all the gains and removed rate control for bebop, as rate control over datalink is very difficult due to lag.

The prescaler values should also be default 1 for stabilization quat_int, as they already are for float_quat. However, this will mean updating a large amount of airframes, which will require some scripting. I might have time for this later.

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_course2016_bebop_avoider.xml
+++ b/conf/airframes/TUDELFT/tudelft_course2016_bebop_avoider.xml
@@ -137,25 +137,6 @@
     <define name="DEADBAND_R" value="50"/>
   </section>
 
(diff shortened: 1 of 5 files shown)


flixr:
As already mentioned in #1482, I think it would make more sense to convert the really simple rate control to float, or do you see any advantage of keeping fixed point here?

EwoudSmeur:
I agree that float would be simpler. Though that can well be a separate pull request. Can we first merge this, before new airframe files with rate gains are added?

PR #1603, opened 2016-04-17
[ground] Fix FBW battery level

fvantienen:
The GCS now combines the FBW battery with the AP battery which is confusing.
This fixes the problem and doesn't show the FBW battery by default.


flixr:
Looks good to me...

fvantienen:
@gautierhattenberger if you're ok with this can you merge this or comment?

Issue #1475, opened 2015-12-12
convert subsystems to modules


flixr:
The goal is to use the same xml description format we have for modules for subsystems (or rather convert subsystems to modules).
This is sort of a meta issue to track progress of the various sub issues...

This will yield the following advantages:
- Better documentation
- auto settings (#23)
- possibility to use periodic and event functions that are automatically called
- hook into datalink parsing
- make dependencies between subsystems more clear/explicit

With #1471 we are on track to do that, still TODO:
- [x] possibility to have firmware specific parts (extra defines, etc), e.g. add firmware attribute to makefile section of module (done with #1534)
- [x] improve dependency system as per #940

actually convert the subsystem to modules:
- [x] gps #1625
- [x] ins #1740
- [x] imu #1788
- [x] ahrs #1859
- [x] telemetry #1835
- [x] radio_control
- [x] actuators
- [x] stabilization/control
- [x] traffic_info #1638
- [x] rest

Related issues to also keep in mind:
- #656
- #246
##

flixr:
Elaboration on periodic/event function usage:
Quite a few subsystems have periodic and event functions that are called from main and have/need some sort of #ifdef protection: imu, ins, gps...
But not all implementations actually need these periodic and event functions.
Converting those to modules let's you easily add those only for the cases where needed and cleans up main.

gautierhattenberger:
I consider this solved, except for moving the code from susbsystems folder to modules folder

PR #1604, opened 2016-04-18
replace subsystem by module in all airframes

flixr:
Towards making it easier to actually convert more subsystems to module...


gautierhattenberger:
any idea to merge `modules` section into the correct `firmware` section (except by hand) ?

podhrmic:
This might sound like a stupid question, but why do we want to change everything to modules? It kind of make sense to have some subsystems that do certain things, and modules that are sort of "smaller subsystems"? Improving subsystems etc yes, but everything as a module seems to be confusing? Just making sure we are on the same page...

gautierhattenberger:
For now, modules are only used as "super" makefile with embedded documentation and settings. The internal static scheduling is the same as before (and will probably stay like this for some time).
So the produced code is the same as before.

Now, regarding the explicit distinction of subsystem and modules in the airframe, I don't think it is really helping new users:
- it is not obvious why you have to different ways to configure things
- some elements like INS, IMU or GPS exist as modules and subsystems
- configuration files are in many different places (modules, firmware shared/FW/RC)

Also, we are planning to implement a dependency mechanism that allow to automatically load required modules, detect some inconsistent configurations, allow shorter firmware section with meta-modules, load modules in correct order and so on. All this things are much easier when you work with the same type of element.
Again, the internal code will stay the same as long as the modules description doesn't allow to produce a code with a precise scheduling if needed (like current subsystems, in some cases at least).

flixr:
@podhrmic see https://github.com/paparazzi/paparazzi/issues/1475 for the main issue tracking/explaining subsystem->module conversion

flixr:
@gautierhattenberger I don't have a good idea on how to merge the modules section automatically...

flixr:
From my side this is good to merge...

gautierhattenberger:
I guess everyone will need to update their files manually for the last part.
Let's merge this then.

PR #1605, opened 2016-04-18
Updates for Elle0 v1_2 airframes

hooperfly:
New airframes for Elle0 v1_2. The `b` version currently does not simulate well after updating the calibration settings. It's real world performance is rock solid.


hooperfly:
Went back and ran the simulations against the latest pull and both frame `A` and frame `B` now seem to behave identically. Frame `B` no longer diverges as each subsequent waypoint was visited.

PR #1606, opened 2016-04-19
[heli][flight_plans] Heli Spinup routine and flight plan mode manual

fvantienen:
- Functionality for mode "manual" in flight plan without stabilization
- Navigation routines for spinning up rotor for helicopters in flight plan
- Added example flightplan

### DIFF
--- a/conf/modules/nav_heli_spinup.xml
+++ b/conf/modules/nav_heli_spinup.xml
@@ -1,0 +1,19 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="nav_heli_spinup" dir="nav">
+  <doc>
+    <description>Helicopter Rotor spinup
+      This module controls the spinup of the main rotor from a Helicopter.
+      - First make sure you are in the correct throttle curve
+      - Next call nav_heli_spinup_setup with a duration and end throttle level
+      - Then call nav_heli_spinup_run
+    </description>
+  </doc>
+  <header>
+    <file name="nav_heli_spinup.h"/>
+  </header>
+  <makefile>
+    <file name="nav_heli_spinup.c"/>
+  </makefile>
+</module>
+
--- a/sw/airborne/firmwares/rotorcraft/navigation.c
+++ b/sw/airborne/firmwares/rotorcraft/navigation.c
@@ -479,6 +479,20 @@ void nav_home(void)
   nav_run();
 }
 
+/** Set manual roll, pitch and yaw without stabilization
+ *
+ * @param[in] roll The angle in radians (float)
+ * @param[in] pitch The angle in radians (float)
+ * @param[in] yaw The angle in radians (float)
+ */
+void nav_set_manual(float roll, float pitch, float yaw)
+{
+  horizontal_mode = HORIZONTAL_MODE_MANUAL;
+  nav_roll = ANGLE_BFP_OF_REAL(roll);
+  nav_pitch = ANGLE_BFP_OF_REAL(pitch);
+  nav_heading = ANGLE_BFP_OF_REAL(yaw);
+}
+
 /** Returns squared horizontal distance to given point */
 float get_dist2_to_point(struct EnuCoor_i *p)
 {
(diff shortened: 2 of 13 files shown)


flixr:
Personally I'm not in favour of adding ever more functions that always return false only to be able to `call` them from the flight plan without `loop="false"`.
See also old discussion about this: https://github.com/paparazzi/paparazzi/pull/830

fvantienen:
Aha yes I forgot that one already.. will change that small thing ;)

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -480,6 +481,15 @@ void nav_home(void)
   nav_run();
 }
 
+/** Set manual roll, pitch and yaw without stabilization */
```
flixr: please add what unit the arguments are supposed to be in

sw/airborne/modules/nav/nav_heli_spinup.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include "modules/nav/nav_heli_spinup.h"
+#include "navigation.h"
+#include "paparazzi.h"
+
+static struct nav_heli_spinup_t nav_heli_spinup;
+
+bool nav_heli_spinup_setup(uint16_t duration, float throttle)
+{
+  nav_heli_spinup.duration = duration;
```
flixr: check that duration is not zero!

flixr:
Please try to not put different features/additions in the same commit...

flixr:
In flight plan: why allow different vmodes in manual instead of requiring throttle?

flixr:
why add a FMODE command to `stabilization_cmd`?

fvantienen:
- I didn't find the time to split up the commits, feel free to split it up into 2 different PR's
- The `vmode` is like this, because this was the easiest way to implement it and it works exactly the same as `attitude`.
- I added the `FMODE` command because eventually either the stabilization or the outer loop should set the correct throttle curve.

sw/airborne/firmwares/rotorcraft/navigation.h
```diff
@@ -48,10 +48,11 @@ extern uint8_t horizontal_mode;
 extern struct EnuCoor_i nav_segment_start, nav_segment_end;
 extern struct EnuCoor_i nav_circle_center;
 extern int32_t nav_circle_radius, nav_circle_qdr, nav_circle_radians;
-#define HORIZONTAL_MODE_WAYPOINT  0
```
flixr: this means changing the meaning of the already existing modes, probably better to append the manual mode at the end.
Also needs to be updated in `server_global.ml` accordingly

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -479,6 +479,20 @@ void nav_home(void)
   nav_run();
 }
 
+/** Set manual roll, pitch and yaw without stabilization
+ *
+ * @param[in] roll The amount of roll between -1.0 and 1.0
```
flixr: really? I don't think so...
the nav_roll/pitch/heading should be in radians in fixed point

fvantienen:
For now I changes it to `commands` instead of `stabilized_cmd`, and if I find a nicer solution later we will make a separate PR.

conf/modules/nav_heli_spinup.xml
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="nav_heli_spinup" dir="nav">
+  <doc>
+    <description>This module controls the spinup of the main rotor from a Helicopter</description>
```
flixr: Please try to remember that the first line of the description is taken as name (very brief descrption) in the generated http://docs.paparazziuav.org/latest/onboard_modules.html

Maybe we should really introduce a `<brief>Foo module</brief>` or so?

flixr:
Looks good to me... @gautierhattenberger if the manual mode is fine with you as well, feel free to merge.

fvantienen:
@gautierhattenberger any feedback?

gautierhattenberger:
It seems that if someone tries to use this with a fixed-wing it will fail without proper error message.
I suggest to at least hide the function `nav_set_manual` in a macro like `NavSetManual` that triggers a `#error` message at compilation for FW firmware.

fvantienen:
Is this ok @gautierhattenberger ?

PR #1607, opened 2016-04-19
[modules] Update ADC temperature to be configurable

fvantienen:
Make temperature module through ADC more configurable.

### DIFF
--- a/sw/airborne/modules/sensors/temp_adc.c
+++ b/sw/airborne/modules/sensors/temp_adc.c
@@ -31,7 +31,8 @@
 #include "subsystems/datalink/downlink.h"
 #include BOARD_CONFIG
 
-float temp_c1, temp_c2, temp_c3;
+bool temp_adc_sync_send = false;
+static float temp_c1, temp_c2, temp_c3;
 
 #if PERIODIC_TELEMETRY
 #include "subsystems/datalink/telemetry.h"
@@ -48,26 +49,28 @@ float temp_c1, temp_c2, temp_c3;
 #endif
 #endif
 
+#ifdef TEMP_ADC_CHANNEL1
+static struct adc_buf temp_buf1;
+
 #ifndef TEMP_ADC_CHANNEL1_TYPE
 #define TEMP_ADC_CHANNEL1_TYPE LM35
 #endif
-#ifndef TEMP_ADC_CHANNEL2_TYPE
-#define TEMP_ADC_CHANNEL2_TYPE LM35
-#endif
-#ifndef TEMP_ADC_CHANNEL3_TYPE
-#define TEMP_ADC_CHANNEL3_TYPE LM35
-#endif
-
-#ifdef TEMP_ADC_CHANNEL1
-static struct adc_buf temp_buf1;
 #endif
 
 #ifdef TEMP_ADC_CHANNEL2
 static struct adc_buf temp_buf2;
+
+#ifndef TEMP_ADC_CHANNEL2_TYPE
+#define TEMP_ADC_CHANNEL2_TYPE LM35
+#endif
 #endif
 
 #ifdef TEMP_ADC_CHANNEL3
 static struct adc_buf temp_buf3;
+
+#ifndef TEMP_ADC_CHANNEL3_TYPE
+#define TEMP_ADC_CHANNEL3_TYPE LM35
+#endif
 #endif
 
 #ifndef TEMP_ADC_NB_SAMPLES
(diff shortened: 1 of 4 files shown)


sw/airborne/modules/sensors/temp_adc.c
```diff
@@ -81,19 +83,25 @@ static struct adc_buf temp_buf3;
...
+static float calc_ntc(int16_t raw_adc, uint16_t pull_up_r, float a, float b, float c)
 {
-  float temp_c;
-  //calc for NTC
-  temp_c = log(((10240000 / raw_temp) - 10000) * 10);
-  //temp_c = 1/(0.001129148+(0.000234125*temp_c)+(0.0000000876741*temp_c*temp_c*temp_c));
-  temp_c = 1 / (0.000603985662844 + (0.000229995493730 * temp_c) + (0.000000067653027 * temp_c *
-                temp_c * temp_c));
-  temp_c = temp_c - 273.15; //convert do celcius
+#ifdef TEMP_ADC_12BIT
```
flixr: What is this define about?

fvantienen: If you have 12 bits ADC versus a 10 bits ADC.
Originally this was written for a 10 bits, so I added this define to support 12 bits as well.

fvantienen:
By the way the usage of this module will be in different pull request, but I'm currently busy getting everything to master etc.

sw/airborne/modules/sensors/temp_adc.c
flixr: Then this should at least be documented, or even better automatically take the correct value depending on what ADC resolution your mcu has.

fvantienen: I will add some documentation, bit in my board header file I set the correct value ;)

gautierhattenberger:
I have never really looked at this module, but I think it is not really properly named. It only supports 2 type of sensor and it is not even possible to configure for a random sensor (like with an offset and sensitivity).
What do you think about supporting a generic sensor, or change the module name ?

fvantienen:
It supports a generic sensor, namely NTC. This is the default temperature sensor used.
Note that the NTC is highly non-linear, that's why you cannot find your simple offset or sensitivity. You instead have an Alpha and Beta value from the datasheet.

sw/airborne/modules/sensors/temp_adc.h
```diff
@@ -28,12 +28,6 @@
 
 #include "std.h"
 
-/// flag to enable sending every new measurement via telemetry
-bool temp_adc_sync_send;
```
flixr: why remove this? IMHO it should rather be added to settings in the module xml

fvantienen: Because in my opinion this was only added for debugging and instead I made the define for that.

flixr: I think it's meant to allow enabling/disabling the reporting at runtime, which IMHO is quite useful.

fvantienen:
I updated everything as requested, is it ok to merge?

sw/airborne/modules/sensors/temp_adc.c
```diff
@@ -31,7 +31,8 @@
 #include "subsystems/datalink/downlink.h"
 #include BOARD_CONFIG
 
-float temp_c1, temp_c2, temp_c3;
+bool temp_adc_sync_send = false;
```
flixr: Should be initialized with the param TEMP_ADC_SYNC_SEND

flixr:
The doc for TEMP_ADC_12BIT was accidentally removed again?
Which boards have 10 bit? AFIAK the stm32 boards have all 12bit, right?
So it would be good to default to the correct one.

sw/airborne/modules/sensors/temp_adc.c
fvantienen: That is done in the initialization function.

fvantienen:
Fixed the default to be for STM32

sw/airborne/modules/sensors/temp_adc.c
flixr: sorry, missed that...

flixr:
It would be better to default to 12bit if STM32 and to 10 bit otherwise (for lpc21), e.g.
```
#ifndef ADC_RESOLUTION
#if defined(STM32F1) || defined(STM32F4)
#define ADC_RESOLUTION 4096
#else
#define ADC_RESOLUTION 1024
#endif
#endif
```

fvantienen:
Ok I added this to `adc.h` because that seemed appropriate.

gautierhattenberger:
Even better could be to have this in adc_arch.h files, but I guess it can be done later.

gautierhattenberger:
Side question out of curiosity: in which application are you using these sensors ? meteo or something else ?

fvantienen:
We are using it for measuring the temperature of the main Motor and the Battery.

PR #1610, opened 2016-04-19
Dual video thread

dewagter:
anyone a better idea? [and don't say: c++ ]


flixr:
Make `video_thread` and array... and initialize/handle all threads in the same module...

Care has to be take for some functions that access shared resources though, e.g. `video_thread_save_shot` shouldn't try to open/save the same files in both threads.
That can also go wrong in the state of this PR with a basically duplicated video_thread2 module...

rmeertens:
Extend the struct video_config_t ? To include a linked list with functions to call after each image?
What do you think about this idea?

rmeertens:
Also a suggestion: extend video_config_t with the parameters of the camera (FOV_W,FX, etc). What do you think about it?

dewagter:
makes sense to me

-Christophe

wilcoschoneveld:
I guess it would make sense to close this pull request?

PR #1611, opened 2016-04-19
Added optical flow landing module

guidoAI:
I added an optical flow landing module. It uses the size divergence from the optical flow algorithm to keep a constant divergence (if set negative the drone will land, 0 = hover, and positive is going up). Advantage with respect to a GPS landing is that it will take into account the landing surface, even if it is higher than the ground level (so landing on a roof for instance). The module does not use sonar for the vertical control, so it can be used by tiny drones that do not carry sonar. The module uses a novel theory on seeing distances with just a single camera (no extra sensors) to detect when it should land. So it can also be used by drones to estimate distance with a single camera.


conf/modules/optical_flow_landing.xml
```diff
@@ -0,0 +1,53 @@
...
+        <dl_setting var="of_landing_ctrl.dgain_adaptive" min="0" step="0.01" max="5.0" module="ctrl/optical_flow_landing" shortname="dgain_adaptive"/>
+      </dl_settings>
+    </dl_settings>
+  </settings>
+
+  <header>
+    <file name="optical_flow_landing.h"/>
+  </header>
+
+  <makefile>
```
flixr: probably only for `target="ap"` or `target="!sim"`?

sw/airborne/modules/ctrl/optical_flow_landing.c
```diff
@@ -0,0 +1,542 @@
...
+#include "subsystems/datalink/telemetry.h"
+
+// used for calculating velocity from height measurements:
+#include <time.h>
+long previous_time;
+
+// sending the divergence message to the ground station:
+static void send_divergence(void)
+{
+  DOWNLINK_SEND_DIVERGENCE(DefaultChannel, DefaultDevice, &divergence, &divergence_vision_dt, &normalized_thrust, &cov_div, &pstate, &pused, &(of_landing_ctrl.agl));
```
flixr: We don't have this message yet...

sw/airborne/modules/ctrl/optical_flow_landing.c
```diff
@@ -0,0 +1,542 @@
...
+  vertical_ctrl_module_init();
+}
+
+/**
+ * Entering the module (user switched to module)
+ */
+void guidance_v_module_enter(void)
+{
+  int i;
+  printf("ENTERING!");
```
flixr: remove these debug printfs?

flixr:
Do you really need a conf with all those airframes?
It's just that if they are redundant (i.e. already in the TUDelft conf) it's way nicer to leave them out as this increases the time required for the tests a lot.

conf/modules/optical_flow_landing.xml
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="optical_flow_landing" dir="ctrl">
+  <doc>
+    <description>
+        Optical flow landing.
+        
+        This module implements optical flow landings in which the divergence is kept constant. When using a fixed gain for control, the covariance between thrust and divergence is tracked, so that the drone knows when it has arrived close to the landing surface. Then, a final landing procedure is triggered. It can also be set to adaptive gain control, where the goal is to continuously gauge the distance to the landing surface. In this mode, the drone will oscillate all the way down to the surface.
```
flixr: Would probably be good to mention here that this is for rotorcrafts/multicopters and not fixedwings?

conf/modules/optical_flow_landing.xml
```diff
@@ -0,0 +1,53 @@
...
+        <dl_setting var="of_landing_ctrl.igain_adaptive" min="0" step="0.01" max="1.0" module="ctrl/optical_flow_landing" shortname="igain_adaptive"/>
+        <dl_setting var="of_landing_ctrl.dgain_adaptive" min="0" step="0.01" max="5.0" module="ctrl/optical_flow_landing" shortname="dgain_adaptive"/>
+      </dl_settings>
+    </dl_settings>
+  </settings>
+
+  <header>
+    <file name="optical_flow_landing.h"/>
+  </header>
+
```
gautierhattenberger: It seems init, enter and run functions are missing. They don't seem to appear in the module xml nor the flight plan.

flixr: As far as I can see they shouldn't be needed, as it's an implementation of the "guidance module" and hence called automatically.

guidoAI:
All requested changes made, and did a pull request for the new message.

PR #1612, opened 2016-04-19
[flight plans] add call_once and run aliases

flixr:
- `<call_once fun="x"/>` is an alias for `<call fun="x" loop="false"/>`
- `<run fun="x"/>` is an alias for `<call fun="x" loop="true">` (hence same as call without loop explicitly specified)

The idea is to make this distinction more clear and encourage the use of `run` and `call_once` instead of `call`.

The previous discussion in #830 and #1606

### DIFF
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -27,14 +27,14 @@
 <!ELEMENT exceptions (exception*)>
 
 <!ELEMENT blocks (block+)>
-<!ELEMENT block (exception|while|heading|attitude|manual|go|xyz|set|call|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|home|path)*>
+<!ELEMENT block (exception|while|heading|attitude|manual|go|xyz|set|call|call_once|run|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|home|path)*>
 
 <!ELEMENT include (arg|with)*>
 <!ELEMENT arg EMPTY>
 <!ELEMENT with EMPTY>
 
-<!ELEMENT while (exception|while|heading|attitude|manual|go|xyz|set|call|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|path)*>
-<!ELEMENT for (exception|while|heading|attitude|manual|go|xyz|set|call|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|path)*>
+<!ELEMENT while (exception|while|heading|attitude|manual|go|xyz|set|call|call_once|run|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|path)*>
+<!ELEMENT for (exception|while|heading|attitude|manual|go|xyz|set|call|call_once|run|circle|deroute|stay|follow|survey_rectangle|for|return|eight|oval|path)*>
 <!ELEMENT exception EMPTY>
 <!ELEMENT heading EMPTY>
 <!ELEMENT attitude EMPTY>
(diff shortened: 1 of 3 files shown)


flixr:
Instead of `call_once` we could e.g. also name it `exec` or so... maybe with a different attribute than `fun`?
Since you can "execute" arbitrary statements (not just function calls)

fvantienen:
I'm ok with `call_once` and for me the argument `fun` as well, because I cannot come up with a better name.

fvantienen:
Could we merge this so that I can fix #1606 even better.

flixr:
I really don't want to hurry this one as naming should be good or we regret it again later (like with `call`).
You can still switch the flight plans from #1606 to an alias later...

flixr:
@gautierhattenberger @dewagter @martinmm any comments or maybe even an idea for better naming than `call_once` or `exec` (or better attribute name than `fun`)?

gautierhattenberger:
I'm fine with both `call_once` and `exec`.
`fun` could be replaced/aliased by `cmd` or `command`, but I don't think it is really necessary.

flixr:
@martinmm @dewagter @podhrmic what do you guys think?

dewagter:
I have no strong opinion here.
I introduced the call_once so I obviously do not object. I believe it can
not be misinterpreted.
Run is less obvious. It could also be called <while>

flixr:
`while` is already taken...

flixr:
Actually the existing while should work the same as `<call fun="foo()" loop="false" break="true"/>` if I see that correctly... so the difference to the proposed `run` is that there break is false by default.

dewagter:
Why should <while> have this break? The last time it is called, the
statement apparently is false, so that means nothing was done, so that
means 1 nav-time-step no navigation code is called if you have a break. Why
break and not continue to the next block? Are you aware of something that
might become worse? I can only think of things that become better.

With <until> it's different: you alway "do something" and then check if it
was the last time you needed this. Then you can break, since you did
something. But with while, you check if it is still needed before doing it.
So if this action was apparently not needed anymore, then why not go the
the next immediately?

-Christophe

flixr:
> Why should <while> have this break?

I'm not saying that it should, just that this is how `while` currently works (and always has)...

Not really sure if we should advocate to use `while` for calling nav_x_run functions (or any other "looping" functions).
While is mainly intended/used for things like
* do nothing as long as there is no GPS fix
```
<while cond="!GpsFixValid()"/>
```
* repeat a sequence of stages until some condition is false
```
<while cond="x > y">
   <go wp="A"/>
   <call fun="foo()"/>
   <go wp="B"/> 
</while>
```

to use while for a navigation routine, you would write
```
<call_once fun="nav_x_setup(1, 2)"/>
<while cond="nav_x_run()"/>
```
where the while has an empty body ("do nothing") and evaluating the condition has the desired effect.
Technically there is nothing wrong with that of course.

Just wondering what we should advocate for calling nav_x_run functions and use in the example flight plans?
- simply use `call` like we have so far
- use `run` alias instead of `call` to make it more clear (like in this PR so far)
- use `while` instead of `call`

dewagter:
1) using while to call functions might be confusing, although it makes it
very clear that the nav function needs to return true while busy. But I can
understand you prefer another name.

2) the other idea that I see no reason that while has a break in the first
place is more related to issues
https://github.com/paparazzi/paparazzi/pull/1606 and
https://github.com/paparazzi/paparazzi/pull/830

3) Concerning the aliases:

   - I'm 100% fine with either <call loop=true/false>, <call_once>, <while>
...
   - I'm 50% fine with the break=true/false (you need to know the code to
understand the importance and most users trying both options would not see
a difference).
   - For me the difference between call and run is not "obvious" ...

-Christophe

flixr:
So how about only adding the `call_once` alias (as that is clearly useful)?
People can then still decide if they prefer to keep using `call` for nav_x_run functions or use `while`.

@gautierhattenberger @martinmm any comments?

flixr:
closing this one: the `call_once` alias was merged with #1675

PR #1615, opened 2016-04-20
[boards] Add the new OPA board

fvantienen:
This adds the newly designed OPA(Oversized Paparazzi Autopilot) board support. It is specially designed for the Outback by @dewagter and includes an FBW and AP in a single board.


podhrmic:
Are the board files available somewhere online, so people can lets say order/build their own?

flixr:
So this also has some changes for pixhawk and rc/spektrum binding/powering... can you please split those into separate commits at least?

PR #1619, opened 2016-04-20
[gps] nmea: send ABI msg on every valid NMEA sentence

flixr:
send GPS ABI message even if position is not available but a new valid sentence was received,
e.g. while no fix but get tow and satellite info

should be a better fix for #1618

same problem is probably present in some other gps drivers...

### DIFF
--- a/sw/airborne/subsystems/gps/gps_furuno.c
+++ b/sw/airborne/subsystems/gps/gps_furuno.c
@@ -55,7 +55,7 @@ static const char *gps_furuno_settings[GPS_FURUNO_SETTINGS_NB] = {
 
 static uint8_t furuno_cfg_cnt = 0;
 
-static void nmea_parse_perdcrv(void);
+static bool nmea_parse_perdcrv(void);
 
 #define GpsLinkDevice (&(NMEA_GPS_LINK).device)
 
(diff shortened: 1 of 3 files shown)


flixr:
Similar problem should be fixed for sirf with b973dd2110537585942d7ccc23d952ea45c76e46

sw/airborne/subsystems/gps/gps_furuno.c
```diff
@@ -125,4 +125,7 @@ void nmea_parse_perdcrv(void)
   struct LtpDef_i ltp;
   ltp_def_from_ecef_i(&ltp, &gps_nmea.state.ecef_pos);
   ecef_of_ned_vect_i(&gps_nmea.state.ecef_vel, &ltp, &gps_nmea.state.ned_vel);
+
+  /* indicate that a new valid sentence is available */
+  gps_nmea.sentence_available = true;
```
dewagter: why is this message sentens_available and not msg_available?

dewagter:
looks great, thanks

sw/airborne/subsystems/gps/gps_furuno.c
flixr: msg_available means that the char parser got a whole msg/sentence, sentence_available here means that it was actually a valid/supported sentence and gps info was updated...

Maybe nicer to get rid of sentence_available and make parse_msg return true if valid...

PR #1622, opened 2016-04-21
[arch] Update udp_arch.c

kirkscheper:
small update


flixr:
Please actually describe the change and why it was made instead of just "small update"

kirkscheper:
Sorry, I thought the change was self explanatory. In the udp read, the code first checks if there is space available in the transport buffer but then reads the maximum amount of data for the transport buffer, I simply changed it such that it reads up to the available limit of the transport buffer. I will try to be more descriptive in the future.

flixr:
Thx. It really does make things easier an quicker if the changes are explained... also when looking up why changes were made later...

PR #1623, opened 2016-04-21
Cleanup cdw airframes

dewagter:


gautierhattenberger:
beside mixing several stuff (cleaning conf, chimu fix) it looks good to me.

dewagter:
Is there a way to trigger a full reintegration server test of all conf
files?

-Christophe

flixr:
Unfortunately neither travis nor semaphore offers something like that, testing all confs is done daily with the semaphore scheduler (but that can only be enabled for one branch, in our case master).

You can run them all locally with `make test_all_confs` though...

dewagter:
I tested all confs locally and they all compile. This can be merged/squashed/reintegrated in the preferred way.

flixr:
ups, that was one click too much... actually wanted to select merge instead of squash...

PR #1624, opened 2016-04-21
converted PI rate controller to floating point

EwoudSmeur:
As requested, the rate controller in floating point. I also changed the integrator scale to be divided by 512.0 to allow easy operation with the (I believe) standard 512.0 Hz control frequency. Anyway, the idea being that you can easily calculate the command after 1 sec 1 rad/s error with the current I gain, for instance compared to the P gain. Now I think about it... maybe dividing by periodic frequency is even better... 

On a side note, the messages are now also in float, so do I have to make a pull request to pprzlink as well? Why is this separate by the way?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate.c
@@ -37,7 +37,7 @@
 #include "subsystems/radio_control.h"
 #include "firmwares/rotorcraft/autopilot_rc_helpers.h"
 
-#define MAX_SUM_ERR 4000000
+#define MAX_SUM_ERR 40000
 
 #ifndef STABILIZATION_RATE_IGAIN_P
 #define STABILIZATION_RATE_IGAIN_P 0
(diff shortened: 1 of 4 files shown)


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate.c
```diff
@@ -161,59 +161,59 @@ void stabilization_rate_read_rc_switched_sticks(void)
...
+  struct FloatRates _error;
+  struct FloatRates *body_rate = stateGetBodyRates_f();
   RATES_DIFF(_error, stabilization_rate_sp, (*body_rate));
   if (in_flight) {
     /* update integrator */
     //divide the sum_err_increment to make sure it doesn't accumulate to the max too fast
-    struct Int32Rates sum_err_increment;
-    RATES_SDIV(sum_err_increment, _error, 5);
+    struct FloatRates sum_err_increment;
+    RATES_SDIV(sum_err_increment, _error, 512.0);
```
flixr: Better use `PERIODIC_FREQUENCY` than hardcoded 512Hz..

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate.h
```diff
@@ -28,23 +28,16 @@
 #ifndef STABILIZATION_RATE
 #define STABILIZATION_RATE
 
-#include "math/pprz_algebra_int.h"
```
flixr: Should include "math/pprz_algebra_float.h" instead...

flixr:
Yes, for the message change, you will need to make a PR in pprzlink.
And it was split out to e.g. finally offer an easier way to integrate communication with Paparazzi vehicles in other code...

You wanted to make the gains the "same" as in attitude stabilization in #1601, is that still the case?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate.c
EwoudSmeur: Good that you agree.

EwoudSmeur:
Float attitude stabilization also has no scaling, so they are compatible. Stab att int still has scaling, so that is not directly compatible.

In this PR I changed the integrator scaling, and changed the gains appropriately in the airframe files. Nothing changed to the P gains.

EwoudSmeur:
So, can we merge this? @dewagter you still want to comment?

flixr:
For me it's fine to merge after https://github.com/paparazzi/pprzlink/pull/12

gautierhattenberger:
also fine for me

dewagter:
This was about  "readability versus efficiency" ?


-Christophe

flixr:
Not only readability and simplification: If you have an MCU with FPU, then the float impl is actually faster as there are less operations...

PR #1626, opened 2016-04-22
Support for classix ?

dewagter:
As far as I know, I must have been the last classix user.

If anyone else in the world still uses this, feel free to keep it.


flixr:
So you are not using yours any more?
Keeping it probably doesn't require much effort, but if you really were the last user and it's unlikely that anyone else will ever want/need this, there is indeed no point in keeping it.

gautierhattenberger:
ours are at the museum since a long time now...

PR #1629, opened 2016-04-24
Utm upgrades

kirkscheper:
I converted traffic_info to a module
Added and upgraded the helper functions for utm structs
added an update step for the gps time_sync in the gps callback so it is always updated
and tried to update the modules that use the traffic info (although apparently none of them are in any conf files).


sw/airborne/subsystems/gps.c
```diff
@@ -342,25 +344,20 @@ void WEAK gps_inject_data(uint8_t packet_id __attribute__((unused)), uint8_t len
...
     // A real UTM position is available, use the correct zone
     utm.zone = gps_s->utm_pos.zone;
     utm.east = gps_s->utm_pos.east / 100.0f;
     utm.north = gps_s->utm_pos.north / 100.0f;
+    utm.alt = gps_s->utm_pos.alt / 1000.f;
   }
   else {
-    struct LlaCoor_f lla;
-    LLA_FLOAT_OF_BFP(lla, gps_s->lla_pos);
-    // Check if zone should be computed
```
flixr: This now means that the utm zone is not computed any more if not set...

sw/airborne/subsystems/gps/gps_sim_nps.c
```diff
@@ -52,6 +52,10 @@ void gps_feed_value(void)
   gps_nps.hmsl        = sensors.gps.hmsl * 1000.;
   SetBit(gps_nps.valid_fields, GPS_VALID_HMSL_BIT);
 
+  utm_of_lla_i(&gps_nps.utm_pos, &gps_nps.lla_pos);
```
flixr: doesn't set utm zone

sw/airborne/math/pprz_geodetic.h
```diff
@@ -70,29 +70,40 @@ extern "C" {
     (_u1).zone = (_u2).zone;     \
   }
 
-
-#define ENU_OF_UTM_DIFF(_pos, _utm1, _utm2) { \
+#define ENU_OF_UTM_DIFF(_pos, _utm1, _utm2) {   \
     (_pos).x = (_utm1).east - (_utm2).east;     \
     (_pos).y = (_utm1).north - (_utm2).north;   \
-    (_pos).z = (_utm1).alt - (_utm2).alt;       \
+    (_pos).z = ((_utm1).alt - (_utm2).alt)/10;  \
```
flixr: Looks wrong to me...

sw/airborne/subsystems/gps.c
```diff
@@ -231,6 +231,8 @@ static void gps_cb(uint8_t sender_id,
   gps = *gps_s;
   AbiSendMsgGPS(GPS_MULTI_ID, now_ts, gps_s);
 #endif
+  gps_time_sync.t0_ticks = sys_time.nb_tick;
```
flixr: This should not be done on any new message, only when tow was actually updated...

sw/airborne/modules/multi/follow.c
```diff
@@ -47,22 +45,32 @@
...
-  new_pos.x = DL_REMOTE_GPS_ecef_x(buffer);
-  new_pos.y = DL_REMOTE_GPS_ecef_y(buffer);
-  new_pos.z = DL_REMOTE_GPS_ecef_z(buffer);
-
-  // Translate to ENU
-  enu_of_ecef_point_i(&enu, &state.ned_origin_i, &new_pos);
-  INT32_VECT3_SCALE_2(enu, enu, INT32_POS_OF_CM_NUM, INT32_POS_OF_CM_DEN);
+  struct ac_info_ * ac = get_ac_info(FOLLOW_AC_ID);
+
+  struct UtmCoor_f my_pos = utm_float_from_gps(&gps, 0);
```
flixr: Better not depend on GPS here... just use position from state interface if actually needed...

sw/airborne/modules/multi/follow.c
```diff
@@ -47,22 +45,32 @@
...
+  my_pos.alt = gps.hmsl / 1000.;
+
+  struct EnuCoor_i enu = stateGetPositionEnu_i();
+  enu.x += POS_BFP_OF_REAL(my_pos.east - ac->utm.east/100.);
```
flixr: Doesn't make sense to me...

sw/airborne/modules/airborne_ant_track/airborne_ant_track.c
```diff
@@ -116,8 +115,8 @@ void airborne_ant_point_periodic(void)
   /*
    * This is for one axis pan antenna mechanisms. The default is to
    * circle clockwise so view is right. The pan servo neutral makes
-   * the antenna look to the right with 0� given, 90� is to the back and
-   * -90� is to the front.
+   * the antenna look to the right with 0� given, 90� is to the back and
```
flixr: please discard those change with invalid characters

sw/airborne/subsystems/gps.c
kirkscheper: This is now done inside the helper function utm_of_lla_i

kirkscheper:
Thanks for the comments. utm_of_lla_i falls furter into either utm_of_lla_f or utm_of_lla_d where the zone is set. 

The stateGet function for UTM doesn't quite work yet, I think it is cause the lla in state isn't set anywhere. Also the conversions from NED are not accurate. I haven;t had time to really look into that just yet. I will add it to my to do list...

sw/airborne/modules/multi/follow.c
```diff
@@ -47,32 +45,44 @@
...
+  // todo replace with stateGet function when they are working for utm pos
+  struct UtmCoor_f my_pos = utm_float_from_gps(&gps, 0);
+  my_pos.alt = gps.hmsl / 1000.;
 
-  // Translate to ENU
-  enu_of_ecef_point_i(&enu, &state.ned_origin_i, &new_pos);
-  INT32_VECT3_SCALE_2(enu, enu, INT32_POS_OF_CM_NUM, INT32_POS_OF_CM_DEN);
+  // get the distance to the other vehicle to be followed
+  struct EnuCoor_f enu = stateGetPositionEnu_f();
+  enu.x += ac->utm.east/100. - my_pos.east;
```
flixr: still don't understand what this is about...
why not simply use the position of the other vehicle and then add the FOLLOW_OFFSET as before?

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -91,32 +90,37 @@ void parse_gps_datalink_small(uint8_t num_sv, uint32_t pos_xyz, uint32_t speed_x
...
-  gps_datalink.ned_vel.y = enu_speed.x;
-  gps_datalink.ned_vel.z = -enu_speed.z;
-  SetBit(gps_datalink.valid_fields, GPS_VALID_VEL_NED_BIT);
+  gps_datalink.gspeed = (int16_t)FLOAT_VECT2_NORM(enu_speed);
+  gps_datalink.speed_3d = (int16_t)FLOAT_VECT3_NORM(enu_speed);
 
   gps_datalink.hmsl = ltp_def.hmsl + enu_pos.z * 10;
   SetBit(gps_datalink.valid_fields, GPS_VALID_HMSL_BIT);
 
+  utm_of_lla_i(&gps_datalink.utm_pos, &gps_datalink.lla_pos);
```
flixr: One of the points in adding the valid_fields was actually to not always calculate every representation...

flixr:
> Also the conversions from NED are not accurate.

Yeah, we need to do the proper conversions here.
UTM != ENU/NED even if the origin coincides as already previously discussed...
This was no problem since so far only fixedwing was using UTM and NED/ENU was simply (ab)used to store UTM coordinates with the local NED origin.

Also IIRC in some other places it is currently assumed that if the UTM origin is initialized, we have a fixedwing firmware and you can actually treat it that way...

flixr:
> utm_of_lla_i falls furter into either utm_of_lla_f or utm_of_lla_d where the zone is set.

sorry, missed that... but that is indeed nicer :-)

flixr:
IMHO it makes more sense to get the UTM stuff sorted out first, before adding the traffic_info and multi module changes...

sw/airborne/modules/multi/follow.c
kirkscheper: I think my idea here was that there was no cross firmware way to set waypoints in utm so I used the utm distance between the acs and added vector to my enu position to get the enu position of the other (approximate, I assume this computation gets more accurate the closer you are) then added the follow offset.

I guess I could alternatively add the offset from the other ac location then convert to lla and set the waypoint with that (or write a helper to set waypoints in utm).

sw/airborne/subsystems/gps/gps_datalink.c
kirkscheper: Fair enough, will remove.

kirkscheper:
And regarding the enu utm stuff I just assumed cause it was used in every other conversion it should also be here but I can remove it. I also don't think that this computation is correct and shouldn't be used, shall I remove it from the other conversions as well? I can additionally add a ecef->lla->utm conversion option in the getUtm.

I also want this to get accepted before the traffic (the traffic pull request actually breaks without it as you can see by the auto tests). Do you want to me to verify the sate get works as well? What was your idea for setting the state? I see that in the code the different state representations are not set everywhere, this results in the utm conversion returning nothing.

flixr:
Added the math additions to master with df91322a19dfcee56118df9e665f733ed0e36000

flixr:
Regarding the state interface, we should first deal with that in a separate PR.

I don't think we can simply remove the current NED<->UTM conversions as they are currently used/needed for fixedwings.
Also do we really want/need to introduce the fixed point UTM representation in the state interface?
It's not like we needed that somewhere so far...

kirkscheper:
I don't have to add the int representation, like I said, the utm_f has a lower precision (I think its like 8cm or something) and its was easier to make conversions but its not critical for me. I will refactor it out.

Should the stateSetLla be called in the gps subsystems?

flixr:
In which case does utm_f have only 8cm precision?

kirkscheper:
Floats can represent integers without rounding errors up to p/m 16777216
and a typical utm north value in Europe is around 5760789 so you get
rounding errors pretty quick. When I was flying with utm as float I had
errors of about 8cm in North and a bit less in East. That's why I do
everything in int representation.

flixr:
right, I didn't really think of the north values
Wondering if we actually shouldn't just go for mm in UtmCoor_i?

kirkscheper:
I'm in favour

flixr:
Ah, not enough range...

kirkscheper:
I split off the gps update and the traffic info to separate PRs. I removed the changes to state for now. Will reconsider this and try again another time. I also need to remember to update the utm_of_gps with stateGet functions once they are working correctly. Check out the PRs and let me know, I will update all things required!

gautierhattenberger:
closing since elements are either merged or reopened in separated PR

PR #1631, opened 2016-04-24
[natnet] update natnet

kirkscheper:
Set reference frame in lla and fix bug to set lla in deg 1e7. 
Generate correct time of week and also send that with the small messages.

### DIFF
--- a/sw/airborne/subsystems/datalink/datalink.c
+++ b/sw/airborne/subsystems/datalink/datalink.c
@@ -195,10 +195,10 @@ void dl_parse_msg(void)
       if (DL_REMOTE_GPS_SMALL_ac_id(dl_buffer) != AC_ID) { break; }
 
       parse_gps_datalink_small(
-        DL_REMOTE_GPS_SMALL_numsv(dl_buffer),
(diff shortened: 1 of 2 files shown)


flixr:
Needs updated REMOTE_GPS_SMALL message, corresponding change in datalink and the math updates from #1629

flixr:
Can you please add the required datalink changes here?

flixr:
Can you please add the required datalink changes here and update the pprzlink submodule to paparazzi/pprzlink@fd4798cae177ca217cfb8683fb7a4a7d93a2148a ?
Then we should have everything needed here together...

kirkscheper:
Not sure what you mean with "Can you please add the required datalink changes" As far as I see it is upto date. Could you be more specific?

flixr:
This changes the REMOTE_GPS_SMALL message that is sent from natnet2ivy, so it needs the changed message in pprzlink and to be updated in the datalink parser...

kirkscheper:
oh, like that. I had planned to just remove the whole thing from the datalink.c in the traffic_info pull request but I will update it here as so that we can close this.

PR #1632, opened 2016-04-24
[state] add utm_i reference frame

kirkscheper:
Should facilitate simple conversions for utm_i structs and more accurate storage of the reference location than the utm_f.

### DIFF
--- a/sw/airborne/state.c
+++ b/sw/airborne/state.c
@@ -310,32 +310,38 @@ void stateCalcPositionLla_i(void)
   SetBit(state.pos_status, POS_LLA_I);
 }
 
+/**
+ * Calculate UTM (float) from any other available representation.
+ * Note that since LLA in float has bad precision but is more accurate
+ * as you move further away from origin.
+ */
 void stateCalcPositionUtm_f(void)
 {
   if (bit_is_set(state.pos_status, POS_UTM_F)) {
     return;
   }
 
-  if (bit_is_set(state.pos_status, POS_LLA_F)) {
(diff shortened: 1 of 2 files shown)


flixr:
IMHO it would make more sense to make one PR with the UTM updates in math and the state interface.
For the additions to math it would also be great if you could add corresponding tests in https://github.com/paparazzi/paparazzi/blob/master/tests/math/test_pprz_geodetic.c

flixr:
I don't understand why you think that the reference location is more accurate as UtmCoor_i

sw/airborne/state.c
```diff
@@ -310,32 +310,38 @@ void stateCalcPositionLla_i(void)
...
     return;
   }
 
-  if (bit_is_set(state.pos_status, POS_LLA_F)) {
-    utm_of_lla_f(&state.utm_pos_f, &state.lla_pos_f);
-  } else if (bit_is_set(state.pos_status, POS_LLA_I)) {
-    /* transform lla_i -> lla_f -> utm_f, set status bits */
-    LLA_FLOAT_OF_BFP(state.lla_pos_f, state.lla_pos_i);
-    SetBit(state.pos_status, POS_LLA_F);
+  if (bit_is_set(state.pos_status, POS_LLA_I)) {
```
flixr: if utm_i is really added, converting from that should be the first choice...

flixr:
Also please note that converting between UTM and ENU/NED with UTM_OF_ENU_ADD is not really correct...

kirkscheper:
Ok, tnx for the feedback, I moved this to #1629

PR #1634, opened 2016-04-24
[fix] fixed some typos and updated some functions

kirkscheper:
Applied some it shift variables instead of hardcoded numbers.
use int conversion in ins instead of float (more accurate)
various typo fixes


kirkscheper:
most important parts imported. Not going to bother with the state utm_i reference frame

PR #1635, opened 2016-04-24
[tools] add the code to run onboard the ARDrone for STDMA a2a coms

kirkscheper:
This adds the capability to run a helper on the ardrone to communicate with a bluetooth dongle in the usb port of the drone. The code implements Self-Organized Time Division Multiple Access protocol which allows the vehicles to broadcast messages directly other drones without the need for a ground station. This dongle should be combined with the extra_pprz_dl module with:
```
<module name="extra_dl.xml">
      <configure name="EXTRA_DL_PORT" value="UDP1"/>
      <define name="UDP1_PORT_IN" value="5000"/>
      <define name="UDP1_PORT_OUT" value="5001"/>
      <define name="UDP1_HOST" value="127.0.0.1"/>
</module>
```
I am not sure if this is the correct place for this so please advise.


flixr:
Nice!
Maybe this would be better put in a separate repo as it doesn't rely on any other pprz stuff?
In any case it would be great if you could add a README.md that explains this a bit.

kirkscheper:
ok, will move this to a repo on the tudelft, or maybe my personal one.

gautierhattenberger:
we can open an "official" repository in paparazzi for this as it may be interesting for many people.

Issue #1199, opened 2015-06-04
Add rotorcraft support for <Follow> flight plan block  [$5]


hooperfly:
The `follow` flight plan block is only supported for `fixed wing` UAS at the moment. It would be nice to have this capability within the `rotorcraft` domain as well. Some design decisions related to how this feature functions across UAS types (i.e. `fixed wing`, `rotorcraft`, `vtol`, etc....) is open for discussion.

Wiki reference: http://wiki.paparazziuav.org/wiki/Flight_Plans#Follow
##

gautierhattenberger:
Main issue is that reference point of rotorcraft are (usually) different, so you can't compare their local position easily. It means that we need to compute a local position from a global one (ecef, lla, utm ?) each time a traffic_info message is received.
Currently this message is sending utm, which is fine for fixedwing, but not really for rotorcraft.

hooperfly:
Can you explain more why UTM is ok for fixedwing vs. rotorcraft? With respect to reference point, I would assume that all aircraft taking part in multi-aircraft exercise would share a common reference point in each of their flight plans. Is this not the case usually? Does reference point mean something other than a (x,y,z)->(0,0,0) reference value?

flixr:
@hooperfly in the fixedwingfirmware the navigation coordinate system is UTM, in rotorcraft firmware it is LTP

hooperfly:
@flixr thanks. I've been reading some archive UTM/LTP fixed/rotor discussion to become more familiar with the issues: https://gitter.im/paparazzi/discuss/archives/2015/01/05

gautierhattenberger:
there is a follow nav module for rotorcraft if needed http://docs.paparazziuav.org/latest/module__follow.html

PR #1638, opened 2016-04-25
[module][refactor] replace traffic_info subsystem with module

kirkscheper:
Splitting #1629. This pull request converts traffic info from subsystem to module and updates the functions that use traffic_info.

Is there any documentation that I should add for this?


flixr:
A detailed description in the module would be great!
The conversion to module really makes sense and is appreciated, although it's a bit more than that ;-)

There are some more things to fix:
- fix waypoint coordinates for fixedwing multi/tcas modules which was broken with #1509 (these modules still expect relative UTM coordinates)

flixr:
Also we should be really sure we actually want to use UTM for traffic_info instead of the more common LLA...
@gautierhattenberger @dewagter what do you guys think?

kirkscheper:
Ah yes, I see that now as well. I will wait till the final decision is made on the utm. I am for lla. Then we don't have to worry about zone extending when in a different zone. I am also fo lla in communications as you have one term less (zone).

gautierhattenberger:
I'm fine with storing the state of other aircraft in lla, but then we also need to provide convenient way to use it with several options:
- direct comparison of two lla coordinates
- convert first in an other frame each time traffic info are used (like "multi" modules)
- make a reduced version of the state interface to access these information in any frame
- probably more options

The painful to implement option 3 is probably the best since you only compute what you need, and even handle local or global coordinates.

kirkscheper:
Thanks for the feedback. Sounds like I will try to use state as an example for the new and improved traffic. I will also take a look at the multi modules to upgrade them to use true global coordinates. I think with this I can also implement a follow nav block for rotorcraft.

I am working on a paper now so would prefer not to change it right now so I will get back to this  in about 2 weeks.

kirkscheper:
after a found break I have found some time to update the traffic. I still have to update the modules with the new protocol but here is a initial preview of what the new traffic info module looks like. I hope this is more in the right direction.

gautierhattenberger:
@kirkscheper it seems you're not pointing to the correct pprzlink version

kirkscheper:
Yea, I know, I am waiting for it to be pulled to master. Currently pointing
to my own branch for testing. This isn't ready for merging yet.

Any comments on the changes? I added an acinfo_lla message, alternatively I
can just replace the utm message. Check changes to server.ml.

I am testing it all right now actually, as soon as I get the thumbs up from
you guys I will clean up the PR and pprlink stuff.

gautierhattenberger:
The overall changes are looking fine. You can probably already make the request for ACINFO_LLA in pprzlink.

sw/ground_segment/tmtc/server.ml
```diff
@@ -381,8 +380,8 @@ let send_aircraft_msg = fun ac ->
         else
           let deg7_of_rad = fun f -> PprzLink.Int32 (Int32.of_float (Geometry_2d.rad2deg (f *. 1e7))) in
           let ac_info_lla = ["ac_id", PprzLink.String ac;
-                             "lat", deg7_of_rad a.pos.posn_lat;
-                             "lon", deg7_of_rad a.pos.posn_long;
+                             "lla_lat", deg7_of_rad a.pos.posn_lat;
```
flixr: I would keep `lat` and `lon` like we have it in the other messages...

kirkscheper:
All cleaned up and ready to go. This does need https://github.com/paparazzi/pprzlink/pull/14 to work fully (weird that all the tests were passed...).

flixr:
Only the configurations in conf_tests.xml are run on every commit. Testing of all confs is only run once per day and not shown in the PR checks...

kirkscheper:
ah, cool. Where can I find the output of the test_all_confs?

flixr:
They are sent to paparazzi-builds@nongnu.org and listed on semaphore: https://semaphoreci.com/paparazziuav/paparazzi (the ones that say "Scheduler rebuilt x" are the ones which compiled all confs)

See also https://wiki.paparazziuav.org/wiki/Builds

sw/airborne/modules/multi/traffic_info.h
```diff
@@ -0,0 +1,383 @@
...
+  if (!bit_is_set(ti_acs[ti_acs_id[ac_id]].status, AC_INFO_VEL_ENU_F)) {
+    acInfoCalcVelocityEnu_f(ac_id);
+  }
+  return &ti_acs[ti_acs_id[ac_id]].enu_vel_f;
+}
+
+/** Get vehicle course (float).
+ * @param[in] aircraft id of aircraft info to get
+ */
+static inline float *acInfoGetCourse(uint8_t ac_id)
```
flixr: I think it would be nicer/better if all get functions returning a scalar actually return a scalar instead of a pointer (like in the state interface).

flixr:
For the future: it makes reviewing much easier if you make separate commits for logical changes and especially separate commits for only formatting changes...

sw/airborne/modules/multi/traffic_info.c
```diff
@@ -0,0 +1,404 @@
...
+{
+  memset(ti_acs_id, 0, NB_ACS_ID);
+
+  ti_acs_id[0] = 0;  // ground station
+  ti_acs_id[AC_ID] = 1;
+  ti_acs[ti_acs_id[AC_ID]].ac_id = AC_ID;
+  ti_acs_idx = 2;
+}
+
+int parse_acinfo_dl(void)
```
flixr: seems like it would make more sense to return a bool here?

sw/airborne/modules/multi/tcas.h
```diff
@@ -54,13 +57,22 @@ extern void tcas_init(void);
...
 
-#define CallTCAS() { if (tcas_status == TCAS_RA) v_ctl_altitude_setpoint = tcas_alt_setpoint; }
+extern void callTCAS(void);
 
-#define ParseTcasResolve() { \
-    if (DL_TCAS_RESOLVE_ac_id(dl_buffer) == AC_ID) { \
-      uint8_t ac_id_conflict = DL_TCAS_RESOLVE_ac_id_conflict(dl_buffer); \
-      tcas_acs_status[the_acs_id[ac_id_conflict]].resolve = DL_TCAS_RESOLVE_resolve(dl_buffer); \
-    } \
+static inline void parseTcasResolve(void)
```
flixr: IMHO it is nicer to make these real functions (they are not speed critical) and move them to the c file. That way we can also minimize the includes needed in the header (only includes that are needed for declarations of things used in the header should be there).
That way we get less coupling and recompilation is faster.

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -675,3 +676,24 @@ void nav_oval(uint8_t p1, uint8_t p2, float radius)
       return;
   }
 }
+
+#ifdef TRAFFIC_INFO
+#include "modules/multi/traffic_info.h"
+
+void nav_follow(uint8_t ac_id, uint32_t distance, uint32_t height)
+{
+  /*  struct EnuCoor_i* target = acInfoGetPositionEnu_i(ac_id);
```
flixr: why is this commented out?
Still some fundamental problem here or simply not yet finished/tested?

sw/airborne/modules/cam_control/cam.c
```diff
@@ -303,13 +302,17 @@ void cam_waypoint_target(void)
...
 {
-#ifdef TRAFFIC_INFO
-  struct ac_info_ * ac = get_ac_info(cam_target_ac);
-  cam_target_x = ac->east;
-  cam_target_y = ac->north;
-  cam_target_alt = ac->alt;
+  struct EnuCoor_f ac_pos *ac = acInfoGetPositionEnu_f(cam_target_ac);
+  cam_target_x = ac->x;
+  cam_target_y = ac->y;
+  cam_target_alt = ac->z;
```
flixr: It looks like cam_target_alt currently expects the alt as hmsl... here you set it to height over local origin.

flixr:
I like the new traffic info, looks well done/structured and nicely commented :+1:

Some general higher level questions and remarks:
- TCAS should also be usable for rotorcrafts now, right?
- Did you test it with a mix of fixedwings and rotorcrafts?
- If I see that correctly TCAS, etc. is now done in ENU...
  Is this "real" ENU or the "fake" ENU that is used in the state interface in the fixedwing case for "local coordinates in UTM"?

@gautierhattenberger regarding the last point: at a glance it looks like this could be problematic (since the ENU position from traffic info is compared to the output of `stateGetPositionEnu_f`. Could you please also take a closer look at this? If there is still an issue with how the state interface handles this we need to open an issue for it...

sw/airborne/modules/multi/traffic_info.h
```diff
@@ -0,0 +1,383 @@
...
+static inline void acInfoSetVelocityEnu_f(uint8_t ac_id, struct EnuCoor_f *enu_vel)
+{
+  VECT3_COPY(ti_acs[ti_acs_id[ac_id]].enu_vel_i, *enu_vel);
+  /* clear bits for all position representations and only set the new one */
+  ti_acs[ti_acs_id[ac_id]].status = (1 << AC_INFO_VEL_ENU_F);
+}
+
+/************************ Get functions ****************************/
+
+extern void acInfoCalcPositionUtm_i(uint8_t ac_id);
```
flixr: how do all these functions handle the case where the info for the requested ac_id is not present?

sw/airborne/modules/multi/tcas.h
kirkscheper: Agree totally.

sw/airborne/firmwares/rotorcraft/navigation.c
kirkscheper: Indeed, not tested yet. I wanted to make sure this was working before I claim that $5 bounty on #1199 :p

sw/airborne/modules/multi/traffic_info.h
kirkscheper: For the set functions, I totally forgot about that. I will update that.

For the get functions, from the current use cases, the position of the other aircraft is used to get a delta which is used to do something so I could return the position of the ownship when the requested ac_id isn't initialised. Velocity is usually used to extrapolate the other ac position with the itow difference so there I would just return zeros. With the current implementation you are likely to get really strange values if the ac isn't initialised yet. Any other ideas?

Now I write this it may also be interesting to have helper functions that just return the relative position of the other vehicle (so do the subtraction for you) cause it is almost always the goal. I was really hoping to be done with this :(

kirkscheper:
Regarding the TCAS stuff, it looks like it only uses relative positions so the ENU vs UTM shouldn't be an issue. For making it work with rotorcraft that needs a little work as it currently uses v_ctl_altitude_setpoint to set the desired alt. I will leave that for another PR.

flixr:
> Regarding the TCAS stuff, it looks like it only uses relative positions so the ENU vs UTM shouldn't be an issue.

Unfortunately that is not true. ENU and UTM coordinate systems don't align perfectly if you are not in the center of a UTM zone. In UTM north always points north and in ENU north is only exactly north at the origin. So if you are further away from the origin the axis don't align...

kirkscheper:
yes yes, this problem I know very well, I have already mentioned before that the conversions in state.c are not correct. you first have to go to lla then to enu. The last time I mentioned this someone mentioned that it was done so as to not have the computation penalty on limited platforms and that the issue would be fixed when the navigation firmware is unified. I really really don't like the hack in state...

kirkscheper:
Oh gosh, I also forgot about ellipsoid to hsml conversions in the frame calculations *face palm*.

kirkscheper:
I think I got most of the comments. ps. if you didn't like what I did in the UTM PR you probably won't like what I did here either. I have a geoid_height variable to convert from ellipsoid to hsml and vise-versa. My plan was that when more accurate methods are developed they could be included in the function which sets the geoid_height variable (traffic_info.c).

sw/airborne/modules/multi/tcas.c
```diff
@@ -219,22 +236,24 @@ void tcas_periodic_task_1Hz(void)
...
         break;
       case RA_LEVEL :
       case RA_NONE :
         tcas_alt_setpoint = nav_altitude;
         break;
+      default:
+        break;
     }
     // Bound alt
-    tcas_alt_setpoint = Max(GROUND_ALT + SECURITY_HEIGHT, tcas_alt_setpoint);
```
flixr: wonder if we shouldn't rather use `ground_alt + SECURITY_HEIGHT` so it takes the geo init changes into account...

gautierhattenberger:
I'm not really sure that the current conversions are really matching. Here, when traffic info UTM position is stored in UTM format, and when ENU is requested, it is computed from UTM->LLA->ENU, which means that it is not compatible for fixed-wing ENU (which are still "local utm / fake enu"). Actually the conversion is not even done since LLA->ENU needs a ned_origin, while only utm_origin is provided in the case of FW.
So if we want to make it work with less efforts (but still some prayers), UTM should be used as it was before, so that when ACINFO is received and used, it is already in the correct format and if it is ACINFO_LLA, LLA->UTM conversion is correct.
If we want to use ENU/NED, we either solve the issue in state (which is not so obvious) or we do the same trick in traffic_info: test utm_origin and ned_origin to decide how the conversion should be done.

sw/airborne/firmwares/rotorcraft/navigation.c
gautierhattenberger: can you add `__attribute__((unused))` in the mean time to silence warnings ?

gautierhattenberger:
It looks okay now, I think it is almost ready to merge.

flixr:
It's still not clear to me in which cases this would now work and in which cases it would be wrong.
Can we please at least try to tease out the remaining cases that are not yet working correctly (and why) before we merge it?
Then we can open appropriate issues for them so it's at least documented and we can more easily track the remaining points and fix them bit by bit...

gautierhattenberger:
It will take same time to analyze and test all of them, but in general, since external positions are received in global coordinates, and that transformation to local ones is done the same (eventually bad) way than state interface and based on the same reference, it should already work in most cases (if not all).

flixr:
OK...

kirkscheper:
With the most recent commit, this is what I think.
When it should work and does: 
Rotorcraft firmware with a gps with LLA messages

When it should work but not tested:
Fixedwing firmware using state interface with UTM = ENU + UTM_ORIGIN
Fixedwing firmware with other vehicles in a different utm zone.

When it doesn't work:
No airframes currently use traffic_info so it might not be tested and updated, (I hope to push a conf with this shortly, maybe even an example!).
Strange situations with uninitialised reference frames...

Is this almost done now? What's missing?

gautierhattenberger:
For points when it doesn't work:
- apparently no example airframes are using "multiple aircraft" functionalities (tcas, formation flight,...) so nothing should be really broken
- this is also true for state interface anyway

I think we are ready to merge. @flixr any objections ?

Issue #1614, opened 2016-04-20
module target selection not working in some cases


flixr:
While target inversion seemed to be fixed with #1583, it still doesn't seem to work as desired/expected in all cases:
- changing `target="ap"` to `target="!sim|nps"` is still supposed to add it to the `ap` target, but that is not the case for e.g. https://github.com/paparazzi/paparazzi/commit/3afcce0737aaf1acbcc4a00503f8382358c88d46
- when a module is in one firmware section (e.g. telelemetry_transparent in rotorcraft firmware), it is also added to all allowed targets in a different firmware (e.g. test_progs). Expected would be that the module is only added for the targets in the firmware section it is part of.

flixr:
Btw, this is why some targets like `demo_ahrs_actuators` are failing...

gautierhattenberger:
It is actually pretty tricky when a module has several makefile sections with different targets (like telemetry_transparent).
When building the list of possible targets, it is loosing the initial logical expression:
target="!sim|nps" target="ap" should be "not(sim||nps) || ap"
but it is seen as the list "!sim, nps, ap" which is understood as "not(sim||nps||ap)"
So I need to find an other way to store and test the list.

flixr:
Should be fixed with #1643

PR #1643, opened 2016-04-28
Fix target filtering

gautierhattenberger:
This should fix the target filtering on modules loading by taking into account the firmware and the list of targets as a real boolean expression (to handle properly the inverted selection)

see #1614

### DIFF
--- a/conf/modules/telemetry_transparent.xml
+++ b/conf/modules/telemetry_transparent.xml
@@ -11,7 +11,7 @@
     <configure name="MODEM_BAUD" value="B57600" description="UART baud rate"/>
   </doc>
   <header/>
-  <makefile target="ap">
+  <makefile target="!sim|nps">
     <configure name="MODEM_PORT" case="upper|lower"/>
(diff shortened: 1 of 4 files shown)


flixr:
Still not quite correct: e.g. for DualBoard_AP_FBW, too much gets added to the fbw target: the files/defines in the makefile section `target="ap"` (without firmware), but only the ones in `target="ap|fbw" firmware="fixedwing"` should be added...

gautierhattenberger:
I don't get it, are you still looking at transparent telemetry module or something else ? There is no `target="ap"` without firmware, I have replaced it with `target="!sim|nps"`, so either the rule is wrong or this part is correctly added. The part with `target="ap" firmware="fw"` is only in the AP part of the makefile. And the part with `target="ap|fbw"` is in both targets (which might be an issue since `fbw_downlink.c` is compiled with AP then?).

flixr:
Ah, sorry... then I was looking at an old version of telemetry_transparent.xml
The "common" target needs to be `target="!sim|nps|fbw"`

Issue #1594, opened 2016-04-04
Compiling with Link Time Optimization flag


gautierhattenberger:
It seems that some boards are short in ROM space to fit the complete autopilot (like the Naze32). The recent Link Time Optimization can reduce the code size (and make it run faster) with the price of a slightly longer compilation time, especially during the link stage.

For instance, a rough test with the quad_flip32 airframe file gives a .bin of 123k (depending of the telemetry and settings enabled) with default settings, and is reduced to 119k just by adding `-flto` flag to `CFLAGS` and `LDFLAGS`.

I think it should be possible to enable this flag when the hardware or the user request it. My question is: where do we put this flag ? In every makefile type, only for embedded stuff, factorized in a makefile like `arm-embeded-toolchain` ?

flixr:
You can should already be able to do something like
`make AIRCRAFT=quad_flip32 ap.compile USER_CFLAGS=-flto`
Maybe we should extend that to USER_LDFLAGS?

Or we create a `ENABLE_LTO` flag or so, which we could then also set by default for some boards?
I wouldn't put something like that in the toolchain file... so probably in Makefile.stm and Makefile.lpc21?
We could also refactor the makefiles even more to pull out common things/flags (warnings, lto flag, etc..)

podhrmic:
Since it is board dependent, shouldn't it be in the board makefile?
How much slower is the compilation? If it is just a few seconds maybe we can have it enabled by default.

flixr:
> Since it is board dependent, shouldn't it be in the board makefile?

It's not really "board dependent" as you could use it for any target, but enabling a flag in the board makefile is what I meant with "set by default for some boards"...

PR #1645, opened 2016-04-30
[build] add boards specific flags when needed

gautierhattenberger:
In particular, activate LTO compilation to reduce code size on small board naze32.
It doesn't seem to be enough, but already helps.

see #1594

### DIFF
--- a/conf/Makefile.chibios-libopencm3
+++ b/conf/Makefile.chibios-libopencm3
@@ -283,7 +283,7 @@ DLIBS = -lm -lopencm3_stm32f4
 #
 
 # List all user C define here, like -D_DEBUG=1
-UDEFS = $($(TARGET).CFLAGS) $(USER_CFLAGS) -DUSE_OCM3_SYSTICK_INIT=0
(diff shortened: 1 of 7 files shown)


flixr:
Since which gcc version is LTO supported? We probably should add a check if the compiler actually supports this? 

Btw, the flip32 aircraft is failing since #1637, which replaced `utm_of_lla_f` in gps.c with `utm_of_lla_i` (which uses the double implementation internally).
This increased coded size by over 5.5kB

kirkscheper:
Wow, that was unintentional. 

I guess a quick fix could be to use USE_SINGE_PRECISION_ TRIG, LLA_UTM and LLA_ECEF for this board.

gautierhattenberger:
It seems that first merge of LTO support in GCC is with version 4.5 (around 2009). Do you think it is really worthy to check compiler version ? It is unlikely to use an older GCC (but we have surprising people in that regard in the community...).

gautierhattenberger:
now the code fits again

flixr:
Thx. In that case I think we don't need a check if it's supported ;-)

PR #1648, opened 2016-05-03
Dual video thread solution

rmeertens:
Hey everybody, 

Last week we had a discussion about the dual video thread (https://github.com/paparazzi/paparazzi/pull/1610). I now solved it using a linked list with listeners for the video. 
This pull request allows you to use both the front camera and the bottom camera on the bebop, and to run optical flow on the bottom camera while running something else on the front camera.


sw/airborne/peripherals/video_device.h
```diff
@@ -33,16 +33,7 @@
...
-struct video_config_t {
-  int w;              ///< Width
-  int h;              ///< Height
-  char* dev_name;     ///< path to device
-  char* subdev_name;  ///< path to sub device
-  uint32_t format;    ///< Video format
-  uint8_t buf_cnt;    ///< Amount of V4L2 video device buffers
-  uint8_t filters;    ///< filters to use (bitfield with VIDEO_FILTER_x)
-};
+#include "../../modules/computer_vision/cv.h"
```
kirkscheper: Shouldn't this just be "modules/computer_vision/cv.h" without the ../..?

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -33,6 +33,10 @@
 #include "modules/computer_vision/viewvideo.h"
 #include "modules/computer_vision/cv.h"
 
+#include "modules/computer_vision/video_thread.h"
+
+#include "boards/bebop.h"
```
kirkscheper: Strange this is added here, would suggest that this only works with the bebop. Is there anyway you can make this more general? Does this also work on the ARDrone?

sw/airborne/modules/computer_vision/cv.h
```diff
@@ -31,10 +31,40 @@
 
 #include "std.h"
 #include "lib/vision/image.h"
+#include "../../peripherals/video_device.h"
```
kirkscheper: This seems like a circular include. I am not sure where else the peripherals/video_device.h is used but it doesn't seem that useful as it is now as it only includes one thing. One idea is to move the #define VIDEO_FILTER_DEBAYER 0x01 to this file and then only include the computer_vision/cv.h in the peripherals/video_device.h and remove this include here.

sw/airborne/modules/computer_vision/colorfilter.c
```diff
@@ -27,8 +27,13 @@
 #include "modules/computer_vision/cv.h"
 #include "modules/computer_vision/colorfilter.h"
 
-#include <stdio.h>
+#include "boards/bebop.h"
```
kirkscheper: Can this be generalized?

sw/airborne/peripherals/video_device.h
rmeertens: Indeed, thanks!

sw/airborne/modules/computer_vision/viewvideo.c
rmeertens: Mmm, you are right that this is a problem. Looks like the cameras of the ARDrone are described in board.c... and the bebop cameras in video.c.... Is there a way to include a specific file depending on the platform you are compiling for?

kirkscheper: The source files are defined in the makefiles in conf/boards. But I think you also have to add the headers correctly here.

flixr:
IMHO there should be no explicit usage of `bottom_camera` or `front_camera` in video_tread.
Also it should still be possible to use a "custom" one...

sw/airborne/modules/computer_vision/viewvideo.c
gautierhattenberger: you can use `#include BOARD_CONFIG` to include the board header file

rmeertens:
One way to go is define every camera you want to use for the video_thread.... but then your airframe file will get huge (and we don't know how many cameras people want to use in the future). 
How would you solve this problem @flixr ?

flixr:
Probably something like an array with all camera/threads, then iterate over each of them instead of using `front_camera` and `bottom_camera` explicitly...

fvantienen:
Or maybe add a new variable to the module that chooses a struct which is then given to each function.
So: `<main_struct name="video_t">` which will then generate a new variable for each module in the airframe file and then gives this variable to every function.

rmeertens:
For working on every platform: I now use this code:
` 
+#ifndef CAMERA_INCLUDE
+#warning "ASSUMING YOU ARE USING A BEBOP! Use the setting CAMERA_INCLUDE to change this"
+#define CAMERA_INCLUDE "boards/bebop.h"
+#endif
+#include CAMERA_INCLUDE`

flixr:
What I meant is that it's not very nice and potentially quite a bit of trouble to have `front_camera` and `bottom_camera` explicitly referenced in video_thread....
It will fail for any other board that doesn't have those two...

Also as @gautierhattenberger already pointed out, you should be able to simply use `#include BOARD_CONFIG` instead of bebop.h, etc...

flixr:
How about something like a `register` function to add camera threads?
Along the lines of:
```C
#define MAX_CAMS 2
struct video_config_t *camera_threads[MAX_CAMS];

struct video_config_t * video_thread_register(struct video_config_t *vid) {
  for (int i; i < MAX_CAMS; i++) {
    if (camera_threads[i] == NULL) {
      // register new camera device/thread
      camera_threads[i] = vid;
      return vid;
    } else if (camera_threads[i] == vid) {
      // camera already registered
      return vid;
    }
  }
  return NULL;
}
```

And call this from the respective module (while checking that the return value is != NULL).
Then go over all entries in `camera_threads` in `video_thread_init`, start, stop, etc..

That way no clumsy define stuff would be needed and you don't start two threads if only one camera is used. A module could also fill in it's own video_config_t and pass that instead of needing to use one of the pre-defined front_camera or bottom_camera.

rmeertens:
> A module could also fill in it's own video_config_t and pass that instead of needing to use one of the pre-defined front_camera or bottom_camera.

I think that this is the best idea! Thanks for the suggestion!

flixr:
Note regarding style/naming, we should try to stick to a common style here:
- variables: under_scored
- structs: CamelCased
- defines: UPPER_CASE

flixr:
IMHO it would be preferable to keep this a bit more modular:
- keep the video_device.h (no reason to include computer vision module headers in board files)
- make video_config and listener part of the video_thread struct instead of the other way round

rmeertens:
> keep the video_device.h (no reason to include computer vision module headers in board files)
What do you mean?

> make video_config and listener part of the video_thread struct instead of the other way round
Don't really agree... You have multiple devices, and each device has a thread attached... Otherwise you could should be able to start multiple threads who start fighting for a device.

flixr:
Before `video_configuration_t` was declared in peripherals/video_device.h, so no dependencies to other things like cv.h, video_thread, etc... that was good.

What I'm trying to say is that the video_configuration should be tied to neither cv.h nor video_thread if we want to use it as a more general camera configuration struct that you can pass from modules.
Just think about how you want to deal with video images you get from some other process via IPC, or the simulator, or some non V4L driver....

> Otherwise you could should be able to start multiple threads who start fighting for a device.

Why? It's just as easy to check if you already have a thread that uses this device....

rmeertens:
Agreed on the video_configuration thing. 

Still don't agree on reversing threads and video devices... It is simply not the obvious object oriented solution....
If I want to use a camera in my module, I either want the front or bottom camera (don't care about the threading behind it)... We COULD make the reverse, but then every time you do something with a camera you have to search through all the threads to find the camera you are specifying.

kirkscheper:
Is it possible to use something like what is used with the datalink devices? Attach a define to each device that is created and use that to interface with the device from your module.

rmeertens:
"Solved" the problem with the settings. 
Any specific suggestions?

flixr:
Um... you accidentally deleted most of conf_example.xml and removed the header from video_device.h

rmeertens:
Still having problems with settings... 
There used to be only one camera, with one FPS slider and one take shot button. 

Now there are N cameras, each one might have a FPS slider and a take shot button. 

I first solved this for the AP case, by defining CAMERA1 and CAMERA2 (people who want to use extra camera's can extend it). For the NPS case this is still difficult, as there are not really cameras in NPS right now...

Any ideas on how to solve this? Define a dummy camera for NPS? Define bottom_camera and front_camera for the bebop NPS? What would the best solution be?

@fvantienen suggested using a slider that selects what camera you are setting something for, but I'm also not a big fan of that solution....

rmeertens:
How about making several sliders, and binding them to a global variable (fps slider 1-4). Using a define the camera can subscribe to this variable by checking if it changed every X frames... 
Still not a painless solution, but I think I like this one the best. What do you guys think?

rmeertens:
Ok, implemented the idea I proposed yesterday. Everything builds and everything works, whoo!
Any remarks?

dewagter:
@rmeertens starts to look quite nice!

what is the idea with video_settings* ? can 2 color filters be started with different settings somehow?

rmeertens:
Nope, currently I made 4 video_settings structs that have sliders and buttons in the GCS. You select what slider you want to use for your camera in the module. 

As for the color_filters: no, simply no. We then need to either:
- adjust the color filter to filter multiple times with different settings
- create color_filter objects (whoo, C++) and find a way to add multiple of them 
- change paparazzi so you can include a module multiple times

sw/airborne/modules/computer_vision/video_thread.c
```diff
@@ -97,21 +69,32 @@ PRINT_CONFIG_VAR(VIDEO_THREAD_SHOT_PATH)
...
+
+// We keep track of each camera device in a linked list
+struct device_linked_list {
+  struct device_linked_list *next;
+  struct video_config_t *camera;
 };
 
-static void video_thread_save_shot(struct image_t *img, struct image_t *img_jpeg)
-{
+struct video_config_t dummy_camera = {
```
fvantienen: why is this here?

sw/airborne/modules/computer_vision/video_thread.h
```diff
@@ -31,19 +31,16 @@
...
-  struct v4l2_device *dev;        ///< The V4L2 device that is used for the video stream
-  uint8_t fps;                    ///< The amount of frames per second
-
-  volatile bool take_shot;      ///< Wether to take an image
-  uint16_t shot_number;           ///< The last shot number
-};
-extern struct video_thread_t video_thread;
 
 // Module functions
+struct video_settings camera1_settings;
```
fvantienen: This is not the way to go in my opinion...

rmeertens: Your way to go was a slider to select a camera right? Still think that this would be a bit more strange... what camera gets what number etc. etc.

fvantienen:
I'm actually with @flixr, which you didn't fix yet.

You should have 2 different structures:
- Video: This contains the settings of a video device, the thread information, a linked list of parsers.
- Video configuration: This should be similar to like `UART1` etc. in board files. So it is just a easy structure containing all the device information to be able to connect. So call them like `VIDEO1`, `VIDEO2` etc.

Then the video_thread will initialize all `VIDEO*` devices if `USE_VIDEO*` is defined. Then in your airframe file you can then add a colorfilter module and give it `VIDEO*` as input and even allow multiple inputs and a specific input for settings if you have multiple.

This way it should be more consistent with the rest of Paparazzi and everything should be possible.

sw/airborne/modules/computer_vision/video_thread.h
fvantienen: No it's more the way all structs are setup etc. look at the general comment

rmeertens:
Video right now has the settings of the device, thread information and a linked list of parsers, right? 

As for video configuration, what do you want exactly? 
Do you mean add this to the bebop.makefile: 
# Cameras, whoo
VIDEO1           ?= front_camera
VIDEO2          ?= bottom_camera
and this to the airframe?
 <module name="video_thread">
      <define name="USE_VIDEO_1" value="1"/>
      <define name="USE_VIDEO_2" value="1"/>
    </module>

Because I still like my solution better... A module says "hey, let me use this camera". At that moment we will start to initialise this camera.... Less defines, less what you need to know what is going on inside for the enduser who makes an airframe, always have the cameras you want available. 

Also, your solution still does not solve the settings problem.

fvantienen:
You should not define `USE_VIDEO2` in video_thread but instead do that in the modules that want to use a camera. So for example:
```
<module name="opticflow">
<makefile>
<configure name="OPTICFLOW_DEVICE1" default="$(OPTICFLOW_DEVICE1)" case="upper|lower"/>
<define name="USE_$(OPTICFLOW_DEVICE1)"/>
<define name="OPTICFLOW_SETTINGS" default="$(OPTICFLOW_DEVICE1)"/>

<configure name="OPTICFLOW_DEVICE2" default="$(OPTICFLOW_DEVICE2)" case="upper|lower"/>
<define name="USE_$(OPTICFLOW_DEVICE2)"/>
</makefile>
</module>
```

This way a module says it wants to use a camera like your method and you can set the device which you want to have settings for ;)

conf/modules/cv_colorfilter.xml
```diff
@@ -28,7 +28,7 @@
   <init fun="colorfilter_init()"/>
   <makefile>
     <file name="colorfilter.c"/>
-    <define name="USE_($COLORFILTER_CAMERA)"/>
+    <define name="USE_$(COLORFILTER_CAMERA)"/>
```
wilcoschoneveld: Unless USE_VIDEO1 is explicitly defined in the video thread module, it doesn't initialize properly. My guess is that the preprocessor resolves video_thread.c before this line?

fvantienen: Thats not possible this adds it to the CFLAGS

wilcoschoneveld:
I refactored the code, removed redundant stuff and fixed the broken tests. Is this sufficient for a merge?

flixr:
Hm... now you have "hardcoded" VIDEO1 and VIDEO2 again... why?
I liked the initial idea of having an array of video devices that the video thread handles and each module can add/register one.

wilcoschoneveld:
I thought that hard-coding the VIDEO devices was proposed in the last set of comments, but I agree that it's not the most elegant way of implementing this feature. I don't mind giving it another iteration.

However, I don't have a clear picture on how to combine the definitions of the module with video settings and a possible 'take shot' function, which can then also be displayed in the GCS settings panel.

I'm thinking about a setup similar to this, but i'm not sure if this is possible in paparazzi:

    <module name="video_thread">
      <thread camera="front_camera" fps="10" path="/internal_000/images/front"/>
      <thread camera="bottom_camera" fps="60" path="/internal_000/images/bottom"/>
    </module>
    
    <module name="obstacle_avoidance">
      <define name="OBSTACLE_CAMERA" value="bottom_camera"/>
    </module>

    <module name="cv_colorfilter">
      <define name="COLORFILTER_CAMERA" value="front_camera"/>
    </module>

    <module name="video_rtp_stream">
      <define name="VIEWVIDEO_CAMERA" value="front_camera"/>
      <define name="VIEWVIDEO_DOWNSIZE_FACTOR" value="1"/>
      <define name="VIEWVIDEO_QUALITY_FACTOR" value="80"/>
    </module>

    <module name="marker_detector"/>

In my view it would make sense to specify in the video thread module which devices to use and what settings you want to apply. This could then also add related settings to the GCS. Any other module can then specify which camera to use, or attach to the first device by default.

Can you give me any pointers in the right direction?

flixr:
Some thoughts:
- @dewagter introduced the possibility to chain image processing in #1573, that is now not the case anymore
  - how you guarantee the order of the image processing calls is still a good question, right now it's just in order that the modules called `cv_add`
- preferably there is no need to specify the cameras in the video thread module, just let modules register one if they need it
  - we could even think about letting each module decide at which fps they want the image and let video_thread grab at the highest requested one (and possibly call modules that require less fps only when that time is up)
- IMHO it would be nicer to put the "save shot" functionality in a separate module
  - so you could possibly also save already processed images

How to deal with settings is probably the biggest urgent question, since in the current system they need to be known beforehand and short of revising the whole settings system I don't have a great idea on how to deal with that right now...
@gautierhattenberger do you have any ideas?

Also taking a broader perspective and just something to keep in mind for the long run: I'm not sure how scalable our current "computer vision framework" approach is:
- we are limited to callbacks that get executed in the thread of the video_thread (ok, a module could copy the image in it's callback and process it in it's own thread)
- we have currently no possibility to process images in a different process, it's all intra-process, but for stability reasons it would be great if we could do some image processing in a completely separate process
- while creating a image processing pipeline is sort of possible with #1573, but rather cumbersome and brittle
- instead of using "custom" callbacks it would be nice to use e.g. some sort of message based pub/sub approach for the images as well
  - similar to ABI (which currently also just does callbacks but could also use messages behind the scenes)

At some point here we have to ask ourselves how much of this we want to do in Paparazzi itself instead of relying on other message/communication frameworks/middlewares...

kirkscheper:
Thanks a lot for the detailed feedback @flixr, I think the problems involved are now much clearer. We have also been discussing the problems with a single threaded vision module amongst ourselves and were thinking of ways to improve that.

@wilcoschoneveld is part of our 2016 IMAV team and will be focused on improving some functionalities (especially with regard to vision processing) in pprz in the coming months leading up to the IMAV (you will probably see more pull requests about that in the coming weeks), but we do need some functionality soon to be able to use both cameras at the same time to make some progress with the various tasks. 

To that end, in your opinion what is required to get this basic functionality pulled into master and what can we spin off into an issue for future development which @wilcoschoneveld (and @rmeertens) can tackle in the coming weeks? @gautierhattenberger and @dewagter your comments are also appreciated.

gautierhattenberger:
Regarding settings, instead of duplicating all settings what can be done is:
* adding a setting to select the video device that will be concerned by the next `take_shot` and other actions
* make a new datalink message to control CV stuff and that specify the device
* do both

First solution can be done easily I think and allow control from GCS, but is a bit error prone if you need to change the device often.
The second option needs a special GUI for user control, but should make things easier when controlling from external programs.

flixr:
So to get something usable ASAP, I would say:
- get rid of VIDEO1, VIDEO2
- create an array to hold all video devices/threads
- when calling `cv_add_to_device` - well - add that video device to the array if not already present and register the cb function
- on start/stop, go over all valid entries in array instead of using hardcoded VIDEO1,VIDEO2
- for fps settings, simply add it to the respective module that uses the video device, e.g.:
  ```<dl_setting var="COLORFILTER_CAMERA.fps" min="1" step="1" max="25" shortname="fps"/>```

rmeertens:
Isn't the something usable ASAP that you are describing simply rmeertens/paparazzi@f6f01c5?

flixr:
IMHO the whole USE_VIDEO1/ and VIDEO1/2 defines are not needed at all and
needlessly restrict it.
Or am I missing something?

rmeertens:
Although I totally agree, I think that @fvantienen wants to have something different. 

So do we... 
- want the array solution https://github.com/rmeertens/paparazzi/commit/f6f01c5
- want to have defines like VIDEO_1 and VIDEO_2 etc.

flixr:
To commit/tree you referenced (and current state of this PR) has the IMHO not needed USE_VIDEO1/VIDEO1 defines...
Simply get rid of those and add an array to hold all requested video devices...

rmeertens:
This going in the right direction?

flixr:
Yes, but honestly I don't see why you use a linked list there, there is no benefit in that case, it just makes it more complex (you need an extra struct for the list, ...)

Also please be careful that you don't accidentally commit the changed submodules as well (no reason to change opencv_bebop here)

rmeertens:
So... like this? 
Also reverted the bebop opencv thingy.

sw/airborne/modules/computer_vision/video_thread.c
```diff
@@ -287,25 +284,28 @@ void video_thread_init(void)
 void video_thread_start()
 {
   // Start every known camera device
-  struct device_linked_list *current_index = &initialised_devices;
-  while (current_index != NULL) {
-    start_video_thread(current_index->camera);
-    current_index = current_index->next;
+  for (int indexCameras = 0; indexCameras < VIDEO_THREAD_MAX_CAMERAS; indexCameras++) {
+    if (cameras[indexCameras] != NULL) {
+      start_video_thread(cameras[indexCameras]);
```
wilcoschoneveld: can also break out of loop

rmeertens:
Anybody has any idea if the NPS view video ever worked? And if so: how did it work?

conf/modules/cv_colorfilter.xml
```diff
@@ -3,6 +3,7 @@
 <module name="ColorFilter" dir="computer_vision">
   <doc>
     <description>ColorFilter</description>
+    <configure name="COLORFILTER_CAMERA" value="VIDEO1|VIDEO2" description="Video device to use"/>
```
flixr: should be `define` not `configure` and probably have front_camera|bottom_camera as values

conf/modules/video_rtp_stream.xml
```diff
@@ -10,6 +10,7 @@
        - Possibility to save an image(shot) on internal or external storage space even in full size, best quality
       Example to add to ARdrone2 airframe with extra USB stick plugger in:
     </description>
+    <configure name="VIEWVIDEO_CAMERA" value="VIDEO1|VIDEO2" description="Video device to use"/>
```
flixr: also a define not a configure option...

flixr:
I don't think that using a webcam or simulated image was ever fully implemented...

sw/airborne/boards/ardrone2.h
```diff
@@ -13,11 +11,15 @@
 #define UARTSPLIT2_DEV /dev/ttyUSB1
 #endif
 
+#include "peripherals/video_device.h"
 
 /* Cameras */
 extern struct video_config_t bottom_camera;
 extern struct video_config_t front_camera;
 
+#define VIDEO1 bottom_camera
```
flixr: not needed...

sw/airborne/boards/bebop.h
```diff
@@ -40,6 +42,9 @@
 extern struct video_config_t bottom_camera;
 extern struct video_config_t front_camera;
 
+#define VIDEO1 bottom_camera
```
flixr: not needed

sw/airborne/boards/pc_sim.h
```diff
@@ -18,5 +18,7 @@
 #define USE_BARO_BOARD 0
 #endif
 #endif
+extern struct video_config_t webcam;
 
+#define VIDEO1 webcam
```
flixr: not needed

sw/airborne/modules/computer_vision/video_thread.h
```diff
@@ -31,19 +31,9 @@
...
-  struct v4l2_device *dev;        ///< The V4L2 device that is used for the video stream
-  uint8_t fps;                    ///< The amount of frames per second
-
-  volatile bool take_shot;      ///< Wether to take an image
-  uint16_t shot_number;           ///< The last shot number
-};
-extern struct video_thread_t video_thread;
-
-// Module functions
+extern void video_thread_initialise_device(struct video_config_t *device);
```
flixr: just for consistency: everywhere else we use the American spelling: initialize

sw/airborne/modules/computer_vision/video_thread_nps.c
```diff
@@ -27,13 +27,14 @@
 // Own header
 #include "video_thread.h"
 #include "cv.h"
+#include "lib/vision/image.h"
+
+// Module functions
+struct video_settings camera1_settings;
```
flixr: not used anywhere

sw/airborne/peripherals/video_device.h
```diff
@@ -1,48 +1,44 @@
-/*
```
flixr: please don't remove the copyright/license and doxygen headers...

sw/ext/opencv_bebop
```diff
@@ -1 +1 @@
-Subproject commit 9eb660dde0f7ec919d476645e136318e74b1a276
+Subproject commit 03cce41a32f43d50e294e7a2e486626b568f4588
```
flixr: changed unrelated submodule again....

conf/modules/cv_colorfilter.xml
wilcoschoneveld: Just to clarify, what is the difference between `define` and `configure`? Why does `configure` allow a default value and `define` does not?

flixr: define creates a C define, configure a makefile variable
see https://wiki.paparazziuav.org/wiki/Modules#In_the_airframe_file

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -46,8 +46,13 @@
 #include "lib/encoding/rtp.h"
 #include "udp_socket.h"
 
-#include BOARD_CONFIG
+#define STR_HELPER(x) #x
+#define STR(x) STR_HELPER(x)
+
+#warning STR(BOARD_CONFIG)
+#pragma message "content of AAA: " STR(BOARD_CONFIG)
```
wilcoschoneveld: @rmeertens are these just debugging statements? if so can I remove them?

sw/airborne/modules/computer_vision/cv.c
```diff
@@ -26,28 +26,52 @@
  */
 
 #include "cv.h"
+#include <stdlib.h> // for malloc
 
-#define MAX_CV_FUNC 10
+void cv_add_to_device(struct video_config_t *device, cvFunction func)
+{
+  // Initialise the device that we want our function to use
+  video_thread_initialize_device(device);
```
wilcoschoneveld: This is currently depending on video_thread.c, which conflicts with the intended structure (video thread depending on cv). It gives a warning during compilation time. Any suggestions on how to refactor this?

wilcoschoneveld:
I implemented all the suggested changes. What do you think, @flixr?

conf/airframes/CDW/cdw_bebop.xml
```diff
@@ -39,11 +39,13 @@
     </module-->
     <module name="video_thread">
       <define name="VIDEO_THREAD_FPS" value="4"/>
-      <define name="VIDEO_THREAD_CAMERA" value="bottom_camera"/>
+      <define name="VIDEO_THREAD_SETTINGS_CAMERA" value="bottom_camera"/>
```
flixr: not used anywhere...

flixr:
I think it would also be better to call `initialize_camera` in `start_video_thread`, so that you could possibly stop a thread, change the camera params and restart it...
Or did you eventually want to pass the return value through to `cv_add_to_device`?

sw/airborne/modules/computer_vision/cv.c
flixr: The only sort of nice solution I can think of is to declare a `add_video_device` function in cv.h and rename `video_thread_initialize_device` to `add_video_device` (and remove the declaration from video_thread.h)
Then we could e.g. also write a different implementation than video_thread, but for the other modules it would work the same way without any code changes...

wilcoschoneveld:
I did indeed add a bool return value to `add_video_device` for that purpose, but I like the idea of initializing the device in start_video_thread (and closing it in stop_video_thread) even more. Is it okay for you if this is implemented in a different pull request?

I updated the cv_blob_locator module and a lot of airframes to work properly.

dewagter:
Just make VERY sure that initialize and close capture are fast asynchronous
functions before moving them to the autopilot flight thread (start/stop)

wilcoschoneveld:
I agree. Currently, it is already possible to dynamically change the desired fps by setting `video_config_t.fps`. 

Is the current state of this pull request good enough for a merge?

rmeertens:
Any updates?

sw/airborne/boards/ardrone/board.c
```diff
@@ -31,32 +31,39 @@
 #include <unistd.h>
 #include "mcu.h"
 
+#include "std.h"
+#include "modules/computer_vision/lib/v4l/v4l2.h"
```
flixr: these includes are not needed, keep the previous #include "peripherals/video_device.h" instead

sw/airborne/boards/bebop.h
```diff
@@ -25,6 +25,8 @@
 
 #define BOARD_BEBOP
 
+#include "std.h"
+#include "modules/computer_vision/lib/v4l/v4l2.h"
```
flixr: why include this here?

sw/airborne/modules/computer_vision/video_thread.c
```diff
@@ -39,7 +39,6 @@
 #include "lib/vision/image.h"
 #include "lib/vision/bayer.h"
 #include "lib/encoding/jpeg.h"
```
flixr: not needed anymore...

sw/airborne/peripherals/video_device.h
```diff
@@ -20,29 +20,44 @@
  */
 
 /**
- * @file peripherals/video_device.h
- * @brief v4l2 video device settings interface
- * Works on Linux platforms
+ * @file modules/computer_vision/cv.h
```
flixr: wrong doxygen header, the original one was correct...

flixr:
Looks good after some minor cleanup: please run fix_code_style on the added files.

Regarding initializing the cameras: yes, this can be improved in a later PR, but we now have to be careful that `video_thread_init` is called **before** any init functions of other cv modules!

Also note that the digital_cam_video.xml module is now broken.

flixr:
Please try to follow our coding style: http://docs.paparazziuav.org/latest/styleguide.html
To fix indentation, etc. you can also run the fix_code_style.sh script on the files you changed/added.

wilcoschoneveld:
Thanks for the feedback @flixr

flixr:
tests are failing now for the nps target as the cameras are not define there...
We could either build these modules only for the ap target or provide dummies for the cameras for NPS...
What do you guys think/want?

wilcoschoneveld:
@flixr is your last comment still relevant?

flixr:
Well... I removed the nps target for all relevant modules for now....
But if you want to use these modules in simulation  (e.g. with a webcam or simulated images) it still has to be updated/implemented for that case....

PR #1649, opened 2016-05-04
[gps] Ublox U-Center module baud rate fix

EwoudSmeur:
We tried to make the autobaud function more reliable for gps modules that standard have a low baud rate and and therefore don't respond in time with the acknowledge to a poll. It seems to work pretty well

### DIFF
--- a/sw/airborne/modules/gps/gps_ubx_ucenter.c
+++ b/sw/airborne/modules/gps/gps_ubx_ucenter.c
@@ -59,8 +59,19 @@ static bool gps_ubx_ucenter_configure(uint8_t nr);
 #define GPS_UBX_UCENTER_REPLY_VERSION     3
 #define GPS_UBX_UCENTER_REPLY_CFG_PRT     4
 
+#define GPS_PORT_DDC        0x00
+#define GPS_PORT_UART1      0x01
+#define GPS_PORT_UART2      0x02
+#define GPS_PORT_USB        0x03
+#define GPS_PORT_SPI        0x04
+#define GPS_PORT_RESERVED   0x05
+#define GPS_PORT_UNK        0x06
+
 // Target baudrate for the module
 #define UBX_GPS_BAUD (UBX_GPS_LINK).baudrate
+
+// Minimum amount of valid packets that need to be received
+#define GPS_UBX_UCENTER_MIN_VALID 1
 
 // All U-Center data
 struct gps_ubx_ucenter_struct gps_ubx_ucenter;
@@ -75,10 +86,12 @@ void gps_ubx_ucenter_init(void)
   gps_ubx_ucenter.status = GPS_UBX_UCENTER_STATUS_AUTOBAUD;
   gps_ubx_ucenter.reply = GPS_UBX_UCENTER_REPLY_NONE;
   gps_ubx_ucenter.cnt = 0;
+  gps_ubx_ucenter.cnt_valid = 0;
 
   gps_ubx_ucenter.baud_init = 0;
   gps_ubx_ucenter.baud_run = 0;
   gps_ubx_ucenter.baud_target = UBX_GPS_BAUD;
+  gps_ubx_ucenter.port_id = GPS_PORT_UNK;
 
   gps_ubx_ucenter.sw_ver_h = 0;
   gps_ubx_ucenter.sw_ver_l = 0;
@@ -163,6 +176,7 @@ void gps_ubx_ucenter_event(void)
   if (gps_ubx_ucenter.status == GPS_UBX_UCENTER_STATUS_STOPPED) {
     return;
   }
+  gps_ubx_ucenter.cnt_valid++;
 
   // Read Configuration Reply's
   switch (gps_ubx.msg_class) {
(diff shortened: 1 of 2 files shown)


flixr:
You changed this based on an older version, right?
Please always double check the introduced changes after rebasing... this seems to revert the fix to still have NMEA enabled so you can use the Parrot app again: 89560dc84fb75e9335d609aa114dd2f7036c14e3

sw/airborne/modules/gps/gps_ubx_ucenter.c
```diff
@@ -258,56 +272,62 @@ static bool gps_ubx_ucenter_autobaud(uint8_t nr)
...
       gps_ubx_ucenter.reply = GPS_UBX_UCENTER_REPLY_NONE;
       uart_periph_set_baudrate(&(UBX_GPS_LINK), B38400); // Try the most common first?
+      gps_ubx_ucenter.cnt_valid = 0;
       gps_ubx_ucenter_config_port_poll();
       break;
     case 3:
-      if (gps_ubx_ucenter.reply == GPS_UBX_UCENTER_REPLY_ACK) {
-        gps_ubx_ucenter.baud_init = gps_ubx_ucenter.baud_run;
+      if (gps_ubx_ucenter.cnt_valid > GPS_UBX_UCENTER_MIN_VALID || gps_ubx_ucenter.reply == GPS_UBX_UCENTER_REPLY_ACK) {
+        gps_ubx_ucenter.baud_init = 38400;
```
flixr: so the `baud_run` set by a reply to gps_ubx_ucenter_config_port_poll() is not used anymore at all?

flixr:
Looking at it, the whole way the ucenter module is set up seems brittle to me:
- `gps_ubx_ucenter_periodic()` is called at 4Hz
  - if it's in `GPS_UBX_UCENTER_STATUS_AUTOBAUD` state, `gps_ubx_ucenter_autobaud` will be called
    - set uart baudrate to next guess and request port configuration via `gps_ubx_ucenter_config_port_poll()`
- `gps_ubx_ucenter_event()` is called from gps_ublox on each valid new message
  - this sets `gps_ubx_ucenter.reply` according to the message that was received
- next 4Hz periodic step:
  - in autobaud check `if (gps_ubx_ucenter.reply == GPS_UBX_UCENTER_REPLY_ACK)`
  - but we actually requested a `CFG_PRT` message before, so we probably want to check for `GPS_UBX_UCENTER_REPLY_CFG_PRT`?
  - the ack reply was in response to???

EwoudSmeur:
Thanks for your reply Felix. I switched to master and did all testing etc on there, so no issues there. We switched off the NMEA messages because they add a lot of extra data. Perhaps we can think of a different way to keep compatible with Parrot.

How do you think we can improve the flow of the module and improve reliability?

fvantienen:
The ACK I think you get everytime you send a CFG message no matter if it is an GET or SET. This is what the protocol description says:
>  Any messages in the CFG class sent to the receiver are either acknowledged (with message UBX-ACK-ACK) if processed successfully or rejected (with message UBX-ACK-NAK) if processing unsuccessfully.

flixr:
For parrot compatibility it might be sufficient to just keep NMEA enabled for the input protocol but disable the outgoing messages?

flixr:
So for the flow I think it would make sense to check the reply in the (kind of misnamed)  gps_ubx_ucenter_event function instead of some time later in periodic, and maybe then use the baud rate from the CFG_PRT reply (currently saved to baud_run)?
In any case just checking for ACK in autobaud is not great...

fvantienen:
@flixr That problem that you noted is indeed in there and stupid. I also noticed it. But I think this whole module needs a rewrite.
Only I got a question of what would be the best approach. My idea is as following:
- First try the baudrate from the the airframe file (And maybe send a message to force a response)
- Check if you receive xx valid UBX messages
- Change baudrate and repeat last step untill got correct baudrate. (instead of all the switch cases a list would be nice)
- Then send an UBX_PORT change to change to the airframe baudrate. (Maybe send this on all ports by default, so that it doesn't require a response to CFG_PORT)
- Switch to airframe baudrate
- Send config (also for all ports?)

Or is there something against changing the baudrate on all ports?

For the bebop 4 things can be done:
- If the NMEA enable for receive only works as fix for that it would be the best solution. Last time someone tested and it didn't work.
- Disable config saving for the bebop.
- Fix an option in GCS to enable NMEA also
- Enable NMEA as output also (which causes a lot of data and prolly also partly caused these problems we have now).

flixr:
Sounds good to me, although I think it would be good to only change the config of the port that is actually used (so don't send/set for all ports).

fvantienen:
That is more difficult if for example the GPS is in B9600 and the port is totally flooded.. The response can take second then.

flixr:
How about using the port_id as returned by CFG_PRT by default, but allow to explicitly define a GPS_UBX_PORT_ID and directly use that if defined? Then we could set this for bebop as it's fixed there anyway... but it would still work nicely on other ublox modules.

fvantienen:
That helps indeed, but we have M8N GPS's in the lab which also have these problems, but have an UART1 or UART2 port.

flixr:
@fvantienen any updates here?

flixr:
@EwoudSmeur @fvantienen it would be great if we could merge a fix for this ASAP...
Any updates here?

flixr:
ping @EwoudSmeur @fvantienen

EwoudSmeur:
pong :)

Sorry I have been quite busy, and will be busy for the coming weeks, so I am not sure if I will have time to look into this.

dewagter:
this was a fundamental problem:

 - you need a reply to know which internal port the uart is connected too
(and at which baudrate)
 - if the device overloads the uart (like some ublox8 with by default give
glonas-gps-beidou-gallileo SV info at high rate), messages are lost

solution:

 - when no reply is received in all normal baudrates:
 - bang-config all ublox-internal-ports on all baudrates to the required
baud, then start over again

-Christophe

flixr:
any updates here? seems it would be quite important to have a working solution for this...

podhrmic:
@EwoudSmeur @dewagter any updates here?

dewagter:
- [x] status: check if baud-rate is correct by counting the correctly
decoded packages instead of checking only for an ACK message solves a lot
of the problems and was implemented in this pull request.
 - [ ] config-banging the ublox in case the previous did not work is an
future idea, not required to make this pull request useful.

-Christophe

On Fri, Nov 18, 2016 at 7:46 PM, Michal Podhradsky <notifications@github.com

OpenUAS:
Only assigned self for now to test on Bebop, Bebop2, Lisa M, Pixracer, and Twog so that the fix can be merged. IMHO honor the pull already...we test master more ;)

podhrmic:
@OpenUAS any news here?

OpenUAS:
No news is good news ;) Flown some, but due to coming magneto fix waited for mor tests. WV was busy this week. Next week some days off to test again.

OpenUAS:
Other fixes @wv-tud, @EwoudSmeur and some other Devs coming on 20170616 validation flights and tests that day, then this issue can be given some TLC. After conflicts resolved, retested that it all works on BB2, and two othe AC with ublox GPS a possiblity to merge.

wv-tud:
@OpenUAS I dont think the M8P on the bebop2s we have suffer from this,I wasn't aware of any issue

OpenUAS:
@EwoudSmeur We have all the hardware to test. Like to finish pull request, any change we can make it happen together?

EwoudSmeur:
I am pretty busy and don't have intense feelings with this pull request. I leave it up to you guys!

gautierhattenberger:
@EwoudSmeur @OpenUAS Do you have a working and up-to-date solution for that ? Otherwise, I would like to close this since nobody will probably work on it.

OpenUAS:
Sorry, will not fix ATM, feel free to reopen and assign to self

PR #1651, opened 2016-05-04
Rename values to cmd_values

fvantienen:
`values` in the `command_laws` section is renamed to `cmd_values`.
This is done in order to be able to use `radio_control.values` inside `command_laws`.
Replace command is: `find conf/airframes -type f -print0 | xargs -0 sed -i 's/,values/,cmd_values/g'`

### DIFF
--- a/conf/airframes/AGGIEAIR/aggieair_ark_quad_lisa_mx.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_ark_quad_lisa_mx.xml
@@ -56,7 +56,7 @@
   </section>
 
   <command_laws>
-    <call fun="motor_mixing_run(autopilot_motors_on,FALSE,values)"/>
+    <call fun="motor_mixing_run(autopilot_motors_on,FALSE,cmd_values)"/>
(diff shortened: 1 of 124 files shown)


flixr:
Hm... `values` is indeed not the greatest name in that macro, but that would break all existing user rotorcraft airframe files...

As an alternative we could change the `SetActuatorsFromCommands` macro to a static inline function (would be nicer anyway) and then using `radio_control.values` should work just fine.

fvantienen:
That also seems like a good solution ;) Didn't think of that

fvantienen:
A static inline gives some problems with the includes.. You for example suddenly need "acutators.h" for the `actuators` array. Also if you include swashplate mixing or motor mixing you suddenly need also those to be included in the `airframe.h`.
@flixr or @gautierhattenberger you know any better fix?

gautierhattenberger:
You can make (yet another) macro like `RadioControlValues`

fvantienen:
Ok I will go for the `RadioControlValues` define.. will close this 😉

PR #1654, opened 2016-05-05
Make half turn in poly survey optional

podhrmic:
I.e. sometimes we dont want the half width turn at the end of the survey (because we want consistent spacing between images). 

Also the poly survey now can be run multiple times, you just have to reset the counter in GCS.


podhrmic:
@flixr any feedback on this? Can we merge?

podhrmic:
@gautierhattenberger any comments on this or can we merge it?

flixr:
When adding new define/configure options, please also add them in the documentation section of the module xml file...

PR #1656, opened 2016-05-06
Skid landing

podhrmic:
Added skid landing routine. It is similar to the landing blocks in the flight plans, but it is a single routine so it is easier to use. Meant for fixedwings that land on their belly, see video: https://www.youtube.com/watch?v=aYrB7s3oeX4

Includes improved flare handling, which makes the last section of the landing much smoother.

### DIFF
--- a/sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
+++ b/sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
@@ -332,7 +332,7 @@ void v_ctl_landing_loop(void)
     land_alt_i_err += land_alt_err / CONTROL_FREQUENCY;  // integrator land_alt_err, divide by control frequency
     BoundAbs(land_alt_i_err, 50);  // integrator sat limits
 
-    // set pitch based on airspeed error
+    // set pitch based on ground speed error
     nav_pitch -= land_speed_err * v_ctl_landing_pitch_pgain / 1000
         + land_speed_i_err * v_ctl_landing_pitch_igain / 1000;  // 1000 is a multiplier to get more reasonable gains for ctl_basic
     Bound(nav_pitch, -v_ctl_landing_pitch_limits, v_ctl_landing_pitch_limits);  //max pitch authority for landing
--- a/sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
+++ b/sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
@@ -30,6 +30,7 @@
 #include "firmwares/fixedwing/nav.h"
 #include "generated/airframe.h"
 #include "firmwares/fixedwing/autopilot.h"
+#include "firmwares/fixedwing/stabilization/stabilization_attitude.h" //> allow for roll control during landing final flare
 
 /* mode */
 uint8_t v_ctl_mode;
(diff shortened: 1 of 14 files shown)


podhrmic:
Looks like this is failing with airframes that use the _new_ control loops.
@gautierhattenberger what is the difference from the _standard_ control loops? Maybe it would make sense to add the new landing control into the _new_ control loops too?
Otherwise there would be an #ifdef guard around v_ctl_landing_loop() I suppose...?

gautierhattenberger:
I don't really like adding new modes, but I guess the nav loop is not fast enough and the mode generation is still not operational.
There is not much difference between old and new loops, but not sure adding this landing loop in all vertical control is the way to go. For instance, it is not working either with the energy control even if compilation is failing on our examples because the common header is not so common.
For now I suggest to properly guard the new mode. @flixr any better idea ?

podhrmic:
This should fix it - not the prettiest, but I tried to make it as legible as possible.

sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
```diff
@@ -451,7 +451,12 @@ inline static void h_ctl_pitch_loop(void)
     h_ctl_elevator_of_roll = 0.;
   }
 
-  h_ctl_pitch_loop_setpoint =  h_ctl_pitch_setpoint + h_ctl_elevator_of_roll / h_ctl_pitch_pgain * fabs(att->phi);
+  if (v_ctl_mode == V_CTL_MODE_LANDING) {
+    h_ctl_pitch_loop_setpoint =  h_ctl_pitch_setpoint;
```
gautierhattenberger: is this making a really big difference ?

sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
```diff
@@ -251,6 +308,50 @@ void v_ctl_climb_loop(void)
...
+  } else {
+    // set throttle based on altitude error
+    v_ctl_throttle_setpoint = land_alt_err * v_ctl_landing_throttle_pgain
+        + land_alt_i_err * v_ctl_landing_throttle_igain;
+    Bound(v_ctl_throttle_setpoint, 0, v_ctl_landing_throttle_max * MAX_PPRZ);  // cut off throttle cmd at specified MAX
+
+    land_alt_i_err += land_alt_err / CONTROL_FREQUENCY;  // integrator land_alt_err, divide by control frequency
+    BoundAbs(land_alt_i_err, 50);  // integrator sat limits
+
+    // set pitch based on airspeed error
```
gautierhattenberger: here, ground speed is used, not airspeed, but it is maybe only an error in the comment

gautierhattenberger:
It indeed looks the minimum to make it compatible, but feels unpleasant somehow. @flixr @dewagter , what do you think ?
It mostly show once again that mode handling should be greatly improved.

sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
podhrmic: It seems to work better like this.

sw/airborne/firmwares/fixedwing/guidance/guidance_v.c
podhrmic: yes it was a typo - fixed to ground speed.

podhrmic:
So I updated it, but now it says that the branch has conflicts? Do you know what happened?

flixr:
Please add the SKID_LANDING defines to the documentation section of the module xml

Also coding convention wise we encourage using under_scored instead of CamelCased variable names.

Also why change all the AC IDs of the test conf?
For the conflict, it is probably easiest to squash and rebase it on top of master and fix the conflicts...

sw/airborne/modules/nav/nav_skid_landing.c
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+#include "generated/airframe.h"
+#include "state.h"
+#include "modules/nav/nav_skid_landing.h"
+#include "firmwares/fixedwing/autopilot.h"
+#include "firmwares/fixedwing/nav.h"
+#include "firmwares/fixedwing/stabilization/stabilization_attitude.h"
+
+struct Point2D
```
flixr: why add Point2D instead of using FloatVect2?

sw/airborne/modules/nav/nav_skid_landing.h
```diff
@@ -0,0 +1,54 @@
...
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+/**
+ * @file modules/nav/nav_skid_landiungr.h
+ * @brief Landing on skidpads
+ * See video of the landing: https://www.youtube.com/watch?v=aYrB7s3oeX4
```
flixr: this description would also be nice in the modue xml

podhrmic: Fixed

sw/airborne/modules/nav/nav_skid_landing.c
podhrmic: Fixed

podhrmic:
Closing and moving into a new pull request #1669

flixr:
Btw, you don't have to close a pull request, you can also just rebase/squash and force push to the same branch and the pull request will be updated...

podhrmic:
Ah, good to know...

PR #1657, opened 2016-05-06
[modules] Add OPA controller

fvantienen:
Special controller for the OPA (Oversized paparazzi Autopilot), which enables the full functionally of the the board.
This includes:
- Power of AP from FBW
- Power of Motor from FBW
- Power of Charging of FBW battery from FBW
- Disabling power of Motor from AP settings trough FBW
- Power of Vision computer trough AP

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_conf.xml
+++ b/conf/airframes/TUDELFT/tudelft_conf.xml
@@ -216,7 +216,7 @@
    telemetry="telemetry/rotorcraft_with_logger.xml"
    flight_plan="flight_plans/dummy.xml"
    settings="settings/rotorcraft_basic.xml settings/control/stabilization_att_int.xml settings/control/rotorcraft_guidance.xml"
-   settings_modules="modules/geo_mag.xml modules/air_data.xml modules/temp_adc.xml modules/logger_sd_spi_direct.xml modules/gps_ubx_ucenter.xml"
+   settings_modules="modules/opa_controller.xml modules/geo_mag.xml modules/air_data.xml modules/temp_adc.xml modules/logger_sd_spi_direct.xml modules/gps_ubx_ucenter.xml"
    gui_color="#ffffdffac31f"
   />
   <aircraft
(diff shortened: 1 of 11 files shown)


podhrmic:
@fvantienen did you add the hardware source files (schematics and layout) of the board into pprz-hardware repo? Because otherwise it is not very useful to have support for a board other people cannot get/make/use...

fvantienen:
Not yet, this board is specifically made for the Outback challenge 2016. Even if we share it probably no one is going to use it.
Nevertheless we want to have it in master, else we always have to redo a lot of work if someone updates something this is depending on. Eventually we want to fly the challenge on the latest master.

gautierhattenberger:
It looks fine for me. Since you are now using a `cmd_status` bit mask, maybe documenting it and add some macros (at least bit shift) to access each elements.

podhrmic:
That is understandable, but maybe somebody can benefit from the design you made and use it for some other competition / application. Otherwise it sets a precedent and people will start uploading their custom boards that are not useful and just clutter the repository. So I would really appreciate if you shared it.

podhrmic:
@esden what do you think?

fvantienen:
I'm not the one who designed it.. so it's not my decision @dewagter.
But if these things not get merged I don't think Paparazzi will get any updates from the Outback anymore, because I don't like to backport everything. So also useful stuff I will then not spend anymore time on to backport it to master, and we will then fly with our own branch of Paparazzi which is never going to get merged.
So for me it is either everything or nothing...

dewagter:
The board for outback is actually 2x lisa-mx on a single PCB, with a lot of
power control (including self-power-down), line drivers, ESD protection,
temperature, rpm & current sensing, logging, backup IMU etc. Any pin from
Lisa-MX is kept the same (which is why this pull request is so small by
reusing mx defines), but many power options are added (e.g. self-power-on).

Most power options (like EStop, Arming, Off) are requirements from the
outback challenge. I expect those to be needed for many more years and even
other boards. And the power control uses defines so if the power control
hardware is added to/beside other autopilots, it should be directly usable
already.

-Christophe

fvantienen:
I added the dox ;)

fvantienen:
@gautierhattenberger or @flixr can we merge this?

flixr:
Please always add all define/configure options in the documentation section of the module xml

PR #1658, opened 2016-05-07
Fixed wrong derivative term in pitch control loop

podhrmic:
Incorrect derivative term in pitch control, replaced it with the correct one.

### DIFF
--- a/sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
+++ b/sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
@@ -477,10 +477,10 @@ inline static void h_ctl_pitch_loop(void)
 #endif
 
 #if USE_GYRO_PITCH_RATE
-  float d_err = h_ctl_ref.pitch_rate - stateGetBodyRates_f()->q;
+  float d_err = stateGetBodyRates_f()->q;
 #else // soft derivation
   static float last_err = 0.0;
-  float d_err = h_ctl_ref.pitch_rate - (err - last_err) / H_CTL_REF_DT;
+  float d_err = (err - last_err) / H_CTL_REF_DT;
   last_err = err;
 #endif
 
--- a/sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
+++ b/sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
@@ -444,8 +444,9 @@ inline static float loiter(void)
 
 inline static void h_ctl_pitch_loop(void)
 {
-  static float last_err;
   struct FloatEulers *att = stateGetNedToBodyEulers_f();
+  struct FloatRates *body_rate = stateGetBodyRates_f();
+
   /* sanity check */
   if (h_ctl_elevator_of_roll < 0.) {
     h_ctl_elevator_of_roll = 0.;
@@ -477,9 +478,8 @@ inline static void h_ctl_pitch_loop(void)
 #endif
 
 
-  float d_err = err - last_err;
-  last_err = err;
-  float cmd = -h_ctl_pitch_pgain * (err + h_ctl_pitch_dgain * d_err);
+
+  float cmd = -1*(h_ctl_pitch_pgain * err + h_ctl_pitch_dgain * body_rate->q);
 #ifdef LOITER_TRIM
   cmd += loiter();
 #endif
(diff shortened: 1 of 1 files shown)


sw/airborne/firmwares/fixedwing/stabilization/stabilization_attitude.c
```diff
@@ -472,9 +473,8 @@ inline static void h_ctl_pitch_loop(void)
 #endif
 
 
-  float d_err = err - last_err;
-  last_err = err;
-  float cmd = -h_ctl_pitch_pgain * (err + h_ctl_pitch_dgain * d_err);
+
+  float cmd = -1*(h_ctl_pitch_pgain * err + h_ctl_pitch_dgain * body_rate->q);
```
gautierhattenberger: This is actually breaking all airframe files using the legacy control loops. Before the pgain was in factor of the angle error and the derivative term, as described here: http://wiki.paparazziuav.org/wiki/Control_Loops#Pitch_loop.
With your modification, it makes all previous pitch damping factor ineffective: pgain is usually in the range of a few thousands (since `cmd` is between [-9600, 9600]) and dgain around 1).
The reason to have this, is that in this form the dgain represent the time constant associated with the derivator, and is independent of the overall gain (pgain here). As a side note, the "new/adaptive" control is using the other usual form kp*e + kd*de/dt as you suggest here.
So finally, I'm not really in favor of changing the control structure now.

Regarding the choice of a soft derivate instead of using pitch rate, it is just that this piece of code is older than the generalized usage of IMU and hence gyro data on the pitch axis.
This is how it is handled in the adaptive loop: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/fixedwing/stabilization/stabilization_adaptive.c#L600 with a USE_GYRO_PITCH_RATE default to true.

podhrmic:
Interesting, I didn't know that. Oddly enough, when I run the simulation (NPS) I couldn't see any difference in pitch control performance, maybe because the NPS model is too simple for that?

I guess the question is then - do we want to eventually "standardize" the control loops (i.e. use the standard form), or is it good "as is"?

podhrmic:
This is what I have from our engineer who actually found this:

> For the legacy airframe files - you just have to rescale the derivative terms to be correct.  The reason I think the old code should be updated is your derivative term is a function of the p_gain.   So as the p_gain goes from 5000-> to 6000, and your derivative term is 3 -  it just went from 15000 to 18000 - difference of 3000 vs 1000.   Also, all the airframe files I have seen have very very very very very low or zero derivative terms since it didn't response like a classical derivative gain.  I feel like this is due to two reasons: 1: Using an improper derivative term  2: The slider for tuning was way too insensitive  (1 or 2 per division) but multiplied by the p_gain caused large swings in an already sensitive tuning parameter.

Maybe it is worth unifying the derivative term with the adaptive loop and retuning the airframes while notifying users about the update?

podhrmic:
Is there a policy for kind of "hanging" pull requests? I am just wondering if it makes sense to close this since there is not much enthusiasm to change the control loops right now... @gautierhattenberger ?

gautierhattenberger:
In general, I don't really like changing control gain structures, even if in this case, the actual effect would be limited.
After looking at the setting file, I completely agree that the `PITCH_DGAIN` range for the basic control is completely stupid. It should be between 0 and 3 at most and with a small step.
@dewagter do you have an opinion on this ?

dewagter:
Personally, I like this change since this unifies control with outerloops
and rotorcraft (and control theory like it is done in most of the world).
But a lot of users might hate "a change to stabilization of their pitch
loop". My personal suggestion would be to ask this on the mailing list.

-Christophe

dewagter:
And I presume with this pull-request all fixedwing airframes on master would get updated to D_new = D_old * P_old

podhrmic:
I updated the pitch control as discussed. However, I changed the pitch D gain as suggested (D_new = D_old * P_old) the airframe doesn't fly well any more. When I left the original D gain it still works fine (tested with Minion_Lia) - so I am hesitant to update the D_gain in the config files... @gautierhattenberger did I miss something?

OpenUAS:
Fine with me if it brings us closer to a unified Rotorcraft and Fixwing. A big fat warning when building airframe would be welcomed. Will risk losing own airframes for sake of progress ;)

EwoudSmeur:
It seems indeed that the gains for airframes should be different. But as you leave the old controller in anyway, why not take the old controller as default?

podhrmic:
Closing for now, as I would like to rather refactor and modularize the control modules before making changes into the "legacy" controllers.

podhrmic:
Again, keeping the branch for now as a reference.

Issue #1616, opened 2016-04-20
Improve static sectors by lifting clockwise and convex hull restrictions


podhrmic:
Hi, 
as mentioned in the [wiki](http://wiki.paparazziuav.org/wiki/Flight_Plans#Sectors) the static sectors have to be defined **clockwise** and be **convex** which is actually really limiting from the user standpoint (I just ran into that issue).

The dynamic sectors seem to work fine no matter on the order their corners are defined, and no matter whether they are convex or not (since the used [algorithm](http://alienryderflex.com/polygon/) is really general).

Would be possible to fix the static sector function generator to work the same way as for dynamic sectors (i.e. to be able to deal with non-clockwise and non-convex polygons)?

gautierhattenberger:
The issue is named properly I think. The static algorithm is more efficient but also more limited (need to be convex on the north-sud axis, cannot be modified at runtime).
The dynamic version is potentially slower, but use the actual waypoint position at runtime.

If you want a static sector that works like the dynamic one, why not using a dynamic sector with hidden waypoints (prefixed by `_`) ?
Their is no real need to make the static version less efficient I think.

flixr:
> The issue is named properly I think.

I think you mean NOT properly named...

gautierhattenberger:
yes indeed...

podhrmic:
@gautierhattenberger thanks for the idea, I haven't realized that. That should be sufficient for our application (at least for now). Just FYI I am adding a "geofencing" feature - essentially a sector and an exception checking whether you are inside the sector, the difference is that it is required and can't be omitted from the flight plan. Hence being more flexible with defining static sectors would be helpful.

Anyway, would you be interested in replacing/enhancing the distance_to_home with the "compulsory" flight area sector?

gautierhattenberger:
I think it can be useful when the system is meant to be used by end-users as a commercial product, but for developers or testers, I personally don't want to think too much about this before flying, the protection circle is good enough. So if you have something, maybe you can share it to discuss the implementation, but for now I would prefer an optional solution.

podhrmic:
OK, sounds good - I will push what I have hopefully in next few days.

gautierhattenberger:
I suggest to close this. Geofincing and non-convex sectors have already been improved.

podhrmic:
I agree. What do you mean by non-convex sectors improvements? Last time I checked only convex sectors worked for geofencing, but it turned out to be satisfactory for all known use-cases.

[...]

PR #1660, opened 2016-05-09
First attempt for geofencing

podhrmic:
Hi folks, this is related to #1616 - and is my first attempt for more "professional" / "industrial" geofencing feature.

What it does?
1) instead of HOME and MAX_DIST_FROM_HOME the user has to define FlightArea (convex polygon)
2) there is now a compulsory MAX_ALTITUDE attribute in the flight plan
3) if the plane is out of flight area or higher than max altitude, it goes to STANDBY mode
4) STANDBY mode is essentially a modified HOME mode - except instead of going to HOME it goes to STANDBY waypoint (name negotioable)
5) STANDBY waypoint cannot be moved
6) we want to check that STANDBY is defined inside flight area
7) there is also INIT mode of the autopilot, which is triggered upon startup - basically it tells you a reset happened, otherwise it is equivalent to AUTO2/NAV mode locked to circle around STANDBY waypoint.
8) If datalink is lost for more than 30s the plane goes to HOME/STANDBY mode (replaces an optional exception)
9) eventually kills the plane if more than 500 meters outside the flight area (until then try to steer it back)

I made small modifications in the OCaml code, but more guidance here is welcome (especially about how to check if a waypoint is inside flight area). This is more to support a use case for lets say industrial/professional users of pprz, who need better safety than max_dist_from_home circle.

let me know what you think.

Thanks.

### DIFF
--- a/sw/airborne/firmwares/fixedwing/autopilot.c
+++ b/sw/airborne/firmwares/fixedwing/autopilot.c
@@ -174,7 +174,7 @@ void autopilot_send_mode(void)
 
 void autopilot_init(void)
 {
-  pprz_mode = PPRZ_MODE_AUTO2;
+  pprz_mode = PPRZ_MODE_INIT;
   kill_throttle = false;
   launch = false;
   autopilot_flight_time = 0;
--- a/sw/airborne/firmwares/fixedwing/autopilot.h
+++ b/sw/airborne/firmwares/fixedwing/autopilot.h
@@ -50,9 +50,10 @@ extern void autopilot_init(void);
 #define  PPRZ_MODE_MANUAL 0
 #define  PPRZ_MODE_AUTO1 1
 #define  PPRZ_MODE_AUTO2 2
-#define  PPRZ_MODE_HOME 3
+#define  PPRZ_MODE_STDBY 3
 #define  PPRZ_MODE_GPS_OUT_OF_ORDER 4
-#define  PPRZ_MODE_NB 5
+#define  PPRZ_MODE_INIT 5
+#define  PPRZ_MODE_NB 6
 
 #define PPRZ_MODE_OF_PULSE(pprz) \
   (pprz > THRESHOLD2 ? PPRZ_MODE_AUTO2 : \
--- a/sw/airborne/firmwares/fixedwing/main_ap.c
+++ b/sw/airborne/firmwares/fixedwing/main_ap.c
@@ -321,7 +321,7 @@ void handle_periodic_tasks_ap(void)
 #if defined RADIO_CONTROL || defined RADIO_CONTROL_AUTO1
 static inline uint8_t pprz_mode_update(void)
 {
-  if ((pprz_mode != PPRZ_MODE_HOME &&
+  if ((pprz_mode != PPRZ_MODE_STDBY &&
        pprz_mode != PPRZ_MODE_GPS_OUT_OF_ORDER)
 #ifdef UNLOCKED_HOME_MODE
       || TRUE
@@ -397,9 +397,9 @@ static inline void telecommand_task(void)
   uint8_t really_lost = bit_is_set(fbw_state->status, STATUS_RADIO_REALLY_LOST) &&
     (pprz_mode == PPRZ_MODE_AUTO1 || pprz_mode == PPRZ_MODE_MANUAL);
 
-  if (pprz_mode != PPRZ_MODE_HOME && pprz_mode != PPRZ_MODE_GPS_OUT_OF_ORDER && launch) {
-    if (too_far_from_home) {
-      pprz_mode = PPRZ_MODE_HOME;
+  if (pprz_mode != PPRZ_MODE_STDBY && pprz_mode != PPRZ_MODE_GPS_OUT_OF_ORDER && launch) {
+    if (too_far_from_home || (datalink_time>30) || (GetPosAlt() > MAX_ALTITUDE)) {
+      pprz_mode = PPRZ_MODE_STDBY;
       mode_changed = true;
     }
     if (really_lost) {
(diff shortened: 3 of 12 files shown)


martinmm:
To me Paparazzi was always the industry standard for professionals :-)

Have not looked into the code but it sounds as if you turned existing optional features into mandatory ones. The good thing about the HOME mechanism is that it is simple and robust. It is easy to understand for a UAV pilot (just a circle) and the source code is easy to check and verify. There are no tricky "is in area" checks needed and you can move the HOME waypoint (which happens from time to time).

Maybe it would be good to have that features in a more easy to adapt way, like just define an optional DATALINK_LOST=30 to activate a fly back instead of adding an exception. But I would keep the simple HOME mechanism as it is.

rmeertens:
For point 9: can we have a mandatory define that sets the max distance from the flight area in meters? For some drones 500 meters is ok, in an urban environment 500 meters is way too far.

gautierhattenberger:
On the general principle, I don't agree to enforce safety things this way, but we could make thinks more configurable (and easier to use eventually).

In details:
1. It breaks all existing flight plans; it makes them a bit more complex to prepare for a limited benefit in the general case (I think); how local/global coordinates are handled ?
2. I think it is a good idea to add this as an option
3. What is the benefit to change the name HOME with STANDBY ? it does exactly the same
4. see point 3
5. I really don't want my STDBY point to be fixed; is it "locally" or "globally" fixed anyway ?
6. so was the HOME point already
7. it looked safer to init the autopilot with the failsafe values... which is also an indication that a reset happen, isn't it ?
8. not an option, we have projects where planes are flying several minutes (or even hours) without telemetry; but I agree we could have a build-in option to parametrize this more easily since it is a common exception in flight plans
9. much more difficult to check this for complex polygon than for a circle, an the distance outside the protection circle is already configurable, so it can be a fixed distance instead of being proportional to the radius

I have also seen that the fixed home radius is replaced a the variable `nav_radius`, which is I think pretty dangerous since it can be modified at run-time to something not flyable by the plane.

podhrmic:
Thanks for the feedback! The reason for introducing the flight area is that sometimes the circle too generous - imagine you are restricted to fly over a particular property / land that has a shape of a rectangle, or there is a restricted airspace nearby you don't want to go to. Or you are flying near obstacles such as buildings etc... Hence the need for a polygon instead of a circle. But I understand that for most users this makes no difference. Maybe some kind of configurable settings - a circle or a flight area would be a good start.

@gautierhattenberger Good point, I guess HOME and STANDBY does the same - the naming convention just came from our field practice. The only difference is that when you have a flight area, you want to check whether your HOME waypoint is _inside_ (the other waypoints don't matter I think). And you dont want the user to move it since they can move it potentially _outside_ of the flight area.

About the reset - in case the reset happens during flight, the init/nav mode will be better at keeping the plane in the air rather than failsafe (at least that is the idea).

@rmeertens hmm, that is a good point - I suppose you could define something like 0.05...

@martinmm essentially yes. I am not trying to have everyone to use these "stricter" and more cumbersome rules, but eventually (at least in US) a strict geofencing will be required by FAA, so it would be good to be prepared.

So in summary:

- optional MAX_ALTITUDE: Y
- add DATALINK_LOST time to make this exception easier to configure: Y
- an optional flight area instead of circle Y/N? 

Anything else?

gautierhattenberger:
About the optional flight area, there is already a lot of examples were one or several layers of fly-zones are used. The protection circle is just bigger than that. The only difference is that when the plane goes out of the zone, it says in nav mode and the rest of the flight is still controlled by the flight plan.
What kind of extra feature do you need then ? entering a specific mode when crossing a border ? This is probably something that can be done from the flight plan actually. How would you like to use/configure this feature ?

PS: if the AP restart in flight, your AHRS will be completely messed and maybe not able to align correctly anymore. So you can probably only do MANUAL mode, and starting from failsafe value is not so bad I think.

podhrmic:
The idea is that (at least in our use case) we want pprz user interaction to be more streamlined -- so it makes sense to hide as much exceptions/safety checks inside the code, rather than having it configurable in the flight plan.

I know it can be already configured, but the drawback of this approach is that you can just delete your carefully designed flight area and other exceptions from the flight plan (even accidentally), and then the code still compiles but won't behave the same (i.e. the plane is limited by the home circle only). Which is potentially dangerous if you are making adjustments in the flight plan just before the flight (which is often the case).

I would like to have some kind of compilation flag that either compiles ground segment tools in normal mode (like now), or in "safe" ("enhanced"? "streamlined"?) mode - which would make the flight area compulsory (basically only gen_flighplan.ml would change I suppose). But maybe that is too complicated?

PS: In our case this would be with an external INS (Vectornav) which should be able to align itself in-flight (although I haven't tested it) - so maybe that is a very special use-case that doesn't concern other users.

gautierhattenberger:
I guess it would be possible to pass some `-safe` option to some of the generators (maybe only gen_flightplan) in order to do some extra checks at compile time. I won't have the time to work on this thought, but it is never too late to learn ocaml !

I case you didn't noticed it (I forgot it myself), if you define a sector with the name `Airspace` it will be checked together with the protection circle (so no need of flight plan exception to enter home mode)
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/navigation/common_nav.c#L57

podhrmic:
@gautierhattenberger thanks for the pointer - I saw that Airspace define before, but haven't realized it could be used this way. So I fixed it, and together with the optional max_altitude and optional datalink_lost_time exception made a new pull request, see #1664 

That should be good for a while, the optional _safe_ version of the flight plan generators might come later. Closing this now, thanks again for feedback.

PR #1662, opened 2016-05-11
Stereocam sendarray fix

knmcguire:
I cleaned up the stereocam module a bit, but more importantly I fixed a issue with sending messages from from lisa-s -> uart -> stereoboard. First it skipped indicators of the start and ending of the messages, but with these added while loops it does not do that anymore. Any suggestions to make it even better are welcome ofcourse

### DIFF
--- a/sw/airborne/modules/stereocam/state2camera/state2camera.c
+++ b/sw/airborne/modules/stereocam/state2camera/state2camera.c
@@ -32,8 +32,14 @@ static int frame_number_sending = 0;
 float lastKnownHeight = 0.0;
 int pleaseResetOdroid = 0;
 
+#ifndef STATE2CAMERA_SEND_DATA_TYPE
+#define STATE2CAMERA_SEND_DATA_TYPE 0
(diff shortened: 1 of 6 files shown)


fvantienen:
Looks good 👍

sw/airborne/modules/stereocam/state2camera/state2camera.c
```diff
@@ -32,8 +32,14 @@ static int frame_number_sending = 0;
 float lastKnownHeight = 0.0;
 int pleaseResetOdroid = 0;
 
+#ifndef STATE2CAMERA_SEND_DATA_TYPE
```
flixr: A (doxygen) comment on what this define means would be nice...

knmcguire: Hi Felix,

I tried to look for pull request but Freek already accepted it I saw. I will add the comments on my next pull request, thanks for notifying me.


________________________________

PR #1663, opened 2016-05-11
Added opencv example

rmeertens:
Title says it all... simple sobel filter.

### DIFF
--- a/conf/modules/cv_opencvdemo.xml
+++ b/conf/modules/cv_opencvdemo.xml
@@ -16,9 +16,9 @@
     <file name="cv_opencvdemo.c"/>
     <file name="opencv_example.cpp"/>
 
-    <flag name="CXXFLAGS" value="I$(PAPARAZZI_HOME)/sw/ext/opencv_bebop/install/include"/>
+    <flag name="CXXFLAGS" value="I$(PAPARAZZI_SRC)/sw/ext/opencv_bebop/install/include"/>
 
-  	<flag name="LDFLAGS" value="L$(PAPARAZZI_HOME)/sw/ext/opencv_bebop/install/lib" />
+  	<flag name="LDFLAGS" value="L$(PAPARAZZI_SRC)/sw/ext/opencv_bebop/install/lib" />
 	<flag name="LDFLAGS" value="lopencv_imgcodecs" />
 	<flag name="LDFLAGS" value="lopencv_imgproc" />
 	<flag name="LDFLAGS" value="lopencv_core" />
--- a/conf/modules/cv_opencvdemo.xml
+++ b/conf/modules/cv_opencvdemo.xml
@@ -2,7 +2,11 @@
 
 <module name="cv_opencvdemo" dir="computer_vision">
   <doc>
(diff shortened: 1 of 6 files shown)


conf/modules/cv_opencvdemo.xml
```diff
@@ -0,0 +1,33 @@
...
+  </doc>
+  <header>
+    <file name="cv_opencvdemo.h"/>
+  </header>
+  <init fun="opencvdemo_init()"/>
+  <makefile>
+    <file name="cv_opencvdemo.c"/>
+    <file name="opencv_example.cpp"/>
+
+    <flag name="CXXFLAGS" value="I$(PAPARAZZI_HOME)/sw/ext/opencv_bebop/install/include"/>
```
gautierhattenberger: this submodule is not part of paparazzi/master yet

rmeertens: Ok. will add it tomorrow!

rmeertens:
Ok, added the submodule

flixr:
Please add a useful description in the module xml, especially for a demo module this is quite important (what does it do? what is it intended for? how are you supposed to use/modify/extend this?)

flixr:
One more general question: do you guys like the C++ OpenCV interface better than the C interface or what was the reason for not using the C interface (since Paparazzi itself is C)?

rmeertens:
We like the C++ interface better. One huge advantage that we see right now is the fact that students can try something on their laptop, and then immediately try it on their quadcopter.

wilcoschoneveld:
Also, I believe the C api has been dropped since OpenCV 1.x. I can't find a proper source right now other than http://answers.opencv.org/question/17546/opencv-will-drop-c-api-support-soon/ (2013).

flixr:
Was just curious, I'm not against C++ ;-)

rmeertens:
> Was just curious, I'm not against C++ ;-)

Whoo, road open to more C++ in Paparazzi!

flixr:
Re C++ support see also https://github.com/paparazzi/paparazzi/pull/1561

rmeertens:
Nice

rmeertens:
I think the description will now work out! 
Maybe I will add a more elaborate example using more OpenCV functions in a later pull request.

conf/modules/cv_opencvdemo.xml
flixr: Also the path should use PAPARAZZI_SRC instead of PAPARAZZI_HOME

sw/airborne/modules/computer_vision/opencv_example.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+using namespace std;
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+using namespace cv;
+
+
+int opencv_example(char* img, int width, int height)
+{
+	// Create a new image, using the original bebop image.
```
flixr: indentation...

rmeertens:
Ok, fixed!

PR #1664, opened 2016-05-11
Updated safety behaviors

podhrmic:
Based on the discussion in #1660 I made these changes:

1. fixed InsideAirspace check - so now it is easy to use a sector together with the home circle
2. optional max_altitude in the flight plan
3. optional DATALINK_LOST_TIME exception

### DIFF
--- a/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
@@ -35,7 +35,7 @@ RP3 Lisa MX
 
     <define name="AGR_CLIMB" />
     <define name="POLY_OSAM_HALF_SWEEP_ENABLED" value="FALSE"/>
-    <define name="HOME_MODE_DATALINK_LOST_TIME" value="45"/>
+    <define name="GEOFENCE_DATALINK_LOST_TIME" value="45"/>
   </firmware>
 
   <modules>
--- a/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
@@ -35,7 +35,6 @@ RP3 Lisa MX
 
     <define name="AGR_CLIMB" />
     <define name="POLY_OSAM_HALF_SWEEP_ENABLED" value="FALSE"/>
-    <define name="AIRSPACE"/>
     <define name="DATALINK_LOST_TIME" value="45"/>
   </firmware>
 
--- a/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
@@ -35,6 +35,8 @@ RP3 Lisa MX
 
     <define name="AGR_CLIMB" />
     <define name="POLY_OSAM_HALF_SWEEP_ENABLED" value="FALSE"/>
+    <define name="AIRSPACE"/>
+    <define name="DATALINK_LOST_TIME" value="45"/>
   </firmware>
 
   <modules>
--- a/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
@@ -35,7 +35,7 @@ RP3 Lisa MX
 
     <define name="AGR_CLIMB" />
     <define name="POLY_OSAM_HALF_SWEEP_ENABLED" value="FALSE"/>
-    <define name="DATALINK_LOST_TIME" value="45"/>
+    <define name="HOME_MODE_DATALINK_LOST_TIME" value="45"/>
   </firmware>
 
   <modules>
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -66,7 +66,8 @@ ground_alt CDATA #REQUIRED
 security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
 qfu CDATA #IMPLIED
-home_mode_height CDATA #IMPLIED>
+home_mode_height CDATA #IMPLIED
+max_altitude CDATA #IMPLIED>
 
 <!ATTLIST waypoints
 utm_x0 CDATA #IMPLIED
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -67,7 +67,8 @@ security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
 qfu CDATA #IMPLIED
 home_mode_height CDATA #IMPLIED
-max_altitude CDATA #IMPLIED>
+max_altitude CDATA #IMPLIED
+airspace CDATA #IMPLIED>
 
 <!ATTLIST waypoints
 utm_x0 CDATA #IMPLIED
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -67,8 +67,8 @@ security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
 qfu CDATA #IMPLIED
 home_mode_height CDATA #IMPLIED
-max_altitude CDATA #IMPLIED
-airspace CDATA #IMPLIED>
+home_mode_max_alt CDATA #IMPLIED
+home_mode_sector CDATA #IMPLIED>
 
 <!ATTLIST waypoints
 utm_x0 CDATA #IMPLIED
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -67,8 +67,8 @@ security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
 qfu CDATA #IMPLIED
 home_mode_height CDATA #IMPLIED
-home_mode_max_alt CDATA #IMPLIED
-home_mode_sector CDATA #IMPLIED>
+geofence_max_alt CDATA #IMPLIED
+geofence_sector CDATA #IMPLIED>
 
 <!ATTLIST waypoints
 utm_x0 CDATA #IMPLIED
--- a/sw/airborne/firmwares/fixedwing/main_ap.c
+++ b/sw/airborne/firmwares/fixedwing/main_ap.c
@@ -108,6 +108,7 @@ PRINT_CONFIG_MSG_VALUE("USE_BARO_BOARD is TRUE, reading onboard baro: ", BARO_BO
 #define COMMAND_YAW_TRIM 0
 #endif
 
+bool DatalinkLost(void);
 
 #ifdef DATALINK_LOST_TIME
 /*
@@ -121,8 +122,9 @@ bool DatalinkLost(void){
 bool DatalinkLost(void){
   return false;
 }
-#endif
+#endif /* DATALINK_LOST_TIME */
 
+bool HigherThanMaxAltitude(void);
 #ifdef MAX_ALTITUDE // user defined max_altitude in the flight plan
 bool HigherThanMaxAltitude(void){
   return (GetPosAlt() > MAX_ALTITUDE);
--- a/sw/airborne/firmwares/fixedwing/main_ap.c
+++ b/sw/airborne/firmwares/fixedwing/main_ap.c
@@ -108,7 +108,8 @@ PRINT_CONFIG_MSG_VALUE("USE_BARO_BOARD is TRUE, reading onboard baro: ", BARO_BO
 #define COMMAND_YAW_TRIM 0
 #endif
 
-#include "modules/nav/nav_home_mode.h"
+/* Geofence exceptions */
+#include "modules/nav/nav_geofence.h"
 
 /* if PRINT_CONFIG is defined, print some config options */
 PRINT_CONFIG_VAR(PERIODIC_FREQUENCY)
--- a/sw/airborne/firmwares/fixedwing/main_ap.c
+++ b/sw/airborne/firmwares/fixedwing/main_ap.c
@@ -108,6 +108,11 @@ PRINT_CONFIG_MSG_VALUE("USE_BARO_BOARD is TRUE, reading onboard baro: ", BARO_BO
 #define COMMAND_YAW_TRIM 0
 #endif
 
+/* Datalink lost time exception [s] */
+#ifndef DATALINK_LOST_TIME
+#define DATALINK_LOST_TIME 30
+#endif
+
 /* if PRINT_CONFIG is defined, print some config options */
 PRINT_CONFIG_VAR(PERIODIC_FREQUENCY)
 PRINT_CONFIG_VAR(NAVIGATION_FREQUENCY)
@@ -398,13 +403,13 @@ static inline void telecommand_task(void)
     (pprz_mode == PPRZ_MODE_AUTO1 || pprz_mode == PPRZ_MODE_MANUAL);
 
   if (pprz_mode != PPRZ_MODE_HOME && pprz_mode != PPRZ_MODE_GPS_OUT_OF_ORDER && launch) {
-    if (too_far_from_home) {
+    if (too_far_from_home || (datalink_time>DATALINK_LOST_TIME) || (GetPosAlt() > MAX_ALTITUDE)) {
       pprz_mode = PPRZ_MODE_HOME;
-      mode_changed = true;
+      mode_changed = TRUE;
     }
     if (really_lost) {
       pprz_mode = RC_LOST_MODE;
-      mode_changed = true;
+      mode_changed = TRUE;
     }
   }
   if (bit_is_set(fbw_state->status, AVERAGED_CHANNELS_SENT)) {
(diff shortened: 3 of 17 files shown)


sw/airborne/subsystems/navigation/common_nav.c
```diff
@@ -54,8 +54,8 @@ void compute_dist2_to_home(void)
   float ph_y = waypoints[WP_HOME].y - pos->y;
   dist2_to_home = ph_x * ph_x + ph_y * ph_y;
   too_far_from_home = dist2_to_home > (MAX_DIST_FROM_HOME * MAX_DIST_FROM_HOME);
-#if defined InAirspace
-  too_far_from_home = too_far_from_home || !(InAirspace(pos_x, pos_y));
+#if defined AIRSPACE
```
fvantienen: Can't you just check `InsideAirspace`, because then you don't need to define this in your airframe file.

sw/tools/generators/gen_flight_plan.ml
```diff
@@ -920,6 +920,11 @@ let () =
       and alt = ExtXml.attrib xml "alt" in
       security_height := get_float "security_height";
       ground_alt := get_float "ground_alt";
+      (* Absolute Max altitude that can be reached *)
+      let max_alt = 30000.0 in
```
fvantienen: Why is there an absolute maximum, and why is it `30000.0`?

sw/airborne/firmwares/fixedwing/main_ap.c
```diff
@@ -398,13 +403,13 @@ static inline void telecommand_task(void)
     (pprz_mode == PPRZ_MODE_AUTO1 || pprz_mode == PPRZ_MODE_MANUAL);
 
   if (pprz_mode != PPRZ_MODE_HOME && pprz_mode != PPRZ_MODE_GPS_OUT_OF_ORDER && launch) {
-    if (too_far_from_home) {
+    if (too_far_from_home || (datalink_time>DATALINK_LOST_TIME) || (GetPosAlt() > MAX_ALTITUDE)) {
       pprz_mode = PPRZ_MODE_HOME;
-      mode_changed = true;
+      mode_changed = TRUE;
```
fvantienen: should be small letters `true`, since we changed the convention a while ago ;)

sw/airborne/firmwares/fixedwing/main_ap.c
```diff
@@ -108,6 +108,11 @@ PRINT_CONFIG_MSG_VALUE("USE_BARO_BOARD is TRUE, reading onboard baro: ", BARO_BO
 #define COMMAND_YAW_TRIM 0
 #endif
 
+/* Datalink lost time exception [s] */
+#ifndef DATALINK_LOST_TIME
+#define DATALINK_LOST_TIME 65534 // 45 days - i.e. no exception
```
fvantienen: I don't like these default values.. Not that we ever reach this number, but it would be better to just add the `#ifndef` around the `if` below.

conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
```diff
@@ -35,6 +35,8 @@ RP3 Lisa MX
 
     <define name="AGR_CLIMB" />
     <define name="POLY_OSAM_HALF_SWEEP_ENABLED" value="FALSE"/>
+    <define name="AIRSPACE"/>
+    <define name="DATALINK_LOST_TIME" value="45"/>
```
fvantienen: Since there is already a check of datalink loss I think this naming should be more clear that it is for the home mode.

sw/airborne/subsystems/navigation/common_nav.c
podhrmic: My bad, I haven't noticed how it is generated. Fixed in new commit.

sw/airborne/firmwares/fixedwing/main_ap.c
podhrmic: Fixed.

sw/airborne/firmwares/fixedwing/main_ap.c
podhrmic: Fixed.

conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
podhrmic: There is an exception that can be defined in the flight plan, this is if you want to enforce datalink check independent on the flight plan. Should be more clear with the updated commit.

conf/flight_plans/flight_plan.dtd
```diff
@@ -65,7 +65,9 @@ ground_alt CDATA #REQUIRED
 security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
 qfu CDATA #IMPLIED
-home_mode_height CDATA #IMPLIED>
+home_mode_height CDATA #IMPLIED
+max_altitude CDATA #IMPLIED
+airspace CDATA #IMPLIED>
```
gautierhattenberger: This attribute doesn't seem to be used at the moment. Also I would suggest `home_mode_max_alt` to make it clear that going higher will make the plane enter HOME mode, and `home_mode_sector` instead of airspace (if I get it correctly, this was suppose to specify the actual sector used for home protection).

conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
gautierhattenberger: Even if "controlled" from the airframe file, it is still a setting that controls when the aircraft will enter HOME mode, so HOME_MODE_DATALINK_LOST_TIME is more appropriate I think.
Although it is nice to make it configurable from an airframe independently of the flight plan, I think it can also be useful to be able to specify it in a flight plan, so it works even if not defined in airframe. If nothing defined -> no check, if defined in airframe and/or FP, use FP unless airframe already defines it.
What do you think ?

conf/flight_plans/flight_plan.dtd
podhrmic: Renamed as you suggested. If you mean `home_mode_height` that one is used in https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/fixedwing/nav.c#L429

conf/airframes/AGGIEAIR/aggieair_rp3_lia.xml
podhrmic: I am thinking about the HOME_MODE_DATALINK_LOST_TIME  as a safety feature you want to bound to the airframe, rather than a particular flight plan. In which case you can still define an exception in the flight plan (with different values for different flights lets say), but if you define HOME_MODE_DATALINK_LOST_TIME in the airfame config file then you will have sort two independent checks. So I don't think it makese sense to define this in the flight plan (since you can use an exception to do the same, and this was kind of a way to avoid using exceptions in the first place). Does that make sense?

flixr:
Since these additions are the same for fixedwing and rotorcraft, I think they should be added to a header that is included in both...

naming convention wise, functions are usually also under_scored...

podhrmic:
Unified in one file - this way it should be way cleaner.

sw/airborne/modules/nav/nav_home_mode.h
```diff
@@ -0,0 +1,59 @@
...
+
+bool datalink_lost(void);
+bool higher_than_max_altitude(void);
+
+#ifdef HOME_MODE_DATALINK_LOST_TIME
+/*
+ * from the airfame config file:
+ * go to HOME mode if datalink lost for HOME_MODE_DATALINK_LOST_TIME
+ */
+bool datalink_lost(void){
```
flixr: since these are only in the header file it would make sense to make them ```static inline```

sw/airborne/modules/nav/nav_home_mode.h
```diff
@@ -0,0 +1,59 @@
...
+  return (datalink_time>HOME_MODE_DATALINK_LOST_TIME);
```
flixr: plz format according to our style guide, i.e. use the fix_code_style.sh script on this file

podhrmic:
Done!

podhrmic:
Should I merge this then?

flixr:
If I see that correctly, the Airspace/home_mode_sector is missing for rotorcrafts...

Also regarding the naming: we shoud choose good/descriptive names here. Are home_mode_max_alt and home_mode_sector really good names?
E.g. when I read `home_mode_max_alt` I think of the max altitude when already are in home mode (analogous to `home_mode_alt`)...
Maybe something like `geofence_max_alt` and `geofence_sector` would be easier to understand for more people?
Or whatever... but I think finding good names is important here...

flixr:
Does this also work with dynamic sectors?

gautierhattenberger:
@flixr I actually suggest `home_mode_max_alt` to clearly state the this is related to the HOME mode, same for sector.
@podhrmic unfortunately `common_nav` is not so common yet, you should modify `firmwares/rotorcraft/navigation.c` as well.

podhrmic:
> Does this also work with dynamic sectors?

@flixr Yes, it works with any sector. It just makes a macro that utilizes _InsideSectorName_ autogenerated functions.

> Maybe something like geofence_max_alt and geofence_sector would be easier to understand for more people?

Naming - I don't have a strong opinion about the name, should I rename it then?

>  unfortunately common_nav is not so common yet, you should modify firmwares/rotorcraft/navigation.c as well.

@gautierhattenberger OK.

podhrmic:
Added the insideHomeSector check for rotorcraft too. @flixr are you cool with the names as they are? `home_mode_max_alt` and `home_mode_sector` ? If so I think we are good to merge...

flixr:
Hm.. I think the naming is not great and changing it later is a pain.
From my perspective, the specification of the geofence(flight area sector) limits and the action to take when you breach that are two separate things. Now a breach means going to HOME mode, but what if we want to make the action configurable in the future? Then having names like `home_mode_max_alt` and `home_mode_sector` will not make that much sense...

Some other things:
- I would keep the rotorcraft_basic simple i.e. without the extra sector and rather make another example with this extra geofence (similar to rotorcraft_basic_safety)
- In the generator, the max_altitude should be checked to be above the ground_alt + security_height
- max_alt is a fixed altitude, so if you take your standard flight plan and go to another (higher) location and run geo_init, all (local/relative) waypoint altitudes will be adjusted accordingly. The max_alt is not adjusted though and could even be lower than ground_alt now!

flixr:
Regarding the last point, maybe it would be nicer to add a `max_height` instead of `max_alt`?

podhrmic:
I change the name to `geofence_xxx` which should better reflect the purpose.
Also:

> I would keep the rotorcraft_basic simple i.e. without the extra sector and rather make another example with this extra geofence (similar to rotorcraft_basic_safety)

Done

> In the generator, the max_altitude should be checked to be above the ground_alt + security_height

Done

> max_alt is a fixed altitude, so if you take your standard flight plan and go to another (higher) location and run geo_init, all (local/relative) waypoint altitudes will be adjusted accordingly. The max_alt is not adjusted though and could even be lower than ground_alt now!

So here I don't really know how to approach it. First, I added optional `geofence_max_agl` together with `geofence_max_alt` - in one case you specify max altitude (absolute), in other case you use AGL (relative). Unfortunately fixedwing and rotorcraft have different navigation interface - for example rotorcraft doesn't have `ground_alt` variable, instead it uses `GetAltRef()`. So it is not possible to have the same check for both fixedwing and rotorcraft. I could split the `nav/nav_geofence.h` into two files - one for fixedwing and one for rotocraft, but that is exactly the opposite of what we want. 

So instead I added a warning that the absolute max altitude is used, so if you change a location you should update the `geofence_max_alt` accordingly. If in the future we will have a unified navigation interface then I can revisit the idea and have the max altitude updated with new ground altitude.

flixr:
> So here I don't really know how to approach it. First, I added optional geofence_max_agl together with geofence_max_alt - in one case you specify max altitude (absolute), in other case you use AGL (relative). Unfortunately fixedwing and rotorcraft have different navigation interface - for example rotorcraft doesn't have ground_alt variable, instead it uses GetAltRef(). So it is not possible to have the same check for both fixedwing and rotorcraft. I could split the nav/nav_geofence.h into two files - one for fixedwing and one for rotocraft, but that is exactly the opposite of what we want.

Using `GetAltRef()` should work for both, that is why it was introduced after all...

flixr:
Hm... you already merged it... I would have preferred to have a consensus on how we do that first...

flixr:
IMHO it would be nicer to introduce a `geofence_max_height` instead of the `geofence_max_alt`... that should be easier and more natural to use an not have the problem with GeoInit.
@gautierhattenberger @dewagter @martinmm what do you guys think?

dewagter:
I think waypoints (or any other place storing height) should have an enum
attribute storing their reference type: gps-geiod, gps-ellipsiod, baro-msl,
baro-qfe, flight-level (baro  standard atmosphere), sonar ...
Conversions are easy. Problems only occur when you forget how they were
originally defined, and then convert them. If you store the reference I
believe all problems like geo-init, flight-plan alt, geofences etc are
easily solved.

But sticking to the current setup and question:

First: I thought only height had geoinit problems, not alt.

I would prefer to be able to enter both absolute or relative limits on z,
but having to pick one, an absolute AMSL altitude limit (not changing with
geoinit) makes most sense to me in case of geo-fences.

flixr:
Attaching an enum that defines the reference for each waypoint would indeed be nice in a lot of cases, but wouldn't change anything for this issue...

> First: I thought only height had geoinit problems, not alt.

I'm not sure what you mean by "problems"...

> I would prefer to be able to enter both absolute or relative limits on z

We could of course also provide both...

> but having to pick one, an absolute AMSL altitude limit (not changing with geoinit) makes most sense to me in case of geo-fences.

I would have thought that in most cases people would rather want to restrict how high you can fly relative to your takeoff altitude (ground_alt).

Right now the `geofence_max_altitude` is really fixed at compile time. While the waypoint altitude (and ground alt) can change after resetting your ground reference (e.g. in GeoInit).
Consider this case:
- flightplan: `<flight_plan alt="120" ground_alt="100" geofence_max_alt="130">`
- now if you go to another location that is at e.g. 120m altitude and do a GeoInit:
- ground_alt will now be at 120m, waypoint alt at 140m
- but the `geofence_max_alt` is still at 120m and it will trigger immediately!! And you can't change it apart from changing the flight plan and recompiling!

So to me it seems that in most cases it would make more sense to rather specify a `geofence_max_height` that would "adapt" to the takeoff altitude (ground_alt)....

This is even already wrong from the start for the example: [rotorcraft_basic_geofence.xml](https://github.com/podhrmic/paparazzi/blob/0742427cf52b3aa395762c37d38ca09d2d2acb56/conf/flight_plans/rotorcraft_basic_geofence.xml#L3)
Here the `geofence_max_alt` is 30m, while the `flight_plan_alt` and even `ground_alt` is higher...
As I mentioned before we should already catch this during code generation...

dewagter:
I agree that an AGL limit is easy and very useful. It would be important in
several cases to also be able to set absolute limits (those are on aviation
maps).

IMHO, an enum for ALL altitudes/heights/z variables (not just waypoints)
would also help this issue: if you define geofence_max_vert="1200ft MSL" as
limit, it should not change with geo-init, while if you define
geofence_max_vert="400ft
AGL" then it is only applicable after geo-init. If you forget the first was
MSL and convert it to AGL using the flightplan ground height and then
geo-init you end up with ... euh... 1200ft MSL, while FP ground as at 300m,
that is 984ft, yields 216ft above flightplan ground, and geo-init occured
at the top of the hill at 345m, so the 216ft AGL are added to ... you still
follow?

I more and more believe that converting alt/height to a vertical-struct
with its reference and an XML parser for vertical (like with GPS positions
ecef/llh/ltp/... in state.h) and converting on the go would solve all
currently known problems and confusion.

struct vertical_struct {
   int32_t vert;

flixr:
Pull requests implementing all this welcome!

But that still means we should implement a `geofence_max_height` right?
Also do we want to be able to enable/disable or change the geofence at runtime?

flixr:
@dewagter also feel free to create an issue/feature request about this where we can track and discuss it separate from the geofence functionality.

podhrmic:
Hi, sorry I've been busy last few days. The reason I went with max_alt initially comes from our typical use case - i.e. if you want to fly somewhere, you need to get a Ceritficate Of Authorization (COA) from FAA, and the COA is issued to a specific place with specific max altitude (and an area) you can fly at. So you can't legally take your UAV and fly it somewhere else, unless you have a new COA. As a result, we have flight plans that are for a specific place only. 

In my understanding, other users using geofencing should (at least in theory) follow a similar practice and then having max_alt is probably more useful than max_agl.

Also (but that is a separate issue) we are not using geo_init in the flight plan, all our waypoints have absolute position.

I ll add a possibility to specify either `max_alt`  or `max_agl` in the flight plan - if you use:
1) `max_alt` - it will be used as is
2) `max_agl` - it will add max_agl to the ground altitude as specified in the flight plan
3) both - there will be an error, you need to use only one
In all cases there are checks at the compile time, but the final max_altitude value will be hard-coded, so you can't change it with geo_init

That should be the most useful. As @flixr said, pull requests implementing the alt struct are welcome, but should be in a separate PR.

dewagter:
Will investigate the implications of a change in https://github.com/paparazzi/paparazzi/pull/1700.
Ideas/comments are welcome.

flixr:
@podhrmic if we indeed introduce a `max_height` (I would rather call it that than `max_agl` since it is not really always above agl, but rather above the set reference), it might be nice to allow both `max_height` and `max_alt` and trigger as soon as one of the two limits is reached?

Another option might be to allow changing geofence_max_alt at runtime...

PR #1665, opened 2016-05-12
[modules] New magneto and pitot

fvantienen:
This adds the functionality for a new type of Magneto + pitot. The module transmits PPRZ messages which contain magneto and airspeed information.

### DIFF
--- a/conf/modules/mag_pitot_uart.xml
+++ b/conf/modules/mag_pitot_uart.xml
@@ -3,6 +3,11 @@
 <module name="mag_pitot_uart" dir="sensors">
   <doc>
     <description>Remotely located magnetometer and pitot tube over uart (RS232) communication</description>
+    <configure name="MAG_PITOT_PORT" value="UART5" description="select which uart it is connected to"/>
+    <configure name="MAG_PITOT_BAUD" value="125000" description="set the baudrate of the uart"/>
+    <define name="IMU_TO_MAG_PHI" value="0" description="rotation of magneto compared to the magneto in roll axis"/>
+    <define name="IMU_TO_MAG_THETA" value="0" description="rotation of magneto compared to the magneto in pitch axis"/>
+    <define name="IMU_TO_MAG_PSI" value="0" description="rotation of magneto compared to the magneto in yaw axis"/>
   </doc>
   <header>
     <file name="mag_pitot_uart.h"/>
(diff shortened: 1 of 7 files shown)


conf/modules/mag_pitot_uart.xml
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="mag_pitot_uart" dir="sensors">
+  <doc>
+    <description>Remotely located magnetometer and pitot tube over uart (RS232) communication</description>
```
flixr: missing description of the configure options...

conf/modules/mag_pitot_uart.xml
```diff
@@ -0,0 +1,22 @@
...
+  <header>
+    <file name="mag_pitot_uart.h"/>
+  </header>
+  <init fun="mag_pitot_init()"/>
+  <event fun="mag_pitot_event()"/>
+  <makefile>
+    <configure name="MAG_PITOT_PORT" default="uart5" case="upper|lower"/>
+    <define name="USE_$(MAG_PITOT_PORT_UPPER)"/>
+    <define name="USE_$(MAG_PITOT_PORT_UPPER)_TX" value="FALSE"/>
+    <define name="$(MAG_PITOT_PORT_UPPER)_BAUD" value="$(MAG_PITOT_BAUD)"/>
```
flixr: missing configure default for baud rate

sw/airborne/modules/sensors/mag_pitot_uart.c
```diff
@@ -0,0 +1,132 @@
...
+  /* Parse the mag-pitot message */
+  uint8_t msg_id = mp_msg_buf[1];
+  switch (msg_id) {
+
+  /* Got a magneto message */
+  case DL_IMCU_REMOTE_MAG: {
+    struct Int32Vect3 raw_mag;
+
+    // Read the raw magneto information
+    raw_mag.y = - DL_IMCU_REMOTE_MAG_mag_x(mp_msg_buf);
```
flixr: hardcoded sign conventions??
Isn't that what imu_to_mag would be for?

sw/airborne/modules/sensors/mag_pitot_uart.c
```diff
@@ -0,0 +1,132 @@
...
+#endif
+
+/* Initialize the magneto and pitot */
+void mag_pitot_init() {
+  // Initialize transport protocol
+  pprz_transport_init(&mag_pitot.transport);
+
+  // Set the Imu to Magneto rotation
+  struct FloatEulers imu_to_mag_eulers =
+    {IMU_TO_MAG_PHI, IMU_TO_MAG_THETA, IMU_TO_MAG_PSI};
```
flixr: IMU_TO_MAG missing in the documentation

fvantienen:
Fixed the comments ;)
One note on the rotation is that it cannot change the axis system of the magneto, but found a `MAG_*_SIGN` define which can solve this.

flixr:
IMHO it makes much more sense to ensure a correct right handed magneto coordinate system on the sender side.
But setting mag signs (which is applied when scaling) also works...

flixr:
Do you guys have the hw/sw for this external module available as well?
Plan to add it in paparazzi? Or maybe link to where the hw/sw for this external module can be found....

fvantienen:
I think this could be released as soon as we tested it fully. Currently only done some static testing.
We made both the HW and SW, and it is pretty trivial so @dewagter will prolly release both.

Indeed the axis system could better be fixed on the sensor side, but for now I keep it like this and as soon as we update the firmware we will update the airframe file 👍

dewagter:
Sure: you can have the external magneto sensor hardware and firmware (using
ppzlink on an atmega).
Should I put it on the hardware git?

flixr:
That would be great.

PR #1666, opened 2016-05-12
[modules] Add RPM sensor

fvantienen:
Simple RPM sensor which measures the time between two pulses.


gautierhattenberger:
Would have been nicer to use the `pwm_meas` module (reuse existing code, makes it work for lpc as well).
You can find example in `aoa_pwm` module.

fvantienen:
Didn't knew that existed.. but I looked a bit at the code and indeed it is partly the same, except for the fact that this RPM module doesn't have an PWM input.
It measures the time between 2 rising or falling edges, instead of the time between a rising and falling edge. So you do not need to measure the duty cycle for the RPM.

gautierhattenberger:
The module provides both duty and period, for instance the `meteo_stick` module is reading a frequency with only the period : https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/meteo/meteo_stick.c#L301

fvantienen:
Is this better @gautierhattenberger ?

PR #1669, opened 2016-05-16
Skid landing take #2

podhrmic:
Fixes from  #1656
I renumbered (sequentially) the conf examples so it is easier to add more examples in the future (I had a hard time finding an unused ID).

### DIFF
--- a/conf/modules/nav_skid_landing.xml
+++ b/conf/modules/nav_skid_landing.xml
@@ -1,0 +1,34 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="nav_skid_landing" dir="nav">
+  <doc>
+    <description>
+    Landing on skidpads
+    See video of the landing: https://www.youtube.com/watch?v=aYrB7s3oeX4
+    Standard landing procedure:
+    1) circle down passing AF waypoint (from left or right)
+    2) once low enough follow line to TD waypoint
+    3) once low enough flare
+
+   Use this in your airfame config file:
+<!--
+Block example:
+@verbatim
+     <section name="LANDING" prefix="SKID_LANDING_">
+       <define name="AF_HEIGHT" value="50" unit="m"/>
+       <define name="FINAL_HEIGHT" value="50" unit="m"/>
+       <define name="FINAL_STAGE_TIME" value="10" unit="s"/>
+     </section>
+</block>
+@endverbatim
+-->
+    </description>
+  </doc>
+  <header>
+    <file name="nav_skid_landing.h"/>
+  </header>
+  <makefile target="ap|sim|nps">
+    <file name="nav_skid_landing.c"/>
+  </makefile>
+</module>
+
--- a/conf/modules/nav_skid_landing.xml
+++ b/conf/modules/nav_skid_landing.xml
@@ -19,16 +19,54 @@ Block example:
        <define name="FINAL_HEIGHT" value="50" unit="m"/>
        <define name="FINAL_STAGE_TIME" value="10" unit="s"/>
      </section>
-</block>
 @endverbatim
 -->
+
+Also define 
+<!--
+Block example:
+@verbatim:
+  <section name="VERTICAL CONTROL" prefix="V_CTL_">
+    <define name="LANDING_THROTTLE_PGAIN"           value="600.0"/>
+    <define name="LANDING_THROTTLE_IGAIN"           value="10.0"/>
+    <define name="LANDING_THROTTLE_MAX"           value="0.65"/>
+    <define name="LANDING_DESIRED_SPEED"           value="18" unit="m/s"/>
+    <define name="LANDING_PITCH_PGAIN"           value="0.1"/>
+    <define name="LANDING_PITCH_IGAIN"           value="0.1"/>
+    <define name="LANDING_PITCH_LIMITS"           value="0.2" unit="rad"/>
+    <define name="LANDING_PITCH_FLARE"           value="0.06"/>
+    <define name="LANDING_ALT_THROTTLE_KILL"     value="15" unit="m"/>
+    <define name="LANDING_ALT_FLARE"           value="5" unit="m"/>
+  </section>
+@endverbatim
+-->
+to properly use landing control loop
     </description>
   </doc>
+
+  <settings>
+    <dl_settings>
+      <dl_settings NAME="NavSkidLanding">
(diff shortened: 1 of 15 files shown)


sw/airborne/modules/nav/nav_skid_landing.c
```diff
@@ -0,0 +1,199 @@
...
+#ifndef SKID_LANDING_FINAL_HEIGHT //> final height [m]
+#define SKID_LANDING_FINAL_HEIGHT 5
+#endif
+#ifndef SKID_LANDING_FINAL_STAGE_TIME //> final stage time [s]
+#define SKID_LANDING_FINAL_STAGE_TIME 5
+#endif
+
+static inline float distance_equation(struct FloatVect2 p1,struct FloatVect2 p2)
+{
+  return sqrt((p1.x-p2.x)*(p1.x-p2.x)+(p1.y-p2.y)*(p1.y-p2.y));
```
flixr: for float values use `sqrtf` or even directly something like
```
struct FloatVect2 diff;
VECT2_DIFF(diff, p1, p2);
return float_vect2_norm(diff);
```
or even add a distance function in math?

flixr:
Please add the new defines in the module xml documentation section

flixr:
Like @gautierhattenberger I don't really like adding new modes like this, but currently don't see any nice alternative...

flixr:
regarding renumbering the AC_IDs in conf_tests: I think it is good to keep the same AC_IDs for the aircrafts as in conf_example

flixr:
Also could you please describe what the benefits of this skid landing module are compared to the usual flight plan flare landing (easier usage is great, but what are the behavior/performance differences)?

sw/airborne/modules/nav/nav_skid_landing.c
podhrmic: Changed to sqrtf

podhrmic:
@flixr to your comments:

> Please add the new defines in the module xml documentation section

it is already there? https://github.com/paparazzi/paparazzi/pull/1669/files#diff-8fa01c82a9a3fc563b913bc4d4ebb5b2
I am trying to group the defines in a section (i.e. ` <section name="LANDING" prefix="SKID_LANDING_">` so it is more organized. How else do you want it?

 

> regarding renumbering the AC_IDs in conf_tests: I think it is good to keep the same AC_IDs for the aircrafts as in conf_example

Should I renumber the ACs in conf_example too? There is no reason for having them randomly numbered - plus now you at least know what IDs are free so you can add more examples/tests (like I did).

> Also could you please describe what the benefits of this skid landing module are compared to the usual flight plan flare landing (easier usage is great, but what are the behavior/performance differences)?

Let me try:

1. It is an alternative to the existing landing procedure that is defined in the flight plan (for example [here](https://github.com/paparazzi/paparazzi/blob/master/conf/flight_plans/versatile.xml#L126)
2. it is easier to use and gives you consistent behavior - because you only call setup/run fucntions instead of defining the landing procedure in the flight plan (so you can't make a mistake there)
3. It has been used for many landings during various missions so it is field proven.
4. I think it behaves similarly (i.e. circle down, follow a line, flare), it is just nicely wrapped in a module.

flixr:
Regarding the docs for defines: since you commented that part it will not be part of the generated docs.
If you add the `<section>` to the doc node (and add a description for each define) it will also be part of the generated docs, like e.g. in http://docs.paparazziuav.org/latest/module__optical_flow_landing.html

> Should I renumber the ACs in conf_example too?

IMHO that would be preferable to only renumbering them in one...

Thanks for the explanation, it's just that from your previous comment "improved flare" sounded like it does it a bit better/differently...

flixr:
Regarding the V_CTL_LANDING gains: they should probably default to the corresponding V_CTL_AUTO gains?

podhrmic:
> Regarding the V_CTL_LANDING gains: they should probably default to the corresponding V_CTL_AUTO gains?

Some gains can be used from V_CTL_AUTO gains, but what about for example V_CTL_LANDING_DESIRED_SPEED or V_CTL_LANDING_ALT_FLARE? That has no counterpart in the V_CTL_AUTO world - and since it heavily depends on the airframe used there seem to be no good default values.

Should I 

1. define them as zero which will make the V_CTL_LANDING not really useful unless you define your own gains (and have a warning or so printed in such case)>
2. keep them on some default value that is probably not great, but at least it will somewhat work with default gains (and again print a warning)
3. dont define them, so if the user doesn't define the proper gains the code wont compile?

flixr:
I didn't look at the landing control loops in detail, but if they are basically the same as the "normal" loops, it would probably make sense to default to those gains as they work (whereas the fixed default might not).
For the DESIRED_SPEED and ALT_FLARE, either way is fine with me... provide a default or require it to be defined.

flixr:
Also what happens if someone tries to use this skid landing module with the new/adaptive control loops?

podhrmic:
Well the skid landing is using the new mode V_CTL_LANDING - and at least for our aircrafts the gains are different. V_CTL_LANDING controls the pitch during flare, and I guess that because that is outside of normal flight envelope (close to the stall speed, high angle of attack), the normal flight gains won't work very well. 

For adaptive control - it wont compile, because the adaptive controls don't have V_CTL_LANDING

flixr:
Well, if you think that users should provide separate gains for V_CTL_LANDING, then maybe it doesn't make much sense to have defaluts (for all those defines)?

flixr:
But if the flight plan flare landing works well with the normal gains and someone just wants to use this module to make things easier, it would make sense to default...

podhrmic:
I ll look into it.

conf/modules/nav_skid_landing.xml
```diff
@@ -19,16 +19,54 @@ Block example:
        <define name="FINAL_HEIGHT" value="50" unit="m"/>
        <define name="FINAL_STAGE_TIME" value="10" unit="s"/>
      </section>
-</block>
 @endverbatim
 -->
+
+Also define 
+<!--
```
flixr: if you comment these things out, they won't appear in the generated docs: http://docs.paparazziuav.org/latest/module__nav_skid_landing.html

You can instead make the complete `<section>` part of the `<doc>` node...

podhrmic: fixed with 6c19fe313d59775528ffc6fd5b3083022cc35cbf

PR #1185, opened 2015-05-24
[can] start support of CAN for STM32F4

gautierhattenberger:
compiling but not tested
see #1005

PR #1670, opened 2016-05-17
added can support for stm32f4

masierra:


sw/airborne/arch/stm32/mcu_periph/can_arch.c
```diff
@@ -193,4 +214,24 @@ void usb_lp_can_rx0_isr(void)
 
   can_fifo_release(CAN1, 0);
 }
+#elif STM32F4
```
flixr: would be sufficient to #ifdef the function name as in #1185

sw/airborne/arch/stm32/mcu_periph/can_arch.c
```diff
@@ -78,6 +78,21 @@ void can_hw_init(void)
   nvic_enable_irq(NVIC_USB_LP_CAN_RX0_IRQ);
   nvic_set_priority(NVIC_USB_LP_CAN_RX0_IRQ, NVIC_USB_LP_CAN_RX0_IRQ_PRIO);
 
+#elif STM32F4
+
+/* Enable peripheral clocks. */
+rcc_periph_clock_enable(RCC_GPIOB);
```
flixr: please indent correctly and would be nicer to use the CAN1_GPIO_PORT_RX etc. defines instead of fixed gpio port/pin as #1185

flixr:
Did you test this? It seems to be essentially the same as #1185 (except that this also changes some arguments for cam_init for F4)

podhrmic:
Hi, I just looked through the CAN driver as defined in mcu_periph: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/mcu_periph/can.h

Is there some specific reason why this driver uses a callback for receive? Wouldn't be better to have transmit and receive functions (something like for UART driver), so CAN can be used in similar fashion as other drivers?

flixr:
@masierra is this tested and working? If yes, it's good to merge.

> Is there some specific reason why this driver uses a callback for receive? Wouldn't be better to have transmit and receive functions (something like for UART driver), so CAN can be used in similar fashion as other drivers?

IIRC @esden did the initial implementation, maybe he can comment on that.

flixr:
@masierra it looks good, but I don't have anything to test this here... so is it tested and working?

masierra:
Sorry. Yeah it's tested and working.

PR #1671, opened 2016-05-17
[pprz_center] also use GCS program args for Simulation session

### LINKED ISSUE #1668: GCS arg of section programs not working for Simulation session
This does NOT work for me.

```
  <section name="programs">
    <program name="GCS" command="sw/ground_segment/cockpit/gcs">
      <arg flag="-layout" constant="bottom_settings.xml"/>
    </program>
  </section>

```
The following DOES work (except for REPLAY and SIMULATION)

```
  <section name="sessions">
    <session name="FOO">
      <program name="GCS">
        <arg flag="-speech" />
        <arg flag="-layout"  constant="bottom_settings.xml" />
      </program>
    </session>
  </section>

```

flixr:
and Replay session.
This allows to e.g. set the desired default options for the GCS (like layout, speech, etc..) in the programs section of the control_panel.xml file
and they will also be used for the Simulation and Replay sessions and not only when started from the tools menu.

should close #1668

### DIFF
--- a/sw/supervision/pc_control_panel.ml
+++ b/sw/supervision/pc_control_panel.ml
@@ -194,11 +194,25 @@ let get_simtype = fun (target_combo : Gtk_tools.combo) ->
       | choice -> List.nth targets (choice-1)
 
 let supervision = fun ?file gui log (ac_combo : Gtk_tools.combo) (target_combo : Gtk_tools.combo) ->
+  let get_program_args = fun program ->
+    let args = ref "" in
+    List.iter
+	  (fun arg ->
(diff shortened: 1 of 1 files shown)


flixr:
@gautierhattenberger does this look OK to you? Seems to work as expected for me...

gautierhattenberger:
yes, looks good to me.

dewagter:
Great! Small change but very visible impact. https://github.com/paparazzi/paparazzi/commit/6c8663aa0467c3737f0ed7fbfc86315b18df35f4.

PR #1675, opened 2016-05-19
[flight plans] add call_once alias

flixr:
- `<call_once fun="x"/>` is an alias for `<call fun="x" loop="false"/>`

The idea is to make it easier and more clear how to call functions that do not return a bool and should only be called once.

Same as #1612 without the run alias


flixr:
@gautierhattenberger good to merge?

PR #1677, opened 2016-05-20
Opencv stuff

rmeertens:
Now it is easy to convert back to the normally used yuv422 arrays of paparazzi. 

Any place better by the way for these conversion functions?

### DIFF
--- a/sw/airborne/modules/computer_vision/opencv_image_functions.h
+++ b/sw/airborne/modules/computer_vision/opencv_image_functions.h
@@ -1,0 +1,9 @@
+
+#ifndef OPENCV_IMAGE_FUNCTIONS_H
+#define OPENCV_IMAGE_FUNCTIONS_H
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+using namespace cv;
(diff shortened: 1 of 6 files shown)


fvantienen:
the `lib/image.h` and `lib/image.c`?

flixr:
I would put it in separate files, you don't want to have the OpenCV dependency if you "just" need the paparazzi specific stuff in image.h

rmeertens:
Ok, will make seperate files out of them.

rmeertens:
Fixed!

flixr:
Can you plz also add the license and doxygen headers to the new files?

sw/airborne/modules/computer_vision/opencv_image_functions.h
```diff
@@ -0,0 +1,9 @@
+
+#ifndef OPENCV_IMAGE_FUNCTIONS_H
+#define OPENCV_IMAGE_FUNCTIONS_H
+#include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+using namespace cv;
```
flixr: Also it's good style to not use namespaces in the header files, instead use `cv::Mat` below.

rmeertens:
Done.

PR #1682, opened 2016-05-22
[nav] rewrite nav_catapult module

gautierhattenberger:
- state machine more readable
- don't use imu struct directly
- more configuration options
- initial module was not working well anyway (was setting full throttle all the time unless high_freq function manually stopped)


flixr:
LGTM

PR #1683, opened 2016-05-22
Make the target device leave DFU mode after upload.

esden:
The added leave command tells the DFU target to leave the DFU mode after downloading or uploading a binary. Especially useful for devices that don't have a reset button.

### DIFF
--- a/conf/Makefile.stm32-upload
+++ b/conf/Makefile.stm32-upload
@@ -69,7 +69,7 @@ upload: $(OBJDIR)/$(TARGET).bin
 ifeq ($(VERIFY),1)
 	@echo "Reading flashed bin file of size $(DFU_SIZE) for verification"
 	$(Q)rm -f $(OBJDIR)/verify.bla
-	$(Q)dfu-util -d 0483:df11 -c 1 -i 0 -a 0 -s $(DFU_ADDR):$(DFU_SIZE) -U $(OBJDIR)/verify.bla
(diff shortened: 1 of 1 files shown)


flixr:
That sounds useful!
Do you know since what version this is supported?

gautierhattenberger:
Does that mean that the program will start right after upload ? When we do SD logging on Apogee, plugging the USB cable after startup is mounting the board as mass storage. I would like to avoid this to happen each time a program is flashed.

esden:
@flixr This is supported from the version that added support for DfuSe, it is part of the -s string.

@gautierhattenberger We can make the reset optional, if you want. On the other hand in case of Apogee this will not boot the payload firmware due to the way you have your USB circuit set up. You are pulling the boot1 pin all the time as long as USB is plugged in. This means that even when you reset you will go back to the stm32 bootloader mode. I have changed the circuit on Elle0 so that the boot1 pin is only asserted for 4ms after plugging in the USB cable. So when you are done flashing the boot1 pin is not asserted any more and the STM32 can reboot into the payload firmware.

flixr:
So if I see that correctly this is then only supported from version 0.7 upwards and hence should only be added to the part that also does the verify (since that is called if version >=0.7)

esden:
DUH I misunderstood somehow what the VERIFY does. I thought it was just a switch to turn the "verify if flashing was successful" part of the execution.

esden:
This is an updated set of patches. The first one only adds the leave command to the verify protected section. The second one adds the option to disable leave.

gautierhattenberger:
Thanks for the info. Even if it doesn't hurt for the Apogee, I think it is better to have it optional like you did.

PR #1684, opened 2016-05-23
added the possibility to add a yaw rate limit to stabilization indi

EwoudSmeur:
the yaw of a typical quadrotor easily saturates. One of the problems that occur when asking for a large change in yaw angle, is that the quadrotor can build up so much angular velocity in the yaw axis, that it can not physically slow down before reaching the yaw setpoint due to saturating. This will result in overshoot. 

This can be helped by adding a yaw rate limit, which this PR does for the INDI controller.

### DIFF
--- a/conf/settings/control/stabilization_indi.xml
+++ b/conf/settings/control/stabilization_indi.xml
@@ -15,7 +15,8 @@
       <dl_setting var="indi.g1.r" min="0" step="0.001" max="10" module="stabilization/stabilization_indi" shortname="ctl_eff_r" param="STABILIZATION_INDI_G1_R" persistent="true"/>
       <dl_setting var="indi.g2" min="0" step="0.01" max="10" module="stabilization/stabilization_indi" shortname="g2" param="STABILIZATION_INDI_G2_R" persistent="true"/>
       <dl_setting var="indi.adaptive" min="0" step="1" max="1" module="stabilization/stabilization_indi" shortname="use_adaptive" values="FALSE|TRUE" param="STABILIZATION_INDI_USE_ADAPTIVE" type="uint8" persistent="true"/>
-      <dl_setting var="indi.max_rate" min="0" step="0.0001" max="80.0" module="stabilization/stabilization_indi" shortname="max_rate"/>
+      <dl_setting var="indi.max_rate" min="0" step="0.01" max="80.0" module="stabilization/stabilization_indi" shortname="max_rate" param="STABILIZATION_INDI_MAX_RATE" unit="rad/s" alt_unit="deg/s"/>
+      <dl_setting var="indi.attitude_max_yaw_rate" min="0" step="0.01" max="80.0" module="stabilization/stabilization_indi" shortname="max_yaw_rate_attitude" param="STABILIZATION_INDI_MAX_R" unit="rad/s" alt_unit="deg/s"/>
     </dl_settings>
 
   </dl_settings>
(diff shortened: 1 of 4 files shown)


flixr:
Nice! But I think it makes sense to define the yaw rate limit in rad/s and then use `<define name="MAX_R" value="120.0" unit="deg/s"/>`
Then it is consistent with the rest and you save the onboard conversion to radians...

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -206,27 +206,36 @@ static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct I
...
+
+  indi.angular_accel_ref.q = indi.reference_acceleration.err_q * QUAT1_FLOAT_OF_BFP(att_err->qy)
+                             - indi.reference_acceleration.rate_q * rates_for_feedback.q;
+
+  //This separates the P and D controller and lets you impose a maximum yaw rate.
+  float rate_ref_r = indi.reference_acceleration.err_r * QUAT1_FLOAT_OF_BFP(att_err->qz)/indi.reference_acceleration.rate_r;
+
+#ifdef STABILIZATION_INDI_MAX_R
+  //define STABILIZATION_INDI_MAX_R in deg/s
+  Bound(rate_ref_r, -RadOfDeg(STABILIZATION_INDI_MAX_R), RadOfDeg(STABILIZATION_INDI_MAX_R));
```
flixr: You can simply use `BoundAbs(rate_ref_r, STABILIZATION_INDI_MAX_R)` here...

flixr:
Maybe it would also be useful to provide a default for `STABILIZATION_INDI_MAX_R` and assign it to a variable which can be exposed via settings to make it configurable at runtime?

EwoudSmeur:
thanks for your comments felix!

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.h
```diff
@@ -83,6 +83,7 @@ struct IndiVariables {
 
   bool adaptive;             ///< Enable adataptive estimation
   float max_rate;            ///< Maximum rate in rate control
+  float attitude_max_yaw_rate; ///< Maximum yaw rate in atttiude control
```
flixr: always good to specify the units in the doxygen comments...

conf/settings/control/stabilization_indi.xml
```diff
@@ -15,7 +15,8 @@
       <dl_setting var="indi.g1.r" min="0" step="0.001" max="10" module="stabilization/stabilization_indi" shortname="ctl_eff_r" param="STABILIZATION_INDI_G1_R" persistent="true"/>
       <dl_setting var="indi.g2" min="0" step="0.01" max="10" module="stabilization/stabilization_indi" shortname="g2" param="STABILIZATION_INDI_G2_R" persistent="true"/>
       <dl_setting var="indi.adaptive" min="0" step="1" max="1" module="stabilization/stabilization_indi" shortname="use_adaptive" values="FALSE|TRUE" param="STABILIZATION_INDI_USE_ADAPTIVE" type="uint8" persistent="true"/>
-      <dl_setting var="indi.max_rate" min="0" step="0.0001" max="80.0" module="stabilization/stabilization_indi" shortname="max_rate"/>
+      <dl_setting var="indi.max_rate" min="0" step="0.01" max="80.0" module="stabilization/stabilization_indi" shortname="max_rate"/>
+      <dl_setting var="indi.attitude_max_yaw_rate" min="0" step="0.01" max="80.0" module="stabilization/stabilization_indi" shortname="max_yaw_rate_attitude"/>
```
flixr: minor additions:
- if you add `param="STABILIZATION_INDI_MAX_R"`, you can also directly save the current value to the airframe file from the settings tool or GCS
- adding `unit="rad/s" alt_unit="deg/s"` will make it clear that the value itself is in rad/s and display it to the user in deg/s in settings

PR #1686, opened 2016-05-24
Guided mode

rmeertens:
Hey guys, 

Working on this the next couple of weeks, I would like this airframe to be in master right now so I can keep updating it all the time.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot.c
+++ b/sw/airborne/firmwares/rotorcraft/autopilot.c
@@ -573,6 +573,40 @@ bool autopilot_guided_move_ned(float vx, float vy, float vz, float heading)
   return false;
 }
 
+bool autopilot_guided_set_vertical_velocity(float vz)
+{
+  if (autopilot_mode == AP_MODE_GUIDED) {
+    guidance_v_set_guided_vz(vz);
+    return true;
+  }
+  return false;
+}
+
+
+bool autopilot_guided_set_horizontal_velocity_ned(float vx, float vy)
+{
+  if (autopilot_mode == AP_MODE_GUIDED) {
+    guidance_h_set_guided_vel(vx, vy);
+    return true;
+  }
+  return false;
+}
+
+extern bool autopilot_guided_set_horizontal_velocity_body_relative(float vx, float vy)
+{
+  if (autopilot_mode == AP_MODE_GUIDED) {
+    float psi = stateGetNedToBodyEulers_f()->psi;
+    float x = cosf(-psi) * vx + sinf(-psi) * vy;
+    float y = -sinf(-psi) * vx + cosf(-psi) * vy;
+    guidance_h_set_guided_vel(x, y);
+    return true;
+  }
+  return false;
+}
+
+
+
+
 void autopilot_check_in_flight(bool motors_on)
 {
   if (autopilot_in_flight) {
@@ -628,11 +662,9 @@ static uint8_t ap_mode_of_3way_switch(void)
 {
   if (radio_control.values[RADIO_MODE] > THRESHOLD_2_PPRZ) {
     return autopilot_mode_auto2;
-  }
-  else if (radio_control.values[RADIO_MODE] > THRESHOLD_1_PPRZ) {
+  } else if (radio_control.values[RADIO_MODE] > THRESHOLD_1_PPRZ) {
     return MODE_AUTO1;
-  }
-  else {
+  } else {
     return MODE_MANUAL;
   }
 }
@@ -651,16 +683,15 @@ static uint8_t ap_mode_of_two_switches(void)
   if (radio_control.values[RADIO_MODE] < THRESHOLD_1_PPRZ) {
     /* RADIO_MODE in MANUAL position */
     return MODE_MANUAL;
-  }
-  else {
+  } else {
     /* RADIO_MODE not in MANUAL position.
      * Select AUTO mode bassed on RADIO_AUTO_MODE channel
      */
     if (radio_control.values[RADIO_AUTO_MODE] > THRESHOLD_2_PPRZ) {
       return autopilot_mode_auto2;
-    }
-    else
+    } else {
       return MODE_AUTO1;
+    }
   }
 }
 #endif
--- a/sw/airborne/firmwares/rotorcraft/autopilot.c
+++ b/sw/airborne/firmwares/rotorcraft/autopilot.c
@@ -526,6 +526,7 @@ void autopilot_set_mode(uint8_t new_autopilot_mode)
 
 }
 
+
 bool autopilot_guided_goto_ned(float x, float y, float z, float heading)
 {
   if (autopilot_mode == AP_MODE_GUIDED) {
@@ -533,6 +534,26 @@ bool autopilot_guided_goto_ned(float x, float y, float z, float heading)
     guidance_h_set_guided_heading(heading);
     guidance_v_set_guided_z(z);
     return true;
+  }
+  return false;
+}
+
+bool autopilot_guided_goto_ned_optional(float x, float y, float z, float heading, bool use_x, bool use_y, bool use_z, bool use_yaw)
+{
+  if (autopilot_mode == AP_MODE_GUIDED) {
+	if(use_x){
+		guidance_h_set_guided_x(x);
+	}
+	if(use_y){
+		guidance_h_set_guided_y(y);
+	}
+	if(use_yaw){
+		guidance_h_set_guided_heading(heading);
+	}
+	if(use_z){
+		guidance_v_set_guided_z(z);
+	}
+	return true;
   }
   return false;
 }
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
@@ -651,6 +651,26 @@ void guidance_h_set_igain(uint32_t igain)
   INT_VECT2_ZERO(guidance_h_trim_att_integrator);
 }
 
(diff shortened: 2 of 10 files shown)


rmeertens:
Added bonus (hope you guys don't mind that I did not split it up in different commits). 
Guidance mode now has more options, and is slightly improved by fixing the yaw when you enter guidance mode.

gautierhattenberger:
Looks like the new flags of the guided mode are not coherent with the description of the message https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml#L2470

flixr:
And also please do make separate pull requests for new features like that and describe the changes/additions appropriately so we can then link the PR in the changelog later...

sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -573,6 +573,40 @@ bool autopilot_guided_move_ned(float vx, float vy, float vz, float heading)
   return false;
 }
```
kirkscheper: Why do you want to separate the commands here when the message has x,y,z and heading? Are there scenarios you will be updating the horizontal and vertical set-points at different times/frequencies? I guess if you only want to command x,y you can command zeros in z and heading, similar for only setting z.

Also, looking at other functions, vertical velocity is typically referred to at vz so autopilot_guided_set_vz.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -228,6 +228,13 @@ void guidance_h_mode_changed(uint8_t new_mode)
     case GUIDANCE_H_MODE_RC_DIRECT:
       stabilization_none_enter();
       break;
+    case GUIDANCE_H_MODE_GUIDED: ;
```
kirkscheper: why not fall into case GUIDANCE_H_MODE_HOVER:

sw/airborne/firmwares/rotorcraft/rotorcraft_datalink.c
```diff
@@ -94,6 +94,20 @@ void firmware_parse_msg(void)
           /* local NED with x/y/z as velocity and yaw as absolute angle */
           autopilot_guided_move_ned(x, y, z, yaw);
           break;
+        case 0x71:
+          /* local NED with x/y/z as velocity and yaw as absolute angle */
+          autopilot_guided_set_vertical_velocity(z);
+          break;
+        case 0x72:
```
kirkscheper: what's the difference with 0x71 and 0x72, it also doesn't seem to be just in python script

sw/ground_segment/python/guided_mode_example.py
```diff
@@ -126,10 +126,49 @@ def move_at_vel(self, north=0.0, east=0.0, down=0.0, yaw=0.0):
         msg['yaw'] = yaw
         print("move at vel NED: %s" % msg)
         self._interface.send_raw_datalink(msg)
+    def move_at_vel_vertical(self, down=0.0):
+        """
+        move at specified velocity in meters/sec with absolute heading (if already in GUIDED mode)
```
kirkscheper: comment says you move with absolute heading but you don't set the heading here.

sw/ground_segment/python/guided_mode_example.py
```diff
@@ -126,10 +126,49 @@ def move_at_vel(self, north=0.0, east=0.0, down=0.0, yaw=0.0):
...
+        msg['x'] = 0.0
+        msg['y'] = 0.0
+        msg['z'] = down
+        msg['yaw'] = 0.0
+        print("move at vel NED: %s" % msg)
+        self._interface.send_raw_datalink(msg)
 
+    def move_at_vel_horizontal(self, north=0.0, east=0.0):
+        """
+        move at specified velocity in meters/sec with absolute heading (if already in GUIDED mode)
```
kirkscheper: comment says you move with absolute heading but you don't set the heading here.

sw/ground_segment/python/guided_mode_example.py
```diff
@@ -126,10 +126,49 @@ def move_at_vel(self, north=0.0, east=0.0, down=0.0, yaw=0.0):
...
+        msg['flags'] = 0x73
+        msg['x'] = north
+        msg['y'] = east
+        msg['z'] = 0.0
+        msg['yaw'] = 0.0
+        print("move at vel NED: %s" % msg)
+        self._interface.send_raw_datalink(msg)
+    def move_at_vel_horizontal_body_relative(self, north=0.0, east=0.0):
+        """
+        move at specified velocity in meters/sec with absolute heading (if already in GUIDED mode)
```
kirkscheper: comment says you move with absolute heading but you don't set the heading here.

flixr:
As @gautierhattenberger the flags should be used as described in the message: http://docs.paparazziuav.org/latest//paparazzi_messages.html#GUIDED_SETPOINT_NED

If you want to add more options, we could squeeze some more info in the first 4 bits of flags (used only for frame right now)

sw/airborne/firmwares/rotorcraft/autopilot.c
rmeertens: Added the function 
`bool autopilot_guided_goto_ned_optional(float x, float y, float z, float heading, bool use_x, bool use_y, bool use_z, bool use_yaw)`
Would this be ok for you?

sw/airborne/firmwares/rotorcraft/rotorcraft_datalink.c
```diff
@@ -76,12 +76,27 @@ void firmware_parse_msg(void)
...
+
+
+      bool in_body_frame = bit_is_set(flags,0);
+      bool use_offset = bit_is_set(flags,1);
+      bool set_velocity = bit_is_set(flags,2);
+      bool use_x = bit_is_set(flags,4);
+      bool use_y = bit_is_set(flags,5);
+      bool use_z = bit_is_set(flags,6);
+      bool use_yaw = bit_is_set(flags,7);
+
```
rmeertens: What do you think about these flags? I will have to update them in the documentation, but that is easy. 
So now it is: 

- Bit 0: is in body frame
- Bit 1: is in global NED instead of local
- Bit 2: is controlling velocity
- Bit 3: kill all humans (free)
- Bit 4: set X
- Bit 5: Set Y
- Bit 6: Set Z
- Bit 7: set yaw

rmeertens:
Check my new comments and let me know what you think about them

flixr:
With your proposal you can't mix pos and vel commands anymore e.g. command a x,y position and z vel, or yaw rate

flixr:
Is it really needed to be able to specify x,y,z separately?
What is the use case for that? When using guided mode, you have an "external guidance controller" anyway, why should you not simply set them all?

kirkscheper:
I don't think that is what he wanted to do, I will take a look at this and try in a different way.

@flixr what did you want to do with the first 4 bits of the flag, currently you only use the first 2. Is it that the first 2 bits define the axis frame for the position command and the next 2 for the speed? or was it just an error that you said bits 0-3?

kirkscheper:
@flixr do you want to be able to all four parameters as position or speed separately? Example setting x position, y speed, z speed and heading angle?

I would imagine that the x,y pair should always be set in the same way (either speed or position) where if you want to hold position you set speed to 0.

flixr:
Regarding the flag bits: I could have also written: bits 0-1: frame, bits 2-3: reserved, bits 4-7: x,y,z,heading as velocity/rate
Basically left it open whether we want to add more coordinate frames (or if we want to use the remaining bits for something else)...

To be honest I didn't think of the case where you might want to use different coordinate frames for the pos and vel: e.g. command relative velocity in body frame but absolute altitude.

> I would imagine that the x,y pair should always be set in the same way (either speed or position) where if you want to hold position you set speed to 0.

For the current implementation this has to be the case, yes... but didn't want to restrict it in the message definition at that point... We could change that if needed though.

If we want to enable/disable certain axes, we could either try to cram that into the current flags field or add another one...
See also https://pixhawk.ethz.ch/mavlink/#SET_POSITION_TARGET_LOCAL_NED which has a lot more options, but is hence also a lot bigger than needed in almost all cases.

kirkscheper:
What we could do is separate the flags for axis system and the command to look something like this:

axis_flag:
bit 0-1 position 
bit 0: set = Local, clear = Body
bit 1: set = Offset, clear=absolute

bit 2 velocity
bit 0: set = Local, clear = Body

flag:
set = use, clear = ignore
bit 0: x pos
bit 1: y pos
bit 2: z pos
bit 3: x vel
bit 4: y vel
bit 5: z vel
bit 6: heading
bit 7: yaw rate

I guess we would have to put in a check that paired bits (e.g bits 0 & 3) are not both set at the same time. 

Alternatively I can make one flag with size uint16.

flixr:
I don't think it makes that much sense to have separate bits to enable x pos and x vel, since as you said we would have to check that both are not set at the same time.
IMHO that would make sense if we also add both pos and vel fields...

Maybe something like this for the flags:
bit 0: xy as offset coordinates
bit 1: xy in body coordinates
bit 2: z as offset coordinates
bit 3: yaw as offset coordinates
bit 4: free?
bit 5: x and y as vel
bit 6: z as vel
bit 7: yaw as vel/rate

I think this should pretty much cover all sensible combinations (that set all x,y,z,yaw)
and if we want to set only some axes add an extra mask field where bits set disable the respective axis.

kirkscheper:
I can live with that. @rmeertens, is this ok? If so I will implement something and push soon.

rmeertens:
Looks good!

rmeertens:
Other pull request is merged!

PR #1687, opened 2016-05-24
[doc] module description update

dewagter:


sw/tools/doxygen_gen/gen_modules_doc.py
```diff
@@ -94,11 +94,14 @@ def module_page(filename, module):
 
 def get_xml_example(filename, module):
     opts = module.findall(".doc/define") + module.findall(".doc/configure")
+    deps = module.findall(".doc/depends")
```
flixr: This should be `module.findall(".depends")`

flixr:
we should probably update the generated docs to not use
```
<modules>
  <load name="foo.xml"/>
</modules>
```
anymore and directly use the new ```<module name="foo"/>`

PR #1688, opened 2016-05-26
Fixed aircraft IDs for new airframe

podhrmic:
I missed the new airframe in the conf_tests - this fixes the problem. Sorry about that.


flixr:
@podhrmic no need to create a PR for stuff like this, you can commit it directly...

PR #1692, opened 2016-05-29
Added an initial version of a texton module that represents the appea…

guidoAI:
…rance of an image with a texton distribution. In past work, we have used texton distributions for instance for obstacle detection. If the entropy of the texton distribution becomes low, this is typically an indication of a close obstacle. This "appearance variation cue" can be a good complement to using optical flow measurements for obstacle avoidance. There are may other uses for texton distributions though, while they are computationally quite cheap.

### DIFF
--- a/sw/airborne/modules/computer_vision/textons.c
+++ b/sw/airborne/modules/computer_vision/textons.c
@@ -1,0 +1,540 @@
+/*
+ * Copyright (C) 2016, Hann Woei Ho, Guido de Croon
+ *
+ * This file is part of Paparazzi.
+ *
+ * Paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
please fix indentation or use the `fix_code_style.s` script.
also there are some comments that can be removed (e.g. in module xml file).

sw/airborne/modules/computer_vision/textons.c
```diff
@@ -0,0 +1,540 @@
...
+uint8_t dictionary_number = 0;
+
+// status variables
+uint8_t dictionary_ready = 0;
+float alpha = 0.0;
+
+// File pointer for saving the dictionary
+static FILE *dictionary_logger = NULL;
+#ifndef DICTIONARY_PATH
+#define DICTIONARY_PATH /data/video/
```
flixr: should be added to the doc section in the module xml file...

guidoAI:
I fixed the code style and added the documentation.

flixr:
@guidoAI thx!
If you tested this and it compiles/works it looks good to me to merge...

PR #1694, opened 2016-05-30
[guidance] Expansion of the guided protocol

kirkscheper:
We expanded the message flag definition to facilitate updating the horizontal and vertical setpoints in different reference frames as well as separate definitions of whether the setpoints are position or velocity commands. Also added a heading rate setpoint. 

This is a draft pull request, I still need to test it (especially the new rate setpoint) before it can be approved. Feedback is welcome @rmeertens.

continues/replaces https://github.com/paparazzi/paparazzi/pull/1686

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot.c
+++ b/sw/airborne/firmwares/rotorcraft/autopilot.c
@@ -526,7 +526,6 @@ void autopilot_set_mode(uint8_t new_autopilot_mode)
     //if switching to rate mode but rate mode is not defined, the function returned
     autopilot_mode = new_autopilot_mode;
   }
-
 }
 
 bool autopilot_guided_goto_ned(float x, float y, float z, float heading)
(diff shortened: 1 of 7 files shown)


sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -573,6 +572,77 @@ bool autopilot_guided_move_ned(float vx, float vy, float vz, float heading)
...
+  // handle x,y
+  if (bit_is_set(flags, 5)) { // velocity setpoint
+    if (bit_is_set(flags, 1)) { // set velocity in body frame
+      float psi = stateGetNedToBodyEulers_f()->psi;
+      x =  cosf(-psi) * x + sinf(-psi) * y;
+      y = -sinf(-psi) * x + cosf(-psi) * y;
+    }
+    guidance_h_set_guided_vel(x, y);
+  } else {  // position setpoint
+    if ((flags & 3) == 0) {   // set absolute position setpoint
```
rmeertens: use bit_is_set

kirkscheper: :( I wanted to check the two bits at the same time... ok, wilco

Kirk

rmeertens: I know ;) 
Kudo's for knowing this trick :+1:

sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -573,6 +572,77 @@ bool autopilot_guided_move_ned(float vx, float vy, float vz, float heading)
...
+      guidance_h_set_guided_pos(x, y);
```
rmeertens: Also, it has to be offset AND in body right??

kirkscheper: yes and no. Here, the offset bit flag is used to indicate it is offset in NED frame and the body flag is used to say it is offset in body frame. I think this is obvious as you can't have absolute in body frame. Are the comments I wrote not sufficient? 

I can alternatively change it that if the body flag is set ignore the other flags and just set the offset body. If the body flag isn't set then I check if it should be offset or absolute in NED... Would that be more reasonable?

rmeertens: Mmm, but if you want offset AND body... you get NOT offset and NOT body: 
`  guidance_h_set_guided_pos(x, y);`

Either do nothing, or choose one of the two

kirkscheper: Currently, when you have offset and body nothing is done. As I said offset is in NED and body is offset in body, you cant' have both. 

To simplify it I will update it with the alternative in my previous comment. If body flag is set I ignore the other flags.

kirkscheper:
@rmeertens thanks for the comment, think its cleaner this way. Do you have time to test this this tomorrow with me as I don't know how the python scripts work?

rmeertens:
line 606: bit is set needs to have the flags ;)

kirkscheper:
Ok, this seems to be good now. Tested and working as we would expect. @flixr how does it look to you?

sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -576,6 +575,77 @@ bool autopilot_guided_move_ned(float vx, float vy, float vz, float heading)
   return false;
 }
 
+/* Set guided mode setpoint
```
flixr: please use doxygen style comments so it actually ends up in the generated documentation....

flixr: ah, ok, doxygen comment in the header file...

flixr:
Looks good to me :-)
Haven't really checked the flags values in the python implementation though... so if it's tested and the pprzlink submodule included here as well, we can merge it.

Also we should probably add `heading_rate` to the `HOVER_LOOP` message...?

kirkscheper:
I just updated the python script slightly and tested it as is on a parrot ardrone and it works pretty well!

kirkscheper:
oh, oops you already merged it, haha. I will make a new request to update the script.

flixr:
no, this is not merged yet... only the changes in pprzlink are merged...
Can you please also update the pprzlink submodule to the latest version with the updated message?

PR #1699, opened 2016-06-02
[chibios] Add support for a full ChibiOS arch

gautierhattenberger:
This is replacing the previous mix chibios/libopencm3:
- all drivers are ported to the new arch
- it was hard to go forward and take advantages of the RTOS, now it is much easier
- switch to v3 of ChibiOS
- for now, it is functionaly the same than before
- the previous variable length array mechanism used internaly by the
  sdlog is replaced by a more efficient library called TLSF (allow
  constant time malloc/free)

It also comes as a replacement of the rt_chibios branch (https://github.com/paparazzi/paparazzi/tree/rt_chibios) and the pull request #1504 (although the approach is different)


podhrmic:
Cool, that is exciting! From looking at it the commit, it doesn't add the arch/chibios dir - is that something that you don't need for the sdlogger?

gautierhattenberger:
@podhrmic Thanks, I didn't even realized that all the new files were missing from the commit

podhrmic:
@gautierhattenberger  looking at the driver implementation - you have threads for IO, and then running the whole autopilot in one thread (which is the different approach). But this would mean locking us to using ChibiOS this hybrid way, instead of using the full potential of multithreading (or haivng a choice):-/ I still think there is a better solution - are you willing to wait with this?

gautierhattenberger:
As mentioned, it mostly provides the same functionality than the current code, but allow to move forward, whatever the direction.
For instance, it is easy to make payload modules running their own threads.
Regarding the implementation of the drivers, I have modified the UART driver so that when it is used to send messages, the mutex are only needed once per message instead of once per character in the previous implementation. So it is much more efficient, and the other ones (SPI, I2C) were already working like this thanks to the transactions queues.
In the end, the regular FW autopilot only needs 6% of the MCU on Apogee.
I also would like to make AP and FBW parts running in separated thread soon, since the design was made for that purpose.
With the time, more elements could be split into threads, but making it nicely compatible with older bare-metal code is not so obvious to me.
So, except for the issue of which ChibiOS Git repository we should follow, I really think getting rid of the mixed chibios/locm3 is a necessary step.

podhrmic:
I definitely agree with removing the chibios-libopecm3 stuff.
So I guess as long as you are open to slowly improving the drivers/threading over time I am cool with that.
Also, what is the github issue? Isn't this the repo we want? https://github.com/ChibiOS/ChibiOS/tree/stable_16.1.x

gautierhattenberger:
Yes, but since they apply a AUTHORS file to change the "svn" name to a full name+email git format, this tree is not compatible anymore with the one we are currently using in master (https://github.com/mabl/ChibiOS). So the command `git submodule sync` needs to be applied at some point for the new tree to be fetched, otherwise the submodule checkout fails.
It's fine for new users, not for people updating from an older branch. This is also why the CI servers are failing. Hopefully @flixr as an idea about how to sort this properly, I'm not really good with git submodules.

podhrmic:
I see. Well I guess since this is in the development branch, it shouldn't be a problem for users to just do a fresh `git clone` (in case it turns out to be too complicated to fix)

flixr:
added the git submodule sync to semaphore, shouldn't be needed for travis as it does a fresh git clone...
The tests are failing because of e.g. duplicated AC_IDs...

podhrmic:
The ID's are fixed in the master - `git rebase`?

flixr:
Currently no idea what's up with the semaphore build, investigating that...
But there is e.g. logger_sd_chibios.makefile missing as can be seen in the travis builds

gautierhattenberger:
@flixr thanks for fixing CI config!
@podhrmic @flixr  do you see something else to fix/improve here ?

conf/Makefile.chibios
```diff
@@ -95,6 +105,16 @@ ifeq ($(USE_LINK_GC),)
   USE_LINK_GC = yes
 endif
 
+# Linker extra options here.
+ifeq ($(USE_LDOPT),)
+  USE_LDOPT =
```
flixr: minor, but I think that writing `USE_LDOPT ?=` is easier to understand than using ifeq...

conf/modules/tlsf.xml
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="tlsf">
+  <doc>
+    <description>
+      Dynamic memory allocator
+      Can be used as a replacement of malloc with a O(1) allocation time
```
flixr: Would be nice to add link to the https://github.com/mattconte/tlsf here...

sw/ext/tlsf/readme.txt
```diff
@@ -0,0 +1 @@
+see http://tlsf.baisoku.org
```
flixr: should either clearly state which version of https://github.com/mattconte/tlsf that is or use it as submodule...

podhrmic:
I ll look through it in detail over the weekend.

conf/Makefile.chibios
gautierhattenberger: This is because it is based on original ChibiOS examples, makes it easier to upgrade afterwards. Should I still change it ?

conf/modules/tlsf.xml
gautierhattenberger: This is modified version, so I should have made a dedicated repository for this, and I though it was not much needed for only two files.

flixr: Then it would be good to describe in the readme in sw/ext/tlsf which version it was based on and what changes were made and why...

sw/ext/tlsf/readme.txt
gautierhattenberger: also, when we did this, the code was not yet published on GitHub, only tarballs and maybe SVN...

conf/Makefile.chibios
```diff
@@ -20,66 +19,77 @@
...
-
-
 #
 # This is the common Makefile for target using chibios
+#
+# chibios directory
+CHIBIOS =  $(PAPARAZZI_SRC)/sw/ext/chibios
+
+# directory with board defines for chibios platforms (board specific)
+BOARD_DIR ?= $(BOARD)/chibios
```
flixr: here you set BOARD_DIR to $(BOARD)/chibios, but for apogee you use `BOARD_DIR=$(BOARD)/v$(BOARD_VERSION)`...
Maybe we should stick to one schema?

sw/airborne/firmwares/fixedwing/main_chibios.h
```diff
@@ -19,13 +19,11 @@
...
  */
 
-/*
- * @file subsystems/chibios-libopencm3/usbStorage.h
- *
+/**
+ * @file firmwares/fixedwing/main_chibios.h
  */
+
 #pragma once
```
flixr: should probably add the include where thread_t is defined...

flixr:
Haven't looked at the drivers in detail, but I would be very happy if we can merge this as it is a lot cleaner than the chibios-locm3 mix... :-)

Then when we have this in master we can continue and pull in the separate main threads that @podhrmic already uses in his chibios branch.

Or maybe even better: we should try to abstract threads and mutexes and build a main that we can use for both Linux OS and ChibiOS (and ...?)

conf/Makefile.chibios
```diff
@@ -20,66 +19,77 @@
...
+CHIBIOS_BOARD_DIR = $(PAPARAZZI_SRC)/sw/airborne/boards/$(BOARD_DIR)
+
+# directory with board and OS configuration (project specific)
+PROJECT_DIR ?= demo
```
podhrmic: I am not sure about the project directory - it is `demo` by default, but in the board makefile we set it to ` ` (empty string), right?

Maybe we can get rid of it / streamline it?

conf/Makefile.chibios
```diff
@@ -20,66 +19,77 @@
...
 
 # Launch with "make Q=''" to get full command display
 Q=@
 
+# Debugging options
+RTOS_DEBUG ?= 1
+
+ifeq (,$(findstring $(RTOS_DEBUG),0 FALSE))
+$(info DEBUGGING ON)
+$(TARGET).CFLAGS += \
```
podhrmic: Not all debug options need to be set to TRUE for debugging, because then the generated files have several hundreds of megabytes. We are using this:
```
# Use only gddb2 for debugging unless otherwise specified
DEBUG_LEVEL ?= 2

ifeq (,$(findstring $(RTOS_DEBUG),0 FALSE))
$(info DEBUGGING ON)
$(info DEBUG_LEVEL = -ggdb$(DEBUG_LEVEL))
$(TARGET).CFLAGS += \
  -DCH_DBG_STATISTICS=FALSE \
	-DCH_DBG_SYSTEM_STATE_CHECK=TRUE \
	-DCH_DBG_ENABLE_CHECKS=TRUE \
	-DCH_DBG_ENABLE_ASSERTS=FALSE \
	-DCH_DBG_ENABLE_TRACE=FALSE \
	-DCH_DBG_ENABLE_STACK_CHECK=TRUE \
	-DCH_DBG_FILL_THREADS=FALSE \
	-DCH_DBG_THREADS_PROFILING=TRUE
$(TARGET).CFLAGS += -DRTOS_DEBUG=1

USE_OPT = -std=gnu99 -O0 -ggdb$(DEBUG_LEVEL) -fomit-frame-pointer -falign-functions \
	 -W -Wall -DUSE_CHIBIOS_RTOS

else
$(info DEBUGGING OFF)
$(TARGET).CFLAGS += -DCH_DBG_THREADS_PROFILING=TRUE
endif
```
Indeed it would be nice to be able to quickly set min/medium/max debugging levels - maybe merge as is, I can make a pull request with this later.

conf/Makefile.chibios
podhrmic: Probably better to make it consistent with the original ChibiOS examples, at least for now

conf/Makefile.chibios
```diff
@@ -131,49 +168,40 @@ endif
...
-
-CHIBIOS = $(CHIBIOS_EXT)
-OPENCM3 = $(OPENCM3_EXT)
-OPENCM3_INC = $(OPENCM3)/include
-OPENCM3_LIB =  $(OPENCM3)/lib/
-
-include $(CHIBIOS_BOARD_DIR)/board.mk
-include $(CHIBIOS)/os/hal/platforms/STM32F4xx/platform.mk
+# Startup files.
+include $(CHIBIOS)/os/common/ports/ARMCMx/compilers/GCC/mk/startup_stm32f4xx.mk
```
podhrmic: this makes it STM32F4xx only - better to have this configurable.

An example from our (older) solution:
In board makefile:
```
# Project specific files and paths (see Makefile.chibios for details)
CHIBIOS_BOARD_PLATFORM = STM32F4xx/platform.mk
CHIBIOS_BOARD_PORT = ARMCMx/STM32F4xx/port.mk
CHIBIOS_BOARD_LINKER = STM32F407xG.ld
```

in chibios makefile:
```
include $(CHIBIOS)/os/hal/platforms/$(CHIBIOS_BOARD_PLATFORM)
include $(CHIBIOS)/os/ports/GCC/$(CHIBIOS_BOARD_PORT)
```

so I guess something along that line should work (so we can test this in Lisa M, Discovery board etc.)

podhrmic: Oh I see, it is already there - just the first include is weird (the startup script).

conf/Makefile.chibios
```diff
@@ -296,7 +298,7 @@ UINCDIR = $(patsubst -I%,%,$(INCLUDES))
 ULIBDIR =
 
 # List all user libraries here
-ULIBS = $(BOARD_LDFLAGS)
+ULIBS = -lm $(BOARD_LDFLAGS)
```
podhrmic: What is `-lm` for?

conf/boards/apogee_1.0_chibios.makefile
```diff
@@ -57,3 +75,14 @@ SBUS_PORT ?= UART2
 # e.g. <servo driver="Ppm">
 #
 ACTUATORS ?= actuators_pwm
+
```
podhrmic: probably nice to remove the commented out lines?

conf/firmwares/test_chibios.makefile
```diff
@@ -0,0 +1,143 @@
...
+
+#
+# common test
+#
+# configuration
+#   SYS_TIME_LED
+#   MODEM_PORT
+#   MODEM_BAUD
+#
+PERIODIC_FREQUENCY ?= 500
```
podhrmic: I would add `RTOS_DEBUG = 1` so the tests always compile with full debug suite.

sw/airborne/arch/chibios/STM32F107xC.ld
```diff
@@ -0,0 +1,167 @@
...
+ * Note: probably not the cleanest solution, but it works
+ *
+ * __process_stack_size__ determines stack size of main() thread
+ */
+__main_stack_size__     = 0x0400;
+__process_stack_size__  = 0x0400;
+
+MEMORY
+{
+    flash : org = 0x08002000, len = 248k
```
podhrmic: Is the custom linker script for the bootloader?

sw/airborne/arch/chibios/STM32F407xG_no_luftboot.ld
```diff
@@ -0,0 +1,161 @@
...
+
+/*
+ * ST32F407xG memory setup.
+ */
+__main_stack_size__     = 0x0400;
+__process_stack_size__  = 0x0400;
+
+MEMORY
+{
+    flash : org = 0x08000000, len = 1M
```
podhrmic: I remember having two separate linker scripts for bootloader/no bootloader was the only way to make it work with ChibiOS - I am wondering if with ChibiOS 3.0 this can be somehow merged in one script?

sw/airborne/arch/chibios/chconf-old.h
```diff
@@ -1,21 +1,23 @@
 /*
-    ChibiOS/RT - Copyright (C) 2006,2007,2008,2009,2010,
-                 2011,2012 Giovanni Di Sirio.
+    ChibiOS/RT - Copyright (C) 2006-2013 Giovanni Di Sirio
 
-    This file is part of ChibiOS/RT.
+    modified by: AggieAir, A Remote Sensing Unmanned Aerial System for Scientific Applications
```
podhrmic: Probably instead of renamimg the old file can be just removed.

sw/airborne/arch/chibios/halconf.h
```diff
@@ -62,6 +69,14 @@
 #define HAL_USE_CAN                 FALSE
```
podhrmic: would be nice to add #if USE_CAN1 | USE_CAN2 as a first step towards using CAN in ChibiOS

sw/airborne/arch/chibios/mcu_periph/sys_time_arch.c
```diff
@@ -0,0 +1,145 @@
...
+  sys_time.nb_sec = sec;
+  sys_time.nb_sec_rem = sys_time.nb_tick - sys_time_ticks_of_sec(sys_time.nb_sec);
+
+  /* advance virtual timers */
+  for (unsigned int i = 0; i < SYS_TIME_NB_TIMER; i++) {
+    if (sys_time.timer[i].in_use &&
+        sys_time.nb_tick >= sys_time.timer[i].end_time) {
+      sys_time.timer[i].end_time += sys_time.timer[i].duration;
+      sys_time.timer[i].elapsed = TRUE;
+      /* call registered callbacks, WARNING: they will be executed in the sys_time thread! */
```
podhrmic: Certainly not ideal, but probably good for now (callbacks in the timer thread).

sw/airborne/arch/chibios/mcu_periph/uart_arch.c
```diff
@@ -0,0 +1,681 @@
...
+{
+  // check if more data to send
+  struct SerialInit *init_struct = (struct SerialInit*)(p->init_struct);
+  chSemWait (init_struct->tx_sem);
+  while (p->tx_insert_idx != p->tx_extract_idx) {
+    uint8_t data = p->tx_buf[p->tx_extract_idx];
+    p->tx_running = true;
+    sdWrite((SerialDriver *)p->reg_addr, &data, sizeof(data));
+    p->tx_running = false;
+    // TODO send by block (be careful with circular buffer)
```
podhrmic: Would it make sense in this case to internally use ChibiOS UART drivers instead of Serial drivers? Just a thought, but because UART driver uses DMA directly it is much more efficient than Serial driver. It is more complex to set up, but maybe the performance gains would be worth it - instead of sending blocks of bytes.

sw/airborne/arch/chibios/modules/core/rtos_mon_arch.c
```diff
@@ -0,0 +1,126 @@
...
+  // FIXME: add finer resolution than seconds?
+  rtos_mon.cpu_load = (1 - (float)(idle_counter - last_idle_counter) / CH_CFG_ST_FREQUENCY) * 100;
+  last_idle_counter = idle_counter;
+
+}
+
+static uint16_t get_stack_free(const thread_t *tp)
+{
+  int32_t index = 0;
+  // FIXME this is for STM32F4 only
```
podhrmic: Shouldn't we add checks that allow this module to compile for STM32F4 only?

sw/airborne/arch/chibios/modules/core/rtos_mon_arch.c
```diff
@@ -0,0 +1,126 @@
...
+  const uint8_t *max_ram_addr = (uint8_t*) (0x20000000 + (128*1024));
+  const int32_t internal_struct_size = (CH_KERNEL_MAJOR == 2) ? 80 : 120;
+  unsigned long long *stk_addr = (unsigned long long *) ((uint8_t *)tp + internal_struct_size);
+
+  while ((stk_addr[index] == 0x5555555555555555) && (((uint8_t *) &(stk_addr[index])) < max_ram_addr))
```
podhrmic: Not a big fan of this - wouldn't `chCoreGetStatusX()` be sufficient to get stack info?
Also, the ChibiStudio plugins already get this info somehow, so maybe it is stored internally in ChibiOS already?

sw/airborne/arch/chibios/subsystems/radio_control/ppm_arch.c
```diff
@@ -0,0 +1,98 @@
+/*
```
podhrmic: I remember we were having issues with PPM on Lisa M because the ICU driver in ChibiOS 2.6 didn't support Timer Channel 3 which was in this case the channel used for PPM rx. Do you know if this is fixed in new ChibiOS?

sw/airborne/firmwares/fixedwing/main_chibios.c
```diff
@@ -43,54 +48,60 @@
...
 
-  while (!chThdShouldTerminate()) {
+  while (!chThdShouldTerminateX()) {
     Fbw(handle_periodic_tasks);
     Ap(handle_periodic_tasks);
     Fbw(event_task);
     Ap(event_task);
-    chibios_chThdSleepMilliseconds(1);
+    chThdSleepMicroseconds(500);
+    //chThdYield();
```
podhrmic: Remove the commented lines?

sw/airborne/modules/loggers/sdlog_chibios/usbStorage.c
```diff
@@ -205,16 +191,15 @@ static const USBDescriptor* getDescriptor(USBDriver* usbp, uint8_t type, uint8_t
...
+static void usbEvent(USBDriver *usbp, usbevent_t event)
 {
   (void) usbp;
 
-  switch (event)
-  {
+  switch (event) {
     case USB_EVENT_CONFIGURED:
-      chSysLockFromIsr();
+      chSysLockFromISR();
```
podhrmic: Since we are using ChibiOS functions here we should have checks that prevent compiling this module without ChibiOS.
Also, if I understand it correctly, we want to have the core code universal (i.e. no RTOS specific code), but the modules can be defined for specific architectures only? Is that correct @flixr ?

sw/airborne/subsystems/gps/gps_ubx.c
```diff
@@ -189,8 +189,7 @@ void gps_ubx_read_message(void)
 }
 
 #if LOG_RAW_GPS
-#include "sdLog.h"
-#include "subsystems/chibios-libopencm3/chibios_sdlog.h"
+#include "modules/loggers/sdlog_chibios.h"
```
podhrmic: Shouldn't we make this optional?

podhrmic:
OK, so I finally have time to look over it in more detail - looks good, just few minor things I would recommend changing (see my comments).

>  Or maybe even better: we should try to abstract threads and mutexes and build a main that we can use for both Linux OS and ChibiOS (and ...?)

Sounds like a good idea. Although I am not very familiar with Linux threads so hard to say how much different it would be. But I definitely support multithreaded solution:)

sw/airborne/subsystems/gps/gps_ubx.c
flixr: it is, depending on `LOG_RAW_GPS`

sw/airborne/modules/loggers/sdlog_chibios/usbStorage.c
flixr: Yes, module can be for a specific arch only... e.g. we have lots of vision modules that only compile on the linux arch...
I guess in most cases there are no checks and it will simply fail if you try to use it on another arch.
It would be better to check a define in the c file and throw a nice error if the arch doesn't match...

conf/Makefile.chibios
gautierhattenberger: I choose not to use BOARD_VERSION since it may not be provided or wanted to separate config files by default.

conf/Makefile.chibios
gautierhattenberger: somehow I needed to link by hand the math library

sw/airborne/arch/chibios/STM32F107xC.ld
gautierhattenberger: file left by mistake

sw/airborne/arch/chibios/STM32F407xG_no_luftboot.ld
gautierhattenberger: file left by mistake

sw/airborne/arch/chibios/chconf-old.h
gautierhattenberger: file left by mistake

sw/airborne/arch/chibios/mcu_periph/sys_time_arch.c
gautierhattenberger: Sure, but actually not used in current AP code. This was initially done for some specific firmwares and test programs.
So keeping it for now, but would indeed special protection.

sw/airborne/arch/chibios/mcu_periph/uart_arch.c
gautierhattenberger: I think this option is out of the scope of this pull request for now.

sw/airborne/arch/chibios/modules/core/rtos_mon_arch.c
gautierhattenberger: I don't think core status function gives information about individual stack info of each threads. If this can be improved, I suggest to do it later, as it is not the most important part.

sw/airborne/arch/chibios/subsystems/radio_control/ppm_arch.c
gautierhattenberger: It is working now

sw/airborne/modules/loggers/sdlog_chibios/usbStorage.c
gautierhattenberger: As stated by the module's name, it only works with chibios. There is no arch indep API for mass storage. If it is ever needed, I'll improve this part as well.

gautierhattenberger:
I think this should cover most of the comments. Thanks.

podhrmic:
Looks good to me!

PR #1701, opened 2016-06-03
[vision] added missing include for ardrone board file

kirkscheper:
Master is breaking ardrone build. I added the required include to the board file for the ardrone and updated the comment of the file location for the video_device file. Builds now.


flixr:
thx!

PR #1702, opened 2016-06-03
Actually do the pullup/pulldown

zathras777:
Inconsistent behaviour between versions of STM code. When used in similar code for STM32F4 the set/clear are used.


flixr:
@zathras777 could you please provide a description of your change? why is this needed?

flixr:
@esden any comment from the libopencm3 side?

flixr:
@zathras777 could you please provide some more rationale for this change?
I don't quite understand why this would be needed....
@esden do you have any insights here?

podhrmic:
Close for feedback timeout? Plus nobody reported a problem with pullups/pulldowns anyway...

gautierhattenberger:
closing until we receive a more detailed error report

Issue #1703, opened 2016-06-03
Strings Deprecated Warning for Ocaml ≥ 4.02


rijeshaugustine:
As of Ocaml v4.02 Strings have been deprecated in favour of Bytes. 
https://ocaml.org/releases/4.02.html

This results in a build warning when compiling Paparazzi.

Changing Strings to Bytes is not sufficient to silence the warnings, as it would break compilation for everyone with OCaml < 4.02.0

@flixr was looing into using the camlp4 preprocessor to IFDEF around this issue... https://github.com/flixr/paparazzi/commit/007deabafe5fecc6ea7a1178566e69de4b804539

flixr:
We could also use `Bytes` everywhere and add `module Bytes = String` for OCaml < 4.02, but that would mean we would have to preprocess every file with camlp4...
So it seems like the nicer option would be to add something like a compat module with `bytes_create` etc. functions and only preprocess the compat.ml and compat.mli with camlp4...

rijeshaugustine:
Is this an ok way to fix the makefile?
https://github.com/rijeshaugustine/paparazzi/commit/af911cd1666cf4d0603f0e7368e5081e63409574

rijeshaugustine:
Ok so I think I fixed the Makefile. I used a shell if statement to only compile files listed in PP_SRC with the preprocessor.  https://github.com/rijeshaugustine/paparazzi/commit/81f9dfb182400af08808f1c31deaf6eb27bb23d4

I started to modify the rest of the .ml files.

PPRZLINK will need to be modified. It compiles first so compat.ml will have to be copied into that repo as well and the Makefile will need to be set up for the ocampl4. Unless there is a better way of doing it?

flixr:
Thanks for the work @rijeshaugustine !
Can you plz make a pull request?
pprzlink will have to be updated separately...

rijeshaugustine:
For pprzlink just adding the compat module to it will cause another warning when building sw/lib/ocaml due to compat.cmo already existing. Can this be rectified by telling the sw/lib/ocaml/Makefile to ignore compat.ml if compat.cmo exits in sw/ext/pprzlink/*/ocaml/ or will that cause other issues in the future?

podhrmic:
[...]

PR #1705, opened 2016-06-07
Ocaml Compatibility, Strings Deprecated Warning

rijeshaugustine:
This should resolve issue #1703 deprecated strings warning.


sw/lib/ocaml/compat.ml
```diff
@@ -0,0 +1,77 @@
...
+  BYTES.contains c s
+
+let bytes_length = fun len ->
+  BYTES.length len
+
+let bytes_make = fun n c->
+  BYTES.make n c
+
+let bytes_copy = fun s->
+	BYTES.copy s
```
gautierhattenberger: please keep the same indentation for all the file

gautierhattenberger:
The `open` instruction is a kind of `using namespace` in C++, it allows to use functions without specifying the module. But here you are using the `Compat.somthing` (and it is the correct way to go) so it should not be necessary.

flixr:
There are still some warnings, e.g.
```
OC http.ml
File "http.ml", line 24, characters 4-44:
Warning 5: this function application is partial,
maybe some arguments are missing.
```

@gautierhattenberger I leave this to you as you have way more OCaml experience and knowledge...
If you think this is a good solution: merge it

rijeshaugustine:
I don't think the warnings are due to strings being deprecated. I was unclear as to how exactly this should be handled. 

The error is due to:
`pipeline # set_proxy_from_environment ();`

method set_proxy_from_environment : insecure:bool -> unit -> unit
Inspect the environment variables http_proxy and no_proxy and set the proxy options from them. If insecure is set, basic authentication is enabled (otherwise only secure auth methods like Digest).

I can look at it further.

gautierhattenberger:
Since Ocamlnet 4.0.4, you need to specify the `insecure` input value. Probably something like:
```
pipeline # set_proxy_from_environment ~insecure:false ();
```
An other candidate for the Compat module...

PR #1706, opened 2016-06-07
Fixed the USB logger

rmeertens:
The USB logger works again! Made it into the current setup where you can specify which camera to log.

### DIFF
--- a/conf/modules/video_usb_logger.xml
+++ b/conf/modules/video_usb_logger.xml
@@ -8,6 +8,8 @@
     </description>
     <define name="VIDEO_USB_LOGGER_PATH" description="Logging path"/>
     <define name="VIDEO_USBLOG_CAMERA" value="front_camera|bottom_camera" description="Video device to log"/>
+    <define name="VIDEO_LOG_WIDTH" value="272" description="Size of the to log images"/>
+    <define name="VIDEO_LOG_HEIGHTH" value="272" description="Size of the to log images"/>
   </doc>
   <depends>video_rtp_stream</depends>
   <header>
@@ -17,10 +19,5 @@
   <periodic fun="video_usb_logger_periodic()" start="video_usb_logger_start()" stop="video_usb_logger_stop()" autorun="TRUE"/>
   <makefile target="ap">
     <file name="video_usb_logger.c"/>
-
-    <define name="VIDEO_DOWNSIZE_FACTOR" value="2"/>
-    <define name="USE_BOTTOM_CAMERA" value="1"/>
-    <define name="LOG_ON_USB" value="1"/>
-    <define name="VIDEO_FPS" value="10"/>
   </makefile>
 </module>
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
it looks fine but I can't really test it

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -28,35 +28,118 @@
...
+#include <sys/stat.h>
+#include <unistd.h>
+
 
 /** Set the default File logger path to the USB drive */
 #ifndef VIDEO_USB_LOGGER_PATH
 #define VIDEO_USB_LOGGER_PATH /data/video/usb
 #endif
 
+#ifndef VIDEO_LOG_WIDTH
```
flixr: please document newly added define options in the module xml

conf/modules/video_usb_logger.xml
```diff
@@ -7,12 +7,14 @@
...
   </doc>
   <depends>video_rtp_stream</depends>
   <header>
     <file name="video_usb_logger.h"/>
   </header>
-  <periodic fun="video_usb_logger_periodic()" start="video_usb_logger_start()" stop="video_usb_logger_stop()" autorun="FALSE"/>
+  
+  <periodic fun="video_usb_logger_periodic()" start="video_usb_logger_start()" stop="video_usb_logger_stop()" autorun="TRUE"/>
   <makefile target="ap">
     <file name="video_usb_logger.c"/>
```
flixr: please remove unused defines....

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -28,35 +28,118 @@
...
+  sprintf(save_name, "%s/img_%05d.jpg", foldername, shotNumber);
+
+  shotNumber++;
+  // Check if file exists or not
+  if (access(save_name, F_OK) == -1) {
+
+    // Create a high quality image (99% JPEG encoded)
+    jpeg_encode_image(img, img_jpeg, 99, TRUE);
+
+#if JPEG_WITH_EXIF_HEADER
```
flixr: this is an option that you should define in your module/airframe file?
Maybe it should be specific to this module?

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -28,35 +28,118 @@
...
+    counter++;
+  }
+
+}
+
+struct image_t *log_image(struct image_t *img);
+struct image_t *log_image(struct image_t *img)
+{
+  save_shot(img, &img_jpeg);
+  return FALSE;
```
flixr: should probably return `img`

rmeertens:
Added all documentation!

flixr:
Thx, but I still think it would be good to use a common prefix for the defines, right now there are
- VIDEO_USB_LOGGER_
- VIDEO_USBLOG_
- VIDEO_LOG_
- and no prefix for JPEG_WITH_EXIF_HEADER

flixr:
also why is it named udb_logger in the first place? It has nothing to do with USB...

rmeertens:
Changed the names of the defines.
Added Christophe as author before my changes (I think it is important to see who worked on it in case of questions)
For now, let's keep it at USB logger. Originally the path was set the the USB stick of the ARDrone.

flixr:
thx!

PR #1707, opened 2016-06-08
[state] allow some conversions of accel

gautierhattenberger:
Conversion between int and float can be done even when ned_origin is not initialized


flixr:
LGTM

PR #1708, opened 2016-06-10
Added a tool for viewing and interacting with the video stream

wilcoschoneveld:
I added a two files:
- rtp_stream.sdp which contains the video stream settings
- rtp_viewer.py which displays the video stream in a window (requires opencv-python)
- a region can be selected in the viewer to send a VIDEO_ROI message via the datalink


flixr:
Nice! Does this have better performance (e.g. concerning lag, etc) than using vlc or mplayer?

Please add a README.md file that lists required dependencies and how to use it (e.g. if you want to change the port).
Also would make sense to make rtp_viewer.py executable and add a shebang line...

rmeertens:
Can you make it send the message VIDEO_ROI when the new opencv-python player is clicked?

wilcoschoneveld:
@rmeertens something like this?

rmeertens:
Looks perfect! Thanks!

sw/tools/rtp_viewer/rtp_viewer.py
```diff
@@ -0,0 +1,69 @@
+#! /usr/bin/python
+
+import cv2
+import sys
+
+from os import path, getenv
+
+PPRZ_SRC = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../')))
+sys.path.append(PPRZ_SRC + "/sw/lib/python")
```
flixr: not needed...

wilcoschoneveld:
Regarding performance, viewing the RTP stream via opencv-python does indeed seem to perform better and reduce the lag somewhat. However, it is apparent from the console warnings that a lot of frame packets are dropped. I don't really have that much knowledge about RTP streams but my guess is that this is a problem of the video_rtp_stream module itself. Maybe someone else has other thoughts about this.

The video delay might be problematic regarding the VIDEO_ROI message, i.e. after 2/3 seconds a UAV could have moved significantly. To resolve this, maybe it is possible to retrieve meta data with the RTP stream? For example, position and attitude or timestamp information, so that the drone can re-use this information to further process the selected region. Anyways I think for simple tasks (like selecting a colored blob to track while a drone is hovering) this won't be too much of a problem.

flixr:
Nice! LGTM... do you want to add/change something else or is it good to merge from your side?

wilcoschoneveld:
I'm happy with it as is, cheers!

PR #1712, opened 2016-06-10
fixed qr code

rmeertens:
the qr code scanner is working again! Also: now printing the upper left corner of the code (@dewagter check this out!)

### DIFF
--- a/sw/airborne/modules/computer_vision/qrcode/qr_code.c
+++ b/sw/airborne/modules/computer_vision/qrcode/qr_code.c
@@ -31,8 +31,8 @@
 
 void qrcode_init(void)
 {
-	// Add qrscan to the list of image processing tasks in video_thread
-  cv_add(qrscan);
+  // Add qrscan to the list of image processing tasks in video_thread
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/computer_vision/qrcode/qr_code.c
```diff
@@ -99,5 +99,5 @@ struct image_t* qrscan(struct image_t *img)
   zbar_image_destroy(image);
   //zbar_image_scanner_destroy(scanner);
 
-  return NULL; // QRCode is not returning a new image.
+  return img;
```
flixr: Actually looks like return NULL is still sensible here, see `cv_run_device`...

rmeertens: See my new update ;)

rmeertens:
Thought it would be nice to allow people to draw a rectangle around found QR codes.

flixr:
Nice!

dewagter:
@roland: thanks for making it compatible with dual-video!
Are you drawing in the grayscale or the YUV?

-Christophe

PR #1714, opened 2016-06-11
[gps] optitrack always send heading

bartslinger:
Until now when flying with gps `datalink` (which we use for optitrack), you always need to define in your aiframe `<define name="AHRS_HEADING_UPDATE_GPS_MIN_SPEED" value="0"/>` to get heading updates.

I cannot imagine that you would want something else than 0, so I put it as default in the makefile.

### DIFF
--- a/conf/firmwares/subsystems/rotorcraft/gps_datalink.makefile
+++ b/conf/firmwares/subsystems/rotorcraft/gps_datalink.makefile
@@ -25,9 +25,6 @@ ap.srcs += $(SRC_SUBSYSTEMS)/gps/gps_datalink.c
 
 ap.CFLAGS += -DUSE_GPS -DGPS_DATALINK
 
-# Always zero to get heading update
-ap.CFLAGS += -DAHRS_HEADING_UPDATE_GPS_MIN_SPEED=0
-
 nps.CFLAGS += -DUSE_GPS
 nps.CFLAGS += -DGPS_TYPE_H=\"subsystems/gps/gps_sim_nps.h\"
 nps.srcs += $(SRC_SUBSYSTEMS)/gps/gps_sim_nps.c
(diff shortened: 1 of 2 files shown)


bartslinger:
Hmm maybe that was too fast, I guess it gives you compilation errors with double defines?

flixr:
I'm not sure if gps type datalink is only used for optitrack or also other cases (as it can of course in principle)....
Maybe it would be nicer to add a gps type optitrack?
Also this addition relies on you choosing ahrs int_cmpl_quat...

bartslinger:
It's only used in TUDELFT airframes. You are not required to use ahrs_int_cmpl_quat, but in case you do, the default is already set.
If you really want to go for gps type optitrack, I suggest to rename everything from datalink to optitrack.

flixr:
I'm not in favour of renaming datalink to optitrack...
And just because the only airframes using gps datalink in the repository are from TU Delft, it doesn't mean that nobody else is using that... and it is really more generic than just for optitrack.

bartslinger:
So you suggest to duplicate the entire module, with only this tiny difference?

flixr:
Basically I'm wondering if it makes sense to set this default for "all gps datalink users".
And maybe it would be more clear to create a gps_optitrack.makefile, which could be as simple as:
```
include $(CFG_ROTORCRAFT)/gps_datalink.makefile
# Always zero to get heading update
ap.CFLAGS += -DAHRS_HEADING_UPDATE_GPS_MIN_SPEED=0
```
(or later a module like in https://github.com/paparazzi/paparazzi/pull/1625)

But maybe it's not worth it...

flixr:
@gautierhattenberger what do you think is the nicer approach?

gautierhattenberger:
@flixr I think your option is the best

PR #1716, opened 2016-06-13
fixed drawing a line, now drawing around any qr code

rmeertens:
The draw line function was.... not drawing lines to the right spot. 
Fixed it by reverting to a previous version. 

Also: now I'm not assuming that a QR code has 4 sides. whoo


flixr:
This now reverts #1574... does it break other things again?
Is it really a YUV422 image?

rmeertens:
Line 649 breaks the current implementation...
Right now we increment two times, once in the for loop and once by incrementing by the pixel size. 

So... we could either revert the previous commit (my preference as that is proven to work) or try to keep the previous commit and fix the for loop (more work, more testing, more bugs). Let's just accept this pull request.

PR #1719, opened 2016-06-13
[ins] add true int computation for vel_cb

kirkscheper:
There seemed to be no integer method implemented in ins_int.c for handling velocity callbacks. Not sure why that was but here is my attempt at an implementation. Any feedback on the float only implementation or on this implementation is welcome.

Note: This compiles but waiting for feedback before testing.

### DIFF
--- a/sw/airborne/subsystems/ins/ins_int.c
+++ b/sw/airborne/subsystems/ins/ins_int.c
@@ -276,15 +276,13 @@ void ins_int_propagate(struct Int32Vect3 *accel, float dt)
   struct Int32Vect3 accel_meas_ltp;
   int32_rmat_transp_vmult(&accel_meas_ltp, stateGetNedToBodyRMat_i(), &accel_meas_body);
 
-  float z_accel_meas_float = ACCEL_FLOAT_OF_BFP(accel_meas_ltp.z);
-
   /* Propagate only if we got any measurement during the last INS_MAX_PROPAGATION_STEPS.
    * Otherwise halt the propagation to not diverge and only set the acceleration.
    * This should only be relevant in the startup phase when the baro is not yet initialized
    * and there is no gps fix yet...
    */
   if (ins_int.propagation_cnt < INS_MAX_PROPAGATION_STEPS) {
-    vff_propagate(z_accel_meas_float, dt);
+    vff_propagate(ACCEL_FLOAT_OF_BFP(accel_meas_ltp.z), dt);
     ins_update_from_vff();
   } else {
     // feed accel from the sensors
@@ -510,10 +508,8 @@ static void vel_est_cb(uint8_t sender_id __attribute__((unused)),
                        float x, float y, float z,
                        float noise __attribute__((unused)))
 {
-
+#if USE_HFF
   struct FloatVect3 vel_body = {x, y, z};
-  static uint32_t last_stamp = 0;
-  float dt = 0;
 
   /* rotate velocity estimate to nav/ltp frame */
   struct FloatQuat q_b2n = *stateGetNedToBodyQuat_f();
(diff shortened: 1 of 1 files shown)


kirkscheper:
@rmeertens what do you think?

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -510,37 +508,43 @@ static void vel_est_cb(uint8_t sender_id __attribute__((unused)),
...
+  QUAT_INVERT(q_b2n, q_b2n);
+  struct Int32Vect3 vel_ned;
+  int32_quat_vmult(&vel_ned, &q_b2n, &vel_body);
 
-  b2_hff_update_vel(vel,  Rvel);
-  ins_update_from_hff();
-#else
-  ins_int.ltp_speed.x = SPEED_BFP_OF_REAL(vel_ned.x);
-  ins_int.ltp_speed.y = SPEED_BFP_OF_REAL(vel_ned.y);
+  ins_int.ltp_speed.x = vel_ned.x;
```
flixr: vel_ned is float, but ins_int.ltp_speed is in fixed point...

kirkscheper: The git diff is messed up, check line 535

flixr: ah right, then I would suggest renaming it `vel_ned_i` or so to make it more readable...

flixr:
Just wondering in general: what prompted you to change it to do the rotation to ned in int?
Apart from that, it looks like the same as before....

kirkscheper:
I just thought if the file is called ins_int that there should be an
integer option. Also I am always sceptical of float implementations now
after the utm debacle :p

I was looking at this file today cause @rmeertens was having this ins walk
issue where the ins ltp position would drift to (0,0) after a while with
opticflow velocity inputs (think he mentioned it on gitter). One theory was
that rounding errors would play a part, not sure. As I said, I first wanted
to get some feedback then I will test the relative accuracy of the
computations for typical operation.

We had a couple other theories for the drift which @rmeertens will
investigate this week (expect more pull requests :) )

flixr:
You will always have drift when you only use velocity measurements... that is unavoidable...

kirkscheper:
Yea... but should you per say always drift to 0,0? It's just a but odd,
we're making sure it isn't a real bug somewhere.

flixr:
No, it should be a rather random drift... if you repeatedly see it drifting towards 0 and more or less staying there, then there is definitely something fishy ;-)

kirkscheper:
I found an error in int32_quat_vmult, it overflows. I will make a PR shortly. I guess that's one reason an int version wasn't used :p

With @rmeertens we determined that the most likely cause of the walk to 0,0 is that the velocity callback is sent at a fixed frequency of 512Hz regardless of if there was an update to the estimate, this makes dt quite small and given the relatively low resolution of the position int, you get some rounding errors. @rmeertens will make a pull request to change this behavior. It is still interesting to note somewhere that if dt gets small that odometry actually becomes less reliable using the int ins.

Additionally, I found that the int rotation has an average error of 0.014% to the value computed by the float rotation. Is this integer method interesting to anyone or shall I just close this? 

Let me know!

flixr:
The velocity callback get's called when there is a new `VELOCITY_ESTIMATE` which should of course only be sent if there is actually a new measurement... which module producing this message are you using?

IMHO using the int version to do the rotation doesn't really bring much benefit (especially since all recent autopilots have a floating point unit)...
Overflows still shouldn't happen of course...

kirkscheper:
Again there is always a difference between what should be and what is in our code... I will leave the details of the vel_est fix to Roland.

PR #1722, opened 2016-06-14
[fix] increase quat size to avoid overflow

kirkscheper:
original rotation would overflow for some rotations. I extended the variable size to ensure it all fits. Simple fix.


flixr:
Thx!

You don't by any chance want to extend the math unit tests to cover things like this?
See [tests/math/test_pprz_math.c, etc](https://github.com/paparazzi/paparazzi/tree/master/tests/math)

kirkscheper:
Sneaky sneaky Felix. Yea, no prob, will do.

Kirk

flixr:
Merging this already...
but having more unit tests for the math lib, state interface, etc... would be really great!
Any additional tests there will be greatly appreciated!

PR #1723, opened 2016-06-14
[xml] Created a simple Bebop2 JSBSim model.

deonblaauw:
Created a simple JSBSim model for the Bebop2 that works in NPS. Might not be 100% accurate (yet), but it does not go unstable at least.


flixr:
Nice! We could really use a better model for the bebop.
It just seems like it has basically no yaw effectiveness any more...
Did you do some sort of simple system identification to get the inertia, etc?
How did you come up with the values for the moment coupling (which seems way too weak now)?

deonblaauw:
So the problem I had was that there was no stable model for the Bebop 2. For the first iteration, I decided to weigh the Bebop 2 for the mass, and to estimate the MOI from equations and papers I found online. For the coupling, I just thumb sucked it. For the statix thrust, I basically tried to match flight test data. For the drag, I used the good old common drag coefficient of 0.02, and just compensated for the smaller fuselage (mostly a guestimate). 

When it comes to iteration #2 of the model, I plan to:
1. Improve the MOI by performing the pendulum test (http://physics.stackexchange.com/questions/177373/approximating-the-moment-of-inertia-of-a-quadcopter)
2. Improve the coupling by matching the NPS dynamic response with flight test data.
3. Improve the drag coefficient from flight test data.
 
I'm more used to coding models like these from 1st principles, and since JSB is quite new to me, I'm going to take this step by step :)

The current model is by NO means perfect, but at least you can validate basic flight plans with it (landing works).

flixr:
:smile: way to go!

The way the yaw moment coupling is set up in the other simple models is to add two forces perpendicualr to the axis with 5cm away from the motor location (so together they produce a moment with one tenth of the thrust).
So far you set both forces for each "moment couple" to the same location as the motors are (zero lever arm), so they cancel each other out and you basically have no yaw moment in the end...

deonblaauw:
Yeah, noticed that. I first wanted something that at least flies around without going crazy, then I'll fine-tune it. I'm not happy with the zero moment coupling at the moment, will try to improve it sometime during the week :+1:

flixr:
@deonblaauw did fix the lever arms for the yaw moment simulation?
Would be really nice to have a better/fixed model for the bebop!

deonblaauw:
Was a little pre-occupied with the whole Hybrid thing, but man, need to get back to this. I was playing around with some values, will focus on it a little this week / next week sometime and will let you know! Thanks for the follow-up man!

flixr:
@deonblaauw ping - any updates here?

flixr:
@rmeertens @kirkscheper and others using the Bebop: anyone willing to pick up where @deonblaauw left off here?

podhrmic:
Anyone up for the task? Otherwise I would close it since it is not going to be finished.

gautierhattenberger:
before closing: is the file proposed here is flying a little or not at all ? Maybe is worth keeping the file if some parameters are already better than in the default model.

PR #1724, opened 2016-06-15
more guidance guided options

rmeertens:
Now it is possible to:

- Keep current position instead of velocity zero when swithing to guided
- Set the max pos and max speed error: resulting in a slower moving drone (whoo, indoor flight!)

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
@@ -250,6 +250,10 @@ void guidance_h_mode_changed(uint8_t new_mode)
       break;
 
     case GUIDANCE_H_MODE_GUIDED:
+#ifdef GUIDANCE_H_GUIDED_KEEP_POS_ON_SWITCH
+      guidance_h_set_guided_pos(stateGetPositionNed_f()->x, stateGetPositionNed_f()->y);
+#endif
+
     case GUIDANCE_H_MODE_HOVER:
 #if GUIDANCE_INDI
       guidance_indi_enter();
@@ -379,7 +383,7 @@ void guidance_h_run(bool  in_flight)
     case GUIDANCE_H_MODE_HOVER:
       /* set psi command from RC */
       guidance_h.sp.heading = guidance_h.rc_sp.psi;
-      /* fall trough to GUIDED to update ref, run traj and set final attitude setpoint */
+    /* fall trough to GUIDED to update ref, run traj and set final attitude setpoint */
 
     case GUIDANCE_H_MODE_GUIDED:
       /* guidance_h.sp.pos and guidance_h.sp.heading need to be set from external source */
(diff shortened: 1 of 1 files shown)


kirkscheper:
Keeping current position when switching to guided mode is already default behavior, see guidance_h_hover_enter.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -379,7 +383,7 @@ void guidance_h_run(bool  in_flight)
     case GUIDANCE_H_MODE_HOVER:
       /* set psi command from RC */
       guidance_h.sp.heading = guidance_h.rc_sp.psi;
-      /* fall trough to GUIDED to update ref, run traj and set final attitude setpoint */
+    /* fall trough to GUIDED to update ref, run traj and set final attitude setpoint */
```
kirkscheper: why the indent? try running fix code style.

kirkscheper:
@flixr what is the convention for describing what defines are for? @rmeertens I would add a one liner which says what the define is used for (in this case trivial but good practice as pprz has a shit load).

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
rmeertens: This is a result of fix code style ;)

rmeertens:
Now explained the parameters.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -489,8 +490,24 @@ static void guidance_h_update_reference(void)
   }
 }
 
-#define MAX_POS_ERR   POS_BFP_OF_REAL(16.)
-#define MAX_SPEED_ERR SPEED_BFP_OF_REAL(16.)
+// GUIDANCE_H_MAX_POS_ERR_METER sets how much distance the drone tries to overcome in the horizontal loop at the same time
```
flixr: Please use doxygen comments so that the description of the define shows up in the generated documentation....

flixr:
IMHO limiting the max pos and speed errors is really a bad hack... even worse than the `gh_set_max_speed`.
You should adjust the reference and limit it to your desired max there.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
rmeertens: Done

rmeertens:
Let me see if I can get your viewpoint: 
Let's say that I have a drone, and I want it to move ten meters forward. I go to guidance mode and say "move 10 meters forward". 
With the current master the drone will quickly go towards this setpoint with a huge speed, so I want to limit that. 
- My solution would be to merge this pull request and limit the max error... now the drone will go slowly towards the setpoint. 
- Would your solution be to slowly set the position in a module every time I want to go forward again for a bit?

kirkscheper:
No, he means to change the reference model that is used to generate the guidance trajectory. for example set GUIDANCE_H_REF_MAX_SPEED.

flixr:
Yes, and you can also change that at runtime via `gh_set_max_speed`

This is also not quite ideal as it is a simple cutoff.. even better would be to adjust the ref model parameters to only yield the desired max velocity.

But it is way cleaner than limiting the max error (which should only be there to prevent overflows, etc..).

rmeertens:
Looks like you are right! Defining REF_MAX_SPEED does indeed fix the problem. 
Shall I close this pull request? Or do we think this addition is a nice to have feature...

flixr:
I think this is an anti-feature.... rather if you run into these error limits in normal operation, they should be raised.

PR #1725, opened 2016-06-15
Fixed config course

rmeertens:
One less failing airframe. Will now look at the others.

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_course2016_bebop_avoider.xml
+++ b/conf/airframes/TUDELFT/tudelft_course2016_bebop_avoider.xml
@@ -6,11 +6,6 @@
   <firmware name="rotorcraft">
     <target name="ap" board="bebop"/> <!-- The type of Autopilot board hardware one uses -->
 
-    <target name="nps" board="pc">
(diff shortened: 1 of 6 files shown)


flixr:
not fixed yet...
Just run `CONF_XML=conf/airframes/TUDELFT/tudelft_course2016_conf.xml prove tests/aircrafts/`

rmeertens:
Mmm, strange. Seems to be caused by the NPS target which does not see the variables of orange avoider in the flightplan.... Any idea what could have caused this?

flixr:
The orange avoider module is currently only for AP target (until the computer vision stuff is implemented/fixed for NPS)...

flixr:
so unless we can get the compute rvision stuff fixed for NPS target soon, I would suggest to remove/comment the nps target for that airframe (as you can't use it anyway)

rmeertens:
This should work, right?

flixr:
probably, you can test it with the command I posted before...

rmeertens:
That command works ;) Now also removed the warnings.

rmeertens:
Also removed the failing NPS target of TUDELFT.

PR #1728, opened 2016-06-15
[modules] autoload modules from an other module

gautierhattenberger:
This is not a dependency mechanism. It is just equivalent to adding modules directly in an airframe file.

This should help to solve the issue in gps_modules branch in particular (#1625).


flixr:
Great, that helps a lot!
So it looks like you should add a `<autoload name="gps_nps"/>` element to the module xml, but I'm wondering if it wouldn't make more sense to use the same syntax as for `<depends>foo,bar</depends>`

flixr:
Hm.. I guess this way it is consistent with how modules are loaded otherwise and probably better to keep the parsing the same here (and it's already done ;-))

PR #1729, opened 2016-06-16
Derotatation as part of the image pipeline

rmeertens:
Shoutout to everybody interested in image processing in Paparazzi!
We often want to derotate our images, to track an object precisely, odometry, optical flow etc. 
Today I ran some experiments: between the image capture and the end of optical flow the euler angles can easily get a difference of 17 degrees off (not that useful for optical flow). Even between the start of an image capture and the end (like in this pull request) there can be an offset of about 7.5 degrees. 

I propose to let people choose how to deal with this interesting problem. One way to make it easy for them is to get the euler angels at the start of the image capture and at the end of the image capture. 

Please let me know what you think about this solution, if you all agree that this is the way to go I will update the image functions in all modules to accept these new arguments.

### DIFF
--- a/sw/airborne/modules/computer_vision/video_thread.c
+++ b/sw/airborne/modules/computer_vision/video_thread.c
@@ -38,7 +38,7 @@
 #include "lib/v4l/v4l2.h"
 #include "lib/vision/image.h"
 #include "lib/vision/bayer.h"
-
+#include "state.h"
 #include "mcu_periph/sys_time.h"
 
 // include board for bottom_camera and front_camera on ARDrone2 and Bebop
@@ -122,6 +122,10 @@ static void *video_thread_function(void *data)
                 vid->w, vid->h, vid->fps, 1000000.f / dt_us);
       }
     }
+    FloatEulers imageStartEulers;
+    imageStartEulers.phi=stateGetNedToBodyEulers_f()->phi;
+    iamgeStartEulers.theta=stateGetNedToBodyEulers_f()->theta;
+    iamgeStartEulers.psi=stateGetNedToBodyEulers_f()->psi;
 
     // Wait for a new frame (blocking)
     struct image_t img;
@@ -138,9 +142,13 @@ static void *video_thread_function(void *data)
       // use color image for further processing
       img_final = &img_color;
     }
+    FloatEulers imageEndEulers;
+    imageEndEulers.phi=stateGetNedToBodyEulers_f()->phi;
+    imageEndEulers.theta=stateGetNedToBodyEulers_f()->theta;
+    imageEndEulers.psi=stateGetNedToBodyEulers_f()->psi;
 
     // Run processing if required
-    cv_run_device(vid, img_final);
+    cv_run_device(vid, img_final,imageStartEulers,imageEndEulers);
 
     // Free the image
     v4l2_image_free(vid->thread.dev, &img);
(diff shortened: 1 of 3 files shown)


dewagter:
1) the current implementation of Euler-Start = the eulers at the end of the last processing
2) the current implementation of Euler-End = only correct if you had to wait for a frame, not if the frame was already available

I think you grab functions must be inside get_frame (in the other thread)

Then I would suggest to send only 1 struct of Frame-Orientation to the computer vision threads with the average, unless you manage to get the timing of the grabbing so accurate that you know when each pixel is grabbed.

rmeertens:
Good points! Will fix!

knmcguire:
Hi Roland,

This is definitely usefull for frame-to-frame computer vision. In optical flow module, this is done too but yes there is a delay. The measurement of the IMU is faster than the capturing of the image, so it must be delayed. 

I see that you are capturing the euler angles after the image is captured... isn't it more useful to do this before  (inside      v4l2_image_get(vid->thread.dev, &img)) ? Capturing IMU data is way faster than capturing the image.

sw/airborne/modules/computer_vision/video_thread.c
```diff
@@ -122,6 +122,10 @@ static void *video_thread_function(void *data)
                 vid->w, vid->h, vid->fps, 1000000.f / dt_us);
       }
     }
+    FloatEulers imageStartEulers;
+    imageStartEulers.phi=stateGetNedToBodyEulers_f()->phi;
```
fvantienen: You need to protect those variables to make it thread safe..
Also it is nicer to fetch the eulers with `stateGetNedToBodyEulers_f()` only once. Next to that I guess you could add it to the `image_t` as a parameter instead of al those seperate variables.

fvantienen:
Also a small note: it would be nice if you code in the same variable name style ;) This way you don't have these different type of camel-cased variables etc.

flixr:
From a higher level perspective it would IMHO make more sense to rather attach a precise timestamp to each image and do the association to whatever vehicle state you (attitude, position, gimbal angle, etc) outside.
To that end we could have a buffer containing the poses for the last x seconds that you can query for the information at a particular timestamp.

flixr:
And yes, code style wise under_scored variables and functions are preferred. If in doubt stick to the style of the file you are editing!

fvantienen:
I'm not sure if I want to go for the option that you can query a specific vehicle state, because it has an extreme overhead. At first I thought it was better, but thinking about the frequency of logging these states and the frequency of the camera, to keep this in sync it is not that easy to optimize it.
Looking at the easier solution we could for now just log the state also in the video thread and save it in each image_t. Since that is the most interesting part, also for stitching etc. And you don't need to determine how much time you want to log all previous states and at which frequency.

flixr:
I'm just a bit afraid that people will want more and more states associated with the image (hence also changing the image_t struct or callbacks each time).
Doing it outside leaves you with more options it lets us circumvent the problem of thread safety.
But whatever works for you guys in the end...

knmcguire:
I don't think adding the rotation to the image structure would be necessary. In principle, it will never exactly match, so that might give the wrong idea as well. 

What was mentioned earlier, the idea of storing the state in a buffer, together with its timestamp, and comparing it to the timestamp of the image to get the most close matching state to image possible, sounds like a more suitable idea.

flixr:
@knmcguire yep, exactly... that's why I proposed it :-)
But it is more work and less "efficient"....

fvantienen:
We close this as the other implementation is already on it's way.

PR #1732, opened 2016-06-16
Added a pprz time to the images

rmeertens:
@fvantienen Can you continue with this?

### DIFF
--- a/sw/airborne/modules/computer_vision/lib/v4l/v4l2.c
+++ b/sw/airborne/modules/computer_vision/lib/v4l/v4l2.c
@@ -39,6 +39,9 @@
 
 #include "v4l2.h"
 
+#include <sys/time.h>
+#include "mcu_periph/sys_time.h"
+
 #define CLEAR(x) memset(&(x), 0, sizeof (x))
 static void *v4l2_capture_thread(void *data);
 
@@ -67,6 +70,8 @@ static void *v4l2_capture_thread(void *data)
 
     // Wait until an image was taken, with a timeout of tv
     int sr = select(dev->fd + 1, &fds, NULL, NULL, &tv);
+    uint32_t now_ts = get_sys_time_usec();
+
     if (sr < 0) {
       // Was interrupted by a signal
       if (EINTR == errno) { continue; }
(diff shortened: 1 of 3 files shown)


fvantienen:
I will continue with this, but for simplicity sake I will start with adding the rotation to the buffer as well. Maybe even GPS coordinates for nicer picture loggings.

flixr:
Is this still relevant with the merge of #1737 ?

fvantienen:
Yes this is especially made for this, because you need to relate this image to a paparazzi compatible time in order to get a pose..

flixr:
You should probably keep the timeval timestamp...

sw/airborne/modules/computer_vision/lib/v4l/v4l2.c
```diff
@@ -92,8 +97,8 @@ static void *v4l2_capture_thread(void *data)
     assert(buf.index < dev->buffers_cnt);
 
     // Copy the timestamp
-    memcpy(&dev->buffers[buf.index].timestamp, &buf.timestamp, sizeof(struct timeval));
-
+    dev->buffers[buf.index].timestamp = buf.timestamp.tv_usec;
```
flixr: keeping timestamp as a timeval struct would make more sense than assigning it the remainder of the second in us...

rmeertens: Mmm, I think I agree with you. @fvantienen , what do you think? At first I thought you wanted to save tv_sec*1000000+tv_usec, but this gives us a flight time of 4294 seconds, just over an hour, before it overflows. 
Shall we just keep the timestamp struct?

flixr: get_sys_time_usec() also overflows after 72min, but that is normally not a problem...

flixr:
Shall I change the timestamp back to the timeval struct or do you guys want to do it?
Just adding the time since startup in usec (pprz_timestamp) looks good to me...

rmeertens:
I will do it next week, that ok?

flixr:
sure

rmeertens:
Thanks!!

PR #1733, opened 2016-06-16
Second order notch filter

bartslinger:
Simple second order notch filter. I use it to attenuate the motor frequency from gyro measurements.


gautierhattenberger:
using float math function is probably enough here (`cosf`, `expf`)

bartslinger:
Did not know that these functions existed. If I'm correct, the cos and exp are for double while cosf and expf are for float. I changed it.

Issue #1698, opened 2016-06-01
set_nice_level in vision_thread


knmcguire:
Hi all,

Since the new vision_thread has been implemented, the functions called by it (cv_add) were running a lot slower before the implementation. I just figured out that the set_nice_level (line 170) in modules/computer_vision/vision_thread.c, is put on 10, which gives the vision thread a very low priority in the timing to reduce CPU usage I presume.

I definitely get the idea of controlling the nice level to protect the rest of the processes of the autopilot, for some vision algorithms this will do just fine, but for optical flow control, the loop rate should be as high as possible. A nice level of 10 drastically reduces the performance.

If I want to make a pull request about this, what is the best approach? To replace the fixed number in set_nice_level with a define (standard: 5, other can be specified in airframe file)?

Thanks in advance

dewagter:
Since this only applies to ardrone and bebop, and both have shown to fly well even with total vision processing overload, and that vision results are used in the control loop so vision becomes part of the (position) stabilization, I would suggest to (also) change the default value.
Obviously I do not mind if it can be changed (setting/airframe)

wilcoschoneveld:
I believe the pull request solved this issue.

PR #1735, opened 2016-06-17
[video_thread] Improvements to video_thread performance

wilcoschoneveld:
The default 'nice level' for the video_threads is 10, but this needs to be lower in order for some critical vision algorithms to work properly and reliably (such as optical flow hover control #1698).

Other possible improvements to video_thread;
- Dynamically change VISION_THREAD_NICE_LEVEL (via GCS setting)
- Being able to have some vision functions run with less fps than others (e.g. use RunOnceEvery)
- Improved fps regulation (e.g. smoothed average)
- Provide the vision functions with access to current frame time and fps

### DIFF
--- a/conf/modules/video_thread.xml
+++ b/conf/modules/video_thread.xml
@@ -9,6 +9,8 @@
 
       - Possibility to save an image(shot) on the internal memory (JPEG, full size, best quality)
     </description>
+
+    <define name="VIDEO_THREAD_NICE_LEVEL" value="5" description="Nice level for each separate video thread"/>
   </doc>
 
   <header>
--- a/conf/modules/video_thread.xml
+++ b/conf/modules/video_thread.xml
@@ -3,11 +3,10 @@
 <module name="video_thread" dir="computer_vision">
   <doc>
     <description>
-      Read video in a thread.
-      Only for Linux devices.
-      To be used in other modules for further processing (e.g. opticflow, QR code, streaming).
-
-      - Possibility to save an image(shot) on the internal memory (JPEG, full size, best quality)
+      Manager module which creates a video processing thread for each video device to be used. Only for Linux devices.
+      To be used in other modules for further processing (e.g. opticflow, QR code, streaming). Using 'cv_add_to_device'
+      from cv.h will register a processing function and initialize the video device if necessary. Thread priority can
+      be changed with VIDEO_THREAD_NICE_LEVEL.
     </description>
 
     <define name="VIDEO_THREAD_NICE_LEVEL" value="5" description="Nice level for each separate video thread"/>
--- a/sw/airborne/modules/computer_vision/video_thread.c
+++ b/sw/airborne/modules/computer_vision/video_thread.c
@@ -81,7 +81,7 @@ static void *video_thread_function(void *data)
   struct video_config_t *vid = (struct video_config_t *)data;
 
   char print_tag[80];
-  sprintf(print_tag, "video_thread-%s", vid->dev_name);
+  snprintf(print_tag, 80, "video_thread-%s", vid->dev_name);
 
   struct image_t img_color;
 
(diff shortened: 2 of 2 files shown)


wilcoschoneveld:
@guidoAI could you please repeat here what functionality is required?

sw/airborne/modules/computer_vision/video_thread.c
```diff
@@ -80,6 +80,9 @@ static void *video_thread_function(void *data)
 {
   struct video_config_t *vid = (struct video_config_t *)data;
 
+  char print_tag[80];
+  sprintf(print_tag, "video_thread-%s", vid->dev_name);
```
flixr: if you already have a fixed size char array: use `snprintf`

flixr:
Can you please also rebase it on master to not have other already merged PRs in here?

conf/modules/video_thread.xml
```diff
@@ -9,6 +9,8 @@
 
       - Possibility to save an image(shot) on the internal memory (JPEG, full size, best quality)
     </description>
+
+    <define name="VIDEO_THREAD_NICE_LEVEL" value="5" description="Nice level for each separate video thread"/>
```
rmeertens: Maybe explain what it is, does, and what values are acceptable?

sw/airborne/modules/computer_vision/video_thread.c
guidoAI: Besides the niceness level, it would be great if the priorities can be set between different vision functions. For example, textons should run at 5Hz and optical flow at 30Hz. With a periodic function this was / (is?) possible, but the vision process is a bit differently arranged now, with all vision functions added with cv_add.

flixr: You still have the option to only copy/save the image in the callback from the video_thread and then run the processing of it in another thread or schedule it as you want...

There is no requirement to run the processing in the callback (and hence the video_thread thread) unless you want to pass a modified version on to the next in the pipeline.

fvantienen:
This looks good 👍

flixr:
After a rebase and change to `snprintf` it also looks good to me...

But for future improvements on individual computer vision functions that should run with higher prio it IMHO makes way more sense to not do those computations in the callback from the video thread, but instead copy the img and run the computation in a separate thread (as I already pointed out above).

sw/airborne/modules/computer_vision/video_thread.c
guidoAI: Copying the reference would be fine, but really copying the image costs time. Can the img ref be used safely at the moment even when the processing will end after a new image comes in? Or is the image memory block freed before?

wilcoschoneveld: @guidoAI the image memory is freed after the computer vision pipeline function is finished, i.e. when all cv_functions are executed

sw/airborne/modules/computer_vision/video_thread.c:149

wilcoschoneveld:
For now I did a rebase to master and implemented the feedback

sw/airborne/modules/computer_vision/video_thread.c
rmeertens: Difficult problem. In this case optical flow fits perfectly in the normal image pipeline, but I can understand that algorithms that take longer might want to only have a reference to this image. 

One solution is to add some kind of variable to the image struct (or list) that indicates for each function that uses the image whether this function still needs the image...
At the moment the image gets read from the camera we can take a new piece of memory when the previous image is not fully processed yet.... 

Note that this: 
- might take a lot of memory (potential problem on the ARDrone). 
- is easy to get bugs in (especially with new students who forget to indicate that they are done). 


Maybe a circular image buffer could also work (solves the memory problem), but then your functions don't have control over when the image is re-used again. 

Anyway, this does not matter for this pull request I think.

conf/modules/video_thread.xml
```diff
@@ -3,11 +3,10 @@
 <module name="video_thread" dir="computer_vision">
   <doc>
     <description>
-      Read video in a thread.
```
flixr: First line of the description (until the dot) is used as brief description and displayed in http://docs.paparazziuav.org/latest/onboard_modules.html

PR #1736, opened 2016-06-17
[filters] delayed first order filter int

bartslinger:
This is a integer implementation of a first-order filter, with some extra's:
* saturation on increment rate
* delay using a buffer

I use this for an INDI implementation for a helicopter, which will follow later in a pull-request.


flixr:
Since this seems to be useful for other things as well, it would make sense to choose a more generic name than "heli_rate_filter"...
And maybe make the frequency a parameter instead of using the fixed `PERIODIC_FREQUENCY`

Out of curiosity: why do you want the delay?

bartslinger:
Do you have a suggestion for the name? I can change the frequency to parameter if you like. I need delay because I use this filter also to make a model of the actuator, which responds +- 18ms after a command is applied. This gives me the estimated position of the actuator.

flixr:
Puh... maybe something like `delayed_low_pass` or so??

bartslinger:
What do you think of `fancy_first_order_lowpass`?

flixr:
lol, then probably rather `delayed_first_order_lowpass` or something else that describes the functionality...

bartslinger:
Ok I'll go for that

bartslinger:
Done

bartslinger:
Done? :-p

gautierhattenberger:
looks good to me

flixr:
@bartslinger We could still give the struct and functions shorter (e.g. abbreviated) names if you want...
But if you are also happy with the long names we can merge it as it is now.

bartslinger:
I'll leave it like this for now. Then I continue with my next PR ;-)

PR #1737, opened 2016-06-17
Pose history

rmeertens:
In https://github.com/paparazzi/paparazzi/pull/1729 @knmcguire and @flixr had the idea of creating a pose history module... So here it is!

Currently it records at 512hz, and remembers its state for two seconds (1024 measurements). Searching in the buffer is not very efficient, but good enough for now to test the optical flow.

### DIFF
--- a/conf/modules/pose_history.xml
+++ b/conf/modules/pose_history.xml
@@ -1,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="pose_history">
+  <doc>
+    <description>Ask this module for the pose the drone had closest to a given timestamp</description>
+  </doc>
+  <header>
+    <file name="pose_history.h"/>
+  </header>
+  <init fun="pose_init()"/>
+  <periodic fun="pose_periodic()" freq="512" autorun="TRUE"/>
+  <makefile>
+    <file name="pose_history.c"/>
+  </makefile>
+</module>
+
--- a/conf/modules/pose_history.xml
+++ b/conf/modules/pose_history.xml
@@ -3,7 +3,7 @@
 <module name="pose_history">
   <doc>
     <description>Ask this module for the pose the drone had closest to a given timestamp</description>
-    <define name="POSE_HISTORY_SIZE" description="Length of the pose buffer"/> 
+    <define name="POSE_HISTORY_SIZE" value="1024" description="Length of the pose buffer"/> 
   </doc>
   <header>
     <file name="pose_history.h"/>
--- a/conf/modules/pose_history.xml
+++ b/conf/modules/pose_history.xml
@@ -3,12 +3,13 @@
 <module name="pose_history">
   <doc>
     <description>Ask this module for the pose the drone had closest to a given timestamp</description>
+    <define name="POSE_HISTORY_SIZE" description="Length of the pose buffer"/> 
   </doc>
   <header>
     <file name="pose_history.h"/>
   </header>
   <init fun="pose_init()"/>
-  <periodic fun="pose_periodic()" freq="512" autorun="TRUE"/>
+  <periodic fun="pose_periodic()" autorun="TRUE"/>
   <makefile>
     <file name="pose_history.c"/>
   </makefile>
--- a/sw/airborne/modules/pose_history/pose_history.c
+++ b/sw/airborne/modules/pose_history/pose_history.c
@@ -70,8 +70,8 @@ timeAndRotation get_rotation_at_timestamp(uint32_t timestamp)
 
   // Save the pose closest to the given timestamp and return this
   timeAndRotation closest_pose;
-  EULERS_COPY(closest_pose.eulers, location_history.ring_data[closestIndex].eulers);
-  RATES_COPY(closest_pose.rates, location_history.ring_data[closestIndex].rates);
+  closest_pose.eulers = location_history.ring_data[closestIndex].eulers;
+  closest_pose.rates = location_history.ring_data[closestIndex].rates;
 
 #if LINUX
   pthread_mutex_unlock(&pose_mutex);
--- a/sw/airborne/modules/pose_history/pose_history.c
+++ b/sw/airborne/modules/pose_history/pose_history.c
@@ -36,11 +36,6 @@
 #endif
 
 typedef struct {
-  uint32_t timestamp;
-  struct FloatEulers rotation;
-} timeAndRotation;
-
-typedef struct {
   uint32_t ring_index;
   uint32_t ring_size;
   timeAndRotation ring_data[POSE_HISTORY_SIZE];
@@ -57,7 +52,7 @@ pthread_mutex_t pose_mutex;
 /**
  * Given a pprz timestamp in used (obtained with get_sys_time_usec) we return the pose in FloatEulers closest to that time.
  */
-struct FloatEulers get_eulers_at_timestamp(uint32_t timestamp)
+struct timeAndRotation get_rotation_at_timestamp(uint32_t timestamp)
 {
 #if LINUX
   pthread_mutex_lock(&pose_mutex);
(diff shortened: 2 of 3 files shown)


sw/airborne/modules/pose_history/pose_history.h
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @file "modules/pose_history/pose_history.h"
+ * @author Roland Meertens
+ * Ask this module for the pose the drone had closest to a given timestamp
+ */
+
+#ifndef POSE_HISTORY_H
+#define POSE_HISTORY_H
+
+#include "state.h"
```
flixr: Should include "math/pprz_algebra_float.h" instead of state.h

flixr:
Nice, that is a good start to test!
Some more comments (e.g. what the timestamp is... unix time, pprz time in what unit...?) would be nice ;-)
plz also fix the code style...

flixr:
Also do you really want Eulers?

rmeertens:
Fixed the includes and comments. 
As for Eulers: you are right, but this is what is used now by the opticflow module. 

Would you like it if I also log FloatQuat in my struct timeAndRotation?

flixr:
Was just wondering... no need to buffer what is not used...
But maybe better call the function `get_eulers_at_timestamp` then...

rmeertens:
Done!

sw/airborne/modules/pose_history/pose_history.c
```diff
@@ -0,0 +1,102 @@
...
+ * Given a pprz timestamp in used (obtained with get_sys_time_usec) we return the pose in FloatEulers closest to that time.
+ */
+struct FloatEulers get_eulers_at_timestamp(uint32_t timestamp)
+{
+  uint32_t index_history = 0;
+  uint32_t closestTimeDiff = abs(timestamp - location_history.ring_data[0].timestamp);
+  uint32_t closestIndex = 0;
+
+  // Seach for the timestamp closes to the timestamp argument of this function.
+  for (index_history = 0; index_history < location_history.ring_size; index_history++) {
```
fvantienen: Would be nicer and more efficient to do a:
```
uint16_t i = 0;
uint32_t diff_min;
uint32_t diff_cur = abs(timestamp - location_history.ring_data[0].timestamp);
do {
    diff_min = diff_cur;
    diff_cur = abs(timestamp - location_history.ring_data[++i].timestamp);
} while(diff < diff_min);

struct FloatEulers closest_angels;
EulersCopy(closest_angels, location_history.ring_data[--i].rotation);
```

This will stop searching as it reaches it minimum. This is true because the ring buffer is ordered by time.

fvantienen: Note that this function is not thread safe(also your implementation). This should be fixed to make it useful.

sw/airborne/modules/pose_history/pose_history.c
```diff
@@ -0,0 +1,102 @@
...
+void pose_periodic()
+{
+  uint32_t now_ts = get_sys_time_usec();
+  timeAndRotation *current_time_and_rotation = &location_history.ring_data[location_history.ring_index];
+  current_time_and_rotation->rotation.phi = stateGetNedToBodyEulers_f()->phi;
+  current_time_and_rotation->rotation.theta = stateGetNedToBodyEulers_f()->theta;
+  current_time_and_rotation->rotation.psi = stateGetNedToBodyEulers_f()->psi;
+  current_time_and_rotation->timestamp = now_ts;
+
+  increase_index_location_history();
```
fvantienen: why is this a separate function?

sw/airborne/modules/pose_history/pose_history.c
```diff
@@ -0,0 +1,102 @@
...
+ * @author Roland Meertens
+ * Ask this module for the pose the drone had closest to a given timestamp
+ */
+
+#include "modules/pose_history/pose_history.h"
+#include <sys/time.h>
+#include "mcu_periph/sys_time.h"
+#include "state.h"
+
+#define POSE_HISTORY_SIZE 1024 // 1024 is two seconds
```
fvantienen: Maybe add an ifndef around this ;) And it is not always 2 seconds, this depends on the `PERIODIC_FREQ` which can be different than 512Hz.

conf/modules/pose_history.xml
```diff
@@ -0,0 +1,16 @@
...
+
+<module name="pose_history">
+  <doc>
+    <description>Ask this module for the pose the drone had closest to a given timestamp</description>
+  </doc>
+  <header>
+    <file name="pose_history.h"/>
+  </header>
+  <init fun="pose_init()"/>
+  <periodic fun="pose_periodic()" freq="512" autorun="TRUE"/>
```
fvantienen: Is it possible to set the frequency to `PERIODIC_FREQUENCY`?

fvantienen:
Roland you should add some mutex's to make it thread safe, but please add some `#if LINUX` or something around them so that this module still works for other platforms as well.

conf/modules/pose_history.xml
flixr: if you omit the freq attribute it will run at the full modules frequency (which is normally PERIODIC_FREQUENCY).

sw/airborne/modules/pose_history/pose_history.c
flixr: You should use the generated `POSE_HISTORY_PERIODIC_FREQ` to "compute" the history size from the desired buffer length in seconds.

sw/airborne/modules/pose_history/pose_history.c
rmeertens: Not sure if your proposed function is better. The ring buffer is indeed ordered by time, but index zero can be way after our timestamp that we want to find... this means that we might increase our diff for a while before we jump back in time and are closer to the point we want to be at.

fvantienen: You can start index_of_insert +1 and then do a round trip.. Note that it must be thread safe, and this new implementation is easier to make thread safe then your previous.

fvantienen:
Could you maybe also add the bodyRates? since I want to use that to de-rotate the image.

rmeertens:
@fvantienen these rates?

fvantienen:
Yes those 👍  Nice!

sw/airborne/modules/pose_history/pose_history.h
```diff
@@ -27,6 +27,13 @@
 #define POSE_HISTORY_H
 
 #include "math/pprz_algebra_float.h"
+
+typedef struct {
+  uint32_t timestamp;
+  struct FloatEulers rotation;
```
flixr: should probably be rather called `orientation` or `eulers`

sw/airborne/modules/pose_history/pose_history.c
```diff
@@ -0,0 +1,133 @@
...
+void float_eulers_copy(struct FloatEulers *to, struct FloatEulers *from);
+void float_eulers_copy(struct FloatEulers *to, struct FloatEulers *from)
+{
+  to->phi = from->phi;
+  to->theta = from->theta;
+  to->psi = from->psi;
+}
+
+void float_rates_copy(struct FloatRates *to, struct FloatRates *from);
+void float_rates_copy(struct FloatRates *to, struct FloatRates *from)
```
fvantienen: Those functions already exist:
`RATES_COPY` and `EULERS_COPY`. https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/math/pprz_algebra.h

flixr: If you copy full structs of the same type, you should simply use a normal assignment...

rmeertens: Done!

rmeertens:
Ok, now using the style guide (years of java programming ;) ) and fixed the assignment.

sw/airborne/modules/pose_history/pose_history.c
```diff
@@ -0,0 +1,115 @@
...
+ * @file "modules/pose_history/pose_history.c"
+ * @author Roland Meertens
+ * Ask this module for the pose the drone had closest to a given timestamp
+ */
+
+#include "modules/pose_history/pose_history.h"
+#include <sys/time.h>
+#include "mcu_periph/sys_time.h"
+#include "state.h"
+#if LINUX
```
flixr: should be `#ifdef __linux__`

conf/modules/pose_history.xml
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="pose_history">
+  <doc>
+    <description>Ask this module for the pose the drone had closest to a given timestamp</description>
+    <define name="POSE_HISTORY_SIZE" description="Length of the pose buffer"/>
```
flixr: missing value...

rmeertens:
Updated conf!

sw/airborne/modules/pose_history/pose_history.c
flixr: @fvantienen you still want something changed here?

Issue #508, opened 2013-08-23
fw: UTM zone derived from flight plan only


martinmm:
Issue: If a fixed wing autopilot uses a flight plan without NavSetGroundReference() and the flight plan has a geo location set within a different UTM zone than the actual one reported by GPS, the GPS message contains non valid data, there is no display in the gcs. Possibly navigation is degraded.

The initial UTM zone is derived from the flight plan. All GPS subsystems get their UTM zone from nav_utm_zone0 with default GPS_USE_LATLONG which is initially set to NAV_UTM_ZONE0 that comes from the flight plan.

The only way to change the UTM zone is by calling nav_reset_reference() through NavSetGroundReference() in the flight plan. (This also calls nav_update_waypoints_alt() and modifies all waypoints.)

It makes sense to use flight plans without a dynamic HOME for repeated flights at the exact same position. There should be a safe way to set the correct UTM zone in those cases.

PR #1741, opened 2016-06-17
replace nav_utm_xxx0 with state.utm_origin_f

kirkscheper:
The utm position seems to use nav_utm_zone0 to compute position when a utm gps isn't present. nav_utm_XXX0 are only updated with unit_t nav_reset_utm_zone(void) which was never called so seems like this won't work when changing zones.

All ins modules set the state.tum_origin_f directly so it seems more logical to use that variable instead of the nav variables.

Additionally, the rotorcraft nav doesn't use similar variables so I thought this may help with future merging of navigation firmware.

Note: This implementation only removes the nav variables, there still isn't an implementation to check the zone and update the utm_origin_f when the zone changes.

### DIFF
--- a/sw/airborne/arch/sim/sim_gps.c
+++ b/sw/airborne/arch/sim/sim_gps.c
@@ -11,9 +11,6 @@
 #include "math/pprz_geodetic_float.h"
 #include "math/pprz_geodetic_int.h"
 
-// currently needed for nav_utm_zone0
-#include "subsystems/navigation/common_nav.h"
-
 #include <caml/mlvalues.h>
 
 
--- a/sw/airborne/firmwares/fixedwing/fixedwing_datalink.c
+++ b/sw/airborne/firmwares/fixedwing/fixedwing_datalink.c
@@ -78,22 +78,22 @@ void firmware_parse_msg(void)
     case DL_MOVE_WP: {
       if (DL_MOVE_WP_ac_id(dl_buffer) != AC_ID) { break; }
       uint8_t wp_id = DL_MOVE_WP_wp_id(dl_buffer);
+      float a = MOfMM(DL_MOVE_WP_alt(dl_buffer));
 
       /* Computes from (lat, long) in the referenced UTM zone */
       struct LlaCoor_f lla;
       lla.lat = RadOfDeg((float)(DL_MOVE_WP_lat(dl_buffer) / 1e7));
       lla.lon = RadOfDeg((float)(DL_MOVE_WP_lon(dl_buffer) / 1e7));
-      lla.alt = MOfMM(DL_MOVE_WP_alt(dl_buffer));
       struct UtmCoor_f utm;
-      utm.zone = nav_utm_zone0;
+      utm.zone = state.utm_origin_f.zone;
       utm_of_lla_f(&utm, &lla);
-      nav_move_waypoint(wp_id, utm.east, utm.north, utm.alt);
+      nav_move_waypoint(wp_id, utm.east, utm.north, a);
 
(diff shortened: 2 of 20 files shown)


flixr:
IIRC this was done because we don't want to deal with zone crossings during flight.
So if we start close to a zone border we keep on using the original zone (i.e. we basically extend it).

@gautierhattenberger correct me if I'm wrong.

kirkscheper:
You don't have to handle once you don't get utm messages in your gps, if you do then you have a problem...

https://github.com/paparazzi/paparazzi/tree/master/sw/airborne/subsystems/gps.c#L311

flixr:
also nav_reset_utm_zone was added so it can be called from the flight plan...

kirkscheper:
Also, I don't really get this... 
https://github.com/paparazzi/paparazzi/tree/master/sw/airborne/subsystems/navigation/common_nav.c#L97

This assumes that the refrence frame is always reinitialised on the ground?

flixr:
Yes, that is currently the assumption... it's what is normally called in GeoInit, see  `NavSetGroundReferenceHere`

kirkscheper:
I am not sure that resetting the zone makes much sense. I think you really need to move your local origin. If you just move your zone won't your waypoints be wrong. You need something like the localize_waypoints from rotorcraft.

I didn't want to actually merge the rotorcraft and fixed wing though, this is just one of those things that I saw and is really scratching at my brain.

If you want to have true zone extension then we need to recompute the zone in the utm_float_from_gps if the GPS differs from state. This will not solve any GCS messaging things though. I would however recommend replacing the update zone with update origin and move all waypoints.

I would still use the state instead of the nav_ stuff

Let me know. I am off till Sunday, so will start again after that.

flixr:
@martinmm you requested the utm zone reset originall (see #508), do you have any feedback here?

kirkscheper:
ok, not quite done yet but current implementation has a function to update the local position of the waypoints when you are in a different zone than in the flight plan: nav_zone_extend_waypoints. I am just not really sure when to call this, I guess you want it when you get your first gps message or something...

I also added zone extending to the utm_int_from gps so that if the zone of the vehicle is different than that of the gps, the gps position will be zone extended into the requested zone.

flixr:
any updates here?

kirkscheper:
I was waiting to get some feedback on if this is useful (to me it seems
that way but I don't fly across utm zones often). If this is an ok approach
I can clean it up and test it.

flixr:
replacing the nav_utmx variables sounds good to me.
I just wouldn't remove the reset_utm_zone at this point...

@gautierhattenberger @martinmm @dewagter any comments here?

podhrmic:
@flixr @kirkscheper do you want to fix the conflicts so we can merge it?

gautierhattenberger:
I'm not really against this idea, but this PR needs some more work, not only solving conflicts.
And it will need some testing in real condition as it can have a few side effect.
Also I would like to have getter functions in the state interface for the reference point, instead of accessing it directly everywhere.

OpenUAS:
If conflict resolved and some adjustment made will test moving from zone 31 to zone 32. Maybe ENAC can test from zone 31 to zone 30? (yes it is a bit of a drive from Tolouse to UTM30. And maybe DLR Munich from 32 to 33?

OpenUAS:
@kirkscheper assigned, but feel free to re-assign. Will support you where we. Can also add more resources from the lab to close the issue. The 2SEAS project could benefit http://www.berisuas.eu/3i-2seas-uav as many more future ones.

kirkscheper:
I don't fly fw so can't test this. I will close the PR but keep the branch alive (just in case).

PR #1743, opened 2016-06-17
Lisa MX 2.1 chibios initial support

podhrmic:
Board files for Lisa MX 2.1
minor updates in the makefile/chibios arch


flixr:
General question:
How do we want the board file organization for chibios files?
Isn't it a bit confusing to have `boards/lisa_mx_2.1.h` and then also `boards/lisa_mx/v2.1/board.h` for chibios?

gautierhattenberger:
For ChibiOS, the file `board.h` is kind of mandatory, that's why I started to arrange it this way as it doesn't impact the libopencm3 version.
Actually, probably only lisa/MX and apogee boards (maybe Krooz?) will be impacted by this double configurations.

flixr:
I'm just wondering if we shouldn't name the directories more clearly... maybe something like
`boards/chibios/lisa_mx/v2.1/board.h` and maybe even `boards/libopencm3/lisa_mx_2.1.h`
Also now sometimes the version has a `v`, sometimes not...

podhrmic:
That sounds reasonable, btw I actually always found it confusing to have board header files in `board` dir, and not in `board/board_name` dir...
Lets settle for `board_name_vX.Y` as a naming convention, it is the same as what we already use for aspirin imu makefiles (the `v` before version number)

flixr:
> btw I actually always found it confusing to have board header files in board dir, and not in board/board_name dir...

Well, if all you need is one header file (as it is the case for most boards and how it was for all in the beginning), then an extra dir is kind of superfluous...

> Lets settle for board_name_vX.Y as a naming convention, it is the same as what we already use for aspirin imu makefiles (the v before version number)

It's just that the board makefiles currently don't have a `v` prefix.. And we probably don't want to change that (and hence what end-users have to put in your airframe file).
Also all existing board header files don't have that... 
But we could do that for the directory names for the `chibios/<board_name>/v<version>/`...

podhrmic:
> It's just that the board makefiles currently don't have a v prefix.. And we probably don't want to change that (and hence what end-users have to put in your airframe file).
Also all existing board header files don't have that...
But we could do that for the directory names for the chibios/<board_name>/v<version>/...

I think if we actually want to change we should change it everywhere - the backwards compatibility is nice, but this is for development branch so it shouldn't be a big deal. And if the names are consistent across the code, it will be actually easier. What do you think?

flixr:
IMHO it is not worth the effort to change it for all board files (and break every ones airframe files)...

podhrmic:
OK. Good to merge in that case?

flixr:
So you want to put the chibios files in a `chibios` subdir in a separate commit/PR?

podhrmic:
No, I thought we didn't want to change the names?

flixr:
I would still put all the chibios stuff in a chibios dir, just not touch the existing libopencm3 files...

podhrmic:
Hmm, but then it would really make sense to have another folder for `libopencm3` (and possibly other for future OSes). What if I move the header files to `lisa_mx/libopencm3/` and leave the names as they are? So have both chibios and libopencm3 folder for the respective boards...

flixr:
For me it would be enough to make a `chibios` subdir as IMHO it would then be clear that everything under that dir is for chibios only ... for all others we have only one arch anyway (either only linux or only libopencm3) so it is not really needed.
But as long as the boards that users have to put into their airframe file stay the same, I don't have a strong opinion if we put the others in `libopencm3` and `linux` dirs or not...

gautierhattenberger:
I don't have a strong opinion about the naming of new boards (or new arch for existing boards), but I would keep the old files in place (better things to do...). I don't mind to have either
- board_name/chibios/vX.X/
- board_name/chibios_vX.X/

if it makes things clear for users

podhrmic:
OK, I ll change it.

PR #1744, opened 2016-06-18
[rpm sensor] compiler error feedback

bartslinger:
Make it slightly more user-friendly. This thing does not work without pwm_meas.

### DIFF
--- a/sw/airborne/modules/sensors/rpm_sensor.c
+++ b/sw/airborne/modules/sensors/rpm_sensor.c
@@ -27,6 +27,10 @@
 #include "mcu_periph/pwm_input.h"
 #include "subsystems/electrical.h"
 
+#ifndef USE_PWM_INPUT
+#error rpm sensor requires module pwm_meas.xml
+#endif
+
 #if PERIODIC_TELEMETRY
 #include "subsystems/datalink/telemetry.h"
 


flixr:
Isn't that already covered by the module dependency?

bartslinger:
Ok I see what you mean, there is this `<depends>` section in the xml. What does this do exactly? I find the following compile error not very descriptive, that's why I suggested this addition.

```
modules/sensors/rpm_sensor.c: In function 'rpm_sensor_get_rpm':
modules/sensors/rpm_sensor.c:56:3: warning: implicit declaration of function 'get_pwm_input_period_in_usec' [-Wimplicit-function-declaration]
   uint32_t period_us = get_pwm_input_period_in_usec(RPM_PWM_CHANNEL) * RPM_PULSE_PER_RND;
   ^
modules/sensors/rpm_sensor.c:56:3: warning: nested extern declaration of 'get_pwm_input_period_in_usec' [-Wnested-externs]
<command-line>:0:17: error: 'PWM_INPUT1' undeclared (first use in this function)
modules/sensors/rpm_sensor.c:56:53: note: in expansion of macro 'RPM_PWM_CHANNEL'
   uint32_t period_us = get_pwm_input_period_in_usec(RPM_PWM_CHANNEL) * RPM_PULSE_PER_RND;
                                                     ^
<command-line>:0:17: note: each undeclared identifier is reported only once for each function it appears in
modules/sensors/rpm_sensor.c:56:53: note: in expansion of macro 'RPM_PWM_CHANNEL'
   uint32_t period_us = get_pwm_input_period_in_usec(RPM_PWM_CHANNEL) * RPM_PULSE_PER_RND;
                                                     ^
CC /home/bart/paparazzi/var/aircrafts/Helisa/ap/peripherals/sdcard_spi.o
make[1]: *** [/home/bart/paparazzi/var/aircrafts/Helisa/ap/modules/sensors/rpm_sensor.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory `/home/bart/paparazzi/sw/airborne'
make: *** [ap.compile] Error 2
make: Leaving directory `/home/bart/paparazzi'
```

flixr:
Ah, right... currently `<depends>` only prints a warning if the dependency is not already added, which you should also see in your case.

I guess we should change that to an error?

From my point of view goal is still to automatically load dependencies (if the target of the parent module matches and is hence not unloaded)...

bartslinger:
I did not get a warning from the `<depends>`

flixr:
Really? Is that config on github? I'd like to reproduce (and fix) that then...

bartslinger:
Oh lol I did get the warning, but as an average user I failed to scroll up and notice it. It went out of the screen quickly since it was followed by a lot of other lines.

fvantienen:
Can we close this with the fix of #1748 ?

bartslinger:
Yes I like that a lot better

PR #1745, opened 2016-06-18
Bebop fw3.2.0 fix

kevindehecker:
Motors got reversed, but are apparently still not the same as with bebop 2 (or I don't understand something). In any case; tested this with a bebop 1 and it flies in att mode.

 Also, pprz center now gives user feedback on some problems during upload. 
- checks on firmware version to be 3.2.0
- tells user to press button 4 times (2 times does not work with bebop 1 @ 3.2.0)
- tells user to check wifi conntection

### DIFF
--- a/sw/tools/parrot/bebop.py
+++ b/sw/tools/parrot/bebop.py
@@ -122,16 +122,22 @@ elif args.command == 'upload_file_and_run':
     # Split filename and path
     f = parrot_utils.split_into_path_and_file(args.file)
 
-    print("Kill running " + f[1] + " and make folder " + args.folder)
-    parrot_utils.execute_command(tn,"killall -9 " + f[1])
-    sleep(1)
-    parrot_utils.execute_command(tn, "mkdir -p /data/ftp/" + args.folder)
-    print('Uploading \'' + f[1] + "\' from " + f[0] + " to " + args.folder)
-    parrot_utils.uploadfile(ftp, args.folder + "/" + f[1], file(args.file, "rb"))
-    sleep(0.5)
-    parrot_utils.execute_command(tn, "chmod 777 /data/ftp/" + args.folder + "/" + f[1])
-    parrot_utils.execute_command(tn, "/data/ftp/" + args.folder + "/" + f[1] + " > /dev/null 2>&1 &")
-    print("#pragma message: Upload and Start of ap.elf to Bebop succesful !")
+    #check firmware version
+    v = parrot_utils.check_version(tn, '').strip()
+    print("Checking Bebop firmware version... " + v )
+    if v != '3.2.0':
+        print("Error: firmware version on bebop is incorrect!")
+    else:
+        print("Kill running " + f[1] + " and make folder " + args.folder)
+        parrot_utils.execute_command(tn,"killall -9 " + f[1])
+        sleep(1)
+        parrot_utils.execute_command(tn, "mkdir -p /data/ftp/" + args.folder)
+        print('Uploading \'' + f[1] + "\' from " + f[0] + " to " + args.folder)
+        parrot_utils.uploadfile(ftp, args.folder + "/" + f[1], file(args.file, "rb"))
+        sleep(0.5)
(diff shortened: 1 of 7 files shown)


flixr:
So they changed the motor ordering with a new firmware version? That is really annoying...

Won't this break it for people still using an older version?

kevindehecker:
Quite annoying yeah. Undoing a firmware upgrade also seems quite hard.

This will break things for older firmware, however, now it is broken for the current firmware. Which is partly why I put the check there, to prevent future unintended crashes.
I have no problem with supporting additional firmware versions, but I did not see an easy way to do so. Supporting the current one (it has already been on there for a while now)  asap is most important I think, so that's the reason for this PR.

flixr:
The really annoying thing is that people who based their own airframe file on the one in the repo before this change won't "get the fix"...
It would be great if we could detect the firmware version onboard (e.g. in board_init) and use the appropriate motor mixing.

kevindehecker:
Right, well I'd say the problem there is that we have all these copy-pasta for air frames that are 99% the same. Why don't we have some kind of possibility to include and extend base air frames?

flixr:
We do have that functionality. But that only reduces the number of cases where it would be a problem, but doesn't eliminate it.
IMHO the only way to "properly" deal with it would be to do this at runtime.

kevindehecker:
Fine. Let's see how long that takes then.

flixr:
I'm not saying we shouldn't merge this in the meantime...
Just that IMHO this is not a good long term solution, and was kind of hoping that one of you guys could implement the runtime detection (as you have the hardware available to actually test it).
So if you don't have time to do that now, I would say we merge it...

sw/tools/parrot/bebop.py
```diff
@@ -122,16 +122,22 @@ def bebop_status():
...
-    parrot_utils.uploadfile(ftp, args.folder + "/" + f[1], file(args.file, "rb"))
-    sleep(0.5)
-    parrot_utils.execute_command(tn, "chmod 777 /data/ftp/" + args.folder + "/" + f[1])
-    parrot_utils.execute_command(tn, "/data/ftp/" + args.folder + "/" + f[1] + " > /dev/null 2>&1 &")
-    print("#pragma message: Upload and Start of ap.elf to Bebop succesful !")
+    #check firmware version
+    v = parrot_utils.check_version(tn, '').strip()
+    print("Checking Bebop firmware version... " + v )
+    if v != '3.2.0':
+        print("Error: firmware version on bebop is incorrect!")
```
flixr: maybe also change that to something like "Error: please upgrade your Bebop firmware to version 3.2.0"?

flixr:
@kevindehecker it would be nice if you could also post an announcement on the mailing list...

kevindehecker:
Thanks for merging and sorry for the late reply, busy busy...

I suppose the goal of the mail would be to inform those who have their Bebop airframe not in Master. But., I'm not on the mailinglist (not planning to be either), and I don't think that is the way to communicate this:
a) this seems more like a release notes type of thing
b) not everyone reads the mailinglist
c) chance is small you read the mail when you cause the problem

 If think it would be better to append the information into the warning message when incorrect firmware is detected:

print("Error: please upgrade your Bebop firmware to version 3.2.0!"
if v != '2.x'
   print("Warning: after updating to 3.2, motors are reversed. If your airframe was not in master, check issue 1745 on github.")  //TODO: remove this message after release 5.x

flixr:
> I suppose the goal of the mail would be to inform those who have their Bebop airframe not in Master. 

Which is basically every one except TU Delft...

> But., I'm not on the mailinglist (not planning to be either), and I don't think that is the way to communicate this:

That's a shame...

> a) this seems more like a release notes type of thing

Sure it will be in the release notes, but informing people that they will have to look out for this problem if they update their bebop would be nice.

> b) not everyone reads the mailinglist

But more than those that read all issues and pull requests...

kevindehecker:
OK, sent the mail.

flixr:
:+1: thanks a lot!

PR #1748, opened 2016-06-19
[gen_modules] exit with error if dependency is not satisfied

flixr:
Exit with an error instead of only printing a warning if a module dependency is not satisfied.

See also #1744


fvantienen:
This seems ok :+1:

PR #1749, opened 2016-06-19
[rotorcraft] default ARRIVED_AT_WAYPOINT to 1m

flixr:
By default mark waypoint as arrived at distance of 1m.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/navigation.c
+++ b/sw/airborne/firmwares/rotorcraft/navigation.c
@@ -110,7 +110,7 @@ static inline void nav_set_altitude(void);
 
 /** minimum horizontal distance to waypoint to mark as arrived */
 #ifndef ARRIVED_AT_WAYPOINT
-#define ARRIVED_AT_WAYPOINT 3.0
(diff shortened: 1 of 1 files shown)


flixr:
sure they can... question is what is the best default?
3m is really a lot!

bartslinger:
If I fly outside with GPS, I only use a PD controller for position (no integrator). The gains are quite low because otherwise it may get unstable. If there is a lot of wind, my helicopter may never fly into this smaller radius. I've actually heard stories about people who had a failed mission because the drone got stuck just outside the radius. Changing the default here is maybe not such a good idea.

bartslinger:
I have no problems with changing the default, but maybe you can add the setting in each airframe at the old value? Then only new airframes are affected, and people will actually notice that it has changed.

podhrmic:
I would also be in favour of keeping the default - since it seems to be working fine (I don't remember how close we get to the waypoints during our fights but the default seems to be a good value). If it is configurable from the aiframe config file then it is easy to customize anyway...

gautierhattenberger:
I currently pretty happy with the 3m default value

Issue #1163, opened 2015-04-29
Camera hardware support for Parrot Bebop


OpenUAS:
This meta issue is to track progress of the Parrot Bebop support for the build in camera hardware
- [x] bottom camera
- [x] front camera
- [x] ISP video processor
- [x] H264 video encoder
- [x] GPU
##

OpenUAS:
Any new insights @fvantienen , if so plz assign it to milestone 5.8 if it is feasable else v6.0, but that would be less fun ;)

dewagter:
- GPU is supported, but what do you want to run on the GPU?
 - ISP and H264 require quite some library files.

flixr:
So ISP is implemented with #1750 thanks to @fvantienen

OpenUAS:
Big fat **THANKS** to @fvantienen for adding all the magic by sleepless nights spells. Superb! The next step, separation and refactoring is already ongoing for an even cleaner solution

PR #1750, opened 2016-06-19
Bebop camera's upgrade

fvantienen:
This upgrades the initialization of both camera's of the Bebop. It also includes an ISP configuration to convert front camera images to UYVY.
This is still a WIP, since my Bebop got bricked and couldn't test anymore. For example the "which" of the subdevice for the front camera must be `1` instead of `0`.


kirkscheper:
related to #1163

wilcoschoneveld:
Currently we cannot get the bottom camera working anymore with master (#1762). This branch does give us a working bottom camera (very high fps, but with very low resolution and incorrect aspect ratio). The front camera does not yet work.

We need something working in order to continue working on the IMAV competition, either by fixing the master branch (based on v4l2) or getting this pull request ready. What do you think?

fvantienen:
I almost have this PR ready, and that strange resolution etc. are just all settings, take a look at `mt9v117.c`. In that camera structure you can set all kind of settings. But if you take a look at my `opticflow_upgrade` there basically are all fixes for this branch already incorporated. I just need to separate them.

rmeertens:
@fvantienen Any progress?

fvantienen:
This is now fully functional and gives much better results for opticflow in the Bebop bottom camera. Also the front camera can now fully be configured :)

Still some documentation could be written about how this all works now, but I think @rmeertens will take a look at that and prolly write this on the wiki.

flixr:
Was it intentional to change the `ts` timestamp for the images again? If yes, why?
See also #1732

fvantienen:
Not really, but it was already in here. Though it is easier for copying and comparing and the overflow doesn't really matter that much.

flixr:
IMHO it makes more sense to keep the exact timeval timestamp...
For fps comparison you could also use the pprz_ts instead (us since startup)

fvantienen:
I could change it back, or maybe do it after this PR. It is indeed nicer to keep the original timeval, since it is more precise and holds more information.

fvantienen:
Can this be merged and then I will fix the timestamp after this?

flixr:
I would prefer if you fix the use of the timestamp here...
The timestamp itself is already correct in master, see https://github.com/paparazzi/paparazzi/commit/789e6eea9506de147dec8c78ab52fecda12185e8
Meaning with this PR you are changing it to the "wrong" version in v4l2.h and image.h

sw/airborne/modules/computer_vision/lib/v4l/v4l2.c
```diff
@@ -179,16 +180,22 @@ bool v4l2_init_subdev(char *subdev_name, uint8_t pad, uint8_t which, uint16_t co
...
   struct v4l2_requestbuffers req;
+  struct v4l2_fmtdesc fmtdesc;
+  struct v4l2_crop crp;
   CLEAR(cap);
   CLEAR(fmt);
   CLEAR(req);
+  CLEAR(fmtdesc);
+  CLEAR(crp);
+
+  printf("Size %d %d, Crop %d %d %d %d\r\n", size.w, size.h, crop.x, crop.y, crop.w, crop.h);
```
flixr: remove the debug printf?

flixr:
how about the bebop_cam_upgrade branch in my personal repo?

fvantienen:
I pushed your branch but also made one small extra commit to remove some warnings ;) Tnx!

wilcoschoneveld:
Awesome!

PR #1751, opened 2016-06-20
Updated contribution guidelines

rmeertens:
Hey everybody, 

Today we had a short discussion in our lab about how open source projects like Paparazzi should work. My proposal is to update our current workflow to get better software earlier in the Paparazzi project, and create issues for "smaller" fixes that are assigned to somebody. 

Examples of what I propose would be done: 
- we merge https://github.com/paparazzi/paparazzi/pull/1745 , then create issues assigned to Kevin to do everything at runtime. 
- we merged https://github.com/paparazzi/paparazzi/pull/1737 three days ago (since it was working perfectly), but created multiple issues (i.e. I used the wrong casing, wrong variable names, that I should add rotation rates, that I made a mistake with the Linux defines etc.). 

I am fully aware that having a pull request merged lowers your priority of working on it, but in my eyes this lower priority is justified by having faster improvements to our always changing autopilot. However, this would be a nice topic to discuss on!

Please tell me what you think about this proposed update.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,8 +45,7 @@ Fixes issue #123
 ## Submitting Changes
 
 * Push your changes to a topic branch in your fork of the repository.
-* Submit a [pull request](https://github.com/paparazzi/paparazzi/compare/) to the repository in the paparazzi organization.
-
+* Submit a [pull request](https://github.com/paparazzi/paparazzi/compare/) to the repository in the paparazzi organization. If this change improves the current code we will merge it immediately, but will create several issues with problems that were in your code. These issues will be assigned to you, and you will be asked to create a new pull request to fix these new issues. 
 # Additional Resources
 
 * [More information on contributing](http://wiki.paparazziuav.org/wiki/Contributing)


kevindehecker:
I agree with the first part of the statement, to a point. I think PR's should be merged asap if they prevent imminent crashes or serious issues. Not crashing is the most important part of an autopilot, I'm invested in that part of the equation.

I don't agree with the second part, at least not the part where the issue gets assigned to the topic opener. I'm not interested in code quality improvement (being honest here, I know it is very important to pprz), therefor the chance is small that I will do it. Assigned or not. I do fully agree we should make issues about the resulting code improvements comments, and more importantly have a good release system (also with short term milestones!) such that making these issues actually makes sense.

kirkscheper:
I don't really agree with this and I don't think this was the result of our discussion today. The proposition would result in double work or in issues that users will never come back to fix as @kevindehecker mentioned. I agree that if it is a vital bug fix that it should be accepted pretty quickly but I don't see that as a big issue currently.

@rmeertens mentioned if the PR improves pprz, I hope that every PR improves the code base else it probably shouldn't be a PR. I think the PRs to your personal airframes doesn't really affect the code base and is infact quickly included so that already goes pretty good. 

What I think is a problem of communication. Communication of up coming developments by core developers and of the on going developments by periphery devels. We need to discuss more often what are the issues we have and how to fix them before/during the fix so that the core developers are part of the discussion earlier leading to less work in the end.

I am a fan of making an issue first to announce the problem you are fixing and why. This should ensure that multiple people are not working on the same issue (seldom) and to get feedback on potential solutions already in pprz. Alternatively, you can make a (semi-) empty PR which describes the issue with a rough draft of the proposed change in order to get early feedback on the work. 

The bigger question we have to answer before we get into solutions is why exactly do some PRs take a while to get accepted. I don't think they are majority issues with formatting...

@flixr and @gautierhattenberger What from your point of view are the main reasons for PRs to be delayed from being merged and what do you think would aid this process.

flixr:
IMHO @kirkscheper summed it up pretty well. And merging half-backed things early is not the solution.

So clearly stating what the goal is, what was changed to achieve that goal and why that solution was chosen helps a lot. As well as scoping the pull requests well, commits in logical units and not changing many different things at once (this works well in most cases).

Reasons for delay are IMHO mostly:
- "uncomplete" PRs, that fix things for some cases, but make it worse for others...
- quality (and style) issues that require re-work and then re-checking the same stuff often...
- more complex changes that just need more time to review and think through all of the cases

And the biggest reason is probably that there are many ways to Rome and it often takes some discussion to tease out which one makes the most overall sense... but that is just a normal (and very vital) part of development.

gautierhattenberger:
Looking at the current opened PR, it seems that the main reason is that they are "still in discussion" (sometimes on major design issues, sometimes on minor formatting points). Then it seems equally split into "feedback timeout", "still incomplete", "broken" or actually "waiting for final approval" (they are not so many).
Most of the PR are validated by @flixr and sometimes by me, but it is a bit difficult to do all of them especially when it concerns code we (at least me) are not using (video stuff at the moment, at least 3 PR could be merged or are almost merge-able). I really don't have the time to get into this right these days, but I guess it is the same for other devs.
Now, if we want to make the waiting time shorter, I guess splitting the PR correctly, following the coding from the start, and targeting relevant devs for each parts would probably help a lot.

flixr:
And of course it saves time if other people look at pull requests as well and either point out things that still need to be fixed/improved or give an honest LGTM.

flixr:
So can we close this?

rmeertens:
Yes, everybody who wanted to give an opinion gave their opinion I think!

PR #1752, opened 2016-06-20
[stabilization] helicopter indi controller

bartslinger:
This is an INDI controller for a small helicopter. It was flight-tested on a walkera genius cp v2. When it is merged, I can add documentation to the wiki on how to use it. It is quite different from the other INDI controller, so therefore I suggest to not combine it with the other one.


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,656 @@
...
+
+#endif // PERIODIC_TELEMETRY
+
+static inline void indi_subtract_vect(int32_t _out[], int32_t _in1[], int32_t _in2[])
+{
+  for (uint8_t i = 0; i < INDI_DOF; i++) {
+    _out[i] = _in1[i] - _in2[i];
+  }
+}
+
```
kirkscheper: There seems to be some functions that have this functionally, check sw/airborne/math/pprz_simple_matrix.h. 

If you have time, you could add a math file which has the wrappers for the matrix computations for vectors. I think these functions would be interesting for others not just for indi, you would just make the INDI_DOF the input t the general function. 

If you don't have time (or the interest) maybe make an issue for to do this.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,656 @@
...
+  /* Multiply error with inverse of actuator effectiveness, to get delta u (required increment in input) */
+  indi_matrix_multiply_vector(c->du, c->invG, c->error);
+
+  /* Bitshift back */
+  c->du[INDI_ROLL]  >>= 16;
+  c->du[INDI_PITCH] >>= 16;
+  c->du[INDI_YAW]   >>= 16;
+  c->du[INDI_THRUST] >>= 16;
+
```
kirkscheper: bitshift back from what? Where does the 16 come from? If it is a fixed bit depth can you add a define which describes the value like INT32_QUAT_FRAC

bartslinger: Ah good point. The bitshift forward is hidden in the inverse effectiveness invG, but this cannot be related to any existing defines. Should I add something like INT32_INVG_FRAC ?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
bartslinger: Ok I can change the matrix thing to MAT_MUL_T
What would be the best location to add the vector functions (with length as variable), pprz_algebra.h?

kirkscheper: I think in the pprz_simple_matrix.h. Or maybe in a pprz_vector.h. @flixr?

flixr: There are n-dim `float_vect_x` functions in pprz_algebra_float.h, so I guess a fixedpoint equivalent should go into pprz_algebra_float.h

bartslinger: https://github.com/paparazzi/paparazzi/pull/1758

bartslinger:
I did not use MAT_MUL because it assigns a float zero. I'm not 100% sure about the double pointer stuff here.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,618 @@
...
+{
+  for (uint8_t i = 0; i < INDI_DOF; i++) {
+    _out[i] = update_butterworth_2_low_pass_int(&actuator_lowpass_filters[i], _in[i]);
+  }
+}
+
+static inline void indi_apply_measurement_butterworth_filters(int32_t _out[], int32_t _in[])
+{
+  for (uint8_t i = 0; i < INDI_DOF; i++) {
+    _out[i] = update_butterworth_2_low_pass_int(&measurement_lowpass_filters[i], _in[i]);
```
EwoudSmeur: Why have twice the same function with a different filter struct? can that not be an argument?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,618 @@
...
+  }
+}
+
+/**
+ * @brief stabilization_attitude_heli_indi_set_steadystate_pitch
+ * @param pitch neutral pitch angle [deg].
+ *
+ * Change the neutral pitch angle.
+ */
+void stabilization_attitude_heli_indi_set_steadystate_pitch(float pitch)
```
EwoudSmeur: This function and the next are never used, does it make sense to keep them?

bartslinger: I use it in combination with a settings file, but wanted to keep that for the next PR

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
bartslinger: Good point, will change that

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,618 @@
...
+  /* Transform yaw into a delta yaw while keeping filtered yawrate (kinda hacky) */
+  int32_t filtered_measurement_vector[INDI_DOF];
+  int32_vect_copy(filtered_measurement_vector, c->filtered_measurement[INDI_NR_FILTERS - 1], INDI_DOF);
+  static int32_t previous_filt_yawrate = 0;
+  filtered_measurement_vector[INDI_YAW] = 512 * (c->filtered_measurement[INDI_NR_FILTERS - 1][INDI_YAW] -
+                                          previous_filt_yawrate);  // = approximately yaw acceleration error
+  previous_filt_yawrate = c->filtered_measurement[INDI_NR_FILTERS - 1][INDI_YAW];
+
+  /* Obtain virtual control input with P controller on pitch and roll */
+  int32_t roll_virtual_control  = (heli_indi_gains.roll_p * att_err.qx)  / 4;
```
EwoudSmeur: why divided by 4?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,618 @@
...
+  Bound(c->u_setpoint[INDI_YAW], 0, MAX_PPRZ);
+  Bound(c->u_setpoint[INDI_THRUST], 0.15 * MAX_PPRZ, MAX_PPRZ);
+
+  /* Apply a compensator to the actuator setpoint to obtain actuator command */
+  c->apply_compensator_filters(c->command_out[__k], c->u_setpoint);
+
+  /* At the end, set 'previous' output to current output */
+  int32_vect_copy(c->command_out[__k - 1], c->command_out[__k], INDI_DOF);
+
+  /* Two correction angles, don't rotate but just add.
```
EwoudSmeur: why not rotate? Or why don't you use a non-diagonal control effectiveness matrix? Why else do you have a matrix?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
bartslinger: For a better resolution in the gains

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
bartslinger: The identification procedure gives me pitch_cmd -> pitch_rate and roll_cmd -> roll_rate. These values go directly into the matrix on the diagonal.
Rotating would reduce the control input with a cosine, but I don't want to reduce the control input in these axes. I just need to give some compensation in the other axis proportional to the first.

I agree that the compensations can also be included in the matrix (off diagonal) but these values then would have to be a function of the values on the diagonal to get the same result. I like to have it simply proportional to the command.

The reason I left the matrix in there is:
- If someone continues the project and investigates the couplings better, it might be useful
- If the controller is extended with vertical acceleration control, the effect of collective on yaw acceleration can be easily included.

flixr: would probably be helpful later to have this as comment in the code ;-)

EwoudSmeur: But still... You can just add the pitch command to the roll parameter estimation and vice versa, then it is very easy to fill that matrix. In fact, it should be the same values as what you have now below, if I am not mistaken. Or how did you determine those values?

bartslinger: I tried that first, but that did not give me a good fit on the data. That's why I went for this approach. It's not exactly the same, this is the difference:
In my case: A roll command is added which is proportional to the (delta) pitch command
With values off-diagonal: A roll command is added which is proportional to the pitch rate error.
What I am doing can also be achieved with just the off-diagonal values

Because fitting did not work, the values are determined empirically: Give pitch forward and watch what happens to roll. Change values until it goes straight ahead. It's also flyable without the compensation. Same procedure for roll with pitch compensation.

With respect to the PR, I prefer to leave it like this. It aligns with what I wrote in my thesis and it might be clearer for someone that continues with my project, because I also think that this decoupling needs more work to come up with a nicer solution.

bartslinger:
Ok this is weird, from the failed build log:

```
$ sudo add-apt-repository ppa:paparazzi-uav/ppa -y
$ sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa -y
Cannot add PPA: 'ppa:team-gcc-arm-embedded/ppa'.
Please check that the PPA name or format is correct.
The command "sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa -y" failed and exited with 1 during .
```

flixr:
That happens sometimes in travis, no idea why... maybe a caching problem on their build farm...
Restarting the build usually does the trick...

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
bartslinger: I removed duplicate code for notch filters. Its not really trivial to make the filter struct an argument without reducing the flexibility. Like it is now, its easy to replace one filter with another.

bartslinger:
I think this is now mergable

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,609 @@
...
+  int32_t alpha_thrust = actuator_model[INDI_THRUST].alpha;
+  _out[INDI_THRUST] = ((output_target << DELAYED_FIRST_ORDER_LOWPASS_FILTER_FILTER_ALPHA_SHIFT) - alpha_thrust *
+                       prev_thrust_out) / ((1 << DELAYED_FIRST_ORDER_LOWPASS_FILTER_FILTER_ALPHA_SHIFT) - alpha_thrust);
+  prev_thrust_out = _out[INDI_THRUST];
+
+  /* At the moment, thrust is not fully implemented */
+  //_out[INDI_THRUST] = _in[INDI_THRUST];
+}
+
+static inline void indi_apply_notch_filters(struct SecondOrderNotchFilter *filter, int32_t _out[], int32_t _in[])
```
flixr: all these apply functions should not be inline, since you want to use pointers to these functions later...

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.c
```diff
@@ -0,0 +1,609 @@
...
+
+  /* attitude error */
+  struct Int32Quat att_err;
+  struct Int32Quat *att_quat = stateGetNedToBodyQuat_i();
+
+  /* Add steady-state value to attitude setpoint, because heli has
+   * non-zero roll angle by default
+   */
+  struct Int32Quat corr_att_sp_quat; // Corrected attitude setpoint
+  INT32_QUAT_COMP_INV(corr_att_sp_quat, stab_att_sp_quat, sp_offset);
```
flixr: please use the functions for the deprecated macros, see also #1773

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_heli_indi.h
```diff
@@ -0,0 +1,81 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#ifndef STABILIZATION_ATTITUDE_QUAT_INT_H
```
flixr: plz fix include guard

flixr:
Thanks!

Issue #197, opened 2012-05-22
support for externally mounted mags


flixr:
We should support externally mounted mags, e.g. for the case where you can't (or don't want to) use the mag on the IMU board for interference reasons.
This means 
- [x] separate ABI id for mag subscriptions in AHRS wrappers
- [ ] need to be able to set a separate MAG_TO_BODY orientation
##

OpenUAS:
not having  a MAG to body more and more an issue with all the different supported hardware in Lab so raised level a bit

podhrmic:
Already works on pixhawk (HMC58xx), so closing as resolved.

PR #1755, opened 2016-06-21
[mag] add a MAG_TO_IMU option to HMC58XX drivers

gautierhattenberger:
this allow to choose arbitrary orientation for external mounted hmc mag
and move data back to IMU frame before sending them over ABI.

related to #197
this is the only driver supporting this, but hmc mag are the most
commonly used, especially as external sensor


flixr:
Nice! Can you please also document this in the module xml file?

PR #1764, opened 2016-06-24
Threaded computer vision

wilcoschoneveld:
This pull request addresses improved computer vision functionality as discussed in #1735.

Quick recap; There is a desire to have some vision functions run with less fps than others. It would be great if the priorities can be set between vision functions. For example, textons should run at 5 Hz and optical flow at 30 Hz.

The current setup; For each video device a vision pipeline is created. Modules can attach to this pipeline and process the image in turn, and even modify the image. For example, a color filter can modify the image and one can then view the result in RTP stream. The problem is that, since the vision pipeline is sequential, some modules (like a costly marker detection algorithm) can stall the performance of other vision functions.

I made a proof of concept with @flixr's suggestion in the video_rtp_stream module. The module attaches to the vision pipeline with a routine that copies the image to a buffer if required. The module processes and sends the frames in a separate thread at 10 Hz.

TODO:
- Abstract thread functionality into cv.h
- Improve thread managing (starting/stopping module)

I think overall this is a good approach, but there are some drawbacks;
- Copying the image costs time
- The modified image is lost


wilcoschoneveld:
What are your thoughts? @rmeertens, @flixr, @guidoAI, @fvantienen

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -109,23 +111,41 @@ struct viewvideo_t viewvideo = {
...
-  // Resize image if needed
-  struct image_t img_small;
-  image_create(&img_small,
-               img->w / viewvideo.downsize_factor,
-               img->h / viewvideo.downsize_factor,
-               IMAGE_YUV422);
+struct image_t *viewvideo_function(struct image_t *img) {
+  if (!viewvideo.new_image) {
+    // Free previous image
+    image_free(&img_copy);
```
fvantienen: First time this breaks? And why create a new image every time? Since you can reuse the previous one, which saves a lot of malloc's and free's over time(and also a lot of processing time).

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -186,15 +201,39 @@ struct image_t *viewvideo_function(struct image_t *img)
...
+    if (!viewvideo.new_image) {
+      usleep(5 * 1000); // 5 milliseconds
+      continue;
+    }
+
+    // Send a new frame from this thread
+    viewvideo_send_frame();
+
+    // Sleep to achieve 10 fps
+    usleep(100 * 1000); // 100 milliseconds
```
fvantienen: Maybe not a fixed frequency here?

fvantienen:
This idea is great, but there are some small implementation issues ;)

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -109,23 +111,41 @@ struct viewvideo_t viewvideo = {
...
-struct image_t *viewvideo_function(struct image_t *img)
-{
-  // Resize image if needed
-  struct image_t img_small;
-  image_create(&img_small,
-               img->w / viewvideo.downsize_factor,
-               img->h / viewvideo.downsize_factor,
-               IMAGE_YUV422);
+struct image_t *viewvideo_function(struct image_t *img) {
+  if (!viewvideo.new_image) {
```
fvantienen: This makes it thread safe, but adds also a delay in the showed image.

sw/airborne/modules/computer_vision/viewvideo.c
wilcoschoneveld: I changed it to create an image only once, the only drawback now is that the settings cannot be modified dynamically (image size, downsize factor)

sw/airborne/modules/computer_vision/viewvideo.c
wilcoschoneveld: It also prevents the vision pipeline from copying the image too often if the thread doesn't need it.

I agree that it adds a delay but i'm not sure if there is a better way to get data across threads. Do you know of any?

rmeertens:
Not a fan of copying the image each time. It makes this solution not better than the one already in there, you could already copy the image right now in your module and as soon as you are ready start a new thread with your copy of the image, and return out of your function. 

Would it be an idea for the modules that are allowed to be slow to make an image queue they can attach to? Each module that wants to do something with for example the front camera can throw in a function void (*check_qr_code)(img_t) in this queue... each(??) image we get we pop the longest waiting function from this queue and execute it. This way a module has some control over how often it gets images (add your callback more, or less). Please let me know what you think of this idea.

flixr:
First I want to mention that this code here would IMHO calls for a monitor (lock a mutex and wait on a condition variable) instead of the `new_image` flag and usleep...

In any case it seems to me that it would make sense to have each module process the image in it's own thread, so that they can have different priorities...

Things to consider first for the whole "problem" of how we want to deal with the image processing:
- How important is the pipe-lining? (i.e. one module modifies an image and passes that on to the next)?
- How costly is making a copy really? (if you already have a pre allocated buffer)

If pipe-lining is not needed and making a copy is fast (enough), then it seems to me that the simplest solution would be to have each image processing module allocate the space for one image and copy the image over in the callback given to `cv_add_device` (if you don't want to skip some since you want it to run slower). Then process the image in it's own thread...

To improve performance (by not copying the image in each module), we could think about a creating a buffer/pool that holds a certain amount of captured images (e.g. protected by a reader-writer lock). The "receiving" module could then just use the pointer to the next image (or newest or skip some) after it is done processing the previous one.
That would mean that it should only be a reader (not modify the image) though...

If pipe-lining is needed, we probably would have to make a queue/buffer for each "pipeline" though...

As a final note, we could also consider sending image pointers via ABI instead of supplying a callback function for `cv_add_device`... If a module modifies an image, it could also resend that via ABI and another module could bind on the message with the sender ID of the previous module (that did the modification), etc...

P.S. I nearly dare not say it, but: using C++ (RAII and shared pointers) would make some things a lot easier to use here... (but also require C++11...)

rmeertens:
Good points Felix! 
As for the pipelining: I don't think it would be necessary to have a linear pipeline like we have now... I think I would prefer a system like ROS where you can subscribe to the output of other modules. Your idea of ABI messages would fit right into this. Note that making this takes time, and I don't think creating the perfect image pipeline has a lot of priority now that we can make all our applications with this. 

I would say that having a different fps per module would do the job (less programming yeah), or even modules simply skipping frames (I think Ardupilot does this with their optical flow). In general I think most problems can/should be solved by the designer of the airframe file/modules. Just be smart enough to put your heavy module after your optical flow module, and let it skip frames (this also provides a nice lesson to students ;) ). 

As for making a copy: again I think it is up to the designer of the module... If every vision module makes its own copy the make one copy at the start approach is best... but if one very efficient algorithms has to wait because we do a copy I think this is a shame. 

I love your idea of having a buffer/pool of images.

kirkscheper:
I think this discussion has been very useful, most of the plausible options have been mentioned. Given that we need this functionality pretty soon for a project, what is the best course of action to implement this? Which idea has the greatest utility with least work? We will implement this this week so let us know.

kirkscheper:
@dewagter @fvantienen your input is also greatly appreciated

wilcoschoneveld:
I added a mutex and condition variable, but I believe the `new_image` flag is still needed to deal with spurious wakeups.

wilcoschoneveld:
Copying a 272x272 image with image_copy takes about 600~700 μs, with image_yuv422_downsample it takes about 900~1000 μs.

480x480 image with image_copy about 2000 μs, image_yuv422_downsample 3300 μs.

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -67,15 +70,11 @@ PRINT_CONFIG_VAR(VIEWVIDEO_QUALITY_FACTOR)
 #endif
 PRINT_CONFIG_VAR(VIEWVIDEO_RTP_TIME_INC)
 
-// Default image folder
```
flixr: This is still used in init, so should be kept...

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -186,22 +206,73 @@ struct image_t *viewvideo_function(struct image_t *img)
...
+void *viewvideo_thread(void *args);
+void *viewvideo_thread(void *args) {
+  // Time-keeping variables
+  struct timespec time_now, time_prev;
+  clock_gettime(CLOCK_MONOTONIC, &time_prev);
+  int32_t delta = 0;
+  uint32_t fps_period = 1000000 / VIEWVIDEO_FPS;
+
+  // Request new image from video thread
+  pthread_mutex_lock(&img_mutex);
```
flixr: You acquire the mutex at the beginning of the thread and never release it until the thread is shut down...
You want to do something like: 
```
while (viewvideo.is_streaming) {
  pthread_mutex_lock(&img_mutex);
  while (!viewvideo.new_image) {
    pthread_cond_wait(&img_available_cv, &img_mutex);
  }
  // critical section: do stuff requiring access to image here

  pthread_mutex_unlock(&img_mutex);
}
```

flixr:
The time and sleep in the thread is not needed...
If you want to reduce the fps, it should be done in the video_thread callback: only copy the image if a certain amount of time has passed since the last one (and currently not still computing something in the thread).

sw/airborne/modules/computer_vision/viewvideo.c
wilcoschoneveld: Isn't it already enough that `pthread_cond_wait` releases the mutex and causes the calling thread to block on the condition variable?

flixr: Right, was badly phrased and I didn't have a close enough look (thought you were unlocking the mutex inside the while).
The above would still be the common usage pattern if you don't want/need to lock the mutex for the whole time, e.g if not the whole part is a critical section.

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -186,22 +216,73 @@ struct image_t *viewvideo_function(struct image_t *img)
...
+    // Store timestamp as previous
+    time_prev = time_now;
+
+    // Request new image
+    viewvideo.new_image = false;
+  }
+
+  pthread_mutex_unlock(&img_mutex);
+
+  return NULL;
```
flixr: should probably call `pthread_exit` here...

wilcoschoneveld:
I refactored all the proposed implementation to be part of the cv module. I still have some more cleaning to do but if possible I would like some feedback from your point of view.

@flixr, @fvantienen, @rmeertens, @kirkscheper

flixr:
Nice! I like the idea to create the threads in `cv_add_to_device` as it requires less boilerplate in the modules.
However you have bit less control over the thread (e.g. priority, etc) then...
Since more than setting a prio is probably not needed, it might be an option to either also pass the desired prio in `cv_add_to_device` or create a separate function to add async callbacks with prio.

Also the max_fps is probably also interesting for the synchronous callbacks... and I would check for `> 0` instead of `!= 0` there...

wilcoschoneveld:
Ok I'm happy with the current state, can we merge this?

flixr:
Looks great, thanks!

PR #1767, opened 2016-06-27
[chibios] split FBW and AP threads

gautierhattenberger:
- add protection on ap and fbw state structures
- add RTOS abstraction macros
- update closing of ap/fbw threads with sdlog

### DIFF
--- a/sw/airborne/firmwares/fixedwing/main_ap.c
+++ b/sw/airborne/firmwares/fixedwing/main_ap.c
@@ -252,11 +252,11 @@ void init_ap(void)
   /* set initial trim values.
    * these are passed to fbw via inter_mcu.
    */
-  PPRZ_RTOS_MTX_LOCK(ap_state_mtx);
+  PPRZ_MUTEX_LOCK(ap_state_mtx);
   ap_state->command_roll_trim = COMMAND_ROLL_TRIM;
   ap_state->command_pitch_trim = COMMAND_PITCH_TRIM;
   ap_state->command_yaw_trim = COMMAND_YAW_TRIM;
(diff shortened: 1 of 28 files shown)


flixr:
Looks mostly good to me, just one thing regarding naming:
Strictly speaking mutexes are not RTOS functionality. E.g. under Linux (non-RT OS, at least without patches) we use the pthreads mutexes.
So maybe simply rename the macros to `PPRZ_MUTEX_*`?

Also before we add a lot to the old fixedwing intermcu: what is the path to common usage of the "new" subsystems/intermcu @fvantienen ?

gautierhattenberger:
Do you have an idea about how to rename `pprz_rtos.h` ?

The intermcu subsystem seems to be very specific to PX4 boards and clones. Not sure it can fit in the fixedwing firmware easily.

fvantienen:
It should not be specific for PX4, there is just some code in there for that that needs to be moved out. We mostly use it with the OPA autopilot (which basically is two lisa/mx's).
The new subsystem intermcu seems the way to go except that it could be converted to a module.

flixr:
So what is missing to use it for fixedwing firmware?

flixr:
see also #1771

gautierhattenberger:
It can probably work on without to much effort as long as you have a physical link (uart in this case), but it is currently not really fitted for shared memory operation (mostly based on pprzlink).
So what should I do for this PR ? I won't have the time to do much more than this at the moment.
At some point, I'll have to port the chibios arch for rotorcraft as well, so I may have a look at how to merge intermcu stuff at this time.

flixr:
Ok, so let's merge this as it is now.
But we should really follow up on a unified inter_mcu in #1771

PR #1773, opened 2016-06-28
[math] print warnings when using deprecated macros

flixr:


rmeertens:
Can we print the names of the (inline?) functions that you DO have to use? And when are we going to remove the old macro's?

flixr:
Would it be sufficient to print "FOO_BAR macro is deprecated, use the lower case function instead"?

The functions have been around for a long time now and we tried to point out to use them instead...
But there is no timeline to when we'll actually remove the macros. I would say we still keep them with the warnings for one stable release...

rmeertens:
Perfect and perfect!

PR #1777, opened 2016-07-04
remove nps target mavtec1

kirkscheper:


flixr:
NPS should work just fine if you also use the radio_control spektrum for NPS.
It currently fails because you are using ppm for the nps target, but have an empty dummy radio xml file set...

kirkscheper:
Yea, but I don't use the nps (actually never used it till now) so I won't miss this. Which do you prefer?

flixr:
If you never use it, leave it out (will also make builds faster)...
But then it would make sense to also remove the unused simulator section...

kirkscheper:
Done.

PR #1779, opened 2016-07-06
[rtp_viewer] added possibility to rotate viewer

wilcoschoneveld:
Useful for the viewing the bebop front camera stream, since it is rotated 90 degrees.

For now I disabled the possibility of sending a ROI message when rotated, since this would require remapping the region to the original rotation.


fvantienen:
Seems good! 👍

PR #1780, opened 2016-07-06
[bebop] updated sonar driver to improve accuracy and range

kirkscheper:
- Sonar now has two types of pulses emitted depending on the altitude, short pulse for for low alt and long pulse for high alt. 
- Added a state machine to make the code more readable.
- Updated computation of distance based on time of flight
- Added comments

### DIFF
--- a/sw/airborne/modules/sonar/sonar_bebop.c
+++ b/sw/airborne/modules/sonar/sonar_bebop.c
@@ -22,6 +22,14 @@
 
 /** @file modules/sonar/sonar_bebop.c
  *  @brief Parrot Bebop Sonar driver
+ *
+ *  This file contains the ADC driver for the sonar on the Parrot Bebop
+ *  and Bebop 2 quadrotors. The sonar is both transmitter and receiver.
+ *  An SPI interface is used to communicate with the 12 bit ADC which
+ *  operates at 160kHz. The required waveform of the sonar is sent during
+ *  the SPI transaction, with the returned data the recorded signal from
+ *  the sonar. Two waveforms have been implemented dependent on the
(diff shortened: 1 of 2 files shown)


kirkscheper:
@wilcoschoneveld I tested this shortly yesterday and it worked quite well. Can you test this with optitrack as a ground truth and let me know if it is accurate? Use <define name="SENSOR_SYNC_SEND_SONAR"/> and <define name="SONAR_UPDATE_ON_AGL" value="TRUE"/>. Plot the sonar distance and the ins height, take off, fly up high and then descend again. If it looks good can you post the tracking here?

wilcoschoneveld:
sonar

fvantienen:
This doesn't have the peak finding algorithm like they have in APM yet right? But if it is already much better we can merge this 👍

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -36,22 +44,63 @@
 #include "state.h"
 #endif
 
+enum peak_search_state {INIT, BROADCASTING, SEARCHING_MIN, SEARCHING_PEAK};
+static enum peak_search_state search_state;
+
+/* conversion form index difference to distance based on time of flight
```
flixr: since you already put comments before the defines (:+1:), please use doxygen comments so it will show up in the generated docs...

kirkscheper:
Turns out this is more difficult than it seems. I will have to revise this and re-evaluate my approach and test this on different surfaces and orientations. Will update this once I have something that works well.

flixr:
So do you want to close this already in light of #1909 ?

PR #1786, opened 2016-07-08
In KILL mode the servos are set to preload with 0 pulse, such that th…

EwoudSmeur:
…e last pulse is finished and it is safe to flash the vehicle. This will prevent (stupid) servos to go to the outer position upon flashing. Tested and works.

As a side effect, servos will not move from their current position in kill mode, and can be moved by hand. I don't think this is a problem, it even makes more sense for kill mode.

### DIFF
--- a/sw/airborne/arch/stm32/subsystems/actuators/actuators_pwm_arch.c
+++ b/sw/airborne/arch/stm32/subsystems/actuators/actuators_pwm_arch.c
@@ -28,6 +28,7 @@
 #include "subsystems/actuators/actuators_shared_arch.h"
 #include "subsystems/actuators/actuators_pwm_arch.h"
 #include "subsystems/actuators/actuators_pwm.h"
+#include "autopilot.h"
 
 #include <libopencm3/stm32/gpio.h>
 #include <libopencm3/stm32/rcc.h>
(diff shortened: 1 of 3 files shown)


flixr:
Hm... I dislike the inclusion of autopilot.h in actuator code... those should be independent as you don't have that in all firmwares...

EwoudSmeur:
yeah... that is why it fails... do you have a suggestion? Maybe the problem is that the firmwares are not more uniform ;)

podhrmic:
I noticed this is for stm32 only - does the issue not affect other architectures?

flixr:
You shouldn't expect that there is an autopilot_mode variable available in the actuators code.
E.g. if you write a simple test/setup firmware or use the actuators code for something else you very likely don't have any "autopilot mode".
Rather this should be done at the firmware level, so disabling the actuator output if you are in kill mode...
Then this could also be handled for non pwm actuators (and should also be done for other architectures as @podhrmic pointed out)

EwoudSmeur:
that makes sense, but that means that ActuatorsPwmCommit() needs to take an argument like 'servo_kill' or 'stop_communication'. I think that pretty much means that every actuator needs to take an argument like that, because they are all linked through macros like:
#define ActuatorsDefaultCommit() ActuatorsPwmCommit()

flixr:
Or make a separate function to "disable" the actuators instead of passing it as an argument to ActuatorsXCommit.
That function you could call in KILL mode instead of SetActuatorsFromCommands (which calls actuators commit in turn).

podhrmic:
@EwoudSmeur @flixr any updates here? Is this still an issue?

EwoudSmeur:
Well, it is still an issue, but I think the proper solution is quite complex, If you consider the maze of macros from which the actuator commit function can be called, different architectures and perhaps other servo behaviours that we haven't considered yet. I am fine with closing.

PR #1787, opened 2016-07-08
[opticflow] tuned gains for bebop airframe

wilcoschoneveld:
Additionally, since the bottom camera upgrade, the image is already rotated. Furthermore, derotation of the flow is not working correctly so disabled for now.

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_bebop_opticflow.xml
+++ b/conf/airframes/TUDELFT/tudelft_bebop_opticflow.xml
@@ -10,7 +10,7 @@
       <module name="udp"/>
     </target>
 
-	<!-- Subsystem section -->
+    <!-- Subsystem section -->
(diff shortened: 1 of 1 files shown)


fvantienen:
Derotation doesn't work anymore because I think the axis are incorrect.. could you maybe take a look at that?

fvantienen:
I also see that at the top of `opticflow_calculator.c` there is an error in the `#ifndef CAMERA_ROTATED_180`.

kirkscheper:
@wilcoschoneveld can you update this with the new gains, derotation and current camera orientation after @knmcguire and @rmeertens take a look at the opticflow?

flixr:
@wilcoschoneveld @kirkscheper @fvantienen Shall we close this or can we still expect an update here?

wilcoschoneveld:
Sorry for the late response, an update is on the way

PR #1788, opened 2016-07-08
[imu] convert imu subsystems to modules

flixr:
Also
- get rid of explicit imu init|event|periodic calls in main
- each imu implementation has their own imu_x_init/event/periodic

Reviews and tests appreciated!

TODO:
- [x] Some airframes are still failing because the autoloading of module is not working in all cases. E.g. loading `imu_lisa_m_v2.1` works, but not `imu_lisa_mx_v2.1` which is just intended as an alias and autoloads `imu_lisa_m_v2.1` (fixed with 19731311899a0fb08f094bcbcc5963de069de030)
- [x] aspirin IMUs had different SPI/I2C defaults for lpc21 than for stm32... ~~should we really keep this or~~ update the failing LPC airframes to explicitly set the devices?
- [x] put imu related settings in the module (one common one or in every module?)


flixr:
@gautierhattenberger any ideas on the recursive autoload problem?

gautierhattenberger:
regarding the settings (basically body_to_imu), it is possible to reference a file, but then generated doc is not as informative. Otherwise, a common imu module seems a good idea.

gautierhattenberger:
Tested and working on Apogee.
Not working on ARDrone2.

flixr:
IMU on Ardrone2 should be fixed, although I can't test that...

flixr:
@gautierhattenberger @kirkscheper @fvantienen or anyone else interested:
IMHO this is ready to merge, but tests of more IMUs wouldn't hurt

flixr:
@podhrmic @gautierhattenberger any more tests/comments here?
Otherwise I would merge it...

podhrmic:
OK, so I did some tests. First of all, if I start GCS I get these errors:
```
RUN '/home/fwmav/Paparazzi/master/paparazzi/sw/ground_segment/cockpit/gcs '
Creating '/home/fwmav/Paparazzi/master/paparazzi/var/logs'
A failure occurred while processing aircraft 'ardrone2_raw'
 - Fail with : File not found: /home/fwmav/Paparazzi/master/paparazzi/conf/telemetry/default_ardrone.xml
 - 'ardrone2_raw' will be ignored by the server
 - Please remove it from 'conf.xml' or fix its parameter(s)
A failure occurred while processing aircraft 'ardrone2_sdk'
 - Fail with : File not found: /home/fwmav/Paparazzi/master/paparazzi/conf/telemetry/default_ardrone.xml
 - 'ardrone2_sdk' will be ignored by the server
 - Please remove it from 'conf.xml' or fix its parameter(s)
A failure occurred while processing aircraft 'bebop'
 - Fail with : File not found: /home/fwmav/Paparazzi/master/paparazzi/conf/airframes/bebop.xml
 - 'bebop' will be ignored by the server
 - Please remove it from 'conf.xml' or fix its parameter(s)
A failure occurred while processing aircraft 'krooz_quad'
 - Fail with : File not found: /home/fwmav/Paparazzi/master/paparazzi/conf/airframes/examples/krooz_sd/krooz_sd_quad_mkk.xml
 - 'krooz_quad' will be ignored by the server
 - Please remove it from 'conf.xml' or fix its parameter(s)
```
So that probably should be fixed...

Otherwise:
- Tested on Lisa MX (Bixler, Quad - in both cases IMU works OK)
- Tested on Lisa M (Bixler) - IMU works OK

flixr:
Thanks for testing.

Sounds like you were you using your own outdated conf.xml.
No aircrafts were added/removed in this PR, only the body_to_imu.xml settings file removed...

Issue #1184, opened 2015-05-21
HITL Upgrade


podhrmic:
**Motivation:** Current HITL subsystems is obsolete and not-functional. Last time it was used as is was probably sometime before 2012. HITL is a great way of testing the code, and having HITL functionality in Paparazzi by default would allow developers to test their code on real hardware in simulated flight, which adds another level of safety before new code is committed. Also users could use it for tuning control loops and simulate missions on actual hardware.

The autopilot can be connected over serial or USB port. That means we can either simulate whole packets (in case of an external AHRS or INS, when the autopilot only parses attitude and position information from an external source), or we simulate data coming from other interfaces too (SPI->gyro, i2c->baro etc). The first option is definitely closer to real flight because no code change is needed (simulation<->flight), the latter option is more thorough (and more complex) and allows the users to test paparazzi AHRS algorithms on actual hardware.
### Tasks:

I would imagine working in the following order:
- [x] 1. remove the obsolete hitl files
- [x] 2. update the flightgear interface to the newest version (minor change in network protocol)
- [x] 3. develop brand new HITL program (either in C++ & Boost, or Python)
- [ ] 4. integrate HITL into paparazzi build process and with NPS

The motivation behind the last point is, that NPS already takes care of the simulation part (FDM of the aircraft). Having FDM again in HITL seems to be just a duplication. I guess in a perfect world, it would be possible to run NPS with HITL option, in which case it would receive external inputs from the autopilot and send proper data back. Related to #1183 

At least in the initial phase, I would focus on HITL where we assume external INS (see above the differences), so the onboard AHRS/INS subsystems wouldn't be used.

Any thoughts on this?
##

podhrmic:
@flixr @gautierhattenberger ?

gautierhattenberger:
Fixing HITL sim is probably a good idea. Same comment than #1183 about adding yet another big dependency like boost. I haven't used C++ for a long time, so I might be wrong about the amount of time to do this job.
You described the problem pretty well, so there is not much to add.

podhrmic:
The motivation for boost is that it will cut significantly the development time. Probably even better would be using Python, but that would require making Python bindings for JSBsim - the existing ones seem to be outdated unfortunately: http://sourceforge.net/projects/jsbsim-python/?source=typ_redirect

Adding boost should be as simple as installing one package on Ubuntu, don't know about other systems (Mac OS? Android?)

podhrmic:
[...]

PR #1789, opened 2016-07-08
NPS uses C++11 instead of glib

podhrmic:
On the way towards cleaning up NPS and reusing some functionality for HITL - right now I replaced the glib with C++11 threads, seems to work the same for fixedwing and VTOL.

I ll clean it up more, in the meantime - any comments @flixr?


flixr:
What is the benefit here? We even loose the possibility to stop NPS and run it at different speed again...

podhrmic:
I was thinking that HITL can be included in NPS - and be just run with different arguments since most of the code is identical. Were the signals for Gaia (changing speed)? Because I wasn't able to stop/start it from the console.

flixr:
CTRL-Z pauses the NPS loop and you can enter a new timescale, works fine here and I still use that regularly.
Totally forgot that we can also change the timescale in NPS via gaia for a while now already.
However that doesn't allow for pausing...

podhrmic:
If I keep the CTRL-Z feature (so it does the same thing as now), would you be in favour of the change to C++11?

flixr:
In general where it makes things easier, yes... definitely prefer that over boost.
Although to me it is not clear how converting from glib mainloop usage to c++11 threads helps the hitl integration (since I don't know what is yet to come)...

podhrmic:
@flixr any idea why is Ivy not working in my example? I am calling the right function and the program flow is very simple - the treads are executed etc, but I am suspecting that Ivy has issues in C++ environment....?

flixr:
Because ivy-glib is still used (which integrates into the glib mainloop that you removed)...
See http://www.eei.cena.fr/products/ivy/documentation/ivy-c/index.html for different main loops with ivy-c.
You could try ivy-c++ instead, but we haven't used that so far (so it is also not packaged by us)...

Also the access to the printed values is not thread safe, so you can't just put that in a separate thread without further protection.

Still unclear to me what the benefit is of splitting it into more threads here... maybe you can explain a bit what is to come with HITL so we can get a better picture of your goal and what is needed?

gautierhattenberger:
If you want to make it run, you need to link against the standard ivy lib (not the glib one) and called by hand the IvyMainLoop (blocking) function in one of your thread. And of course protect data access between threads.
But mostly I have the same question than Felix, what is the benefit of this ? If it is a preliminary work, can we have a look at the final objective in an other branch ?

podhrmic:
That makes sense, I ll give it a shot, thanks! 

The motivation here is that HITL uses 90% of NPS code anyway (the extra is two threads for IO, and possibly another thread for logging data). The way I implemented it first was to just copy-paste-modify the functions I needed, and added a new "hitl" target (see #1360 and #1184 ).

But that was a very naive approach with a lot of code duplicity. To implement HITL functionality in NPS (and possibly figure out some better name for both) I need threads because there is asynchronous IO access (serial data to/from the autopilot) and the original NPS has only one big loop.

I wanted to use glib originally, but that has its own problems - mainly glib makes it not portable (GNOME only?), you have to use the g_struct and g_functions that are apparently not best written (for example http://stackoverflow.com/questions/2240154/why-use-glib-functions#2240239 ), not well documented, and you are locked to plain old C.

By switching to C++11 I am hoping to:
1) make the NPS/HITL code more portable (less dependencies - might be important in the future)
2) make it easier to change/maintain/configure/extend the code (faster and easier to write application like that in C++)
3) make it modern (no need for a desktop app in plain C and glib in 2016...)

So yes, in theory I could achieve the same with glib threads, but that is imho as much work as converting it to C++ which will be more useful in the future.

Does it make sense to you like that?

flixr:
yep, it does ;-)

podhrmic:
FYI it turned out that refactoring NPS into C++ (and using classes and all that jazz) would be too much work - especially because some NPS sensors are using glib structs (linked lists?) and are expected to be in plain C. So I am closing this PR and will open a new one wit NPS+HITL using GLIB. Non ideal, but definitely less work. Sorry about that.

PR #1791, opened 2016-07-12
Switched wind directions in nav_smooth.c

podhrmic:


flixr:
For simple fixes like this (which you merge directly anyway) you don't have to create a pull request, you can directly commit that to master.
Bonus points for mentioning the original issue/pull request that it fixes in the commit ;-)

PR #1792, opened 2016-07-12
Coverity fixes vol. II

podhrmic:
This should fix the remaining airborne related issues from Coverity.
Not sure about the `switch` statements without `break`  - @flixr ?

### DIFF
--- a/sw/airborne/math/pprz_algebra_int.c
+++ b/sw/airborne/math/pprz_algebra_int.c
@@ -462,16 +462,16 @@ void int32_quat_of_rmat(struct Int32Quat *q, struct Int32RMat *r)
       uint32_t two_qy = int32_sqrt(two_qy_two << INT32_TRIG_FRAC);
       two_qy = two_qy << (INT32_QUAT_FRAC - INT32_TRIG_FRAC);
       if (two_qy != 0) {
-      q->qi = ((RMAT_ELMT(*r, 2, 0) - RMAT_ELMT(*r, 0, 2)) <<
-               (INT32_QUAT_FRAC - INT32_TRIG_FRAC + INT32_QUAT_FRAC - 1))
-              / two_qy;
-      q->qx = ((RMAT_ELMT(*r, 0, 1) + RMAT_ELMT(*r, 1, 0)) <<
-               (INT32_QUAT_FRAC - INT32_TRIG_FRAC + INT32_QUAT_FRAC - 1))
-              / two_qy;
-      q->qy = two_qy / 2;
-      q->qz = ((RMAT_ELMT(*r, 1, 2) + RMAT_ELMT(*r, 2, 1)) <<
-               (INT32_QUAT_FRAC - INT32_TRIG_FRAC + INT32_QUAT_FRAC - 1))
-              / two_qy;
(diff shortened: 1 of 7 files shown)


flixr:
The missing breaks are intended, it falls through to the common parts...

sw/airborne/math/pprz_algebra_int.c
```diff
@@ -425,38 +425,43 @@ void int32_quat_of_rmat(struct Int32Quat *q, struct Int32RMat *r)
...
+                 (INT32_QUAT_FRAC - INT32_TRIG_FRAC + INT32_QUAT_FRAC - 1))
+                / two_qx;
+      }
     } else if (RMAT_ELMT(*r, 1, 1) > RMAT_ELMT(*r, 2, 2)) {
       const int32_t two_qy_two = RMAT_ELMT(*r, 1, 1) - RMAT_ELMT(*r, 0, 0)
                                  - RMAT_ELMT(*r, 2, 2) + TRIG_BFP_OF_REAL(1.);
       uint32_t two_qy = int32_sqrt(two_qy_two << INT32_TRIG_FRAC);
       two_qy = two_qy << (INT32_QUAT_FRAC - INT32_TRIG_FRAC);
+      if (two_qy != 0) {
       q->qi = ((RMAT_ELMT(*r, 2, 0) - RMAT_ELMT(*r, 0, 2)) <<
```
flixr: indentation?

sw/simulator/nps/nps_main.c
```diff
@@ -356,8 +356,7 @@ static bool nps_main_parse_options(int argc, char **argv)
           case 2:
             nps_main.fg_time_offset = atoi(optarg); break;
           case 3:
-            if (optarg == NULL) {nps_main.js_dev = strdup("0");}
-            else {nps_main.js_dev = strdup(optarg);}
+            nps_main.js_dev = strdup(optarg);
```
flixr: does this still work? there must have been a reason for doing it this way...

sw/simulator/nps/nps_main.c
```diff
@@ -369,24 +368,24 @@ static bool nps_main_parse_options(int argc, char **argv)
...
+          case 9:
             nps_main.fg_port_in = atoi(optarg); break;
         }
         break;
 
-      case 'j':
-        if (optarg == NULL) {nps_main.js_dev = strdup("0");}
-        else {nps_main.js_dev = strdup(optarg);}
-        break;
+          case 'j':
```
flixr: unintended indentation changes?

sw/airborne/math/pprz_algebra_int.c
podhrmic: Fixed

sw/simulator/nps/nps_main.c
podhrmic: Silly Eclipse can't correctly indent nested switches. Fixed.

sw/simulator/nps/nps_main.c
podhrmic: Hmm, it doesn't or at least not the same way - marking as intentional, whatever the original reason here was...

flixr: that's what I though... you can't make a copy of a NULL pointer, so in that case if it was not provided we explicitly use a string with `0` (as that is the intended default).

PR #1794, opened 2016-07-12
G2d translation

podhrmic:
I am getting familiar with `geometry_2d` library and I found it harder to understand with French comments. Naively translated using google translate, and manually fixed where appropriate.

@gautierhattenberger do you think you could look over it briefly and point out if there are some obvious errors?


gautierhattenberger:
close with c0cdb673229b073fcb28d496ac62f14f70cd19a8

PR #1795, opened 2016-07-13
opticflow_module: moved mutex

rmeertens:
Remove the line 
`-  // TODO: why is there a mutex above and not below when changing opticflow_result?
`

And moved the mutex.


flixr:
Thx.
Please try to mention the "component" or module in the commit message and pull request title in the future...

rmeertens:
Thanks. Will do!

PR #1797, opened 2016-07-13
[guided] fixes in guided guidance functions

wilcoschoneveld:
- <del>the flight plan system requires functions to return false in case of no errors</del>
- the vertical guidance functions now also set the guidance reference

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
@@ -682,9 +682,9 @@ bool guidance_h_set_guided_pos(float x, float y)
     ClearBit(guidance_h.sp.mask, 5);
     guidance_h.sp.pos.x = POS_BFP_OF_REAL(x);
     guidance_h.sp.pos.y = POS_BFP_OF_REAL(y);
-    return false;
+    return true;
   }
-  return true;
+  return false;
 }
 
 bool guidance_h_set_guided_heading(float heading)
@@ -693,9 +693,9 @@ bool guidance_h_set_guided_heading(float heading)
     ClearBit(guidance_h.sp.mask, 7);
     guidance_h.sp.heading = ANGLE_BFP_OF_REAL(heading);
     INT32_ANGLE_NORMALIZE(guidance_h.sp.heading);
-    return false;
+    return true;
   }
-  return true;
+  return false;
 }
 
 bool guidance_h_set_guided_body_vel(float vx, float vy)
@@ -712,9 +712,9 @@ bool guidance_h_set_guided_vel(float vx, float vy)
     SetBit(guidance_h.sp.mask, 5);
     guidance_h.sp.speed.x = SPEED_BFP_OF_REAL(vx);
     guidance_h.sp.speed.y = SPEED_BFP_OF_REAL(vy);
-    return false;
+    return true;
   }
-  return true;
+  return false;
 }
 
 bool guidance_h_set_guided_heading_rate(float rate)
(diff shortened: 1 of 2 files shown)


flixr:
The return values are fine as they are (and documented) and should not be changed for calling in the flight plan.
It would not make any sense anyway, as those are not functions that should be called repeatedly until it is done (and returns false).
If you want to call those from the flight plan, you should use `call_once`.

kirkscheper:
It is quite strange that the navigation commands return 0 when successfully complete and not the other functions in pprz.

If the reason for the return is to facilitate repeated calling why can we call functions like nav_reset_reference and others which only get called once can be called with 'call' and not 'call_once'? It would be nice to have one guideline which is used everywhere, this discrepancy caused us quite a lot of wasted time and effort. 

Regarding the return values of the guidance_set functions, the return isn't used anywhere so I assume we can change them to whatever is most common sense. I actually think the method in master makes more sense and the nav return values are very strange, maybe something to reconsider or document well, it isn't stated anywhere in the nav file the reason for the inversed logic of the return values.

@wilcoschoneveld can you revert the function returns and use call_once in the airframe. Still commit the other changes.

flixr:
Regarding calling functions from the flight plan: in most cases you should use `call_once` and only use `call` if you want to "run" a function repeatedly until it returns false.
See the previous discussions on this: #830 #1612 #1675 

That we still have navigation macros that always return false was really a bad hack to make them usable with `call`.
And also https://wiki.paparazziuav.org/wiki/Flight_Plans#Call

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -595,6 +595,8 @@ static void guidance_h_traj_run(bool in_flight)
 
 static void guidance_h_hover_enter(void)
 {
+  /* reset speed setting */
+  guidance_h_set_guided_vel(0., 0.);
```
flixr: IMHO it would be better to directly set
```
guidance_h.sp.speed.x = 0;
guidance_h.sp.speed.y = 0;
```
here instead of resetting vel to zero only if you are in guided mode, set the xy speed bit and then clear it again in the next line...

sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
```diff
@@ -481,8 +481,18 @@ static void run_hover_loop(bool in_flight)
...
+
+    /* set altitude setpoint */
     guidance_v_z_sp = POS_BFP_OF_REAL(z);
+
+    /* reset speed setting */
+    guidance_v_zd_sp = 0;
+
+    /* reset guidance reference */
+    guidance_v_z_sum_err = 0;
+    GuidanceVSetRef(guidance_v_z_sp, guidance_v_zd_sp, 0);
```
flixr: why do you want to reset the ref every time you change the setpoint?
IMHO that is not what you usually want... you might want to reset the ref on entering the guided mode, but not on every update...

kirkscheper: Is it possible to add this to the init function in this file? For the IMAV we are using the flight plan in combination with guided commands for autonomous flight, we startup in guided mode so don't switch to it...

flixr: You are actually completely circumventing the reference here as you set to the target setpoint...
IMHO the thing it makes sense to manually set the reference to is the current position/velocity
```
GuidanceVSetRef(stateGetPositionNed_i()->z, stateGetSpeedNed_i()->z, 0);
```
If it makes sense to do that in every set_guided_x function or not depends on how you want those command to behave in the end... in any case we should clearly state this in the function description.
Doing that in init makes no sense as you neither have a setpoint nor the current position yet...

kirkscheper:
I merged the error of setting vel to 0 at the end of the hove enter. The rest of the issues should not be significant. We will continue testing and update other things later as necessary.

PR #1799, opened 2016-07-14
Spektrum rewrite

fvantienen:
This is a total rewrite of the Spektrum satellite parser, which also include support for DSMX!
Protocol information is available at http://www.spektrumrc.com/ProdInfo/Files/Remote%20Receiver%20Interfacing%20Rev%20A.pdf

Things that need to be done before merging:
- [ ] Add more satellites to xml makefile section of module
- [ ] Fix UART ports in all board files
- [ ] No guessing of TX type when RX sends type
- [ ] Enable settings for binding pulses and bind at boot (or live when power can be reset or RX)
- [ ] Binding at boot function
- [ ] Fix sim/NPS RC
- [ ] Test with different RX/TX.


sw/tools/generators/gen_periodic.ml
```diff
@@ -109,7 +109,7 @@ let write_settings = fun xml_file out_set telemetry_xml ->
   fprintf out_set "<!-- This file has been generated by gen_periodic from %s -->\n" xml_file;
   fprintf out_set "<!-- Version %s -->\n" (Env.get_paparazzi_version ());
   fprintf out_set "<!-- Please DO NOT EDIT -->\n\n";
-  fprintf out_set "<settings>\n";
+  fprintf out_set "<settings target=\"ap|nps\">\n";
```
flixr: what is this about?

flixr:
How are the fbw and intermcu_init changes related to the new spektrum parser?

fvantienen:
Yeah those should be removed here.. but I tested this in a FBW+AP setup. And wanted to try to get settings to work in the FBW(because there this module runs). But this caused other problems with the GCS part not able to set the FBW settings(which I wanted to set through intermcu).

sw/tools/generators/gen_periodic.ml
fvantienen: See the general comment ;) this is btw true, since there is no telemetry in fbw.

flixr: not quite, fixedwing fbw has telemetry, also other targets like tests and maybe some loggers do...

podhrmic:
@fvantienen any updates here? Can you fix the merge conflicts?

OpenUAS:
Would like to see this great and warm welcomed improvement merge a reality so:

This I can fix:
* Add more satellites to xml makefile section of module
* Fix UART ports in all board files

Maybe @kevindehecker can help since he was busy with spektrum bind code a lot for Pixhawk so these:
* nable settings for binding pulses and bind at boot (or live when power can be reset or RX)
* Binding at boot function

Maybe @flixr this one? 
* Fix sim/NPS RC

and maybe you cold add this yourself?
* No guessing of TX type when RX sends type

Then we can:
* Test with different RX/TX., 
We have quite some different DSM2 and X TX's at hand

OpenUAS:
As @fvantienen will be available again as of ~20180629 well go through this one

OpenUAS:
Since @fvantienen will be available again as of ~20180629 we will be validating this one again together.

fvantienen:
In favor of #1799 I close this pull request.

PR #1800, opened 2016-07-14
don't clean opencv_bebop with clean in main folder

kirkscheper:
Today I ran test_all_confs and saw that the autonomous_dronerace airframe failed. It ended up that this was because my opencv_bebop submodule had never been initialised and it is not built automatically with make. It is not that useful for new developers wanting to work with opencv or is it good for the CI tests tht will test airframes that use opencv.

Due to the way the opencv_bebop is made it builds quite slowly the first time it checksout but it also tries to build some files every time you build. What I do with this commit is have it that the first time you build paparazzi it will download and build opencv_bebop but only try to rebuild the project if the install folder is missing from the subfolder. This means that the library will only be rebuilt if the user runs 'clean_opencv_bebop' from sw/ext or 'clean' from 'sw/ext/opencv_bebop' folder. Another pitfall of this is the CI build will likely take longer to run. as opencv will be fetched and built for each pull request.

This works for me but is not ideal, it would be great to only build the project if the 'submodule update' actually did something.

This also fast-forwards the submodule to the recent version.

### DIFF
--- a/sw/ext/Makefile
+++ b/sw/ext/Makefile
@@ -94,9 +94,11 @@ clean:
 	$(Q)if [ -f luftboot/src/Makefile ]; then \
 		$(MAKE) -C luftboot/src clean; \
 	fi
+
+clean_opencv_bebop:
 	$(Q)if [ -f opencv_bebop/Makefile ]; then \
 		$(MAKE) -C opencv_bebop clean; \
 	fi
 
 .NOTPARALLEL: libopencm3 luftboot
-.PHONY: all clean update_submodules libopencm3 luftboot chibios fatfs luftboot_flash libopencm3.build luftboot.build mavlink.build libsbp pprzlink pprzlink.build opencv_bebop opencv_bebop.build
+.PHONY: all clean update_submodules libopencm3 luftboot chibios fatfs luftboot_flash libopencm3.build luftboot.build mavlink.build libsbp pprzlink pprzlink.build opencv_bebop opencv_bebop.build clean_opencv_bebop


flixr:
I think this is already solved by 229647ba37f77df436ff99f86f7695d3761c8cce 3a1624053ca104807178255ce9370c8fbb8c0407 and 10df83ee3d318f6e8abd800d91e9c8b2ba1a27dd?

kirkscheper:
Well I guess it isn't cause it didn't work for me today working with Master
from today.

Kirk

flixr:
Well, I only updated that today... which version exactly did you test?
It should be working with latest master (v5.9_devel-767-g10df83e)

The build of test_all_confs is still running: https://semaphoreci.com/paparazziuav/paparazzi/branches/master/builds/971

flixr:
Also we probably shouldn't build opencv_bebop on general make, as that would need additional dependencies that not everyone might have...
That is why I added building of opencv_bebop as a dependency of test_all_confs` only...

flixr:
Also I think it tries to build some files every time because a patch is applied every time and then the changes removed again (to not get a dirty working dir in the submodule).
If we would make a fork of opencv and commit the patches there, this probably wouldn't happen...

podhrmic:
:+1: @flixr Not everybody uses opencv so I would prefer not to have it built automatically. Maybe have some `make opencv` command you could run after the general `make` ?

kirkscheper:
test_all_conf work for me now. 

Ok, so we don't want the module on as default, can we put a warning in a makefile that if you try to link to opencv but it's not built to run make opencv_bebop in the sw/ext/ folder, it wasn't that obvious to me that this was required anyway... I thought there were just some static libraries hidden somewhere.

Shall I make the fork on the tudelft? 

I still think that opencv_bebop should not be cleaned with the clean in the main pprz folder...

flixr:
> Ok, so we don't want the module on as default, can we put a warning in a makefile that if you try to link to opencv but it's not built to run make opencv_bebop in the sw/ext/ folder, it wasn't that obvious to me that this was required anyway... I thought there were just some static libraries hidden somewhere.

yep, warning would be appreciated

> Shall I make the fork on the tudelft?

sounds good to me

> I still think that opencv_bebop should not be cleaned with the clean in the main pprz folder...

Why?

kirkscheper:
Cause it gets rebuilt from scratch which takes forever. Also if nothing has changed in the files then why would that library have to be rebuilt.

flixr:
ok, I usually expect that the top level clean really cleans all, but I see your point...
Can you update this PR accordingly?

kirkscheper:
If you don't have to build opencv_bebop with each make the rebuilding because of the patch shouldn't be an issue, I will only add that the clean doesn't clean opencv_bebop.

PR #1802, opened 2016-07-14
Updated geofencing

podhrmic:
Based on #1664 - newly added an option to specify `geofence_max_agl` in the flight plan, on top of `geofence_max_alt`. Both values are checked and used, the lower one triggers the safety mode.

Also added a check for negative security height (which is a nonsense).

### DIFF
--- a/conf/conf_tests.xml
+++ b/conf/conf_tests.xml
@@ -5,7 +5,7 @@
    airframe="airframes/AGGIEAIR/aggieair_ark_quad_lisa_mx.xml"
    radio="radios/AGGIEAIR/aggieair_taranis.xml"
    telemetry="telemetry/default_rotorcraft.xml"
-   flight_plan="flight_plans/rotorcraft_basic.xml"
+   flight_plan="flight_plans/rotorcraft_basic_geofence.xml"
    settings="settings/rotorcraft_basic.xml settings/control/rotorcraft_guidance.xml settings/nps.xml settings/control/stabilization_att_float_euler.xml"
    settings_modules="modules/gps.xml"
    gui_color="#ffff954c0000"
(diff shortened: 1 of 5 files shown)


flixr:
Hm... it is actually not really a `max_agl` setting... if we keep the same naming as for waypoints it would be more appropriately named `max_height` as it is the height above the currently used reference and NOT over current ground level (which requires either a terrain model or actual agl measurement).

podhrmic:
@flixr I think `max_agl` is more descriptive because `height` is somewhat ambiguous (height above what?). The use-case is that you set the ground altitude of the flight plan, and you know you cannot go above lets say 400 m AGL. So you say `max_agl=400`, assuming you are flying over a flat terrain and that you are doing GeoInit at the beginning of your flight (so the local reference is the current ground altitude).

flixr:
Well, I still think it would be better to use consistent naming, we already have `security_height` and `home_mode_height` which refer to height above field elevation (`ground_alt`).
To me it seems that naming the geofence limit differently, while it has exactly the same reference and assumes same ground alt everywhere is not the best idea.
Also if we ever actually put a terrain model onboard and use that, how would you call this then?

@gautierhattenberger @martinmm @gautierhattenberger what do you guys think?

podhrmic:
Fair enough - I ll change it to height.
On a side note - SRTM tells you the orthometric height ( GPS ellipsoidal height - geoid height) while GPS altitude (which we are using currently for waypoints) doesn't have the geoid information.

As @dewagter pointed out in #1700 GPS alt is never used in aviation. I know most of the GPS units provide HMSL measurements (which we want). Is there a reason for using GPS alt to represent waypoints instead of HMSL? Or can I just specify `hmsl` attribute in the waypoint tag to give it correct altitude?

podhrmic:
Fixed with 2d23f5fc8723891de60edfe0c6e927eecb8bde0b

flixr:
Waypoint alt on the ground and flightplan is in HMSL. Onboard they are stored wrt the local reference though.

And unfortunately there are quite a few GPS/INS that don't give use hmsl... vectornav, xsens, piksi...

PR #1805, opened 2016-07-15
Fixes to optical flow module

knmcguire:
[opticflow] fix segmentation fault (derotation) in edgeflow, removed camere_rotate_180 (is already done), some added changes to the conf files.

@rmeertens, could you check with your bebop if you still have the segmentation problem with edgeflow and derotation selected?


rmeertens:
Confirmed no crash

rmeertens:
Note that you now changed 6 files, where we could do with 2....

knmcguire:
I know, but I already made some changes to the airframe file and some small fixes, that I wanted to merge anyway, before you notified me on the problem. But yes I could have maybe made more commits to make it more clear

PR #1807, opened 2016-07-15
Configs for the camera parameters

rmeertens:
Hey everybody. 

The bebop has an awesome configurable front camera, I would like to be able to configure this camera somehow. 
My current proposal is adding the parameters in your airframe... this solution is not yet perfect, so I would love to get feedback on this!


rmeertens:
Shoutout to @fvantienen @flixr @kirkscheper and @dewagter

dewagter:
Is the camera able to setup other dimensions that easily now?

fvantienen:
See #1808

rmeertens:
Actually had the wrong names for the parameters. Suggesting we merge this as soon as possible as nobody proposed a better idea for the initial values of these parameters till now.

sw/airborne/boards/bebop/mt9f002.c
```diff
@@ -43,18 +43,18 @@
...
   .output_size = {
-    .w = 2048,
-    .h = 3320
+    .w = MT9F002_OUTPUT_WIDTH,
+    .h = MT9F002_OUTPUT_HEIGHT
   },
   .sensor_size = {
-    .w = 2048,
-    .h = 3320,
+    .w = MT9F002_OUTPUT_WIDTH,
```
flixr: Isn't the sensor size fixed?

flixr:
So does this actually work? Do you really want/need to set all the paramters (output_size, sensor_size and crop) to the same values?
At least sensor size sounds like something that is always fixed...

rmeertens:
Well, I/we don't know yet what values are possible. I think in this case even sensor size is not fixed. For the bebop you can choose:
- what part of the camera you want to read
- what you give to the ISP 
- what we want to extract at runtime. 

Anyways. Right now it is always fixed, this pull request makes that you can toy around. Some sizes must be the same as other sizes (i.e. the board_init and video_confix must have the same values). 
A non-caring user will notice nothing if we merge this... a caring programmer can suddenly play around. Up till now I saw no better ideas for setting the initial parameters.

flixr:
> I think in this case even sensor size is not fixed.

Then it probably shouldn't be called sensor size?

So you didn't test any other sizes so far?

rmeertens:
Tested several sizes, currently only 0.5 en 0.25 of what we have now. For the exact possibilities of the camera we have to talk to @fvantienen  who has the datasheet.

sw/airborne/boards/bebop/board.c
```diff
@@ -33,6 +33,7 @@
 #include "mt9v117.h"
 #include "mt9f002.h"
 #include "mcu.h"
+#include BOARD_CONFIG
```
flixr: any reason to include BOARD_CONFIG here instead of boards/bebop.h ?

rmeertens:
No reason at all. Fixed!

PR #1808, opened 2016-07-15
[Bebop] Camera functions

rmeertens:
I could be mistaken, but I think currently there are no easy functions for the camera at runtime (like: resize, move the cropped part, etc.). 
Do we want this? this pull requests lets you at least set the color gains.

### DIFF
--- a/sw/airborne/boards/bebop/mt9f002.c
+++ b/sw/airborne/boards/bebop/mt9f002.c
@@ -773,7 +773,7 @@ static uint16_t mt9f002_calc_gain(float gain) {
 /**
  * Sets the GreenR, Blue, Red and GreenB gains
  */
-static void mt9f002_set_gains(struct mt9f002_t *mt)
+void mt9f002_set_gains(struct mt9f002_t *mt)
 {
   write_reg(mt, MT9F002_GREEN1_GAIN, mt9f002_calc_gain(mt->gain_green1), 2);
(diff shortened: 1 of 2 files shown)


fvantienen:
Would be nice if we could add these functions to the camera structure. This way everyone camera can set these type of functions if they support it.

rmeertens:
True... but we have a large potential amount of camera's... some which do support the functions we have, some not, some extra... 
Is there some generic camera library we can use? And else I suggest to implement the methods for every camera, as every camera can probably do something different.

rmeertens:
Another idea is to create interfaces (whooo... C++?) and let each camera implement the suitable ones.

fvantienen:
That would be best and the same for the Camera itself. So you have a main class `CameraInterface` from which the `Mt9f002` extends.

The the camera interface has functions like `setSize`, `setColorGains`, `setExposure` which all return by default something like: `CamereInterface::NOT_AVAILABLE` in the `CameraInterface` class. This way each camera can implement the functions they support.

rmeertens:
Sounds pretty good! So for now we can still merge https://github.com/paparazzi/paparazzi/pull/1807 (as there will be some way to set the camera resolution) and close this pull request.

flixr:
@rmeertens so close this then?

rmeertens:
Yep, but with feedback on https://github.com/paparazzi/paparazzi/pull/1817 please ;)

PR #1809, opened 2016-07-16
Plot shapes onto GCS

rijeshaugustine:
This adds a shape plotting feature to the GCS.
It is useful for plotting obstacles or image outlines.

Messages are sent over the ivy bus to the gcs. Currently circles and polygons can be plotted.
Fill color is also available. Passing in "None" will disable fill. 

In the future an optional text flag may be a good addition or a popup.
```xml
  <message name="SHAPE" id="38">
    <field name="id"     type="uint8"/>
    <field name="linecolor"  type="string"/>
    <field name="fillcolor"  type="string" values="color|none"/>
    <field name="shape"  type="uint8" values="Circle|Polygon"/>
    <field name="status" type="uint8" unit="bool" values="create|delete"/>
    <field name="latarr"    type="int32[]"  unit="1e7deg" alt_unit="deg" alt_unit_coef="0.0000001"/>
    <field name="lonarr"    type="int32[]"  unit="1e7deg" alt_unit="deg" alt_unit_coef="0.0000001"/>
    <field name="radius" type="float" unit="m"/>
    <field name="alt" 	 type="int32" unit="mm"/>
  </message>
 ```

### CODE REVIEW

File: sw/ground_segment/cockpit/shapes.ml [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+ *
+ *)
+
+ module G = MapCanvas
+ open Latlong
+ module LL = Latlong
+ open Printf
+
+
+type shdata = {
```
gautierhattenberger: unused type

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,112 @@
...
+  shid : string;
+  shlinecolor : string;
+  shfillcolor : string;
+  shtype : float;
+  shstatus : float;
+  shlatarr : float array;
+  shlonarr : float array;
+  shradius : float}
+
+type circleshape = { mutable circsh : GnoCanvas.ellipse}
```
gautierhattenberger: not really needed since you could directly add the GnoCanvas.* to the hash table

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,112 @@
...
+
+let update_circle = fun id wgs84 color ?fill_color radius (geomap:MapCanvas.widget) ->
+  try
+    let gencircle = geomap#circle ~width:2 ?fill_color ~color:color wgs84.(0) radius in
+    if not (circle_exist id) then
+    let circleshape = {circsh = gencircle } in
+    Hashtbl.add circleshapes id circleshape;
+    else
+    let shape = Hashtbl.find circleshapes id in
+    shape.circsh#destroy ();
```
gautierhattenberger: the geomap object has `move_item` method

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,112 @@
...
+
+let update_shape = fun raw positions geomap->
+  try
+    if raw.shtype = 0. then
+    update_circle raw.shid positions  raw.shlinecolor raw.shradius geomap;
+    if raw.shtype = 1. then
+    update_polygon raw.shid positions raw.shlinecolor geomap
+  with _ -> ()
+
+let update_shape_with_fill = fun raw positions geomap ->
```
gautierhattenberger: could be merge with previous function with optional parameter

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,112 @@
...
+    update_circle raw.shid positions ~fill_color:raw.shfillcolor raw.shlinecolor raw.shradius geomap;
+    if raw.shtype = 1. then
+    update_polygon raw.shid positions ~fill_color:raw.shfillcolor raw.shlinecolor geomap
+  with _ -> ()
+
+let del_shape = fun raw ->
+  try
+    if raw.shtype = 0. then
+    let shape = Hashtbl.find circleshapes raw.shid in
+    Hashtbl.remove circleshapes raw.shid;
```
gautierhattenberger: actually, all drawable item (ellipse, segment, polygon,..) have the type `GnomeCanvas.re_p GnomeCanvas.item` (http://lablgtk.forge.ocamlcore.org/refdoc/GnoCanvas.html#TYPEellipse), so they could be handled in the same hash table.

sw/ground_segment/cockpit/shapes.ml
gautierhattenberger: my bad, it's used in Live.ml

sw/ground_segment/cockpit/live.ml
```diff
@@ -1478,6 +1478,27 @@ let get_intruders = fun (geomap:G.widget) _sender vs ->
...
+    shfillcolor = PprzLink.string_assoc "fillcolor" vs;
+    shtype = i "shape";
+    shstatus = i "status";
+    shlatarr = floatarr "latarr";
+    shlonarr = floatarr "lonarr";
+    shradius = (f "radius")} in
+  new_shmsg data geomap
+
+let listen_shapes = fun (geomap:G.widget) ->
+  safe_bind "SHAPE" (get_shapes geomap)
```
gautierhattenberger: please make the PR in pprzlink as well

gautierhattenberger:
would be nice to add the `Segment` type as well since it already exists in `MapCanvas`

sw/ground_segment/cockpit/shapes.ml
rijeshaugustine: I made separate hashtables for different items for easier indexing by users. They can send send two different shape types with the same id. The idea is sending unique identifiers to different "classes" of shapes.  A polygon is  an image outline while a circle is a path obstacle.

gautierhattenberger: you can make a key by merging the type and the id so you can use the same id for different types, not sure it really makes a big difference, but still looks cleaner to handle a single table for all items.

rijeshaugustine: Rect and ellipse take points individually as floats and have type `GnomeCanvas.re_p GnomeCanvas.item`
Polygon has type `GnomeCanvas.polygon_p item` since it takes in a float array.

I think I will still need two hash tables.

sw/ground_segment/cockpit/shapes.ml
rijeshaugustine: I was having issues setting a new circle with  `#absolute_affinity`

`let update_circle = fun id wgs84 color ?fill_color radius (geomap:MapCanvas.widget) ->
  try
    let gencircle = geomap#circle ~width:2 ?fill_color ~color:color wgs84.(0) radius in
    if (circle_exist id) then
    let shape = Hashtbl.find circleshapes id in
    shape#affine_absolute (geomap#convert_positions_to_points wgs84);
    else
    Hashtbl.add circleshapes id gencircle
  with _ -> ()`

Code like this will generate a circle and allow the change in location the once. Subsequent calls just draw a new circle.

rijeshaugustine: also if an update also wants to change color then a simple move_item will not work.

sw/lib/ocaml/mapCanvas.ml
```diff
@@ -592,15 +604,32 @@ object (self)
...
     let utm = LL.utm_of LL.WGS84 geo in
     let geo_east = LL.of_utm LL.WGS84 (LL.utm_add utm (radius, 0.)) in
     let (xe, _) = self#world_of geo_east in
     let rad = xe -. x in
-    let l = GnoCanvas.ellipse ?fill_color ~props:[`WIDTH_PIXELS width; `OUTLINE_COLOR color] ~x1:(x-.rad) ~y1:(y -.rad) ~x2:(x +.rad) ~y2:(y+.rad) group in
+    let l = GnoCanvas.ellipse ?fill_color ~props:[`WIDTH_PIXELS width; `OUTLINE_COLOR color; `FILL_STIPPLE (Gdk.Bitmap.create_from_data ~width:stpwidth ~height:stpwidth stpstr)] ~x1:(x-.rad) ~y1:(y -.rad) ~x2:(x +.rad) ~y2:(y+.rad) group in
+    l#show ();
+    l
+
+  method stipple_opacity = fun opacity ->
```
rijeshaugustine: @gautierhattenberger  I added this function to vary the fill stipple. It is not very nice looking, but I'm not sure if there is another way to vary it. 0 gives me no fill and 3 is opaque. 

I also added this as a public method. so it is mentioned in mapCanvas.mli. It should only be used by method polygon and circle within the same file. Is there a better way of dealing with this?

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,141 @@
...
+ *
+ *)
+
+open Latlong
+
+type shdata = {
+  shid : string;
+  shlinecolor : string;
+  shfillcolor : string;
+  shtype : float;
```
gautierhattenberger: type and status should not be float, at least int or even better a specific type

sw/ground_segment/cockpit/shapes.ml
```diff
@@ -0,0 +1,141 @@
...
+  Hashtbl.mem lineshapes id
+
+let text_exist = fun id ->
+  Hashtbl.mem textshapes id
+
+let update_circle = fun id wgs84 opacity fill_color color radius (geomap:MapCanvas.widget) ->
+  try
+    let gencircle = geomap#circle ~width:2 ~fill_color ~opacity ~color wgs84.(0) radius in
+    if (circle_exist id) then
+    let shape = Hashtbl.find circleshapes id in
```
gautierhattenberger: please fix indentation (and in other functions as well)

sw/lib/ocaml/mapCanvas.ml
gautierhattenberger: you can just create a function local to the MapCanvas module (e.g. at the top of this file). No need for a method nor a reference to the function in the mli file.

PR #1811, opened 2016-07-17
New paparazzicenter in Python

gautierhattenberger:
This is the result of a student project to develop a new paparazzi center based on Python/Qt.
See http://wiki.paparazziuav.org/wiki/Paparazzi_Center/Evolutions for the main idea and development process.
It can be tested by running the main paparazzi launcher with the `-python` option. I more convenient way might be added later.
This PR is mostly a way to not loose this work, and other people are encourage to improve it in many aspects (HMI, functionalities, ...).


flixr:
Should we clean this up a bit more before merging?

E.g. when I try to start it here it fails because some files are not found (e.g. you have one config somewhere that references some file that doesn't exist in your branch).
Also you need to cd into sw/supervision/python and run main.py in there instead of being able to run it directly. For this it would make sense to set the python pprzcenter up as proper python modules.

gautierhattenberger:
I know the correct handling of launching directory is not done, but I'm not really confident about how to do this properly in Python.
So, is anyone able to do this soon ? Otherwise we can open an issue for that. The point of this PR was avoid loosing this work.

flixr:
The proper directory handling and module setup is not that important right now.
But it would definitely be good if the whole thing wouldn't crash if it can't find a file (e.g. an airframe file referenced in one of the configs)...

flixr:
Hm.. it's a bit better, but now it seems it will simply not update the cache and instead show an old version instead of what is actually in the file...
I guess we could merge it like this, but IMHO it is still quite a far way until we can recommend people to use this...

gautierhattenberger:
Actually, what (should) happen is that if you start from a bad config, it will restore the default config. Otherwise it should work, and wrong filename only produce an error in the console.
I agree it is far from ideal, but we don't have the resources to do better for now.

flixr:
Ok, so we just have to keep that in mind... I just tested what happens if I modify the conf_example so that it is not valid anymore... but that is the "default" config if I see that correctly.

PR #1812, opened 2016-07-18
video usb logger fix

rmeertens:
parameters width and height set at runtime. 
removed a useless dependency. 
fixed warnings. 
now logging more data!

### DIFF
--- a/conf/modules/video_usb_logger.xml
+++ b/conf/modules/video_usb_logger.xml
@@ -12,7 +12,6 @@
     <define name="VIDEO_USB_LOGGER_HEIGHTH" value="272" description="Size of the to log images"/>
     <define name="VIDEO_USB_LOGGER_JPEG_WITH_EXIF_HEADER" value="TRUE" description="Whether to store data in the exif header or not"/>
   </doc>
-  <depends>video_rtp_stream</depends>
   <header>
     <file name="video_usb_logger.h"/>
   </header>
--- a/conf/modules/video_usb_logger.xml
+++ b/conf/modules/video_usb_logger.xml
@@ -12,6 +12,7 @@
     <define name="VIDEO_USB_LOGGER_HEIGHTH" value="272" description="Size of the to log images"/>
     <define name="VIDEO_USB_LOGGER_JPEG_WITH_EXIF_HEADER" value="TRUE" description="Whether to store data in the exif header or not"/>
   </doc>
+  <depends>video_thread,pose_history</depends>
   <header>
     <file name="video_usb_logger.h"/>
   </header>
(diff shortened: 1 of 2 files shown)


conf/modules/video_usb_logger.xml
```diff
@@ -12,7 +12,6 @@
     <define name="VIDEO_USB_LOGGER_HEIGHTH" value="272" description="Size of the to log images"/>
     <define name="VIDEO_USB_LOGGER_JPEG_WITH_EXIF_HEADER" value="TRUE" description="Whether to store data in the exif header or not"/>
   </doc>
-  <depends>video_rtp_stream</depends>
```
dewagter: I think it now depends on video_thread and some module that compiles has the jpeg.c

fvantienen:
You made this awesome `pose_history.xml` which is even thread safe.. So it would be nice to use it as it synchronizes the data more accurately. This can then be done in the video capturing thread ;)

fvantienen:
Btw isn't there an option to append all the images to one file(like a video file). I know you can do that for certain encodings.. Maybe also for jpeg?

conf/modules/video_usb_logger.xml
flixr: jep, depends on video_thread, the needed jpeg.c should be added to the sources here directly...

rmeertens:
fixed the dependencies and use the pose history. Will take the video file suggestion as a future feature request suggestion ;)

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -35,34 +35,30 @@
...
 static FILE *video_usb_logger = NULL;
-struct image_t img_jpeg;
+struct image_t img_jpeg_global;
+bool created_jpeg = FALSE;
 char foldername[512];
 int shotNumber = 0;
 
-void save_shot(struct image_t *img, struct image_t *img_jpeg)
+void save_shot_on_disk(struct image_t *img, struct image_t *img_jpeg);
+void save_shot_on_disk(struct image_t *img, struct image_t *img_jpeg)
```
flixr: If it's only an internal function, just make it static... then you also don't need the declaration before it...

sw/airborne/modules/computer_vision/video_usb_logger.c
```diff
@@ -35,34 +35,30 @@
...
 {
 
   // Search for a file where we can write to
   char save_name[128];
 
   sprintf(save_name, "%s/img_%05d.jpg", foldername, shotNumber);
```
flixr: please never use sprintf and use the save snprintf version instead

rmeertens:
Done.

flixr:
thx, but please fix indentation, codey style...

rmeertens:
done

PR #1813, opened 2016-07-18
[modules] put airspeed over ABI

dewagter:


fvantienen:
Because of some internet error it failed but is eems ok 👍

PR #1815, opened 2016-07-18
[chibios] recover from hard fault

gautierhattenberger:
In case of crash of the autopilot, the mcu is restarted and the FBW thread is restarted (low level peripherals as well) but not the AP thread, assuming it is the source of the problem.
It allows a safety pilot to take back the control of the plane in manual mode.
Telemetry is also restarted, but the configuration of xbee modules (when relevant) is not done (already done, and skip the active for faster reboot).


flixr:
LGTM

PR #1817, opened 2016-07-19
first step towards camera interface

rmeertens:
Hey everybody, 

Creating this pull request to get feedback from you before going further. As this is the first time I'm creating a C++ interface to be used by C I am having a lot of doubts about how to proceed. 

In the end we want to have a camera interface which every camera implements in a c++ class. Next to this we still need a binding with c... so we will call a function mt9f_camera_resize(w,h), which then takes the camera object and calls resize on that one?

As I don't have that much time to spend on this: what would the fastest way be to get a simple working program?

### DIFF
--- a/conf/boards/bebop.makefile
+++ b/conf/boards/bebop.makefile
@@ -36,7 +36,7 @@ $(TARGET).srcs +=  $(SRC_BOARD)/board.c
 
 # Compile the video specific parts
 $(TARGET).CFLAGS += -DI2C_BUF_LEN=56 -DUSE_I2C0
-$(TARGET).srcs +=  $(SRC_BOARD)/mt9v117.c $(SRC_BOARD)/mt9f002.c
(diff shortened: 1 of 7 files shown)


rmeertens:
Ok. Got some more code. Camera is now working with the C++ interface. @fvantienen @flixr could use some initial feedback on the direction i'm now going in. 

@fvantienen you mentioned that setting these parameters can't be done whenever I want. Can you indicate how and when we want to adjust them?

EwoudSmeur:
Is this still relevant, as @fvantienen is working on a vision framework? Should we close this?

flixr:
@rmeertens is this still relevant or can we close it?

rmeertens:
As @fvantienen started on the vision framework again (and I'm not building it anymore) I think it's up to him...

rmeertens:
But knowing Freek I can probably close it now ;)

PR #1818, opened 2016-07-20
more variables to configure and some comments

rmeertens:
Added an example in the comments about what you can and can't do with the camera (as far as I explored up till now).


flixr:
Please try to put the "area" it relates to in the commit message, if you just read that later you have no idea what it is about.....

PR #1820, opened 2016-07-21
Pull request edgeflow noisemeasurement fix

knmcguire:
Hi all,

I noticed that the latest changes of adding a noise measurement to the optical flow module (for the intention of use of the hff filter) were not exactly producing the necessary results. This pull request is to at least make it flyable again in the current master, after I heard several complaints from our lab.

The most important issue was this: The more corners the algorithm tracked, the higher the noise measurement (est velocity error = tracked corners/ max_corners) becomes. However this is not wanted for the ins_hff, since there a higher estimated velocity error results in less trust in that measurementof the Kalman filter. I made a new equation, which actually gives a lower noise measurement with more corners. 

In the future there needs to be a better, with optitrack groundtruth backed up, noise model, however this pull request will hopefully give an intermediate solution. I also added an optional median filter if the hff is not used or if more outliers needed to be filtered out.

### DIFF
--- a/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
+++ b/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
@@ -48,8 +48,8 @@ struct opticflow_t {
   uint8_t method;               ///< Method to use to calculate the optical flow
   uint16_t window_size;             ///< Window size for the blockmatching algorithm (general value for all methods)
   uint16_t search_distance;         ///< Search distance for blockmatching alg.
-  uint8_t derotation;             ///< Derotation switched on or off (depended on the quality of the gyroscope measurement)
-  uint8_t median_filter;          ///< Decides to use a median filter on the velocity
+  bool derotation;             ///< Derotation switched on or off (depended on the quality of the gyroscope measurement)
+  bool median_filter;          ///< Decides to use a median filter on the velocity
 
   uint16_t subpixel_factor;          ///< The amount of subpixels per pixel
   uint8_t max_iterations;           ///< The maximum amount of iterations the Lucas Kanade algorithm should do
--- a/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
+++ b/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
@@ -49,6 +49,7 @@ struct opticflow_t {
   uint16_t window_size;             ///< Window size for the blockmatching algorithm (general value for all methods)
   uint16_t search_distance;         ///< Search distance for blockmatching alg.
   uint8_t derotation;             ///< Derotation switched on or off (depended on the quality of the gyroscope measurement)
+  uint8_t median_filter;          ///< Decides to use a median filter on the velocity
 
   uint16_t subpixel_factor;          ///< The amount of subpixels per pixel
   uint8_t max_iterations;           ///< The maximum amount of iterations the Lucas Kanade algorithm should do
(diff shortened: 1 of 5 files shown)


knmcguire:
This branch has been testflown with a bebop. The additional median filter seemed to be efficient to filter out most of the vibrations, which made it way easier to control in guided mode

rmeertens:
do we need to change tudelft_bebop_opticflow.xml? What was wrong with the camera rotation and how did you now fix it? Line 548: why is it set here to 0.2?

Something not related to this pull request but relevant to optic flow: 
Currently if you suddenly lose opticflow we don't update the estimated velocity (no ABI message is sent). This is a problem when:
- you try to hover stable, your last detected velocity is 0.5, and then you lose opticflow: your drone will keep compensating against this 0.5 meters per second resulting in a fly away while your camera image is bad. 
- you try to keep a forward velocity, your detected velocity is 0.0, and then you lose opticflow: you will keep trying to fly faster and faster. 

We should think about how to handle loss of optic flow (dont know what level would be the best place).

knmcguire:
The camera rotation is already done when the image is grabbed, so this define wasn't necessary any more. It was already removed from the code itself.

the 0.2 is just an experimental number what I've experienced during my research. Right now I couldn't find a suitable quality measurement to fully describe the change in velocity error estimate, so for edgeflow its fixed (for now). There is an equation at the lukas kanade method.

Yes you are right, if the optical flow module shuts down, that will definitely be an issue. But if the hff is used, and no measurement of the velocity comes in for a while, it should handle itself.  But if ins is not on hff, yes the last value will stay there. It's an issue to be fixed but probably for another pull request in the future.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
```diff
@@ -49,6 +49,7 @@ struct opticflow_t {
   uint16_t window_size;             ///< Window size for the blockmatching algorithm (general value for all methods)
   uint16_t search_distance;         ///< Search distance for blockmatching alg.
   uint8_t derotation;             ///< Derotation switched on or off (depended on the quality of the gyroscope measurement)
+  uint8_t median_filter;          ///< Decides to use a median filter on the velocity
```
flixr: If you only use this to enable/disable the median_filter, it should be a `bool`.
Same actually goes for `derotation`

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -128,7 +128,7 @@ void opticflow_module_run(void)
                            opticflow_result.div_size,
                            opticflow_state.agl);
     //TODO Find an appropiate quality measure for the noise model in the state filter, for now it is tracked_cnt
-    if (quality > 0.8) {
+    if (quality < 0.8) {
```
flixr: For the future, quality is not the best term here (at least not for how it's used)... it's actually the noise (so lower is better), but for quality you would expect higher to be better...

knmcguire:
implemented your suggestions felix, thanks for the review.

rmeertens:
modules/computer_vision/opticflow_module.c: In function ‘opticflow_module_run’:
modules/computer_vision/opticflow_module.c:127:28: error: ‘quality’ undeclared (first use in this function) 

Aww;

knmcguire:
Woops, my bad. Will fix it tomorrow

Op 21 jul. 2016 om 18:08 heeft Roland Meertens <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


modules/computer_vision/opticflow_module.c: In function ‘opticflow_module_run’:
modules/computer_vision/opticflow_module.c:127:28: error: ‘quality’ undeclared (first use in this function)

Aww;

PR #1827, opened 2016-07-23
[Bebop] cam extra setting

rmeertens:
@fvantienen @dewagter 
Found that we can skip pixels when reading the image (instead of averaging, what we do now), giving us a higher FPS. Can't test it this or next week as I'm on vacation. Can somebody else quickly look at it?

### DIFF
--- a/sw/airborne/boards/bebop.h
+++ b/sw/airborne/boards/bebop.h
@@ -77,6 +77,14 @@
 #endif
 
 
+#ifndef MT9F002_X_ODD_INC_VAL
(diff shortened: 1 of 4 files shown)


fvantienen:
This seems like a good idea, only this gives problems with the ISP which makes it very complex to implement this correctly. So for now my advice is to not use this.

flixr:
So, shall we close this then?

rmeertens:
Sure, don't have time now to check all possible settings in this case :(

PR #1828, opened 2016-07-26
correction for bebop firmware version check during uploading

hannwoei:
To allow bebop firmware version 3.2.0 and higher, but not just 3.2.0.

### DIFF
--- a/sw/tools/parrot/bebop.py
+++ b/sw/tools/parrot/bebop.py
@@ -125,8 +125,8 @@ elif args.command == 'upload_file_and_run':
     #check firmware version
     v = parrot_utils.check_version(tn, '').strip()
     print("Checking Bebop firmware version... " + v )
-    if v != '3.2.0':
-        print("Error: please upgrade your Bebop firmware to version 3.2.0!")
+    if ((v < '3.2.0') or (v > '3.4.0-RC2')):
+        print("Error: please upgrade your Bebop firmware to version between 3.2.0 and 3.4.0!")
     else:
         print("Kill running " + f[1] + " and make folder " + args.folder)
         parrot_utils.execute_command(tn,"killall -9 " + f[1])


sw/tools/parrot/bebop.py
```diff
@@ -125,8 +125,8 @@ def bebop_status():
     #check firmware version
     v = parrot_utils.check_version(tn, '').strip()
     print("Checking Bebop firmware version... " + v )
-    if v != '3.2.0':
-        print("Error: please upgrade your Bebop firmware to version 3.2.0!")
+    if v < '3.2.0':
```
kirkscheper: Is this correct to compare the string value? It may be more robust to to convert the string values to a decimal and compare the major, minor and increment values individually.

hannwoei: I am new to python. maybe some programming experts can make it better. I have tested with a simple python script and it just works. Some results are:
case1:
v=3.2.0
new_v=3.3.0
v>new_v = False
case2:
v=3.2.0
new_v=1.1.0
v>new_v = True
case3:
v=3.2.0
new_v=4.1.0
v>new_v = False

kevindehecker:
I had a reason that I made it check for a specific firmware. Parrot re-arranged the motor order out of the blue. This led me to crash my PPRZ+Bebop setup, and I know many others did so as well before the problem got fixed in PPRZ. Furthermore, e.g. the 5.8 release still crashes an updated bebop... Yesterday a new student attempted to fly 5.8 + updated Bebop, if it wasn't for our warnings he would probably have crashed.

So, I'm not convinced this PR is a good solution. I do agree the upgrade rate of the Bebop seems rather high, which makes my solution impracticable... but at least safe. In the very least I think PPRZ should not accept to fly a Bebop with an unknown (new) firmware without extensive warning.

sw/tools/parrot/bebop.py
kevindehecker: Google + SO :)
http://stackoverflow.com/questions/11887762/compare-version-strings

hannwoei:
I flew an updated bebop (v3.3.0)  with the current paparazzi master yesterday too and it did not have problems. However, I do agree that flying with untested firmware is not a good idea. 

FYI, paparazzi wiki suggests that the user should use bebop firmware version3.2.0. or higher. I thought this has been tested.

kirkscheper:
@kevindehecker What is the solution here in your opinion? Shall we do a check that it is >3.2.0 && <3.3.0 and update the upper limit as new versions are tested by developers?

kevindehecker:
In my opinion, yes. In practice this means that PPRZ will not work with Bebop's for a few days after a firmware update from Parrot. Arguably this is better than taking the risk of crashing Bebops by uninformed users or (especially with old PPRZ major releases).

EwoudSmeur:
can we include 3.4? I am already flying with this

kirkscheper:
@hannwoei can you update your pull request to consider these comments?

hannwoei:
Yes, I have updated the file. Now we can only use version 3.2 to 3.4.

sw/tools/parrot/bebop.py
```diff
@@ -125,8 +125,8 @@ def bebop_status():
     #check firmware version
     v = parrot_utils.check_version(tn, '').strip()
     print("Checking Bebop firmware version... " + v )
-    if v != '3.2.0':
-        print("Error: please upgrade your Bebop firmware to version 3.2.0!")
+    if ((v < '3.2.0') or (v > '3.4.0')):
+        print("Error: please upgrade your Bebop firmware to version 3.2.0 or higher!")
```
kirkscheper: I would also change the printed error to suit the range

kirkscheper:
@EwoudSmeur can you test that this still works with the recent changes?

EwoudSmeur:
Some of my bebops have version 3.4.0-RC2. This still does what it is supposed for 3.0.0 and 3.3.0

PR #1830, opened 2016-07-28
[rpm_sensor] filter and broadcast rpm via ABI

dewagter:


dewagter:
merged via https://github.com/paparazzi/paparazzi/commit/5f7e2b3381485ed6293d9e6cb5b9e7872007e5c4

PR #1831, opened 2016-07-29
Nps update - proof of concept of HITL+SITL integration

podhrmic:
Continuation of #1789 this time using existing infrastructure (and glib).

The commit is rather large, but there are few separate commits that could be merged immediately:
- 9a9c58bc4424594593a7f45f06a49376c5dde606 Added AggieAir control panel (simple)
- support for higher baudrates & minor fixes in existing AP code (5e1576c27699fb5839ea44413d38acf673c1fa11 27393daecfefc01303b58935d4e72d822080f987 8dd3699f96a50466be965f05eb1c7f14cf10f5fe 2895ec12a5f65926642a3d7502bc6aced51e103c)
- I added a "flight logger" module -> it is crucial for HITL, basically it sends telemetry messages (mainly COMMANDS) over a serial link. We also use it for logging (we have an Overo that saves data on SD card) 576169cdbe4c6e4116d32fcdf491cd5e3d766be6
- Then finally I wanted to be able to use NPS as SITL and HITL and have it modular - so far I have to add a flag to the airframe config file, I would prefer having a separate HITL target eventually (and rename NPS to SITL)
- `nps_main.c` and `nps_hitl_main.c` are using threads and the timing/etc seems to work fine. It is mostly a proof of concept now, I am hoping to get some feedback from you before I clean it up more. Eventually I would like to have support for Xsens and other INS units (not just Vectornav).

I ll be on vacations next week, but I was hoping you could have a look at it in the meantime. @flixr @gautierhattenberger @deonblaauw ?
Thanks.


flixr:
The small fixes and corrected typos you can directly commit to master...

conf/firmwares/subsystems/shared/nps.makefile
```diff
@@ -45,6 +45,40 @@ nps.srcs += $(NPSDIR)/nps_main.c                 \
...
+$(info USE_HITL defined)
+nps.srcs += $(NPSDIR)/nps_hitl_main.c
+
+ifdef AP_DEV
+nps.CFLAGS += -DAP_DEV=\"$(AP_DEV)\"
+else
+nps.CFLAGS += -DAP_DEV=\"/dev/ttyUSB2\"
+endif
+
+ifdef AP_BAUD
```
flixr: `AP_BAUD ?= B3000000` is simpler and more readable than the ifdef

flixr:
The nps threads look good to me, but as you said it needs more cleanup to avoid the code duplication in nps_hitl_main and the uart configuration should be added as nps commandline parameters instead of defines?

podhrmic:
Work in progress, will let you know once it is ready for review.

podhrmic:
OK, I am not quite sure what happened here (I did `git rebase master` in the `nps_update` branch), but the HITL is ready to be merged.

Original NPS functionality is kept, so it works as before. You can define USE_HITL which will trigger the HITL option. I split the common nps stuff into `nps_main_common.c`, the same for nps makefiles.

There are two things which I would like do to but don't know how:
- [ ]  Add new target - `nps_hitl` and `nps_sitl` where the second one would be the current NPS, and the first one would be NPS+HITL (so you don't have to change it in the airframe config every time)
- [ ]  Have vectornav automatically add the nps_ins_vectornav.c when compiled as `nps_hitl` so it is not hardcoded in the nps makefile (that way we can implement more INS sensors, like Xsens etc)

@gautierhattenberger @flixr @deonblaauw Feedback is welcome!

flixr:
You rebased but then merged your old branch in again...
Can you please reset to aad06c8e4b672131ec0de76561ecb30d3546a902 and push again?

flixr:
> After I rebased the latest commits from master, the autopilot doesn't start throttle when launched. I suspect there is some special safety procedure added? Because it worked just fine before the rebase...

Nope, nothing like that added. See https://github.com/paparazzi/paparazzi/compare/33e775acdf4c194f292e34a353c09bdf0581f5bf...e4c791904d9750dc0b6f2674c6df12e04e2d6fca

flixr:
I think you should also run the sensor simulation in HITL and use their values to fake the INS vectornav message. That way you take the body_to_imu into account and you can configure the noise, delay, etc..

flixr:
> I am also not sure why even in HITL mode I have to initialize the nps autopilot - for some reason it doesn't work without the proper (and yet unnecessary) init in nps_autopilot_fixedwing...

Can please be a bit more specific? What exactly do you mean by "doesn't work"?

Also it would make sense to only start the FG thread if a fg_host is configured...

flixr:
And a `nps_hitl` target would definitely make sense, don't think its worth renaming the `nps` target to `nps_sitl` though... IMHO that would just need a lot of additional changes without added benefit.

podhrmic:
OK, this should fix the double merge. I ll do more tests this week.

podhrmic:
Also - for the name of the targets. I would prefer changing it to `sitl` and `hitl` (or possibly `nps_sitl` and `nps_hitl`) because it is more descriptive -> both use nps as a simulation backend so it could be a part of the name. But for new developers/users it is easier to understand what `sitl` and `hitl` is, rather than `nps`.

`nps_sitl` and `nps_hitl` seems to be also most backwards-compatible, because people used to `nps` won't be too confused, and new users will immediately recognize the functionality. The common parts I would put into `nps_common.makefile` or something like that.

There is still the old `sim` target - I know it works and is good for simple stuff, but are people still using it? Maybe it is time to send to the `legacy code` folder...

gautierhattenberger:
I'm still using the basic sim. Since it is pretty independent, although limited, I don't see a reason to suppress it for now.

podhrmic:
@flixr fixed the start issue, now fixedwing SITL/HITL works nicely
fg_thread and ap_init fixed.

> I think you should also run the sensor simulation in HITL and use their values to fake the INS vectornav message. That way you take the body_to_imu into account and you can configure the noise, delay, etc.

I ll make it a TODO later, since it is not really needed for Vectornav (or INS per say)

@gautierhattenberger sure thing, I was just curious.

flixr:
Renaming the nps target is just a bit of a pain... first it breaks most airframe files and needs to be changed in all modules, airframe files, you need to tell people that until v5.9 they need to use `nps` after that `nps_sitl`, etc....
IMHO renaming it to nps_sitl to only make the target name a bit more clear is not worth it, but if more people/users think that this would be very beneficial I'm not strongly opposed to it either...

And what was your idea with RC type NONE? This seem totally superfluous to me... if you don't need it why even include/call it in the first place?

podhrmic:
I wanted to avoid calling `nps_radio_control_init(type_rc, num_rc_script, rc_dev);` because it does nothing in HITL. But for some reason I have to call `nps_autopilot_init()` otherwise the simulation doesn't initialize properly (the problem is around `Fbw_init()` and `Ap_init()` -> `downlink_init()`). 

But now it is breaking things so I ll look into it more:-/

flixr:
It would probably be enough to call `downlink_init()` explicitly instead of `nps_autopilot_init()`

flixr:
Regarding the flight_logger module: I think it would make sense to rename it, as it is really more generic than than and doesn't log anything. It's just an extra telemetry process/link.
We already have extra_dl, so maybe we should name this simply extra_telemetry.

podhrmic:
I need to merge this so I can move on with other stuff that depends on this (and this PR is already quite large). NPS and HITL works as is, but you have to specify in the airframe config file which one you want (NPS or HITL). @flixr @gautierhattenberger unless you have any large objections I will merge this as is.

flixr:
As already noted earlier, I think we should rename the flight_logger module...

Also the file `conf/radios/AGGIEAIR/aggieair_sbus_fakerator.xml` is missing, so your conf won't build...
And in your conf you have `stabilization_att_float_eulers` settings added, but that doesn't match your stabilization subsystem and compilation fails as well...

sw/simulator/nps/nps_ins.h
```diff
@@ -0,0 +1,44 @@
...
+#ifdef CONTROL_FREQUENCY
+#define INS_FREQUENCY CONTROL_FREQUENCY
+#else
+#define INS_FREQUENCY PERIODIC_FREQUENCY
+#endif
+#endif
+
+uint8_t *ins_buffer;
+
+void nps_ins_init(void);
```
flixr: These function declarations should be `external`

sw/simulator/nps/nps_ins.h
```diff
@@ -0,0 +1,44 @@
...
+// if undefined, match with control frequency because that is how it should be used
+#ifndef INS_FREQUENCY
+#ifdef CONTROL_FREQUENCY
+#define INS_FREQUENCY CONTROL_FREQUENCY
+#else
+#define INS_FREQUENCY PERIODIC_FREQUENCY
+#endif
+#endif
+
+uint8_t *ins_buffer;
```
flixr: Should be external and actually created in the c file...

conf/modules/ins_vectornav.xml
```diff
@@ -48,4 +48,10 @@
       endif
     </raw>
   </makefile>
+  
+  <!-- For some reason this doesn't work ?
```
flixr: Of course not: This will load the module (and hence try to call the init and event functions which don't exist in the NPS implementation).

conf/firmwares/subsystems/shared/nps_hitl.makefile
```diff
@@ -0,0 +1,29 @@
+# Hey Emacs, this is a -*- makefile -*-
+
+#
+# NPS HITL Simulator
+#
+# HITL specific makefile
+#
+
+nps.CFLAGS  += -DUSE_HITL=1
```
flixr: This doesn't seem to be needed/used anywhere...

sw/simulator/nps/nps_fdm_jsbsim.cpp
```diff
@@ -597,7 +604,8 @@ static void init_jsbsim(double dt)
   }
 
   // initial commands to zero
-  feed_jsbsim(0.0, 0.0, 0.0, 0.0);
+  double init_commands[5] = {0.0};
```
flixr: This results in a segfault when you only have 4 commands (like for quadrotor models)...
The commands_nb must match the number of actuator names in feed_jsbsim

conf/firmwares/subsystems/shared/nps_hitl.makefile
podhrmic: removed

conf/modules/ins_vectornav.xml
podhrmic: there is not a good way around this - eventually NPS and HITL will be different targets, but for the time being lets leave it as is. The only INS you can use in HITL is Vectornav so doesn't really matter right now.

sw/simulator/nps/nps_ins.h
podhrmic: done, although it makes no difference for compilation

sw/simulator/nps/nps_ins.h
podhrmic: Done.

sw/simulator/nps/nps_fdm_jsbsim.cpp
podhrmic: Fixed

flixr:
Can you please run the new/modified files through the formatter and get rid of the double commits?

sw/simulator/nps/nps_autopilot.h
```diff
@@ -18,8 +18,10 @@
 #ifdef NPS_ACTUATOR_NAMES
 #define NPS_COMMANDS_NB COMMANDS_NB
 #else
```
podhrmic: done

podhrmic:
Done. How do I remove the double commits?

flixr:
You can rebase/cherry pick your stuff onto the part that branched off later... or simply cherry pick your latest changes onto the nps_update branch in my repo where I already removed the double commits...

Regarding the naming of the extra telemetry module: why xgear? What is that supposed to mean?
As I already mentioned further up, it is simply a secondary/extra telemetry and we already have `extra_dl`, so why not name this `extra_telemetry` or something like that?
@gautierhattenberger what do you think about the naming for this module?

flixr:
@gautierhattenberger what would be the advantage of having the extra/secondary telemetry functionality in the `extra_dl` module instead of a separate one?

gautierhattenberger:
Just to avoid having several modules doing similar things but not exactly. Not absolutely needed, I just thought it would be a bit better.

flixr:
So for me calling it `extra_telemetry` (analogous to `extra_dl`) or `secondary_telemetry` is fine.
The usage can further be described in the module doc...

podhrmic:
I actually merged it with the `extra_dl` because after all it seemed natural to be able to do both in one module. Will push an update soon. @flixr btw I don't know how can I push commits into your branch - what if I just push commits here and then when this is merged all small commits get squashed into one anyway....

flixr:
You can't push commits to my branch, but you can reset your branch to mine (and cherry pick or commit on top then)... and force push again

podhrmic:
@flixr can we just squash merge this? Cherry picking and resetting this branch is nice but tedious, and the commits will be merged as one so there is no real point in it...

podhrmic:
ouch, the last commit was unintentional:-/

podhrmic:
@flixr can we merge it?

conf/modules/extra_dl.xml
```diff
@@ -10,6 +10,7 @@
     <file name="extra_pprz_dl.h"/>
   </header>
   <init fun="extra_pprz_dl_init()"/>
+  <periodic fun="extra_pprz_dl_periodic()" freq="PERIODIC_FREQUENCY" autorun="TRUE"/>
```
flixr: Shouldn't this run at TELEMETRY_FREQUENCY instead of PERIODIC_FREQUENCY?

flixr: hm.... but that means that the periods specified in the telemetry file won't be correct...

podhrmic: what would be the correct solution then?

flixr: can't you use TELEMETRY_FREQUENCY and add in your airframe file for the nps target `<configure name="TELEMETRY_FREQUENCY" value="512"/>`?

flixr: or set the telemetry freq to 512 in the nps hitl makefile...

podhrmic: I was using ```<configure name="TELEMETRY_FREQUENCY" value="PERIODIC_FREQUENCY""/>``` for all targets, because this is not only for NPS, has to work like that for AP too...

podhrmic: The AggieAir airframes should have that. Should become clear with more examples.

podhrmic:
@flixr so what else do you want me to change here? This is ready to merge, but I can't do it because it is hanging on some review you requested....:-/ I would really like to have this done asap so we can move forward here...

flixr:
The extra_pprz_dl_periodic() function should still be called at TELEMETRY_FREQUENCY....

podhrmic:
@flixr @gautierhattenberger Can you merge it pls?

gautierhattenberger:
I think it is okay. At least nps is working in sitl for rotorcraft and fixed-wing.

PR #1832, opened 2016-07-29
[rpm_control] throttle curve with constant rpm PI control

dewagter:
- rpm option in throttle curve
 - PI control for constant rpm
 - trim for in flight ajustments

### DIFF
--- a/conf/modules/heli_throttle_curve.xml
+++ b/conf/modules/heli_throttle_curve.xml
@@ -9,7 +9,6 @@
       <dl_settings NAME="ThrottleCurve">
          <dl_setting var="throttle_curve.rpm_fb_p" min="0.00" step="0.005" max="100" shortname="rpm_fb_p" param="THROTTLE_CURVE_RPM_FB_P" />
          <dl_setting var="throttle_curve.rpm_fb_i" min="0.00" step="0.005" max="100" shortname="rpm_fb_i" param="THROTTLE_CURVE_RPM_FB_I" />
-         <dl_setting var="throttle_curve.throttle_trim" min="-9600" step="0" max="9600" shortname="throttle_trim" />
       </dl_settings>
     </dl_settings>
   </settings>
--- a/conf/modules/heli_throttle_curve.xml
+++ b/conf/modules/heli_throttle_curve.xml
@@ -4,6 +4,15 @@
   <doc>
     <description>Throttle Curve Mixers</description>
   </doc>
+  <settings>
+    <dl_settings>
+      <dl_settings NAME="ThrottleCurve">
+         <dl_setting var="throttle_curve.rpm_fb_p" min="0.00" step="0.005" max="100" shortname="rpm_fb_p" param="THROTTLE_CURVE_RPM_FB_P" />
+         <dl_setting var="throttle_curve.rpm_fb_i" min="0.00" step="0.005" max="100" shortname="rpm_fb_i" param="THROTTLE_CURVE_RPM_FB_I" />
+         <dl_setting var="throttle_curve.throttle_trim" min="-9600" step="0" max="9600" shortname="throttle_trim" />
+      </dl_settings>
+    </dl_settings>
+  </settings>
   <header>
     <file name="throttle_curve.h"/>
   </header>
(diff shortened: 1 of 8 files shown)


conf/modules/heli_throttle_curve.xml
```diff
@@ -4,6 +4,15 @@
   <doc>
     <description>Throttle Curve Mixers</description>
   </doc>
+  <settings>
+    <dl_settings>
+      <dl_settings NAME="ThrottleCurve">
+         <dl_setting var="throttle_curve.rpm_fb_p" min="0.00" step="0.005" max="100" shortname="rpm_fb_p" param="THROTTLE_CURVE_RPM_FB_P" />
+         <dl_setting var="throttle_curve.rpm_fb_i" min="0.00" step="0.005" max="100" shortname="rpm_fb_i" param="THROTTLE_CURVE_RPM_FB_I" />
+         <dl_setting var="throttle_curve.throttle_trim" min="-9600" step="0" max="9600" shortname="throttle_trim" />
```
fvantienen: This should not be part of this PR ;)

fvantienen:
Actually strange that this build, since this should break some airframes that use this except for the outback airframe. Because of the commands change (COLLECTIVE, TRHUST) in comparison with before `throttle_curve.throttle` and `throttle_curve.collective`.

sw/airborne/modules/helicopter/throttle_curve.h
```diff
@@ -46,12 +47,21 @@ struct throttle_curve_t {
...
   int16_t collective;                         ///< Output collective of the throttle curve
+  uint16_t rpm;                               ///< Output RPM of the throttle curve
+
+  uint16_t rpm_meas;                          ///< RPM measured
+  bool rpm_measured;                          ///< Whenever the RPM is measured
+  float rpm_err_sum;                          ///< Summed RPM error
+  float rpm_fb_p;                             ///< RPM feedback p gain
+  float rpm_fb_i;
+
+  int32_t throttle_trim;                             ///< RPM feedback i gain
```
fvantienen: Also here it should be removed and this comment should be above.

conf/modules/heli_throttle_curve.xml
dewagter: why not?

sw/airborne/modules/helicopter/throttle_curve.c
```diff
@@ -44,41 +73,104 @@ void throttle_curve_init(void)
...
                                 + ((curve.collective[curve_p + 1] - curve.collective[curve_p]) * x / curve_range);
+    if(curve.rpm[0] != 0xFFFF)
+      throttle_curve.rpm = curve.rpm[curve_p]
+                            + ((curve.rpm[curve_p + 1] - curve.rpm[curve_p]) * x / curve_range);
+    else
+      throttle_curve.rpm = 0xFFFF;
+  }
+
+  // Trim in curve 3 hack
+  if(throttle_curve.mode == 2) {
```
fvantienen: See here....

conf/modules/heli_throttle_curve.xml
fvantienen: This trim was for our forward flight. (I removed throttle with this in a specific curve). See the other comment: "see here"

conf/modules/heli_throttle_curve.xml
```diff
@@ -4,6 +4,14 @@
   <doc>
     <description>Throttle Curve Mixers</description>
```
flixr: please document the new defines here...

flixr:
All airframes files that use this need to be updated, the builds don't fail because there is no airframe in conf_tests.xml that uses this...
Also please add some descriptions to the modules on what exactly they do and how they should be used.
Also the added rpm_control.xml file seems superfluous...
And please fix the code style...

flixr:
Also why change the signature of throttle_curve_run?
Now you don't pass `autopilot_motors_on` as parameter anymore but instead take the global var directly... but add `ap_mode` as parameter instead of using the global var in that case...

dewagter:
closing in favor of newer https://github.com/paparazzi/paparazzi/issues/1878

Issue #243, opened 2012-07-25
Airborne code organized into "process"


gautierhattenberger:
In order to prepare the use of some light RTOS (maybe some day...), the airborne code should be organized as much as possible into "process".
It is already started since the new sys_time has been introduced. More effort should be placed into that. Especially the modules could be organized into such processes.

A possible decomposition of process:
- stabilization
- navigation
- datalink
- estimation
- radio control
- safety/sanity checks ?
- extra stuff such as payload (lowest priority)
##

gautierhattenberger:
the code is already more or less structured into "processes" and RTOS versions even exist for some extra tasks (SDlog, wind estimation, ...)

PR #1835, opened 2016-08-06
Telemetry modules

gautierhattenberger:
Convert telemetry subsystems to modules.

- [x] needs hardware testing
- [x] hard fault recovery need telemetry modules init

The second point is not done yet since modules init are part of the global AP thread. Should I just call all modules init (and downlink init) or change the modules generator to be able to group functions (will be useful when converting FBW subsystems to modules) ?


podhrmic:
@gautierhattenberger IMHO it would make sense if each of the modules had their own thread with a separate recovery mechanism (since eventually everything will be a module).

Should we also mutex critical resources in the telemetry modules? Because I could imagine that now you can easily have multiple telemetries running at the same time - so access to lets say state variables etc. should be protected, right?

gautierhattenberger:
Not sure it is really needed to have each module in a thread, but we can definitely group them (already have some sort of tasks: navigation, control, safety, payload, ...).
Regarding the access of shared resources, we need to discuss that a little bit (and probably not in this thread). I already asked if state interface for instance should be protected in a transparent way like what you did in your branch, or if anyone using it should request for a mutex to use it (second option being more efficient but ask for a lot of work). At least the current implementation allows multiple threads to send messages other the telemetry system (it already the case with AP and FBW).

My real concern is more about the practical implementation and naming:
- should we have all the functions (init, periodic, event) of a module in the same "thread" or should we specify this for all functions independently.
- mix the two options: a global value can be defined, overwritten by local ones if needed
- what is the best name: process (not nice, but we already use this key word in other config files), thread (not really better I think, can be misleading if bare metal version is used), task (I would choose this one), group (not really nice and too generic). Any better idea ?

gautierhattenberger:
see #243

podhrmic:
Just quickly about the mutexes - the current solution (protect the state interface) is IMHO the way to go (at least for now).

About your questions:

> should we have all the functions (init, periodic, event) of a module in the same "thread" or should we specify this for all functions independently.

I would put it all in the same "thread" - like you start a thread, and it initializes what it needs. Should be easier to use and more error-proof.


> mix the two options: a global value can be defined, overwritten by local ones if needed

Can you give an example? I am not sure if I fully follow. 

> what is the best name: process (not nice, but we already use this key word in other config files), thread (not really better I think, can be misleading if bare metal version is used), task (I would choose this one), group (not really nice and too generic). Any better idea ?

Task +1

flixr:
For most modules it probably doesn't make sense to give them their own thread...

Task is probably an OK name

gautierhattenberger:
Tested and working on:
- sim and nps
- pprz and xbee on STM32F4
- ardrone2

anyone for testing superbit and bluegiga ?

podhrmic:
I can test Lisa F1 today (transparent), although I don't expect any problems here.

flixr:
Tested and working:
- transparent and transparent_usb on LisaMX (STM32F4)
- transparent on LisaM 1.0 (STM32F1) (usb_serial does not work on v1.0, but should work on later versions)

flixr:
@fvantienen @kirkscheper could you test superbitrf and bluegiga?

kirkscheper:
I am on vacation :) also I've been working on an old master on a big
project, will update and test when I am back.

Kirk

flixr:
@kirkscheper have you had a chance to test bluegiga yet?

kirkscheper:
No, can test it on Monday though, that ks for the reminder.

Kirk

kirkscheper:
Bluegiga doesn't work with this branch but actually it doesn't currently work with master either. I have been using a branch based on an older master so haven't noticed this. I don't receive any downlink messages... Have there been any changes to SPI lately?

gautierhattenberger:
I don't think so. Can you give us the reference of the working version ?

kirkscheper:
I am trying some older versions so see where it breaks. get back to you soon

kirkscheper:
First break happens at commit Coverity fixes (#1838) b68d1661aa6b09408eaab6cae87c55731af81848. An error in that commit prevents correct messages from being interpreted by the GCS. I still receive some downlink data though.

When the imu refactor is commited however with [imu] convert imu subsystems to modules (#1788) 17d32772c0e65cc738ba905fee4cbd4dd3297c87 the downlink is totally broken.

gautierhattenberger:
Regarding the fix https://github.com/paparazzi/paparazzi/commit/b68d1661aa6b09408eaab6cae87c55731af81848 can you try to revert the line 238 in bluegiga.c (although it should be more dangerous).
For the imu modules, it is really not related to datalink stuff, that's really strange.

kirkscheper:
The imu update does play around with spi and i2c defaults though. I am guessing it has something to do with that.

kirkscheper:
One difference is that the new lisa s imu defines ASPRIN_2_I2C and USE_I2C while the old one didn't. The ap_srcs.list doesn't define those two at all. Any chance that might have any effect?

kirkscheper:
ok, on a hunch I removed those definitions from master and it works again...

kirkscheper:
Same fix fixes this PR so it seems like this refactor also works for Bluegiga.

gautierhattenberger:
can you try to add `case="upper|lower"` to the lines configuring the SPI and I2C dev in the file imu_lisa_s_v1.0.xml ?

kirkscheper:
Nope, doesn't work.

gautierhattenberger:
do you have your patch in a branch somewhere ?

kirkscheper:
https://github.com/kirkscheper/paparazzi/tree/bluegiga_fix

Its is strange the i2c definition was not in the previous xml for aspirin or lisa/s

gautierhattenberger:
@kirkscheper can you test the patch of #1858 ?

flixr:
so I guess this is good to merge now?

gautierhattenberger:
It seems to work. I have only one concern: it is changing a little bit the module generator to have functions split in different "task". This part have not been tested a lot so far. Even if it should be fine, I initially planed to keep this for after the release. If you think it is fine, feel free to merge it.

flixr:
Maybe needs a quick re-test after latest changes?

gautierhattenberger:
re-tested with pprz and xbee over model, and pprz over udp (and it is still working)

gautierhattenberger:
can I merge this one ?

flixr:
LGTM

dewagter:
Nice work, thanks!

PR #1837, opened 2016-08-08
[superbitrf] Removed unused code

podhrmic:
This code is never executed - @fvantienen do you have any more insight in this?


podhrmic:
Any feedback? Will merge it soon.

flixr:
Looks good to merge for me as it just removes the dead code, but would be nice to get feedback from @fvantienen if it is something that would be needed in some cases and should be kept around for reference or so...

PR #1838, opened 2016-08-08
Coverity fixes

podhrmic:
A couple of fixed from latest coverity scan.

### DIFF
--- a/sw/airborne/modules/gps/gps_ubx_ucenter.c
+++ b/sw/airborne/modules/gps/gps_ubx_ucenter.c
@@ -167,10 +167,10 @@ void gps_ubx_ucenter_event(void)
   // Read Configuration Reply's
   switch (gps_ubx.msg_class) {
     case UBX_ACK_ID:
-      if (gps_ubx.msg_id & UBX_ACK_ACK_ID) {
+      if (gps_ubx.msg_id == UBX_ACK_ACK_ID) {
         gps_ubx_ucenter.reply = GPS_UBX_UCENTER_REPLY_ACK;
         DEBUG_PRINT("ACK\n");
-      } else if (gps_ubx.msg_id & UBX_ACK_NAK_ID) {
+      } else if (gps_ubx.msg_id == UBX_ACK_NAK_ID) {
         gps_ubx_ucenter.reply = GPS_UBX_UCENTER_REPLY_NACK;
         DEBUG_PRINT("NACK\n");
       }
--- a/sw/airborne/modules/nav/nav_survey_zamboni.c
+++ b/sw/airborne/modules/nav/nav_survey_zamboni.c
@@ -67,7 +67,7 @@ bool nav_survey_zamboni_setup(uint8_t center_wp, uint8_t dir_wp, float sweep_len
 
   // if turning right leave circle before angle is reached, if turning left - leave after
   if (sweep_spacing > 0) {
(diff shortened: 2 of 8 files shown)


sw/airborne/modules/gps/gps_ubx_ucenter.c
```diff
@@ -167,10 +167,10 @@ void gps_ubx_ucenter_event(void)
   // Read Configuration Reply's
   switch (gps_ubx.msg_class) {
     case UBX_ACK_ID:
-      if (gps_ubx.msg_id & UBX_ACK_ACK_ID) {
+      if (gps_ubx.msg_id == UBX_ACK_ACK_ID) {
```
flixr: This is not the same... is the `==` really correct?

sw/airborne/modules/nav/nav_survey_zamboni.c
```diff
@@ -67,7 +67,7 @@ bool nav_survey_zamboni_setup(uint8_t center_wp, uint8_t dir_wp, float sweep_len
 
   // if turning right leave circle before angle is reached, if turning left - leave after
   if (sweep_spacing > 0) {
-    zs.pre_leave_angle -= zs.pre_leave_angle;
+    zs.pre_leave_angle = 0;
```
flixr: what is this change about?

sw/airborne/subsystems/datalink/bluegiga.c
```diff
@@ -235,7 +235,7 @@ void bluegiga_load_tx(struct bluegiga_periph *p, struct spi_transaction *trans)
     bluegiga_increment_buf(&p->tx_extract_idx, packet_len);
 
     // clear unused bytes
-    for (i = packet_len + 1; i < trans->output_length; i++) {
+    for (i = packet_len + 1; i < BLUEGIGA_BUFFER_SIZE; i++) {
```
flixr: What if `trans->output_length` is smaller than `BLUEGIGA_BUFFER_SIZE`?

podhrmic: p->work_tx[i] has length of BLUEGIGA_BUFFER_SIZE, so you are essentially erasing the whole buffer every time (equivalent to memset)

sw/airborne/modules/nav/nav_survey_zamboni.c
podhrmic: I don't know - but the statement was effectively setting the angle to zero. Maybe NPS would  give the answer?

sw/airborne/modules/gps/gps_ubx_ucenter.c
podhrmic: See a few lines below - gps_ubx.msg_id == xxx is used instead of &
UBX_ACK_ACK_ID = 0x01, UBX_ACK_NAK_ID = 0x00
Before the second comparison was always false, now both should work fine.
Btw the whole module is poorly written..:(

podhrmic:
@flixr any more comments on this? If not I would merge it...

sw/airborne/modules/gps/gps_ubx_ucenter.c
flixr: Yeah, the module could really use an overhaul... see also #1649

flixr:
LGTM

PR #1839, opened 2016-08-10
set nav cmds instead of angles in nav manual mode

EwoudSmeur:
I think having dedicated variables for the commands in nav manual mode makes more sense than abusing the nav angle setpoints for this purpose. 
This can also resolve bugs when using this mode: nav_heading is for instance also set (as a BFP angle) by guidance_h_nav_enter()

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/navigation.c
+++ b/sw/airborne/firmwares/rotorcraft/navigation.c
@@ -82,6 +82,7 @@ bool nav_survey_active;
 
 int32_t nav_roll, nav_pitch;
 int32_t nav_heading;
+int32_t nav_cmd_roll, nav_cmd_pitch, nav_cmd_yaw;
 float nav_radius;
 float nav_climb_vspeed, nav_descend_vspeed;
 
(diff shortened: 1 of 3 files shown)


podhrmic:
@EwoudSmeur I must have missed something, but what is NAV MANUAL mode?

EwoudSmeur:
It is a nav mode you can set from the flight plan, like the nav mode attitude. It allows the user to set the stabilization commands from the flight plan. 

It is useful for helicopters, as they need a few seconds to spin up, in which you may not want a controller to run. So you can set the stabilization commands to zero while the rotor starts spinning.

podhrmic:
ah cool, in that case it looks good to me.

flixr:
Basically looks good to me, but some doxygen comments on what the vars are used for would be nice.
And/or rename them to something like `nav_manual_roll` or so to make this more clear?

EwoudSmeur:
Something like this? 
I think the naming is clear this way, it is a command set from nav. The naming of HORIZONTAL_MODE_MANUAL is more confusing in my eyes.

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -482,16 +486,20 @@ void nav_home(void)
...
- * @param[in] yaw The angle in radians (float)
+ * @param[in] roll command in pprz scale (float)
+ * @param[in] pitch command in pprz scale (float)
+ * @param[in] yaw command in pprz scale (float)
+ *
+ * This function allows to directly set commands from the flight plan,
+ * if in nav_manual mode.
+ * This is for instance useful for helicopters during the spinup
  */
 void nav_set_manual(float roll, float pitch, float yaw)
```
flixr: why have roll, pitch, yaw as floats here when the commands are always in int32_t anyway?

PR #1842, opened 2016-08-16
[gcs] prevent speech spam with 'approaching'

### LINKED ISSUE #1833: GCS spams unfinished sentences
After the successful resolution of https://github.com/paparazzi/paparazzi/issues/1046, I am back with a new issue. The GCS still does not finish some sentences because it starts a new one and can in some cases start to spam messages. This can happen for instance when you turn off a drone, if I remember correctly it can then start to alternate between "all links lost" and "approaching".

Now, the approaching message is actually the one I feel is the least interesting message of them all, though you hear it most often.

gautierhattenberger:
only triggered when close enough but not too much, and flying a long segment
fix #1833

### DIFF
--- a/sw/ground_segment/cockpit/live.ml
+++ b/sw/ground_segment/cockpit/live.ml
@@ -38,6 +38,9 @@ let (//) = Filename.concat
 let gcs_id = "GCS"
 
 let approaching_alert_time = 3.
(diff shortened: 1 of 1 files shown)


flixr:
LGTM
There might still be some improvement needed to prevent repeated messages, but this should already reduce it a lot.

PR #1844, opened 2016-08-18
[ubx_i2c] silence warnings by casting i2c buffer

masierra:
Was getting a warning about using memcpy with volatile buffer since recent changes. Saw this fix used in other i2c modules. Tested and working on ublox i2c with no warnings (the warning about casting volatile is silenced in the makefiles).


flixr:
Thx.

PR #1846, opened 2016-08-18
Sys mon fix

podhrmic:
Only one module (sys_mon) used - it will automatically detect if it is rtos, sim or bare-metal and will send the proper message (RTOS_MON or SYS_MON). Easier to use because the user don't have to change between sys_mon and rtos_mon anymore. Plus it works for SIM arch now.


flixr:
Only drawback that I can see is that now you can't use the "old" sys_mon module on OS targets anymore. I think that it can still be useful to get an idea about the cycle times and jitter.
It would make sense to implement the rtos_mon (which is actually not restricted to RTOSs) for Linux targets as well, at least there the cycle times and jitter were sometimes useful.

podhrmic:
I was thinking about actually merging those two together - so then you would have one struct that combines `Sys_mon` and `Rtos_mon` variables. But I didn't do it becase Rtos_mon takes much more memory and wouldnt be used on bare-metal, which might be a waste for old LPC boards.

Linux expansion is welcome, but probably people using ArDrones etc are better suited for the task.

podhrmic:
Btw I tested the old sys_mon on Lisa MX with ChibiOS and it just shows 100% CPU usage and the other indicators are constant, while the correct CPU usage from RTOS_MON is ~10%. So it is probably safe to say that people wont miss the old sys_mon for RTOS targets...

flixr:
It worked fine on Linux targets... the jitter is sometimes really useful information...

flixr:
by "worked fine" I actually mean that the timing info was working and useful, the cpu_load estimate not...
maybe we should try to add this for the rtos_mon as well?

But since rtos_mon is not implemented currently for linux, it is probably no reason to not merge this...

podhrmic:
this will still run the "regular" sys_mon for linux targets, only Chibios (and nps) uses rtos_mon - so you can use it as before

podhrmic:
see https://github.com/paparazzi/paparazzi/pull/1846/commits/451940fd2f69ac8501659bb6d934b157dab42329#diff-a91ff52aa80c2551abdcfae76b3c93daR52

sw/airborne/modules/core/sys_mon_rtos.h
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright (C) 2016 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ * Copyright (C) 2010  Gautier Hattenberger
```
gautierhattenberger: previous license header looked more correct.

PR #1848, opened 2016-08-19
Disable USB VBUS checking.

esden:
It should not affect other platforms that do have the AF9 pin connected,
this should just end up ignoring the state of the AF9 pin.


flixr:
Merging as it seems to fix telemetry transparent_usb on elle0 and still works on LisaMX

PR #1850, opened 2016-08-25
Seed NES, SNES, and N64 game pad configurations

hooperfly:
Game pad/joystick configurations for use in NAV/GUIDED modes:
  - NES 
  - SNES
  - N64


hooperfly:
Of note, tested on both Mac OS X and Linux. Found that the same NES controller on Mac OS X needs a trim value of +1.0 on both X/Y axis; on Linux, no trim value was required for either X or Y axis.

PR #1855, opened 2016-09-13
Optical flow filter and derotation fix

knmcguire:
Hi all,

Here is some improvements to the optical flow code.
- Kalman filter: I made a generic kalman filter just for the use of vision. The problem of using the HHF filter is that it also effect the reaction of the controls (for some reason there is too much dependency on the process model but I can't change the noise factors). I think that it is wise to filter vision results separate from the state filters. Right now it is implemented for optical flow, but later it can be used for tracking as well. I hope you will agree
- derotation correction factor: there seems to be an error between the calculated derotation flow by the gyro and the calculated flow. Now it has been fitted together and for the bebop a correction term has to be added (credits to @mmoiozo for figuring this out).

This has been test flown with two bebops.

cheers, 

Kimberly


conf/airframes/TUDELFT/tudelft_ardrone2_opticflow_stereo.xml
```diff
@@ -23,8 +23,8 @@
 
   <modules main_freq="512">
     <module name="bat_voltage_ardrone2"/>
-    <module name="cv_opticflow"/>
-    <module name="opticflow_hover"/>
+    <!--module name="cv_opticflow"/>
+    <module name="opticflow_hover"/-->
```
kirkscheper: this doesn't look like this has much to do with the optic flow filter of derotation.

conf/airframes/TUDELFT/tudelft_bebop_opticflow.xml
```diff
@@ -19,7 +19,7 @@
     <module name="gps" type="datalink"/> <!-- Optitrack -->
     <module name="stabilization" type="indi"/>
     <module name="ahrs" type="int_cmpl_quat">
-      <configure name="USE_MAGNETOMETER" value="TRUE"/>
+      <configure name="USE_MAGNETOMETER" value="FALSE"/>
```
kirkscheper: why turn the magneto off? How do you know your heading?

sw/airborne/modules/computer_vision/lib/filters/kalman_filter_vision.h
```diff
@@ -0,0 +1,15 @@
...
+ *  Created on: Sep 12, 2016
+ *      Author: knmcguire
+ */
+
+#ifndef KALMAN_FILTER_VISION_H_
+#define KALMAN_FILTER_VISION_H_
+
+void kalman_filter_linear_2D_float(float *model, float *measurements, float *covariance, float *state
+                                   , float *process_noise, float *measurement_noise);
+
```
kirkscheper: Formatting looks strange here

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.h
```diff
@@ -75,6 +79,13 @@ void calc_fast9_lukas_kanade(struct opticflow_t *opticflow, struct opticflow_sta
                              struct opticflow_result_t *result);
 void calc_edgeflow_tot(struct opticflow_t *opticflow, struct opticflow_state_t *state, struct image_t *img,
                        struct opticflow_result_t *result);
+
+void kalman_filter_opticflow_velocity(float *velocity_x ,  float *velocity_y, float *acceleration_measurement,
+                                      float fps,
+                                      float *measurement_noise, bool reinitialize_kalman);
```
kirkscheper: also some strange formatting

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
```diff
@@ -150,8 +155,28 @@ PRINT_CONFIG_VAR(OPTICFLOW_METHOD)
...
+PRINT_CONFIG_VAR(OPTICFLOW_DEROTATION)
+
 #ifndef OPTICFLOW_MEDIAN_FILTER
-#define OPTICFLOW_MEDIAN_FILTER TRUE
+#define OPTICFLOW_MEDIAN_FILTER FALSE
+#endif
+PRINT_CONFIG_VAR(OPTICFLOW_MEDIAN_FILTER)
+
+#ifndef OPTICFLOW_KALMAN_FILTER
+#define OPTICFLOW_KALMAN_FILTER TRUE
```
kirkscheper: can you have the median filter and the Kalman filter running at the same time?

conf/airframes/TUDELFT/tudelft_ardrone2_opticflow_stereo.xml
knmcguire: true, but it was just that I noticed this airframe file needed some updating, so that's why this quick unrelated change.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
knmcguire: Not by standard, that is why the ifndef statement set to false at the median filter and true with the kalman filter

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -169,8 +169,7 @@ struct image_t *opticflow_module_calc(struct image_t *img)
   * ARdrone and Bebop, however this can be different for other quadcopters
```
flixr: I guess it would make sense to remove the comment here as well...

flixr:
Haven't looked at the Kalman code so far, but please update the doc section in the module xml file according to the new features you added...

knmcguire:
@flixr  The kalman code does not have it's own module.xml file. Should I add doxygen explanation to the kalman_filter_vision.c instead? I could add it to the opticalflow module xml but it is meant as a general filter for vision purposes

knmcguire:
I just pushed commits with the suggested changes. Thanks for the advice!

knmcguire:
@flixr  Here I can't see about which line you are talking about for some reason.

- Could you specify which define in cv_opticflow.xml? OPTICFLOW_CAMERA by any chance?

- Also, in opticflow_module.c, I think you are talking about the memcpy in line 208 and 222. Should I also replace the memcpy with a direct assignment in lines 169 and 178

flixr:
Hm.. weird that github didn't pick up the line of the comment anymore...
I meant the DEROTATION define, but if there are other obsolete/unused defines in the doc please remove them as well.

Yep, if you just copy structs use assignment every time.

knmcguire:
I changed the memcpy to direct assignments. However, the DEROTATION define (also the CAMERA) is still used in the code (opticflow_calculator.c), so it should be kept in the module.xml

flixr:
Instead of renaming `OPTICFLOW_AGL_ID` to `OPTICFLOW_RECEIVE_ABI_ID` and using the same sender ID for AGL, IMU and BODY_TO_IMU, each message callback should get it's own ID.
This is used to select from which sender you want to receive the message and it's not the same sender that publishes all these messages... e.g. AGL will come from a sonar sensor, IMU from an IMU and BODY_TO_IMU from a global config or IMU as well.

flixr:
thx, looks good to me except that the doc section in cv_opticflow.xml needs to be updated again to reflect the latest ABI ID changes...

PR #1858, opened 2016-09-20
[fix] remove unused i2c dev from aspirin_2 and lisa_s conf

gautierhattenberger:
this was produce a strange side effect on bluegiga module (#1835)


gautierhattenberger:
@kirkscheper thanks for testing

PR #1859, opened 2016-09-20
convert AHRS subsystems to modules

flixr:
So far there should be no functional changes, except that the respective settings are now loaded with the module.

TODO:
- [ ] get rid of explicit AHRS calls in main


flixr:
@gautierhattenberger merge this as is and do the "internal" refactoring to get rid of the calls from main later?
Then we already have the most important user facing benefits: real modules with documented defines and autoload of corresponding settings...

gautierhattenberger:
I guess your right, it can be merge like this for the next release.

PR #1861, opened 2016-09-21
Compensate baro for reference frame conversion

EwoudSmeur:
The barometer measures the altitude, but it is used in the NED based vertical filter. 
Will soon test if it gives the correct altitude after driving ~10km

### DIFF
--- a/sw/airborne/subsystems/ins/ins_int.c
+++ b/sw/airborne/subsystems/ins/ins_int.c
@@ -329,7 +329,19 @@ static void baro_cb(uint8_t __attribute__((unused)) sender_id, float pressure)
       vff_realign(0.);
       ins_update_from_vff();
     } else {
-      ins_int.baro_z = -pprz_isa_height_of_pressure(pressure, ins_int.qfe);
+      float baro_up = pprz_isa_height_of_pressure(pressure, ins_int.qfe);
+
+      // Calculate the distance to the origin
+      struct EnuCoor_f *enu = stateGetPositionEnu_f();
+      double dist2_to_origin = enu->x*enu->x + enu->y*enu->y;
+
+      // correction for the earth's curvature
+      const double earth_radius = 6378137.0;
+      float height_correction = (float) (sqrt(earth_radius*earth_radius + dist2_to_origin) - earth_radius);
+
+      // The VFF will update in the NED frame
+      ins_int.baro_z = baro_up - height_correction;
+
 #if USE_VFF_EXTENDED
       vff_update_baro(ins_int.baro_z);
 #else


EwoudSmeur:
With this commit I have a ~factor half altitude scaling, when flying near the origin... Wonder where that is going wrong

flixr:
I don't understand what your issue is here. The function gives you the
height over the given pressure altitude. QFE is the pressure at field
elevation, which is initialized at startup or reference reset. So you get
the height over the origin.
There should be no need to transform that to a global altitude.

Am 21.09.2016 10:30 vorm. schrieb "Ewoud Smeur" <notifications@github.com>:

EwoudSmeur:
The issue occurs when you fly far away. The barometer will give an altitude, basically in the LLA frame. So if you want to put this in the NED frame you need to apply the right conversion.

Example: you have a waypoint 10 km further. That waypoint will have an UP coordinate (ENU) 8 m lower than the default waypoints at the origin, to compensate for the curvature of the earth. If you don't apply this compensation to the baro reading, you will end up 8 meters lower.

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -331,7 +331,20 @@ static void baro_cb(uint8_t __attribute__((unused)) sender_id, float pressure)
       vff_realign(0.);
       ins_update_from_vff();
     } else {
-      ins_int.baro_z = -pprz_isa_height_of_pressure(pressure, ins_int.qfe);
+      float baro_up = pprz_isa_height_of_pressure(pressure, ins_int.qfe);
+
+      // Create LLA coordinate with current position and baro altitude
+      struct LlaCoor_i lla;
+      LLA_COPY(lla,*stateGetPositionLla_i());
+      lla.alt = 1000*baro_up + state.ned_origin_i.lla.alt;
```
flixr: would be better to use `ins_int.ltp_def.lla.alt` here...

flixr:
I meant 1cm error at 10km...

dewagter:
There is a remaining problem with this code: 

 - (as seen in our logfiles when flying 20 km away) when flying to a distant waypoint, the vehicle will instantly descend to the corrected altitude (to arrive at the correct height) instead of following the earth curvature.

flixr:
What do you mean? This will do the correction with every baro update...

EwoudSmeur:
Christophe means that because the waypoints are in the NED frame, if the drone has to fly to a waypoint far away that is at the same altitude, it will immediately descend to the right NED z, instead of following the same altitude above the earth surface.

dewagter:
Only good solution is to save the reference in all altitudes (wp/current): #1700

flixr:
I would be fine to merge this if a fast spherical model is used and the LTP def of the INS is used.
It should at least be better than without it...

EwoudSmeur:
I simplified the calculation, but indeed if you have waypoints far apart you may see a sawtooth in the altitude graph...

EwoudSmeur:
ok like this?

flixr:
thx!

PR #1866, opened 2016-09-30
[iridium] Iridium dial tool

dewagter:
communicate with Iridium Core 9523 Satellite Transceiver Module


sw/tools/iridium/iridium_link.py
```diff
@@ -0,0 +1,220 @@
...
+from socket import SOL_SOCKET, SO_BROADCAST
+import serial
+import pygtk
+import gtk
+import os
+pygtk.require('2.0')
+
+
+
+DEFAULT_DEVICE = "/dev/paparazzi/iridium"
```
gautierhattenberger: do you have a udev rule for this or is it only an example ?

sw/tools/iridium/iridium_link.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python
+from twisted.internet import gtk2reactor
```
gautierhattenberger: license header is missing

sw/tools/iridium/iridium_link.py
```diff
@@ -0,0 +1,220 @@
...
+DEFAULT_DEVICE = "/dev/paparazzi/iridium"
+DEFAULT_BAUD = "19200"
+DEFAULT_NUMBER = "00881111111111"
+UDP_IN = 4243
+UDP_OUT = 4242
+
+# IridiumCtrl interface
+class IridiumCtrl:
+
+  def open(self, widget, data=None):
```
gautierhattenberger: if I'm not wrong, this is not the recommended indentation for python code

flixr: please adhere to pep8 (4 spaces indentation, no tabs)

flixr:
Can you please also add a Readme.md that describes the functionality and lists the dependencies (twisted, iridium_protocol, ..)

dewagter:
All iridium stuff was placed in a separate repository: https://github.com/tudelft/iridium/

Is it an option to submodule it and only download when needed?

flixr:
What do you mean by "download when needed"? As soon as someone does a `git submodule update` it will download the submodule, regardless of what rules we add in our makefiles...
Also I don't think that the whole repo is needed as a submodule (including hw files, etc).

It seems that only `iridium_protocol.py` is needed, why not add that directly here?
Or do you expect lots of changes there and it would be annoying to keep in sync?

flixr:
sorry, missed that `iridium_protocol.py` is already added here...

flixr:
So, fixing python indentation, using PAPARAZZI_SRC instead of HOME for the icon and adding a Readme.md and requirements.txt would IMHO make this complete/mergeable...

sw/tools/iridium/iridium_link.py
```diff
@@ -0,0 +1,220 @@
...
+    self.cb_recv_data = cb_recv_data
+
+  def startProtocol(self):
+    self.transport.socket.setsockopt(SOL_SOCKET, SO_BROADCAST, True)
+
+  def sendData(self, data):
+    self.transport.write(data, ('127.255.255.255', self.out_port))
+
+  def datagramReceived(self, datagram, addr):
+    if self.cb_recv_data != None:
```
flixr: Please not that checking for None should not be done with comparison operators, but with `is` and `is not`.
See https://www.python.org/dev/peps/pep-0008/#programming-recommendations
http://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or

dewagter:
IMHO all comments were now processed

flixr:
Can you please also replace all the other cases of `foo != None` with `foo is not None`?

dewagter:
Sure!

-Christophe

gautierhattenberger:
can you merge or rebase with master and fix conflict, beside that it's all good

dewagter:
Rebased / Fixed.

PR #1868, opened 2016-09-30
[atc] aeronautical altitude information

dewagter:


gautierhattenberger:
Please add license header, and some indentations should be fixed.

Also, I'm wondering if the name `atc` is not a bit misleading.

PR #1870, opened 2016-09-30
[svinfo] Icon

dewagter:


sw/ground_segment/python/svinfo/svinfoviewer.py
```diff
@@ -122,6 +122,9 @@ def __init__(self):
         self.Bind(wx.EVT_SIZE, self.OnSize)
         self.Bind(wx.EVT_CLOSE, self.OnClose)
 
+        ico = wx.Icon(os.getenv("PAPARAZZI_HOME") + "/sw/ground_segment/python/svinfo/gps.ico", wx.BITMAP_TYPE_ICO)
```
gautierhattenberger: Could you make this more robust if paparazzi env are not set ? Something like https://github.com/paparazzi/paparazzi/blob/master/sw/ground_segment/python/settings_app/settingsapp.py#L10
Also, it should be PAPARZZI_SRC, same for most of the other pull requests.

flixr: seems that is fixed now..

flixr:
Ugh... accidentally clicked on "update branch" and it merged master into this... shouldn't matter though if @gautierhattenberger accepts the changes and squashes it...

PR #1871, opened 2016-09-30
[sideslip] measure sideslip from lateral force

dewagter:


conf/modules/sideslip.xml
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
```
podhrmic: Provide some documentation here...

sw/airborne/modules/sideslip/sideslip.c
```diff
@@ -0,0 +1,55 @@
...
+#include "subsystems/imu.h"
+
+float low_pass_filtered_sideslip_value = 0.0f;
+
+#if PERIODIC_TELEMETRY
+#include "subsystems/datalink/telemetry.h"
+
+static void sideslip_send_telem(struct transport_tx *trans, struct link_device *dev)
+{
+  pprz_msg_send_SIDESLIP(trans, dev, AC_ID, &low_pass_filtered_sideslip_value);
```
podhrmic: well you are effectively just sending low-passed accel-y from the IMU - maybe it should be named like `lateral_acc_estimator` since you are not really measuring nor estimating the sideslip angle...

PR #1872, opened 2016-09-30
[emergency-kill] Kill is a setting. To be less xml dependent it shoul…

dewagter:
…d be a message so that several boards can be killed together

### DIFF
--- a/sw/ground_segment/cockpit/live.ml
+++ b/sw/ground_segment/cockpit/live.ml
@@ -253,6 +253,10 @@ let dl_setting = fun ac_id idx value ->
   let vs = ["ac_id", PprzLink.String ac_id; "index", PprzLink.Int idx;"value", PprzLink.Float value] in
   Ground_Pprz.message_send "dl" "DL_SETTING" vs
 
+let dl_emergency_cmd = fun ac_id cmd ->
+  let vs = ["ac_id", PprzLink.String ac_id; "cmd", PprzLink.Int cmd] in
+  Ground_Pprz.message_send "dl" "DL_EMERGENCY_CMD" vs
+
 let get_dl_setting = fun ac_id idx ->
   let vs = ["ac_id", PprzLink.String ac_id; "index", PprzLink.Int idx] in
   Ground_Pprz.message_send "dl" "GET_DL_SETTING" vs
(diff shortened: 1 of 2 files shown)


sw/ground_segment/cockpit/live.ml
```diff
@@ -457,6 +461,10 @@ let create_ac = fun ?(confirm_kill=true) alert (geomap:G.widget) (acs_notebook:G
   let via_http = Str.string_match (Str.regexp "http") af_url 0 in
   let af_xml = ExtXml.parse_file ~noprovedtd:via_http af_file in
 
+  (** Get firmware name *)
+  let firmware = ExtXml.child af_xml "firmware" in
```
gautierhattenberger: you can have several firmwares in a xml airframe file

sw/ground_segment/cockpit/live.ml
```diff
@@ -755,11 +763,15 @@ let create_ac = fun ?(confirm_kill=true) alert (geomap:G.widget) (acs_notebook:G
           let connect = fun ?(warning=true) setting_name strip_connect ->
             try
               let id = settings_tab#assoc setting_name in
-              strip_connect (fun x -> dl_setting_callback id x)
+              if setting_name = "kill_throttle" then
```
gautierhattenberger: If kill connect is made a specific command, I would also make a dedicated `connect` function, there is no real gain here.

sw/ground_segment/cockpit/live.ml
```diff
@@ -253,6 +253,10 @@ let dl_setting = fun ac_id idx value ->
   let vs = ["ac_id", PprzLink.String ac_id; "index", PprzLink.Int idx;"value", PprzLink.Float value] in
   Ground_Pprz.message_send "dl" "DL_SETTING" vs
 
+let dl_emergency_cmd = fun ac_id cmd ->
+  let vs = ["ac_id", PprzLink.String ac_id; "cmd", PprzLink.Int cmd] in
+  Ground_Pprz.message_send "dl" "DL_EMERGENCY_CMD" vs
```
gautierhattenberger: This message doesn't exist yet in pprzlink

dewagter: https://github.com/paparazzi/pprzlink/pull/34

dewagter:
The kill button in the GCS is now sending a setting. For each board/ap/fbw/ac this setting has a different number, which is potentially dangerous. The MD5 check solves a lot of it, but in FBW/AP boards this can still go wrong when the FBW and AP both have a datalink.

In the long run I think it is better to remove the kill setting completely and replace it with a dedicate well documented datalink message. This would also make paparazzi more compatible with other ground stations (e.g. MavLink).

sw/ground_segment/cockpit/live.ml
dewagter: the message should be attached to the kill button of the GCS but we do not want to break kill settings

sw/ground_segment/cockpit/live.ml
dewagter: @fvantienen why can it not be sent to fixedwing?

sw/ground_segment/cockpit/live.ml
gautierhattenberger: still, I think it looks cleaner with a dedicated kill function that also send a setting if it exists

dewagter: @fvantienen could you help with this?

sw/ground_segment/cockpit/live.ml
dewagter: Maybe this part should have been in another pull request then. Is a NO_WIND option in link also an option? Is no rotorcraft ever using wind?

flixr:
As @gautierhattenberger already pointed out, parsing the firmware that way is not robust at all.
Disabling wind info should be in a separate PR anyway...
Also the other changes to parse message are not related to the kill setting/message...

About the kill message, I think it is a good idea. But with this change committed kill won't work anymore until people update the firmware on their aircraft, since the setting is not applied anymore.
That could be potentially dangerous, so it might be a good idea to update the setting and send the message or find some other way to make this backwards compatible.

flixr:
Also the conflicts in pprzlink need to be fixed...

dewagter:
- [x]  setting is still being applied
 - [x] firmware -> wind problem removed

flixr:
LGTM
@gautierhattenberger ?

PR #1875, opened 2016-09-30
[MCU_PWR] bugfix

dewagter:


sw/airborne/mcu.c
```diff
@@ -89,11 +89,19 @@ void mcu_init(void)
...
+  gpio_setup_input(BTN_ON, BTN_ON_PIN);
+  if(gpio_get(BTN_ON, BTN_ON_PIN))
 #endif
   {
-  MCU_PWR_ON(MCU_PWR, MCU_PWR_PIN);
+    MCU_PWR_ON(MCU_PWR, MCU_PWR_PIN);
   }
+#if defined BTN_ON
+  else {
+    while(true) {
```
gautierhattenberger: Can you comment and explain the reason of this `while true` ?

sw/airborne/mcu.c
```diff
@@ -89,11 +89,19 @@ void mcu_init(void)
...
 #if defined BTN_ON
-  if(!gpio_get(BTN_ON, BTN_ON_PIN))
+  gpio_setup_input(BTN_ON, BTN_ON_PIN);
+  if(gpio_get(BTN_ON, BTN_ON_PIN))
 #endif
   {
-  MCU_PWR_ON(MCU_PWR, MCU_PWR_PIN);
+    MCU_PWR_ON(MCU_PWR, MCU_PWR_PIN);
   }
+#if defined BTN_ON
```
gautierhattenberger: It would be more readable I think like this:
```
#ifdef BTN_ON
 ... all button related code here ...
#else
  MCU_PWR_ON(MCU_PWR, MCU_PWR_PIN);
#endif
```

sw/airborne/mcu.c
fvantienen: Else other parts of the software can still turn it on later. This is just to make sure that the on button needs to be pressed in order to turn it on.
Because of our schematic we were able to turn the autopilot on by moving the servo's previously.

PR #1876, opened 2016-09-30
[fbw_gps] GPS in FBW/FTD (e.g. geofencing)

dewagter:


sw/airborne/subsystems/intermcu/intermcu_ap.c
```diff
@@ -144,6 +164,56 @@ static inline void intermcu_parse_msg(void (*rc_frame_handler)(void))
       break;
     }
 
+#if TELEMETRY_INTERMCU
```
flixr: Seems like this should be part of another PR...

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -162,12 +162,16 @@
 #define GPS_SIM_ID 11
 #endif
 
+#ifndef GPS_IMCU_ID
```
flixr: Would be better to add a new ID 14 instead of changing the IDs of existing ones...

sw/airborne/subsystems/intermcu/intermcu_ap.c
```diff
@@ -144,6 +164,56 @@ static inline void intermcu_parse_msg(void (*rc_frame_handler)(void))
...
+      uint8_t size = DL_IMCU_DATALINK_msg_length(imcu_msg_buf);
+      uint8_t *msg = DL_IMCU_DATALINK_msg(imcu_msg_buf);
+      telemetry_intermcu_on_msg(0, msg, size);
+      break;
+    }
+
+#endif
+
+#if IMCU_GPS
+    case DL_IMCU_REMOTE_GPS: {
```
flixr: Is still missing in pprzlink...

sw/airborne/subsystems/intermcu/intermcu_fbw.c
```diff
@@ -66,10 +66,21 @@ static void intermcu_parse_msg(void (*commands_frame_handler)(void));
...
+static abi_event gps_ev;
+static void gps_cb(uint8_t sender_id, uint32_t stamp, struct GpsState *gps_s);
+#endif
+
 void intermcu_init(void)
 {
   pprz_transport_init(&intermcu.transport);
 
+#if USE_GPS
+  AbiBindMsgGPS(ABI_BROADCAST, &gps_ev, gps_cb);
```
flixr: The ABI ID used here should probably be configurable?

podhrmic:
Can you describe how this fits together with existing geofencing features we have in the flight plan? See #1802

dewagter:
@podhrmic Currently FBW does not run any flightplan, so GeoFence must be done in a module.

dewagter:
- [x] GPS driver in FBW
 - [ ] use GPS in FBW
 - [ ] pprzlink update
 - [ ] forward FBW GPS to AP

dewagter:
PPRZLINK update pending: https://github.com/paparazzi/pprzlink/pull/34

dewagter:
Ready

sw/airborne/subsystems/intermcu/intermcu_fbw.c
flixr: Think this should still be changed from ABI_BROADCAST to something like `IMCU_GPS_ID` which should default to `GPS_MULTI_ID`
Then you could also have multiple GPS on the fbw controller...

flixr: And using ABI_BROADCAST here, gps_cb gets actually called twice per new GPS message: once from the actual GPS implementation and once from gps.c with GPS_MULTI_ID

flixr:
Also what about the other gps modules, right now only ublox can be used on fbw...

gautierhattenberger:
When #1835 is merged, we'll have the possibility to group modules functions into groups. Maybe it will help to call only a subpart of the modules in FBW, so all GPS modules would be available in this case.

dewagter:
- [x] ICMU_GPS_ID
 
Are you sure you still want me to convert ALL gps modules to run on FBW? Or await #1835?

flixr:
I don't see how #1835 is related to that...
In any case it would make much more sense to change the target for gps modules from only `ap` to `!nps|sim`, then they are usable for any target except simulation...

gautierhattenberger:
The PR itself is not related, but I was think that the splitting into tasks could help here. After looking again at the code, you're right, it doesn't.

flixr:
To be done here:
- resolve conflicts
- update of module xml of other GPS modules

dewagter:
- [x] conflicts resolved
 - [x] other GPS modules

PR #1877, opened 2016-09-30
[gps_bebop] only enable NMEA on bebop since some modules only provide…

dewagter:
… nmea after firmware updates

### DIFF
--- a/sw/airborne/modules/gps/gps_ubx_ucenter.c
+++ b/sw/airborne/modules/gps/gps_ubx_ucenter.c
@@ -31,9 +31,9 @@
 #include "subsystems/gps/gps_ubx.h"
 #include "ubx_protocol.h"
 #include "subsystems/datalink/downlink.h"
-#include <stdio.h>
 
 #if PRINT_DEBUG_GPS_UBX_UCENTER
+#include <stdio.h>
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/gps/gps_ubx_ucenter.c
```diff
@@ -424,14 +424,25 @@ static inline void gps_ubx_ucenter_config_port(void)
     case GPS_PORT_UART2:
       UbxSend_CFG_PRT(gps_ubx_ucenter.dev,
           gps_ubx_ucenter.port_id, 0x0, 0x0,
+#ifdef BOARD_BEBOP
```
gautierhattenberger: Board specific defines should be avoided here. There should be a default mask with the possibility to redefine it if needed.

dewagter:
I do not know what is worse: Bebop defines in the Gps or Gps defines in the Bebop. The only good solution would be to convince Parrot to use UBX instead of NMEA, and disable NMEA all the time.

Could a more generic and reusable ```DO_NOT_DISABLE_NMEA``` define be better in the mean time?

dewagter:
moved define to board file, removed ifdef and added module documentation

PR #1878, opened 2016-09-30
[rpm-control] throttle curves with rpm control

dewagter:


flixr:
duplicate of #1832 ?

podhrmic:
in general it would be nice to have more comments in the `throttle_curve` module, including some examples...

dewagter:
- [ ] All airframes files that use this need to be updated
 - [ ] No airframe in conf_tests.xml
 - [ ] Add descriptions to the modules on what exactly they do and how they should be used
 - [ ] The added rpm_control.xml file seems superfluous...
 - [ ] Code style
 - [ ] Also why change the signature of throttle_curve_run? Now you don't pass autopilot_motors_on as parameter anymore but instead take the global var directly... but add ap_mode as parameter instead of using the global var in that case

flixr:
any updates here?

podhrmic:
@dewagter any updates here?

OpenUAS:
Work on this re-started by @fvantienen

PR #1879, opened 2016-09-30
[dual_datalink] full duplex dual datalink

dewagter:


gautierhattenberger:
good idea, but still some bug to fix

flixr:
@gautierhattenberger does it look good to you now?

PR #1881, opened 2016-09-30
[ahrs] main speed component is in wing direction

dewagter:


sw/airborne/subsystems/ahrs/ahrs_int_cmpl_quat.c
```diff
@@ -270,7 +270,11 @@ void ahrs_icq_update_accel(struct Int32Vect3 *accel, float dt)
 #define ACC_FROM_CROSS_FRAC INT32_RATE_FRAC + INT32_SPEED_FRAC - INT32_ACCEL_FRAC - COMPUTATION_FRAC
 
     const struct Int32Vect3 vel_tangential_body =
+#ifdef VERTICAL_WING_HYBRID
+      {0, 0, -(ahrs_icq.ltp_vel_norm >> COMPUTATION_FRAC)};
```
gautierhattenberger: can you add some comments about the plane orientation in this case ? is there a way to make this more generic ?

dewagter: Yes, it would be very good to make this more generic.

 - one option is to use the 3D gps speed (this would change the behaviour of the filter)
 - another option is to define the FIXEDWING_FRONT_AXIS
 - another option is to define the IMU2FIXEDWING_ROTATION

Any preference

flixr: I made the change to use vel on all axes a long time ago for @EwoudSmeur , not sure if it has been tested though:
https://github.com/flixr/paparazzi/commit/1c05c0a880bc33612707703fe8b22c5b10349cfc

PR #1882, opened 2016-09-30
[forward_rotorcraft] last step to fly fixedwings with rotorcraft code…

dewagter:
…: forward flightplan logic!

### DIFF
--- a/sw/tools/generators/gen_flight_plan.ml
+++ b/sw/tools/generators/gen_flight_plan.ml
@@ -427,6 +427,7 @@ let rec print_stage = fun index_of_waypoints x ->
         in
         let at = try Some (ExtXml.attrib x "approaching_time") with _ -> None in
         let et = try Some (ExtXml.attrib x "exceeding_time") with _ -> None in
+        let fwd = try Some (ExtXml.attrib x "forward") with _ -> None in
         let at = match at, et with
           | Some a, None -> a
           | None, Some e -> "-"^e
@@ -446,6 +447,10 @@ let rec print_stage = fun index_of_waypoints x ->
         let vmode = output_vmode x wp last_wp in
         if vmode = "glide" && hmode <> "route" then
           failwith "glide vmode requires route hmode";
+        if fwd = Some("true") then begin
+          lprintf "FlightModeFixedwing();\n"
+        end else
+          lprintf "FlightModeRotorcraft();\n";
         left (); lprintf "}\n";
         begin
           try
@@ -462,10 +467,12 @@ let rec print_stage = fun index_of_waypoints x ->
             let wp = get_index_waypoint (ExtXml.attrib x "wp") index_of_waypoints in
             ignore (output_hmode x wp "");
             ignore (output_vmode x wp "");
+            lprintf "FlightModeRotorcraft();\n";
           with
               Xml2h.Error _ ->
                 lprintf "NavGotoXY(last_x, last_y);\n";
-                ignore(output_vmode x "" "")
+                ignore(output_vmode x "" "");
+                lprintf "FlightModeRotorcraft();\n";
         end;
         begin
           try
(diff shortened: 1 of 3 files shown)


sw/tools/generators/gen_flight_plan.ml
```diff
@@ -446,6 +447,11 @@ let rec print_stage = fun index_of_waypoints x ->
         let vmode = output_vmode x wp last_wp in
         if vmode = "glide" && hmode <> "route" then
           failwith "glide vmode requires route hmode";
+        if fwd = Some("true") then begin
+          lprintf "outback_hybrid_mode = HB_FORWARD;\n";
```
gautierhattenberger: isn't this a bit too specific ?

sw/tools/generators/gen_airframe.ml
```diff
@@ -269,14 +269,28 @@ let parse_command = fun command no ->
   let failsafe_value = int_of_string (ExtXml.attrib command "failsafe_value") in
   { failsafe_value = failsafe_value; foo = 0}
 
-let parse_heli_curves = fun heli_surves ->
-  let a = fun s -> ExtXml.attrib heli_surves s in
-  match Xml.tag heli_surves with
+let parse_heli_curves = fun heli_curves ->
```
gautierhattenberger: could have been a different PR...

dewagter:
The control part is work in progress. Can you already comment on the idea to have FORWARD tags in the flightplan to allow fixedwings to fly with rotorcraft code.

gautierhattenberger:
I actually don't get what the forward attribute is suppose to do. Does it say that the rotorcraft nav and guidance should be used instead the fixed-wing code depending if it is set to true or false ?

gautierhattenberger:
In general, I don't really like the idea to have specific firmware calls in the generated flight plan code (just in case we have rover, blimp or spacecraft code someday...).
The lower level should take care of the correct mode or navigation style, but I agree some information might be missing. Especially I don't think the guidance can tell if a `go` or a `stay` is requested. I think that would be a useful information for hybrid vehicles.
What about adding some sort of "user_data" attribute allowing to pass something from the XML flight plan to the lower level guidance without the need for the generator to know what it is ?

dewagter:
Forward/Hover in http://www.delftacopter.com/ was activating a different outerloop controller. ```go control="guidance_v_rotorcraft"```. Even for rotorcraft this outerloop can be interesting: to fly fast forward (or when there is a lot of wind) we use constant throttle with pitch-for-altitude control.  And when considering fixedwings, ```go userdata="22"``` would often be "speed".

dewagter:
@gautierhattenberger in the case of hybrids, the flightplan must be able to select if ```go``` is done hovering (slow) or as fast as possible. How would you suggest to do this?

```<go user_data="v=5"/>``` and ```<go user_data="vmax"/>```
```<go speed="5"/>``` and ```<go speed="max"/>```
```<go/>``` and ```<go forward="true"/>```

```<call fun="setup_hover_mode()"/><go>``` and ```<call fun="setup_forward_mode()"/><go>```

**Comment by gautierhattenberger** [comment]
What about `... user_cmd="some_c_code"/>` or `... user_call=""`. So you can choose:
```
<go wp="P1" user_cmd="forward_mode()"/>
<circle wp="P2" radius="10" user_cmd="nav_speed=2"/>
<oval ... user_cmd="hover_mode(); camera_stuff()"/>
```
Basically allowing the equivalent of `call_once` after (before ?) each stage calls.

dewagter:
Sound good. For us, the ```user_cmd``` should be called every step while doing the circle/go/... etc. In that respect it is different from ```call_once``` and it is a real update, not just the same in another way.

gautierhattenberger:
Yes it is indeed an addition not a simple alias.
Actually, it reminds me that we already have a `pre_call` and `post_call` instruction for the blocks. Maybe you should keep the same name to have this functionality at stage levels?

dewagter:
perfect: clear name + uniformity in the project

-Christophe

dewagter:
closing in favor of #1900, thanks @gautierhattenberger

PR #1883, opened 2016-09-30
[longdistance_nav] overflow in squared distance

dewagter:


sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -328,11 +328,11 @@ bool nav_approaching_from(struct EnuCoor_i *wp, struct EnuCoor_i *from, int16_t
   /* compute distance of estimated/current pos to target wp
    * distance with half metric precision (6.25 cm)
```
gautierhattenberger: comment should be updated

PR #1884, opened 2016-09-30
[dsmx] bind with DSMX

dewagter:


sw/airborne/arch/stm32/subsystems/radio_control/spektrum_arch.c
```diff
@@ -49,8 +49,13 @@ INFO("Radio-Control now follows PPRZ sign convention: this means you might need
 #define ONE_MHZ 1000000
 
 /* Number of low pulses sent to satellite receivers */
+#ifdef USE_DSMX
```
flixr: please use `#if USE_x` for flags, so it can explicitly be set to false as well...

PR #1885, opened 2016-10-01
remember last block so exceptions can continue where they left after the exception condition is over

dewagter:


sw/airborne/subsystems/navigation/common_flight_plan.h
```diff
@@ -55,6 +55,7 @@ void nav_goto_block(uint8_t block_id);
 #define Label(x) label_ ## x:
 #define Goto(x) { goto label_ ## x; }
 #define Return() { nav_block=last_block; nav_stage=last_stage; block_time=0;}
+#define ReturnToStartPrevBlock() { nav_block=last_block; nav_stage=0; block_time=0;}
```
gautierhattenberger: The `Return` instruction can be used from a special flight plan instruction. Shall we do the same for this, or with an attribute of the existing instruction ?

sw/airborne/subsystems/navigation/common_flight_plan.h
```diff
@@ -44,8 +44,8 @@ void nav_goto_block(uint8_t block_id);
 #define InitStage() nav_init_stage();
 
 #define Block(x) case x: nav_block=x;
-#define NextBlock() { nav_block++; nav_init_block(); }
-#define GotoBlock(b) { nav_block=b; nav_init_block(); }
+#define NextBlock() nav_goto_block(nav_block + 1)
```
podhrmic: Are we checking here for overflow in the number of blocks? E.g. I am in the last block and will want to go to the next block?

gautierhattenberger: The flight plan will fall into the default case which currently does nothing. But the instruction is not supposed to be called by hand, only from the generated code.

sw/airborne/subsystems/navigation/common_flight_plan.h
dewagter: Would be great. Any idea which should be the default: return to last_stage or stage=0

dewagter:
Unified: ```<return/>``` and ```<return reset_stage="1"/>``` are proposed instead of ```<call fun="ReturnToStartOfBlock"/>```

sw/tools/generators/gen_flight_plan.ml
```diff
@@ -324,7 +324,7 @@ let rec print_stage = fun index_of_waypoints x ->
     match String.lowercase (Xml.tag x) with
         "return" ->
           stage ();
-          lprintf "Return();\n";
+          lprintf "Return(%s);\n" (ExtXml.attrib x "reset_stage");
```
gautierhattenberger: You should not rely on the DTD to be sure that `reset_stage` attribute will exist. Better use `ExtXml.attrib_or_default` function instead.

sw/airborne/subsystems/navigation/common_flight_plan.h
```diff
@@ -54,7 +54,7 @@ void nav_goto_block(uint8_t block_id);
 
 #define Label(x) label_ ## x:
 #define Goto(x) { goto label_ ## x; }
-#define Return() { nav_block=last_block; nav_stage=last_stage; block_time=0;}
+#define Return(RESET_STAGE) { nav_block=last_block; if (RESET_STAGE==1) {nav_stage=0;} else {nav_stage=last_stage;} block_time=0;}
```
gautierhattenberger: why uppercase here ?

PR #1888, opened 2016-10-03
[module] driver for the TeraRanger One range finder

gautierhattenberger:


conf/modules/teraranger_one.xml
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="teraranger_one" dir="range_finder">
+  <doc>
+    <description>Driver for the TeraRanger One range finder (I2C)</description>
+    <configure name="TERARANGER_ONE_I2C_DEV" value="i2cX" description="Select I2C bus"/>
+    <define name="TERARANGER_ONE_I2C_ADDR" value="0x30" description="Select an alternate I2C address"/>
+    <define name="TERARANGER_ONE_OFFSET" value="0.0" description="Add an offset to the measurments" unit="m"/>
+    <define name="USE_TERARANGER_ONE_AGL" value="TRUE|FALSE" description="Send data as AGL measurments to the system"/>
```
flixr: Does it even make sense to be able to turn this off? If yes, should probably be mentioned that it is TRUE by default.

conf/modules/teraranger_one.xml
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="teraranger_one" dir="range_finder">
+  <doc>
+    <description>Driver for the TeraRanger One range finder (I2C)</description>
+    <configure name="TERARANGER_ONE_I2C_DEV" value="i2cX" description="Select I2C bus"/>
```
flixr: Should probably mention that i2c2 is the default...

sw/airborne/modules/range_finder/teraranger_one.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef TERARANGER_ONE_H
+#define TERARANGER_ONE_H
+
+#include "std.h"
+
+struct TeraRanger {
+  uint16_t raw; ///< raw distance in mm
+  float dist;   ///< scaled distance in m
+  float offset; ///< offset in m
+  bool data_available;
```
flixr: doesn't seem to be read anywhere, is this still supposed to be used somewhere?

conf/modules/teraranger_one.xml
gautierhattenberger: This is for the case where you don't want to use the sensor to detect ground but something else, like obstacle in front of you. In this case you don't want to send the ABI AGL message.

sw/airborne/modules/range_finder/teraranger_one.h
gautierhattenberger: This is used when you don't use the sensor through ABI message but with direct reading (see remark above).

conf/modules/teraranger_one.xml
gautierhattenberger: done

PR #1890, opened 2016-10-05
Added sbus fakerator for simulating SBUS transmitters on your laptop.

podhrmic:
Useful for HITL testing.

### DIFF
--- a/sw/tools/sbus_fakerator/sbus_fakerator
+++ b/sw/tools/sbus_fakerator/sbus_fakerator
@@ -1,0 +1,3 @@
+#!/bin/bash
+python sw/tools/sbus_fakerator/main.py $1 $2
+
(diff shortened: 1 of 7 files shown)


sw/tools/sbus_fakerator/sbus_fakerator
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+python sw/tools/sbus_fakerator/main.py $1 $2
```
flixr: Why not simply add a shebang line to main.py and rename it to sbus_fakerator?

podhrmic: I dont think I follow - can you elaborate? This works just fine so why not use it as is?

flixr: sure it works, just saying there is no need to have an extra file for this....
You could have also just added `#!/usr/bin/env python` to the top of main.py and renamed that to sbus_fakerator...

PR #1893, opened 2016-10-08
Coverity fixes from the latest build

podhrmic:
See https://scan.coverity.com/projects/paparazzi-paparazzi for details of the defects. 
Mostly out-of-bounds access

### DIFF
--- a/sw/airborne/modules/display/max7456.c
+++ b/sw/airborne/modules/display/max7456.c
@@ -400,7 +400,9 @@ static bool _osd_sprintf(char *buffer, char *string, float value)
   if (frac_nb > 3) {
     frac_nb = 3;       // Bound value.
   }
-  y = (sizeof(to_asc) - 2); // Point y to the end of the array.
+  // Point y to one before the end of the array
+  // Because we will increment it with frac_nb-1, so pointing it to the end would result in overflow
+  y = (sizeof(to_asc) - 2);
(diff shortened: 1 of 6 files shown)


sw/airborne/modules/hott/hott_eam.h
```diff
@@ -100,7 +100,7 @@ struct HOTT_EAM_MSG {
 
   uint16_t batt2_voltage;      //#23 #24 battery 2 voltage 0.02V steps
 
-  int8_t temp1;                //#25 Temperature sensor 1. 0�=20, 26�=46
+  int8_t temp1;                //#25 Temperature sensor 1. 0�=20, 26�=46
```
flixr: Please keep utf8 encoding here...

sw/airborne/modules/display/max7456.c
```diff
@@ -396,7 +400,7 @@ static bool _osd_sprintf(char *buffer, char *string, float value)
   if (frac_nb > 3) {
     frac_nb = 3;       // Bound value.
   }
-  y = (sizeof(to_asc) - 1); // Point y to the end of the array.
+  y = (sizeof(to_asc) - 2); // Point y to the end of the array.
```
flixr: not sure about this here... now it doesn't point to the last element of the array anymore...
if this is intended, the comment should be updated...

PR #1894, opened 2016-10-10
[gps_ubx] cast 1234 to correct type instead of 134

dewagter:
After using some of the generated code in an 8-bit micro-controller [which gave total garbage as result], a small issue was found in the UBX_GPS paring. This **probably** does not affect 32bit microcontrollers in most cases, but here is a fix just in case (if anyone else would run it on 8bit).

Any reason behind this? Since it is at 3 places it almost looks like done on purpose.


gautierhattenberger:
I suspect copy-paste to be the main culprit here

PR #1898, opened 2016-10-12
[uart_drop] rename for uniformity: module name='switch' type='uart'

dewagter:
I wanted to rename ```servo_switch``` to ```switch_servo``` as well for uniformity but that might break many airframe files. Can you advise on if this is a good idea or not.

In the modules folder I'd prefer to see all switching modules starting with the same name.


gautierhattenberger:
I don't mind changing the names to have something more coherent

flixr:
Also fine with me...

PR #1900, opened 2016-10-12
[flightplan] add pre and post calls at stage level

gautierhattenberger:
this can help for issues like #1882

### DIFF
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -159,6 +159,8 @@ height CDATA #IMPLIED
 throttle CDATA #IMPLIED
 climb CDATA #IMPLIED
 pitch CDATA #IMPLIED
+pre_call CDATA #IMPLIED
+post_call CDATA #IMPLIED
 until CDATA #REQUIRED>
 
 <!ATTLIST attitude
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
Should I merge this already ?

There is still something not consistent with the current generator when the `until` attribute is used. Some instructions are checking the condition at the beginning (like a `while` loop), while others are testing it after calling nav functions (like a `do ... while` loop). I didn't change the behavior for now. But what would be the best choice ?

flixr:
I don't quite understand what this has to do with the `until` attribute, can you elaborate (e.g with some examples)?

gautierhattenberger:
If you have `<attitude ... until="some_condition"/>`, it will generate
```
if (some_condition) NextStageAndBreak();
NavAttitude(...)
break;
```
But if you do `<circle ... until="some_condition"/>`, you will have
```
NavCircle(...)
if (some_condition) NextStageAndBreak();
break;
```
So, in the first case, if the condition is not true, the nav instruction will not be called, while with the second case, nav instructions are called at least once before changing stage.
This is indeed not related to the current issue, but I just realized it, so I'm wondering if something should be done for this.
For the first case, `heading`, `attitude`, `manual` are concerned. For the second, it's `go`, `stay`, `circle`, `eight`, `oval` and `survey`.

flixr:
Ok. Then I would say we merge this as it is... and make a new issue/PR for consistent `until` usage...

PR #1901, opened 2016-10-12
[bebop] new 3.9.0 firmware from parrot tested OK

dewagter:


sw/tools/parrot/bebop.py
```diff
@@ -125,8 +125,8 @@ def bebop_status():
     #check firmware version
     v = parrot_utils.check_version(tn, '').strip()
     print("Checking Bebop firmware version... " + v )
-    if ((v < '3.2.0') or (v > '3.4.0-RC2')):
-        print("Error: please upgrade your Bebop firmware to version between 3.2.0 and 3.4.0!")
+    if ((v < '3.2.0') or (v > '3.9.0')):
```
flixr: This won't work anymore as soon as a version with two digits e.g. 3.10.x is out...
See previous discussion in #1828

PR #1904, opened 2016-10-13
[pkg_man] mini package manager

dewagter:
store where files are copy-pasted from so they can benefit from downstream and update upstream

### DIFF
--- a/.packages
+++ b/.packages
@@ -1,0 +1,7 @@
+[package "sw/tools/iridium"]
+	url = https://github.com/tudelft/iridium.git
+	lcommit = 9956f97dc51fad3e7110682c967e99b146c29622
+	lpath = sw/tools/iridium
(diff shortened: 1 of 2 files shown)


flixr:
This is only meant to more easily update files that are copied and committed to the paparazzi tree, right?

podhrmic:
If it is for beebop only (judging from the comments?) shouldn't it be in `sw/tools` ?

dewagter:
@flixr yes, but most of all: to store where it came from, which revision etc
@podhrmic  not for bebop (will fix comments) but for small "partial submodules"

flixr:
LGTM

dewagter:
Any suggestion in which folder to put pkgman imported code? Can it be all
over the place like now should they all go in sw/ext or in sw/pkgman to
differentiate with submodules?

flixr:
Does it need to be put in a specific standard directory? It might make more sense to choose the directory based on what is copied...

PR #1907, opened 2016-10-24
PX4FLOW integration

podhrmic:
Fixed the velocity messages (using the correct variables)
Works for PX4FLOW mounted as described here: https://pixhawk.org/modules/px4flow
Option to use i2c or UART to get data from PX4FLOW
Sends debug messages (maybe make optional if needed).
Flight test pending (HOVER mode I suppose).


podhrmic:
Based on what I learned at a test flight, the PX4FLOW data cannot be simply integrated to obtain position (it would work only under ideal condition), so to actually use it for navigation a full state estimation is needed (I am looking into it). Does anyone have experience with OpenCV optical flow on ArDrone or Bebop to comment on this? @flixr @gautierhattenberger

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -30,64 +31,134 @@
...
-#define PX4FLOW_NOISE 0.5
+/*
+ * acceptable quality of optical flow (0-min,1-max)
+ * higher threshold means fewer measurements will be available
+ * but they will be more precise. Generally it is probably a better
+ * idea to have lower threshold, and propagate the noise (1-quality)
+ * to the INS
+ */
+#ifndef PX4FLOW_QUALITY_THRESHOLD
+#define PX4FLOW_QUALITY_THRESHOLD 0.1
```
flixr: Needs to be documented in the module xml file...
Also I don't quite understand why noise would be 1-quality... noise is not something unit-less between 0-1...

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -30,64 +31,134 @@
...
+  static float noise = 0;
+  quality = ((float)optical_flow.quality)/255.0;
+  noise = 1-quality;
+
+  if (quality > PX4FLOW_QUALITY_THRESHOLD) {
+    // flip the axis (if the PX4FLOW is mounted as shown in
+    // https://pixhawk.org/modules/px4flow
+    AbiSendMsgVELOCITY_ESTIMATE(PX4FLOW_VELOCITY_ID,
+                                optical_flow.time_usec,
+                                optical_flow.flow_comp_m_y,
```
flixr: Since this now changes the orientation the sensor has to be mounted, it would make sense to add the possibility to specify the mounting orientation.

flixr:
As long as you don't have any sort of absolute measurements your system will never be fully observable and you will drift no matter what fancy filter you can think of...

sw/airborne/modules/optical_flow/px4flow.c
```diff
@@ -30,64 +31,134 @@
...
+                        &optical_flow_rad.temperature,
+                        &optical_flow_rad.quality,
+                        &optical_flow_rad.time_delta_distance_us,
+                        &optical_flow_rad.distance);
+}
+
+static void decode_heartbeat_msg(struct mavlink_message *msg __attribute__((unused)))
+{
+  //TODO: Check whether we have a right version of firmware etc? i.e. heartbeat.autopilot=MAV_AUTOPILOT_PX4
+  DOWNLINK_SEND_HEARTBEAT(DefaultChannel, DefaultDevice,
```
flixr: I'm not in favor of adding mavlink messages verbatim to the paparazzi messages...

sw/airborne/modules/optical_flow/px4flow.h
```diff
@@ -33,22 +34,71 @@
...
+/** Mavlink optical flow rad structure.
+ *  Using MAVLINK v1.0 generated code:
+ *  Message ID 106
+ *  Fields are ordered to guarantee alignment
+ */
+struct mavlink_optical_flow_rad {
+  // Timestamp (microseconds, synced to UNIX time or since system boot)
+  uint64_t time_usec;
+  // Sensor ID
+	uint8_t sensor_id;
```
flixr: plz fix indentation

podhrmic:
Just for reference - this is how the pos hold with PX4FLOW can look like: https://www.youtube.com/watch?v=yEGzQCG80Z0 (this is sort of my goal). It might drift eventually, but good enough for a couple of minutes of HOVER.

sw/airborne/modules/optical_flow/px4flow.c
podhrmic: It would be great, but then you would have to specify the orientation of the sensor etc. so for the time being the user will have to mount the sensor in a specific way, improvements are indeed welcome later.

sw/airborne/modules/optical_flow/px4flow.c
flixr: No. Normalized noise doesn't make much sense.
This is the measurement noise to be used in the filter, so the variance.

sw/airborne/modules/optical_flow/px4flow.c
podhrmic: removed

podhrmic:
Good to merge. @flixr do you approve my changes?

sw/airborne/modules/optical_flow/px4flow.c
podhrmic: Well do you have a better solution? "scale somehow" is essentially what I am already doing here with the user defined scaling factor.

flixr:
I still think it would be better to try to provide something like a variance in the VELOCITY_ESTIMATE message. E.g. as a user I would like to take some measurements and compare them to my ground truth to get an approximate standard deviation/variance that I can then put as a define into my airframe file.
I have no idea how the "quality" reported scales with the actual variance of the measurement. But assuming I get close to 1 quality measurements, the PX4FLOW_NOISE_SCALE would actually be the variance. So IMHO PX4FLOW_NOISE_SCALE is not a good name for that...
I would suggest PX4FLOW_NOISE_STDDEV or so and scale that with (1-quality), then square it to get the variance and send that.
Standard deviation is something that is easy to understand, e.g. stddev of 0.1 m/s means that 95% of the measurements are within +-0.2m/s of the real value.

podhrmic:
Fixed. Also please note that using the measurements to get some meaningful position estimate is another topic (and PR) and I have to yet figure it out. In short, I might be changing the noise/quality/variance setup later.

flixr:
Ok, if you actually get measurements with quality of 255(normalized to 1) (not sure if this ever happens), this will set the noise to zero. So you could basically say that this value is really/absolutely true and the KF will take exactly that value.

flixr:
So in the end it could be something like: you measure the stddev in a good setup (quality close to 1) and in a bad setup, define the two and somehow scale in between (simplest case: linear).
And by default assign a stddev of e.g. 5 x PX4FLOW_NOISE_STDDEV to the bad case.
So maybe for now something like
```
noise = PX4FLOW_NOISE_STDDEV + (1 - quality) * PX4FLOW_NOISE_STDDEV * 5;
noise = noise * noise;
```
and set a default stddev of e.g. 1cm/s, so 0.01 or whatever you think is reasonable for this sensor.

podhrmic:
In practice you get quality close to 1 (255) fairly often. I think it mostly means that the optical flow was able to match the features between frames accurately. As I said, I ll have to look into the way Pixhawk uses optical flow, and probably use a kalman filter to use the flow measurements. Right now, the values are not really useful (doesn't work with ins_extended). So I would prefer to merge as is, and then revisit this later (rather than keep merging latest master into this branch).

flixr:
Then at least make it something as proposed above, that way you don't have a noise of zero, which is really not what you want as it basically "disables" the KF filter...
And you could easily make a ins_hff_extended module...

podhrmic:
Just to point out, there is currently no ins using the noise in velocity estimate (ins_hff discards it). So no matter what we agree on here is likely to change later once I actually start using the optical flow data for position estimates...

flixr:
It's used in INS with hff: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/ins/ins_int.c#L536-L538

podhrmic:
Ah, I missed that. Well in that case don't merge it yet, I ll keep making changes here and will merge it after it is test flown.

podhrmic:
@flixr This should fix your concerns. The stddev for px4flow can be adjusted in settings.

Issue #1899, opened 2016-10-12
Video thread(s) not running in latest master


wilcoschoneveld:
The video_thread_function (sw/airborne/modules/computer_vision/video_thread.c:83) does not seem to start in the latest master. It still works as of commit fe5500811fa46d51c427cbc7cc83d5e0d1a973e2. 

I looked into the commits but couldn't find a hint of where it broke. Could this be because of the refactored telemetry modules? @gautierhattenberger 

Thanks!

flixr:
You did run a toplevel `make clean && make` or at least a `make` right?

wilcoschoneveld:
Yes, I did

kirkscheper:
This is confirmed on a different bebop using a different pc. Both cameras don't startup for some reason.

flixr:
Please always specify exactly which version you tested...
Can you try to narrow down on the commit that broke this? E.g. try right before and after #1835 was merged with 4316871b6324276d2c3128dbf09d7664beea9142

wilcoschoneveld:
I tested master at commit 183ac5cfe627311a7a78ac277d1cf62a529559d2. I understand that my issue description is not really extensive and this is because I have not investigated this issue as much; we are really strapped for time due to IMAV coming up next week.

flixr:
Closed with 44f15c363d09a066c04c68b6848cc67aa4c732da

PR #1856, opened 2016-09-16
Bebop front camera subsampling

wilcoschoneveld:
This pull request allows for sub-sampling in the bebop front camera to reduce the camera resolution and increase performance. I took the work from @wv-tud and updated it with the latest master. The current settings give a steady 15 fps.

I'm trying to figure out how to correctly combine subsampling with a desired output size in order to get a more conventional image size, but there are a number of programming restrictions to follow; http://www.onsemi.com/pub_link/Collateral/MT9F002-D.PDF

PR #1908, opened 2016-10-27
Dual video stream

kirkscheper:
This PR does the following:
Fixes bebop video not running on master as defined in #1899 (can be cherry picked)
Updates to the ISP configuration which allows for subsampling and image scaling of the bebop front camera, replaces #1856
Adds Auto White Balance and Auto Exposure for front camera on the Bebop (needs a small change to the exposure, found some inconsistencies when performed together with subsampling).
Adds dual video streaming capabilities for the bebop and update the rtp viewer (naming convention may be still changed).

This is almost done so I will push this to get some feedback before it gets to polished. Thanks in advance. @wilcoschoneveld @wv-tud


conf/airframes/examples/bebop.xml
```diff
@@ -32,15 +43,20 @@
...
-      <define name="VIEWVIDEO_CAMERA" value="front_camera"/>
-      <define name="VIEWVIDEO_DOWNSIZE_FACTOR" value="1"/>
-      <define name="VIEWVIDEO_QUALITY_FACTOR" value="20"/>
+      <define name="VIEWVIDEO_CAMERA" value="bottom_camera"/>
+      <define name="VIEWVIDEO_CAMERA2" value="front_camera"/>
+      <define name="VIEWVIDEO_DOWNSIZE_FACTOR" value="2"/>
+      <define name="VIEWVIDEO_QUALITY_FACTOR" value="40"/>
+      
+      <configure name="VIEWVIDEO_BROADCAST" value="FALSE"/>
+      <configure name="VIEWVIDEO_HOST" value="192.168.42.98"/>
```
wilcoschoneveld: In the example airframe file we might want to have broadcast enabled, or a way of detecting the right host address since right now I think it would only work on your IP.

sw/airborne/modules/computer_vision/viewvideo.c
```diff
@@ -72,7 +72,7 @@ PRINT_CONFIG_VAR(VIEWVIDEO_RTP_TIME_INC)
 #ifdef VIDEO_THREAD_SHOT_PATH
 #define VIEWVIDEO_SHOT_PATH VIDEO_THREAD_SHOT_PATH
 #else
-#define VIEWVIDEO_SHOT_PATH /data/video/images
+#define VIEWVIDEO_SHOT_PATH /data/ftp/internal_000/images
```
wilcoschoneveld: maybe we can just remove the entire VIEWVIDEO_SHOT_PATH thing. I don't think we need to create those streamer files and it's not used for anything else. Also, VIDEO_THREAD_SHOT_PATH has already been long gone; video_capture.c is the boss now.

wilcoschoneveld:
LGTM, added some comments. Awesome work on this pull request!

sw/tools/rtp_viewer/rtp_viewer.py
```diff
@@ -10,9 +12,41 @@
...
+\t-p PORT | --port=PORT where PORT is the port number to open for the RTP stream (5000 or 6000)
+\t-s SCALE | --scale=SCALE where SCALE is the scaling factor to apply to the incoming video stream (default: 1)
+\t-r ROTATE | --rotate=ROTATE where ROTATE is the number of clockwise 90deg rotations to apply to the stream [0-3] (default: 0)
+'''
+    print(fmt % lpathitem[-1])
+    
+def GetOptions():
+    options = {'port':[], 'scale':[], 'rotate':[]}
+    try:
+        optlist, left_args = getopt.getopt(sys.argv[1:],'h:p:s:r:', ['help', 'port=', 'scale=', 'rotate='])
```
flixr: I would also advocate argparse...

flixr:
Thx, cherry picked the video_thread fix with 44f15c363d09a066c04c68b6848cc67aa4c732da

sw/tools/rtp_viewer/rtp_viewer.py
```diff
@@ -72,7 +78,7 @@ def on_key(self, key):
             self.mouse['start'] = None
 
     def on_mouse(self, event, x, y, flags, param):
-        if event == cv2.EVENT_LBUTTONDOWN and self.rotate == 0:
+        if event == cv2.EVENT_LBUTTONDOWN and self.rotate == 0 and False:
```
flixr: This completely disables the mouse event...

wilcoschoneveld: @kirkscheper the 'and False' part was added in IMAV to disable the mouse event, since it has no use. In order to make the VIDEO_ROI message usable I believe we also need to account for stream downsize, opencv scale & rotation and stream port (?to know aircraft and camera). Maybe this is for a different PR though. What do you think?

kirkscheper: Defo a different PR. @wilcoschoneveld  can you look at this? I think I will turn it back on for now and make an issue for this.

flixr:
LGTM, just wondering if it wouldn't be more clear to name the `cv_ae_awb` module something like `bebop_ae_awb` since it's specific to the bebop anyway.

flixr:
@kirkscheper can you rename the module to something more descriptive as posted before?

kirkscheper:
Sorry for the long delay, got swamped by end of year tasks.

conf/airframes/examples/bebop.xml
```diff
@@ -10,6 +10,17 @@
       <module name="udp"/>
     </target>
 
+    <define name="MT9F002_OUTPUT_HEIGHT" value="640" />
```
flixr: I forgot to ask: in which cases is it necessary to change the defaults as you have done here?

kirkscheper: I think the original resolution is 2048x3320 over the 180 degree FOV but the user can select the size of the crop and subsampling they like using these settings.

flixr: Ok. So the output width/height is just the cropped part?? Or downsampled?

In any case most people won't be able to set all these defines appropriately.
So I guess it would be nice if we had a couple of useful pre-defined combinations that you can then select (define).

kirkscheper: Good point, it isn't that straight forward or obvious. @wilcoschoneveld is it possible for you to make a couple of useful combinations that help a new user? I guess this is also useful for the course.

wilcoschoneveld: Yes I can look into this. What about something like 3 different combinations (1 high fps, 1 wide angle, 1 in between) and put them in the airframe commented out?

kirkscheper: @flixr where is a good place to put this, the module config maybe. I guess we want it to be added to the doxygen as well, not just the example airframe.

flixr: These 3 sound good to me.
But it would be nicer to add those to the file where the defaults are defined.
So that the user can add something like
`<define name="BEBOP_CAMERA_CONFIG" value="HIGH_FPS"/>`
or similar.
If you omit that, you will get the defaults or you can define each option explicitly as done here...

flixr: Not sure to which module this define should then be added for documentation purposes... as IIRC there is no bebop specific module...

wilcoschoneveld: We are working on a separate vision library which will make the paparazzi vision implementation obsolete.

fvantienen: If anyone is interested it is currently available and under development at https://github.com/tudelft/tudelft_vision.

PR #1909, opened 2016-10-28
added outlier detector to bebop sonar

kirkscheper:
This looks for large changes in sonar readings and delays sending the new measurements till the rate of change of the new measurement compared to the last sent measurement drops below a certain fixed threshold. This was found to be necessary to prevent large outliers from being propagated through the ins (the sonar is typically trusted quite heavily).

This outlier detector will also delay the propagation of true vertical height transitions for eg when flying over a table, the gradient threshold has been chosen to remove temporary outliers and not delay true transitions for very long.

### DIFF
--- a/sw/airborne/filters/median_filter.h
+++ b/sw/airborne/filters/median_filter.h
@@ -24,49 +24,54 @@
 #ifndef MEDIAN_H
 #define MEDIAN_H
 
-#ifndef MEDIAN_DATASIZE
-#define MEDIAN_DATASIZE 5
-#endif
+#define MAX_MEDIAN_DATASIZE 13
+#define MEDIAN_DEFAULT_SIZE 5
 
 #include "std.h"
 #include "math/pprz_algebra_int.h"
 
 struct MedianFilterInt {
-  int32_t data[MEDIAN_DATASIZE], sortData[MEDIAN_DATASIZE];
-  int8_t dataIndex;
+  int32_t data[MAX_MEDIAN_DATASIZE], sortData[MAX_MEDIAN_DATASIZE];
+  uint8_t dataIndex;
+  uint8_t size;
 };
 
-inline void init_median_filter(struct MedianFilterInt *filter);
-inline int32_t update_median_filter(struct MedianFilterInt *filter, int32_t new_data);
-inline int32_t get_median_filter(struct MedianFilterInt *filter);
+inline void init_median_filter_i(struct MedianFilterInt *filter, uint8_t size);
+inline int32_t update_median_filter_i(struct MedianFilterInt *filter, int32_t new_data);
+inline int32_t get_median_filter_i(struct MedianFilterInt *filter);
 
-inline void init_median_filter(struct MedianFilterInt *filter)
+inline void init_median_filter_i(struct MedianFilterInt *filter, uint8_t size)
 {
-  int i;
-  for (i = 0; i < MEDIAN_DATASIZE; i++) {
+  uint8_t i;
+  if (size > MAX_MEDIAN_DATASIZE){
+    filter->size = MAX_MEDIAN_DATASIZE;
+  } else if ((size % 2) == 0) {
+    // force filter to have odd number of entries so that
+    // returned median is always an entry and not an average
+    filter->size = size + 1;
+  } else {
+    filter->size = size;
+  }
+  for (i = 0; i < filter->size; i++) {
     filter->data[i] = 0;
     filter->sortData[i] = 0;
   }
   filter->dataIndex = 0;
 }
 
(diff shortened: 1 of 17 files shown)


gautierhattenberger:
Did you consider using a median filter https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/filters/median_filter.h ? It is usually very efficient at removing peeks while introducing very low delay.

kirkscheper:
@gautierhattenberger nice idea, didn't think of that. A median filter would have the added benefit of continuously propagating a measurement unlike the current solution. If no one has a problem with that I can change the outlier detector to median filter.

podhrmic:
I like the idea with the median filter - certainly useful for other sonar readings (such as px4flow).

kirkscheper:
I made the median filter variable size and expanded it to handle floats this was not strictly necessary for this application, I could have filtered the index rather than the distance but yea... If anyone has a big issue with that I can change it). 

This also now replaces #1780.

sw/airborne/filters/median_filter.h
```diff
@@ -25,6 +25,7 @@
 #define MEDIAN_H
 
 #define MAX_MEDIAN_DATASIZE 128
+#define MEDIAN_DEAFULT_SIZE 5
```
podhrmic: typo here ;-/

gautierhattenberger:
This also should be merged or rebased on master to fix conflict in sonar file.

sw/airborne/modules/lidar/lidar_lite.c
```diff
@@ -62,7 +62,7 @@ void lidar_lite_init(void)
   lidar.distance = 0;
   lidar.distance_raw = 0;
 
-  init_median_filter(&lidar_filter);
+  init_median_filter(&lidar_filter, 7);
```
podhrmic: This should be configurable from the airframe config file - maybe add a variable in lidar_lite.xml that specifies the size? Lets say in noisier applications I might want to change this value to something else.

podhrmic:
@kirkscheper Could you fix the doxygen comments as @flixr mentioned and also the user-adjustable filter size (see my comment)? Otherwise it looks good to me.

flixr:
@kirkscheper any updates here?

kirkscheper:
Again, sorry for the delay

flixr:
@kirkscheper can you plz fix the build errors and conflicts? Otherwise it looks good to me :+1:

podhrmic:
@kirkscheper any updates here? Can you fix the conflicts?

kirkscheper:
After quite a long wait (sorry again) this is finally done. I tested this on a bebop 2 and the sonar part works well however the ins filter was quite strange (with sonar on or off). I didn't look deeply into what was going on but the estimated velocity was always really large, even when the baro reading was relatively steady. This was made way worse when doing a geo_init (NavSetGroundReferenceHere).

Anywho, this should be done now.

sw/airborne/filters/median_filter.h
```diff
@@ -75,61 +76,175 @@ inline int32_t update_median_filter(struct MedianFilterInt *filter, int32_t new_
...
   struct MedianFilterInt mf[3];
 };
 
-#define InitMedianFilterVect3Int(_f) {  \
-    for (int i = 0; i < 3; i++) {         \
-      init_median_filter(&(_f.mf[i]));    \
-    }                                     \
+#define InitMedianFilterVect3Int(_f) {                        \
+    for (int i = 0; i < 3; i++) {                             \
+      init_median_filter_i(&(_f.mf[i]), MEDIAN_DEFAULT_SIZE); \
```
gautierhattenberger: why not passing the size parameters in this macro as well ? same for the others of this kind.

sw/airborne/filters/median_filter.h
```diff
@@ -75,61 +76,175 @@ inline int32_t update_median_filter(struct MedianFilterInt *filter, int32_t new_
...
+  // return data value in middle of sorted array
+  return get_median_filter_i(filter);
 }
 
-inline int32_t get_median_filter(struct MedianFilterInt *filter)
+inline int32_t get_median_filter_i(struct MedianFilterInt *filter)
 {
-  return filter->sortData[(MEDIAN_DATASIZE) >> 1];
+  if (filter->size % 2 == 0){
+    return (filter->sortData[filter->size / 2] + filter->sortData[filter->size / 2 - 1]) / 2;
```
gautierhattenberger: I'm not completely happy this, because it goes again one of the property of the median filter which is that the output value is always one of the input elements. It is not the case here anymore. Not sure it is the most useful property, but still. I can see two options here:
- the size parameter is always set to the closest upper odd value
- the size parameter is only the "half-size", so the final size is (2*size+1)

PR #1910, opened 2016-10-28
added agl estimator when assuming non flat ground

kirkscheper:
Typically, in the Vertical Float Filter, the sonar is assumed to always be correct and is used to estimate the baro offset from the ground altitude, this is only true when we assume that the ground is flat with no obstacles. This PR proposes that if so defined by the user, the filter can be inversed allowing the VFF to estimate the obstacle height from the difference between the baro and sonar measurements allowing the vehicle to fly at constant barometric altitude over obstacles. The sonar measurements help to filter the barometer measurements and provide more consistent vertical speed estimates. This filter has already been test flown in an obstacle filled environment to good effect.

One observed side effect is that the barometric pressure in an environment changes over time which would cause the vehicle to have a biases altitude. To fix this I propose to augment the Kalman filter with an additional state to estimate both the sonar and barometer offset using time scale separation, fast changes in sonar alt are likely obstacles but time averaged readings are likely the ground truth (pun intended) and should be added as a baro offset. I will implement and test this shortly. Any feedback is appreciated.


sw/airborne/state.h
```diff
@@ -580,6 +582,14 @@ static inline void stateSetPositionUtm_f(struct UtmCoor_f *utm_pos)
   state.pos_status = (1 << POS_UTM_F);
 }
 
+/// Set altitude above ground level (float).
+static inline void stateSetAltAgl_f(float agl)
+{
+  state.alt_agl_f = agl;
+  /* clear bits for all position representations and only set the new one */
```
flixr: I don't think that it is a good idea to clear all other bits here as you don't want to invalidate the rest of the position...

sw/airborne/state.h
```diff
@@ -691,6 +701,15 @@ static inline struct UtmCoor_f *stateGetPositionUtm_f(void)
   return &state.utm_pos_f;
 }
 
+/// Get altitude above ground level (float).
+static inline float *stateGetAltAgl_f(void)
```
flixr: Should directly return a float instead of a pointer to it (like the other functions returning a primitive type).

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -145,6 +141,15 @@ static void gps_cb(uint8_t sender_id, uint32_t stamp, struct GpsState *gps_s);
 static abi_event vel_est_ev;
 static void vel_est_cb(uint8_t sender_id, uint32_t stamp, float x, float y, float z, float noise);
 
+/** ABI binding for AGL.
+ * Usually this is coming from opticflow.
```
flixr: This sounds a bit weird, optiflow doesn't give you an AGL measurement... it's still the sonar (or something similar)

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -315,30 +319,50 @@ void ins_int_propagate(struct Int32Vect3 *accel, float dt)
...
-    ins_int.baro_initialized = true;
-  }
-
-  if (ins_int.baro_initialized) {
-    if (ins_int.vf_reset) {
-      ins_int.vf_reset = false;
-      ins_int.qfe = pressure;
-      vff_realign(0.);
-      ins_update_from_vff();
+  static uint8_t num_points = 0;
```
flixr: This should either be factored out into a filter or one of the existing lowpass, mean or median filters used...

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -54,19 +54,15 @@
 #include "math/pprz_geodetic_int.h"
 #include "math/pprz_isa.h"
 
+#ifndef INS_BARO_AGL_OFFSET
```
flixr: What is this offset supposed to be?
If it is supposed to be the static offset between the reported AGL and the real AGL, this should not be done in the INS, but the offset directly applied in the module creating the AGL message.

kirkscheper: oh, maybe the name isn't very good but it is supposed to be the initial baro offset, it is currently actively estimated in the vff but it should be possible for the user to define a static offset value of the baro from the ground for eg when the vehicle has long legs or there is some large physical offset of the barometer from the ground. This is also used when resetting the altitude, when reset the baro altitude doesn't go to zero but is reset to this static value. Was useful to us recently when using a quad with long landing legs.

sw/airborne/subsystems/ins/ins_int.c
kirkscheper: haha, typo. tnx

sw/airborne/subsystems/ins/ins_int.c
kirkscheper: Will do. Is this a good idea though? I had implemented this cause the baro on the bebop is HORRIBLY noisy and changing the certainty of the Kalman filter wasn't enough to filter it out. I can imagine that this may not be very beneficial for other baros. I think I may move this to only apply to the bebop baro.

sw/airborne/subsystems/ins/ins_int.c
flixr: This is basically a question of coordinate system...
Right now most other parts of Paparazzi assume that if you are on the ground, you are at 0 AGL and not at 0.3m if your baro is mounted 30cm from the landing legs.

kirkscheper:
This is mostly done, just needs some refactoring after #1909 and a flight test to confirm performance.

podhrmic:
@kirkscheper any updates here?

podhrmic:
@flixr @kirkscheper what should we do with this PR? Potentially useful functionality, but if it is not finished...

podhrmic:
@flixr should we keep it open? @kirkscheper can you update the PR?

OpenUAS:
@kirkscheper , looks useful, can offer to re-test in various scenarios ( after finished and conflicts are fixed) .Since it is an extension that is not enabled by default, if reviewed and tested and docs added of usage and example, a lower risc fix IMHO and we love to see it in master...

OpenUAS:
Self assigned in the hope we can find time to work with @kirkscheper over this to get it retested, and merged finally into upsteram

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -435,33 +466,37 @@ void ins_int_update_gps(struct GpsState *gps_s)
...
 #if USE_SONAR
-static void sonar_cb(uint8_t __attribute__((unused)) sender_id, float distance)
-{
-  static float last_offset = 0.;
-
-  /* update filter assuming a flat ground */
-  if (distance < INS_SONAR_MAX_RANGE && distance > INS_SONAR_MIN_RANGE
+    if (distance > 0
+        && vff.z > -INS_SONAR_MAX_RANGE
+        && vff.z < -INS_SONAR_MIN_RANGE
```
kirkscheper: So I am reviving this PR but have a couple of quips. Firstly, This line basically assumes that the vertical filter is converged and close to the true value but what if it isn't... What if the sonar is measuring the correct distance but the vertical filter is not correct? 

What is a better way to test that the measured sonar value is correct?

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -553,9 +583,25 @@ static void vel_est_cb(uint8_t sender_id __attribute__((unused)),
...
+  last_stamp = stamp;
+  prev_vel.x = vel_ned.x;
+  prev_vel.y = vel_ned.y;
+
+  if (stamp - last_pos_stamp >= 0.05) {
+    ins_int.ltp_pos.x += POS_BFP_OF_REAL(sum_x * 1e-6);
+    ins_int.ltp_pos.y += POS_BFP_OF_REAL(sum_y * 1e-6);
+    sum_x = 0.; sum_y = 0.;
+    last_pos_stamp = stamp;
   }
```
kirkscheper: I think I will put this in a separate PR

OpenUAS: Thx for reviving

kirkscheper:
ok, so I made a few changes to this. The biggest one is that I got rid of the USE_SONAR and SONAR_UPDATE_ON_AGL. 

IMHO this was super confusing. The USE_SONAR would activate the sensor and the SONAR_UPDATE_ON_AGL would allow this value to be used by the ins but that is only true if you are using the vff_extended module. In addition, IMHO you shouldn't use the vff_extended if you are not using sonar/agl because that is the entire point of the vff_extended. If you don't want SONAR_UPDATE_ON_AGL then just use vff. This removes a lot of confusion and removes a redundant setting/option.

I can image that if you have multiple agl sensors that you may want to disable the sonar. But I think that the sonar should be default on and you can disable them by disabling the abi messages for the sensors that you don't want. 

Secondly, I got rid of the SONAR_MIN_RANGE and SONAR_MAX_RANGE cause I am not really sure how they are supposed to work. I would image they are the limits of where you want to use the sonar, so if you are higher than 5m you don't want to use it any more. The problem here is if the sonar stops working st 5m and is at 6m the sensor won't say 6m, it will say 5m or 0m or some random number in between so this check wouldn't prevent you from using values outside of the range of sensor. If I am misunderstanding this please let me know. 

Additionally, the max values are probably only dependent on the sensor so these limit checks should happen there before the value is sent over ABI to ensure that the data is valid.

Also, these parameters are generally just copy-pasted from airframe to airframe without being explicitly used.

There are also some bug fixes here. I have also tested the obstacle estimator on a bebop and it looks ok. The values in the filter result in a 20cm jump in the state estimate when traversing a table of 80cm.

I would love some feedback here so I'm looking forward to your comments!!

kirkscheper:
After a very long wait I was finally able to finish this PR. It is flight tested and ready for review. This affect the ins_int so requires a close review for potential risks.

On that, I saw that if your flight plan is initialized far from the real location of the vehicle some strange behavior can happen due to the hight correction applied to the curvature of the earth. The fix I made seems to help but check if this makes sense.

Another is in the bebop sonar. Here I reduced the sensitivity of the detection, this helped with removing false positive reflections when out of the range of the sensor (above 4m). This may cause some false negatives if the surface flown over is not overly reflective. I was not able to test it over different floors but I think that this setting is fine. I did not make it a settable variable as it is not very intuitive value to set.

This should also fix  #2223, which is an error I accidentally introduced a while back (sorry). The baro setup here has also been confirmed working on a disco.

sw/airborne/boards/baro_board_ms5611_i2c.c
```diff
@@ -64,7 +65,15 @@
 PRINT_CONFIG_VAR(BB_MS5611_TYPE_MS5607)
 
 struct Ms5611_I2c bb_ms5611;
+Butterworth2LowPass lowpass_pressure;
```
gautierhattenberger: Is it reasonable to filter the raw values inside the sensor drivers ? For instance some filters might better handle unfiltered data rather than filtered (and slightly delayed) data. I think this should be the task of the INS to pre-filter or not the sensor data. Also we have more sensors than navigation filters.
An exception would be the case of a sensor with know "issue" like peaks where a median filter for instance would help to remove outliers.

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -448,33 +471,45 @@ void ins_int_update_gps(struct GpsState *gps_s)
...
+ * This callback handles all estimates of the height of the vehicle above the ground under it
+ * This is only used with the extended version of the vertical float filter
+ */
+#if USE_VFF_EXTENDED
+static void agl_cb(uint8_t sender_id, float distance) {
+  if (distance <= 0 || !(ins_int.baro_initialized)) {
+    return;
+  }
 
+if (sender_id == AGL_SONAR_ADC_ID ||
```
gautierhattenberger: Would be better to avoid the explicit list of sonar ID I think. Do you imagine the case where you have to two types of sonar/agl sensors at the same time ?

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -330,39 +335,56 @@ void ins_int_propagate(struct Int32Vect3 *accel, float dt)
...
 
-  if (ins_int.baro_initialized) {
-    if (ins_int.vf_reset) {
-      ins_int.vf_reset = false;
-      ins_int.qfe = pressure;
-      vff_realign(0.);
-      ins_update_from_vff();
-    } else {
-      float baro_up = pprz_isa_height_of_pressure(pressure, ins_int.qfe);
+  if (!ins_int.baro_initialized && pressure > 1e-6) {
```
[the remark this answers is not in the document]
kirkscheper: So the 1000 there is actually up for discussion. I have seen some strange things here with the baro on the bebop where there are some really small values returned but is quite sporadic so was not able to log the actual value. What is a reasonable number to check for a bad measurement (instead of 1000 which is quite arbitrary).

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -330,39 +335,56 @@ void ins_int_propagate(struct Int32Vect3 *accel, float dt)
...
+#define press_hist_len 10
+    static float press_hist[press_hist_len];
+    static uint8_t idx = 0;
+
+    press_hist[idx] = pressure;
+    idx = (idx + 1) % press_hist_len;
+    float var = variance_f(press_hist, press_hist_len);
+    if (var < 1){
```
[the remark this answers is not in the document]
kirkscheper: yes, but 1 is already quite large here. I will add a define.

sw/airborne/boards/baro_board_ms5611_i2c.c
kirkscheper: I did this here cause this baro is horrible, the noise is really bad. The default configuration is already on maximum oversampling ratio so I was not sure what else to do. I can replace the lp filter with a median filter but I think that this raw signal is not reliable for any use...

sw/airborne/subsystems/ins/ins_int.c
kirkscheper: This was done since the previous define was USE_SONAR and the defines to use the value depending on the altitude was also specific to sonar. I can replace the sonar defines with AGL defines but indeed, what if you have multiple sensors that can measure agl. We have had drones with sonar, laser and downward facing stereo camera, all which can provide agl.

Advice on how best to handle that config?

sw/airborne/subsystems/ins/ins_int.c
```diff
@@ -424,7 +446,8 @@ void ins_int_update_gps(struct GpsState *gps_s)
 
   if (ins_int.hf_realign) {
     ins_int.hf_realign = false;
-    hff_realign(gps_pos_m_ned, gps_speed_m_s_ned);
+    const struct FloatVect2 zero = {0.0f, 0.0f};
+    hff_realign(gps_pos_m_ned, zero);
```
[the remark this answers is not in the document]
kirkscheper: tbh, I don't remember why i did this... don't even know why it is in this pr. Will remove.

sw/airborne/subsystems/ins/vf_extended_float.c
```diff
@@ -178,61 +213,41 @@ void vff_propagate(float accel, float dt)
...
+  vff.offset      += K[3] * y;
+  vff.obs_height  += K[4] * y;
 
-  update_baro_conf(z_meas, conf);
+  for (i = 0; i < VFF_STATE_SIZE; i++) {
+    for (j = 0; j < VFF_STATE_SIZE; j++) {
+      vff.P[i][j] -= K[i] * P[j];
+    }
+  }
+#ifdef VFF_EXTENDED_NON_FLAT_GROUND
```
gautierhattenberger: This new flag is not documented in the module xml I think.
Also, in this case, is it really`idef` or `ifndef`. This part seems to drive the obstacle height to zero. Should this be called when the feature is activated or not ? When not activated, what prevents the internal state to diverge ?

OpenUAS:
Retry to get it fixed, tested and merged

sw/airborne/subsystems/ins/vf_extended_float.c
kirkscheper: The internal state should not be updated when the VFF_EXTENDED_NON_FLAT_GROUND is not set since vff.P[4][4] = 0.f;. As such the entire row and column of the P matrix is never updated. I confirmed this on-board.

I removed the ifdef for safety, who knows...

I also added some documentation and updated the vff message with some extra info. This is dependent on https://github.com/paparazzi/pprzlink/pull/91.

gautierhattenberger:
update pprzlink for the new message and it is good to merge

OpenUAS:
Great, I can test on Bebop, Bebop2 and my master based Disco Branch if needed

Issue #1857, opened 2016-09-19
Adapted Throttle causes drop in guidance height


knmcguire:
Hi, I have an issue with the vertical guidance with ADAPT_THROTTLE_ENABLED on true. 

If I switch from AP_MODE_ATTITUDE_RC_CLIMB to AP_MODE_HOVER_Z_HOLD with the Bebop 1 at 1-2 meters, it drops to the ground and slowly rises to the correct height. However I want it to stay at that height from the beginning 

I had the same problem with AP_MODE_ATTITUDE_DIRECT  to hover, but that is because it didn't adapt the intergral factor by the adapt throttle. However, with RC_CLIMB, it is already using the adapt_throttle in guidance... 

Anybody have any thoughts on how to fix this? Thanks in advance.

agressiva:
Some time i have the same problem when using adaptive_throttle
With adaptive disabled the quad never drop altitude.

flixr:
Do you have any logs that contain this? It's kind of hard to guess where the issue is without any actual data...

We should probably not run the adaptive throttle estimator if we are in DIRECT mode... but it *should* work fine when switching from any other mode that internally runs the hover loop.

flixr:
Also if the acceleration is over a threshold (`GUIDANCE_V_ADAPT_MAX_ACCEL`) the estimator is not updated... and since the acceleration in the state interface is fed by the INS, it also depends what INS you use and how "clean" the estimated current acceleration is.

kirkscheper:
For us, this is 'solved' with 1917...

PR #1911, opened 2016-10-28
[guided mode] expand guided mode in vertical guidance

kirkscheper:
This refactors the speed/position setting in the vertical loop in guided mode as well as added a thrust setting option which should facilitate autonomous takeoff in guided mode care of @dewagter .

I am also sneaking in a fix to the vertical adaptive loop which should address #1857 , basically if the user defines a nominal throttle setting as well as adaptive throttle, use the user definition as the initial guess for the adaptive estimator. Tested and verified improvement.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
@@ -103,7 +103,13 @@ int32_t guidance_v_delta_t;
 
 float guidance_v_nominal_throttle;
 bool guidance_v_adapt_throttle_enabled;
-bool guidance_v_guided_vel_enabled;
+
+
+#define GUIDANCE_V_GUIDED_MODE_ZHOLD      0
+#define GUIDANCE_V_GUIDED_MODE_CLIMB      1
+#define GUIDANCE_V_GUIDED_MODE_THROTTLE   2
+
+int guidance_v_guided_mode;
 
 /** Direct throttle from radio control.
  *  range 0:#MAX_PPRZ
@@ -118,6 +124,7 @@ int32_t guidance_v_rc_zd_sp;
 
 int32_t guidance_v_z_sp;
 int32_t guidance_v_zd_sp;
+int32_t guidance_v_th_sp;
 int32_t guidance_v_z_ref;
 int32_t guidance_v_zd_ref;
 int32_t guidance_v_zdd_ref;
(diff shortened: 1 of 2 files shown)


wilcoschoneveld:
This PR also fixes an issue where the guided speed reference would be reset when switching to guided mode, resulting in a vertical drop in altitude.

sw/airborne/firmwares/rotorcraft/guidance/guidance_v_adapt.c
```diff
@@ -36,12 +36,17 @@
...
- *  finds the vehicle needs more thrust.
+ *  If the nominal hover throttle is defined use it otherwise it is better
+ *  to start with low thrust and let it rise as the adaptive filter finds
+ *  the vehicle needs more thrust.
  */
 #ifndef GUIDANCE_V_ADAPT_INITIAL_HOVER_THROTTLE
+#ifdef GUIDANCE_V_NOMINAL_HOVER_THROTTLE
+#define GUIDANCE_V_ADAPT_INITIAL_HOVER_THROTTLE GUIDANCE_V_NOMINAL_HOVER_THROTTLE
+#else
 #define GUIDANCE_V_ADAPT_INITIAL_HOVER_THROTTLE 0.3
```
podhrmic: It might make sense to change the default to 0.5 (ideally you want to hover at ~50% throttle)?

podhrmic:
The guided part looks good, the nominal throttle fix should be a separate PR though. I believe you that it improves the hover performance, although I am not sure that the description of issue in #1857 is what is being fixed here.

sw/airborne/firmwares/rotorcraft/guidance/guidance_v_adapt.c
flixr: I would advise against raising the default to 50%
- there is no reason why it should be ideal to hover at 50% (it's even nicer if you need less, i.e. have a better thrust/weight ratio)
- it's better to start at a lower hover throttle and let the adaptive throttle increase it than to start with too much and have it shoot up

gautierhattenberger: I confirm Felix comments, the initial value is set low on purpose to avoid a rotorcraft that hover at low throttle to jump during takeoff.

kirkscheper:
About the fix to #1857, what we observed was that when flying in ATT mode the adaptive throttle is not active so when you switch to guided mode (where it is active) the estimator starts estimating the nominal throttle at the default low throttle setting which is (by design) too low so the vehicle descends. This simply allows the user to set a best guess of the nominal throttle which should reduce the observed drop. It does not strictly 'fix' the problem but it give the user a way to minimise the problem. For us it works well.

I will make a separate PR for that 'fix'.

gautierhattenberger:
I think it looks good to merge. The newly added direct throttle control in guided mode can't be used from the GUIDED message at the mode, only as an internal command. Any plan to add this possibility (if needed) ?

PR #1913, opened 2016-10-28
[blob_finder] upgraded pixel_cnt to 32 bit integer

wilcoschoneveld:
The unsigned range of integer values that can be stored in 16 bits is 0 through 65,535. This is not sufficient enough for blob detection with e.g. a bebop front camera.


podhrmic:
Does this affect the calculation along the line? For example here: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/cv_blob_locator.c#L185 
(we are dividing by int32 now)

wilcoschoneveld:
I don't think so. The mentioned line calculates the cg of a blob and according to the logic the numerator is always substantially greater than the denominator.

PR #1914, opened 2016-10-28
[video_capture] recovered saving of exif information

wilcoschoneveld:
This PR recovers the video_exif functionality that was left out in the computer vision pipeline update.

### DIFF
--- a/conf/modules/video_capture.xml
+++ b/conf/modules/video_capture.xml
@@ -3,12 +3,15 @@
 <module name="video_capture" dir="computer_vision">
   <doc>
     <description>
-      Capture images from video device on the internal memory (JPEG, full size, best quality)
+      Capture images from video device on the internal memory (JPEG, full size, best quality). Images can be saved by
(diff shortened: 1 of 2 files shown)


flixr:
Nice, please add the new define to the doc section of the module...

wilcoschoneveld:
I believe the intended usage of the video_exif module is to add it to the airframe and it will place the define itself, so that the video_capture module responds accordingly. I'm not sure if there are guidelines for this kind of behavior but I added a doc line anyways.

dewagter:
I do not think adding a non-existing define to docs is a good idea. I'd suggest to add the description to the docs instead: "To save EXIF data, include the exif module", of have the define (probably a configure then?) auto-include the module if that is at all possible.

flixr:
The define exists... but yes it is automatically set if you add the video_exif module.
So I agree with @dewagter here that a comment in the description makes more sense.
Also a comment in the code where this define is checked would be nice (to indicate that the user is not expected to set this define, but that it is added if the video_exif module is loaded.

wilcoschoneveld:
Thanks for the pointers.

PR #1916, opened 2016-10-28
update INDI tu use full control effecitvens matrix

EwoudSmeur:
Instead of using commands (roll, pitch, yaw, thrust), give the
controller direct control over all actuators.
This allows for online individual actuator control effectiveness
estimation and inversion, as well as better saturation handling.
The INDI controller as it was previous is renamed to 'simple'.

I think this is an essential step in making a more complete controller.
This will allow us to do much better control allocation, for which we already did successful experiments.
I will still update the code to make it look a bit better, for instance the hardcoded matrices should go of course.
But perhaps you can already see if you like how the controller is interfacing with the actuators instead of commands, or if there is a better way.


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -35,139 +35,149 @@
...
+  stabilization_indi_second_order_filter_init(&indi_rate, STABILIZATION_INDI_FILT_OMEGA, STABILIZATION_INDI_FILT_ZETA, STABILIZATION_INDI_FILT_OMEGA_R, 3);
+  stabilization_indi_second_order_filter_init(&actuator_state_filt, STABILIZATION_INDI_FILT_OMEGA, STABILIZATION_INDI_FILT_ZETA, STABILIZATION_INDI_FILT_OMEGA_R, ACTUATORS_NB);
+  stabilization_indi_second_order_filter_init(&estimation_input, IDENTIFICATION_INDI_FILT_OMEGA, IDENTIFICATION_FILT_ZETA, IDENTIFICATION_INDI_FILT_OMEGA, ACTUATORS_NB);
+  stabilization_indi_second_order_filter_init(&estimation_output, IDENTIFICATION_INDI_FILT_OMEGA, IDENTIFICATION_FILT_ZETA, IDENTIFICATION_INDI_FILT_OMEGA, 3);
 #if PERIODIC_TELEMETRY
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_STAB_ATTITUDE_INDI, send_att_indi);
 #endif
 }
 
+#define VECT_ASSIGN(_v, _b, _n) {  \
```
podhrmic: aren't these functions already defined in pprz_math folder?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -35,139 +35,149 @@
...
+                                   &G1[1][0],
+                                   &G1[1][1],
+                                   &G1[1][2],
+                                   &G1[1][3],
+                                   &G1[2][0],
+                                   &G1[2][1],
+                                   &G1[2][2],
+                                   &G1[2][3],
+                                   &G1[3][0],
+                                   &G1[3][1]);
```
podhrmic: Maybe better to send an array (and then have user to parse the values) instead of sending individual elements? But indeed this is probably very minor.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -205,96 +215,94 @@ void stabilization_indi_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t heading)
...
-#endif
-#if STABILIZATION_INDI_FILTER_YAW_RATE
-  rates_for_feedback.r = indi.rate.x.r;
-#endif
+  float avg_u_in = (indi_u[0] + indi_u[1] + indi_u[2] + indi_u[3])/4.0;
 
-  indi.angular_accel_ref.p = indi.reference_acceleration.err_p * QUAT1_FLOAT_OF_BFP(att_err->qx)
-                             - indi.reference_acceleration.rate_p * rates_for_feedback.p;
+#warning "saturation at 8100 instead of max_pprz because of bebop motors"
+  Bound(stabilization_cmd[COMMAND_THRUST],0, 8100);
```
podhrmic: This shouldn't be here - we want INDI to work on other quads too...

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -0,0 +1,387 @@
...
+#include "subsystems/datalink/telemetry.h"
+
+static void send_att_indi(struct transport_tx *trans, struct link_device *dev)
+{
+  //The estimated G values are scaled, so scale them back before sending
+  struct FloatRates g1_disp;
+  RATES_SMUL(g1_disp, indi.est.g1, INDI_EST_SCALE);
+  float g2_disp = indi.est.g2 * INDI_EST_SCALE;
+
+  pprz_msg_send_STAB_ATTITUDE_INDI(trans, dev, AC_ID,
```
podhrmic: Make sure the message is consistent with the one defined in new INDI, or use a different message

podhrmic:
@EwoudSmeur can you provide some use-case in which this controller is better than the simple-INDI (which I happen to quite like:)? Then it will be easier for me to see the difference.

And two additional questions:
1) can you provide some script that will estimate the initial values of the control matrix G (in the paper you mention it is calculated from the flight log)
2) can this controller be easily made more generic - i.e. to work for hexacopters too? Otherwise there should be some warning against compiling it for rotorcrafts with `ACTUATORS_NB != 4`

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
EwoudSmeur: We have vector functions specifically for vectors with 2 or 3 elements, but not for vectors of arbitrary length. I can also put these macros in the math folder

gautierhattenberger: we have at least for float vectors: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/math/pprz_algebra_float.h#L501

EwoudSmeur: oh nice, had not seen those yet

EwoudSmeur:
@podhrmic Thanks for the review! Right now the advantage that this controller would offer is twofold:
1) You can estimate the effectiveness of each actuator online, including the inversion. So you don't depend on the inversion you have to do yourself with the motor mixing matrix before you fly.
2) You can use actuator feedback.
However, I think the biggest advantage is that you can do a lot better control allocation in case of saturation. Since the control effectiveness for yaw is typically quite low, a lot of control effort is put in to it. If this leads to saturations, the pitch and roll may suffer, while these are more important than yaw. I will make a pull request for that once this is finished.

- Yes, I will provide a script.
- Indeed I want to make it such that it also works for multirotors with more actuators.

EwoudSmeur:
I still want to add/change some more stuff, but this PR will get quite big... Should we merge this first and then I make a second PR, or shall I add everything to this PR?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -447,82 +446,80 @@ void lms_estimation(void) {
...
-        // The yaw axis typically needs slower learning, perhaps due to large inputs
-        mu_scaling = 0.3;
-        ddx_error = ddx_error + G2[0]*ddu_estimation[0] + G2[1]*ddu_estimation[1] + G2[2]*ddu_estimation[2] + G2[3]*ddu_estimation[3];
-        for(int8_t j=0; j<4; j++) {
-          calc_g2_element(ddx_error,j,mu2);
-        }
-      }
-
-      // Calculate the row of the G1 matrix corresponding to this axis
+  for(int8_t i=0; i<3; i++) {
```
podhrmic: hardcoded value of 3?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -447,82 +446,80 @@ void lms_estimation(void) {
...
+    float mu_scaling = 1.0;
+    float ddx_error = G1[i][0]*du_estimation[0] + G1[i][1]*du_estimation[1] + G1[i][2]*du_estimation[2] + G1[i][3]*du_estimation[3] - estimation_rate_dd[i];
+
+    // when doing the yaw axis, also use G2
+    if(i==2) {
+      // The yaw axis typically needs slower learning, perhaps due to large inputs
+      mu_scaling = 0.3;
+      ddx_error = ddx_error + G2[0]*ddu_estimation[0] + G2[1]*ddu_estimation[1] + G2[2]*ddu_estimation[2] + G2[3]*ddu_estimation[3];
       for(int8_t j=0; j<4; j++) {
```
podhrmic: Maybe have a define G_NUMBER _OF_ROWS or so instead of 4 so it is more clear?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -62,6 +62,8 @@ struct ReferenceSystem reference_acceleration = {
 
 //only 4 actuators supported for now
 #define INDI_NUM_ACT 4
```
podhrmic: shouldn't we use ACTUATORS_NB instead?

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -578,3 +578,62 @@ void float_eulers_of_quat(struct FloatEulers *e, struct FloatQuat *q)
   e->theta = -asinf(dcm02);
   e->psi = atan2f(dcm01, dcm00);
 }
+
+/*
+ * 4x4 Matrix inverse.
+ * obtained from: http://rodolphe-vaillant.fr/?e=7
+ */
+float float_mat_minor_4d(float m[16], int r0, int r1, int r2, int c0, int c1, int c2);
```
podhrmic: wouldn't be better to have an algorithm for arbitrary size of matrix?

podhrmic:
Imho OK with a bigger PR - that way the new INDI is added in one merge:) Also, try to rebase on the master branch, that should fix the compilation errors.

EwoudSmeur:
I rebased on master, but the build fails because my message is not in pprzlink yet... Not sure if I should first try to merge it in pprzlink...

gautierhattenberger:
yes, we need to merge your PR in pprzlink and then you'll be able to update this one.

podhrmic:
@EwoudSmeur is this now ready to be merged?

EwoudSmeur:
I think so, but maybe I can still ask one of my colleagues to take a look at it...

podhrmic:
OK, was it flight tested?

EwoudSmeur:
Yes, I have flown this with a Bebop1

podhrmic:
@EwoudSmeur could you resolve the conflicts (I think mostly deleting those airframe configs)? Then I think it is good to merge. Thanks.

podhrmic:
Merging since semaphore failed only because it run out of execution time (see the build).

PR #1918, opened 2016-10-29
added mission index to mission module

rijesha:
It would be nice to associate a specific id to each mission that is sent to the mission module.
Currently all we get back is an array of the type of missions in the queue. 

By associating an ID number there is no confusion as to what missions are in the queue.

### DIFF
--- a/sw/airborne/modules/mission/mission_common.c
+++ b/sw/airborne/modules/mission/mission_common.c
@@ -100,16 +100,14 @@ struct _mission_element *mission_get(void)
 // Report function
 void mission_status_report(void)
 {
-  // build task list
-  uint8_t task_list[MISSION_ELEMENT_NB];
+  // build index list
   uint8_t index_list[MISSION_ELEMENT_NB];
   uint8_t i = mission.current_idx, j = 0;
   while (i != mission.insert_idx) {
-    task_list[j] = (uint8_t)mission.elements[i].type;
     index_list[j++] = mission.elements[i].index;
     i = (i + 1) % MISSION_ELEMENT_NB;
   }
-  if (j == 0) { task_list[j] = 0; index_list[j++] = 0; } // Dummy value if task/index list is empty
+  if (j == 0) { index_list[j++] = 0; } // Dummy value if index list is empty
   //compute remaining time (or -1. if no time limit)
   float remaining_time = -1.;
   if (mission.elements[mission.current_idx].duration > 0.) {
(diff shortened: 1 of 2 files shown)


rijesha:
pprzlink branch is here https://github.com/rijesha/pprzlink/tree/mission_ind

rijesha:
paparazzi/pprzlink#39

gautierhattenberger:
This is a very nice improvement, but at the moment, it is not possible to have two variable-length arrays in a message, at least when the C interface is used to access the elements. Since this mission system might be used by a C program, we need to choose if the arrays are change to fixed-length (not really nice, will always be ever too short or two long), or only keep one (the new index array is probably more useful).

podhrmic:
Would it help to combine the two arrays together and provide length of `taks_list` and length of `index_list` co the receiver can parse it correctly? Something like:
```
<field name="task_list_len" type="uint8"/>
<field name="index_list_len" type="uint8"/>
<field name="task_and_index_array" type="uint8[]"/>
```

rijesha:
The task_list_len and index_list_len would be the same length. From my understanding the ivy code would have to be rewritten to look for an array length. Also if you are sending one array it is fine. So you could have an array that has 2x the number of tasks on the autopilot and just alternate between type and index.

gautierhattenberger:
The "task" list only gives the type of pattern, and it was a fast and lazy debugging tool I choose when doing this interface. In particular you can't tell if two tasks have been swapped during transmission and you can't really associate for sure the parameters of each tasks.
With a list of index, you can check the correct order of tasks and you can associate them with the original parameters without errors (if you are the sender of the original commands at least).
So I would be in favor of keeping only the index. A future addition (unless someone wants to do it now) would be to add a uplink message to request the parameters of a specific task based on its index, and a downlink message to send the type and the parameters (in an array) for a given task.

rijesha:
Yes that would be a great addition. I'll look into it when I have a bit of free time. It would be nice to include a python parsing function for the new message as well. For now I think we should push what we have.

gautierhattenberger:
I'm merging now to unlock over PR that needs to update pprzlink.

flixr:
@gautierhattenberger not quite sure what you meant... in any case, pprzlink submodule is not updated yet and hence builds are failing...

gautierhattenberger:
My bad, I thought this PR was up to date regarding the correct pprzlink revision. Should be fixed now.

PR #1919, opened 2016-10-29
Guidance Vector Field

noether:
A guidance module based on vector fields: https://wiki.paparazziuav.org/wiki/Module/guidance_vector_field (wiki entry still to be improved)

After polishing (and hopefully merging) this commit I will work on the integration of the module in the GCS , right now it can only draw circles using the CIRCLE msg.

So far it only works with fixedwings. The extension to rotorcrafts should be easy to implement, but I have not looked at pprz's rotorcrafts code yet...


sw/airborne/modules/guidance/gvf/gvf.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef GVF_H
+#define GVF_H
+
+#define GVF_GRAVITY 9.806
+
+#include "std.h"
+
+// Control
+extern float gvf_error;
```
gautierhattenberger: This variables could be in a gvf structure

sw/airborne/modules/guidance/gvf/trajectories/gvf_ellipse.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/** \file gvf_ellipse.h
+ *
+ *  Guidance algorithm based on vector fields
+ *  2D Ellipse trajectory
+ */
+
+#ifndef GVF_ELLIPSE_H
+#define GVF_ELLIPSE
```
gautierhattenberger: it should be `GVF_ELLIPSE_H`

sw/airborne/modules/guidance/gvf/trajectories/gvf_ellipse.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "../gvf.h"
+
+extern float gvf_ellipse_a;
```
gautierhattenberger: maybe make a gvf_ellipse structure

sw/airborne/modules/guidance/gvf/trajectories/gvf_line.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+/** \file gvf_line.h
+ *
+ *  Guidance algorithm based on vector fields
+ *  2D straight line trajectory
+ */
+
+#ifndef GVF_LINE_H
+#define GVF_LINE
```
gautierhattenberger: it should be `GVF_LINE_H`

sw/airborne/modules/guidance/gvf/trajectories/gvf_sin.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+/** \file gvf_sin.h
+ *
+ *  Guidance algorithm based on vector fields
+ *  2D sinusoidal trajectory
+ */
+
+#ifndef GVF_SIN_H
+#define GVF_SIN
```
gautierhattenberger: `GVF_SIN_H`

sw/airborne/modules/guidance/gvf/gvf.h
```diff
@@ -0,0 +1,90 @@
...
+extern float gvf_ke;
+extern float gvf_kn;
+extern int8_t gvf_s;
+
+// Trajectories
+// 0 - Straight line
+// 1 - Ellipse
+// 2 - Sin
+
+extern uint8_t gvf_traj_type;
```
gautierhattenberger: I suggest to make an enum for the list of possible types

sw/airborne/modules/guidance/gvf/gvf.c
```diff
@@ -0,0 +1,308 @@
...
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+#include <math.h>
+#include "std.h"
+
+#include "gvf.h"
```
gautierhattenberger: usually, we have "global" path (from airborne code folder), so would be `modules/guidance/gvf/gvf.h`, and the others accordingly.

conf/modules/gvf_module.xml
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="gvf" dir="guidance/gvf">
+  <doc>
+    <description>Guidance Vector Field Module</description>
+  </doc>
+
+  <header>
+    <file name="gvf.h"/>
+	<file name="trajectories/gvf_line.h"/>
```
gautierhattenberger: fix indent (no tabs)

conf/settings/control/gvf.xml
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE settings SYSTEM "../settings.dtd">
+
+<settings>
```
gautierhattenberger: settings should be included in the module

noether:
yep, I will apply these changes. The structs that you suggest makes sense.

I will also split the settings. The essential ones in the module (the gains and direction) and optional ones (parameters of the trajectories) in the current file at /conf

gautierhattenberger:
Actually, you can have several `settings` in the same module that you can then choose to enable separately. See https://github.com/paparazzi/paparazzi/blob/master/conf/modules/cam_point.xml for an example. You just need to give a name to each settings section.

noether:
btw shall I add a gvf_demo.xml flight plan? so people can play with the settings from the GCS, etc.

conf/modules/gvf_module.xml
```diff
@@ -0,0 +1,62 @@
+< !DOCTYPE module SYSTEM "module.dtd" >
+
+<module name = "gvf" dir = "guidance/gvf">
+                           <doc>
```
gautierhattenberger: something went wrong with formatting here...

noether: haha, I see that the script for fixing the style messed the .xml up :P .

sw/airborne/modules/guidance/gvf/trajectories/gvf_ellipse.c
```diff
@@ -0,0 +1,76 @@
...
+  float px = p->x;
+  float py = p->y;
+  float wx = gvf_trajectory.p[0];
+  float wy = gvf_trajectory.p[1];
+  float a = gvf_trajectory.p[2];
+  float b = gvf_trajectory.p[3];
+  float alpha = gvf_trajectory.p[4];
+
+  // Phi(x,y)
+  float xel = (px - wx) * cosf(alpha) - (py - wy) * sinf(alpha);
```
gautierhattenberger: I suggest to make two local variables so that cos(alpha) and sin(alpha) are computed only once. Same for the sin guidance function.

noether: sure, I can do it. 

However (not 100% sure about it) I would say that the compiler takes care of such optimizations with the flag -O2 right? It checks how many times the same operation is done in the same scope and then make the corresponding optimization.

gautierhattenberger: Yeah, maybe we discussed that already. It still hurts a bit to see so many trig functions ;)

noether: haha yep. No worries, it does not take any effort to make this change. I will do it within tomorrow :P.

flixr:
Shouldn't we prefix the module name and the "nav" functions that are supposed to be called from the flightplan with `nav_` like with the other nav modules?

flixr:
The defines that the user is supposed to (optionally) add to the airframe file should be documented in the module xml file. A bit more detailed description in the module xml (maybe with link to the the wiki page) would also be nice.

noether:
True, I have forgotten to add the #defines in the .xml. I will add also an extra description there tomorrow.

I am not sure about the prefix nav_ . Maybe this is another discussion and for sure you know better than me how to proceed. Instead of having nav_circle, nav_gvf_ellipse, nav_foo_circle.... is there any plan for how to organize/call different algorithms for the same purpose? e.g. to follow a circle? what do you think?

gautierhattenberger:
Regarding the `nav_` prefix, this is actually a mix of nav and guidance. Initially we wanted to make a switch to choose between the CARROT guidance and GVF. But I finally agreed with Hector that the best way for now was to integrate as a module.
I think, the main issue comes (again) from the crappy navigation API.

Issue #1781, opened 2016-07-06
Autoload bebop sonar driver


kirkscheper:
Currently, the bebop sonar module has to be explicitly added to airframe files which is a bit unnecessary, it could be autoloaded with the Bebop. The thread to read the data would only be made in the init if USE_SONAR is defined in the airframe file.
- This can be done in the imu_bebop module when IMUs are converted to modules.
- Remove USE_SONAR definition from sonar_bebop module, should be added in airframe if it should be used as is the standard with other boards (e.g ARdrone).
- Additionally, should sonar drivers be in a separate sonar folder or just in the sensor folder?

PR #1920, opened 2016-10-31
autoload bebop sonar, thread only started if USE_SONAR defined

kirkscheper:
unifies functionality with ARDrone. Addresses #1781


gautierhattenberger:
LGTM

PR #1925, opened 2016-11-02
[abi] fix endless loop if AbiBindMsgX is called twice for same event

### LINKED ISSUE #1924: Calling AbiBindMsgRPM twice leads to unresponsive system
Noticed on a bebop1. After telnetting I could see that a printf in the callback gave a lot of output, but I did get any telemetry. When I make sure AbiBindMsgRPM is called only once I have no problem.

I know the function is not supposed to be called twice, but I also think it should not make the system unresponsive.

flixr:
AbiBindMsgX should be called only once per abi_event struct.
Regardless, this tries to make it a bit more robust if you re-use the same abi_event for consecutive AbiBindMsgX calls:
If AbiBindMsgX is called multiple times _consecutively_, the event/callback is now only added if it is different from the last one added...
This should at least fix the endless loop since the event->next can't point to itself anymore...

Should fix #1924 in a rudimentary way... but would still be possible to bind multiple times using the same abi_event if another bind for the same message happens in between.
Bottom line: make sure that your code doesn't call AbiBindMsg with the same abi_event struct multiple times!

### DIFF
--- a/sw/tools/generators/gen_abi.ml
+++ b/sw/tools/generators/gen_abi.ml
@@ -118,6 +118,7 @@ module Gen_onboard = struct
   let print_msg_bind = fun h msg ->
     let name = String.capitalize msg.name in
     Printf.fprintf h "\nstatic inline void AbiBindMsg%s(uint8_t sender_id, abi_event * ev, abi_callback%s cb) {\n" name name;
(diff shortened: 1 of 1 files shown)


EwoudSmeur:
That makes sense, thanks!

PR #1926, opened 2016-11-02
Lidar-Lite v1 module

podhrmic:
Lidar module version 1 (silver label)
https://github.com/pulsedlight3d

### DIFF
--- a/conf/modules/lidar_lite.xml
+++ b/conf/modules/lidar_lite.xml
@@ -1,0 +1,25 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="lidar_lite" dir="lidar">
+  <doc>
+    <description>Lidar-Lite v1 (silver label) from PulsedLight3D sensor connected over i2c bus</description>
+    <configure name="LIDAR_LITE_I2C_DEV" value="i2c2" description="I2C device to use for Lidar-lite"/>
+    <configure name="LIDAR_LITE_I2C_ADDR" value="0x62" description="slave address of Lidar-lite"/>
+  </doc>
+  <header>
+    <file name="lidar_lite.h"/>
+  </header>
+  <init fun="lidar_lite_init()"/>
+  <periodic fun="lidar_lite_downlink()" freq="10"/> <!-- for debug -->
+  <periodic fun="lidar_lite_periodic()" freq="50"/> <!-- poll data, theoretical max is 50Hz (every 20ms) -->
+  <makefile>
+    <configure name="LIDAR_LITE_I2C_DEV" default="i2c1" case="lower|upper"/>
+    <configure name="LIDAR_LITE_I2C_ADDR" default="0xC4"/>
+    <define name="LIDAR_LITE_I2C_DEV" value="$(LIDAR_LITE_I2C_DEV_LOWER)"/>
+    <define name="USE_$(LIDAR_LITE_I2C_DEV_UPPER)"/>
+    <define name="LIDAR_LITE_I2C_ADDR" value="$(LIDAR_LITE_I2C_ADDR)"/>
+    <file name="lidar_lite.c"/>
+  </makefile>
+</module>
+
+
--- a/conf/modules/lidar_lite.xml
+++ b/conf/modules/lidar_lite.xml
@@ -5,19 +5,22 @@
     <description>Lidar-Lite v1 (silver label) from PulsedLight3D sensor connected over i2c bus</description>
     <configure name="LIDAR_LITE_I2C_DEV" value="i2c2" description="I2C device to use for Lidar-lite"/>
     <configure name="LIDAR_LITE_I2C_ADDR" value="0x62" description="slave address of Lidar-lite"/>
+    <configure name="USE_LIDAR_LITE_AGL" value="1" description="use Lidar-lite for AGL measurements. On by deafult"/>
   </doc>
   <header>
     <file name="lidar_lite.h"/>
   </header>
   <init fun="lidar_lite_init()"/>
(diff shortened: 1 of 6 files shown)


conf/modules/lidar_lite.xml
```diff
@@ -0,0 +1,25 @@
...
+  <doc>
+    <description>Lidar-Lite v1 (silver label) from PulsedLight3D sensor connected over i2c bus</description>
+    <configure name="LIDAR_LITE_I2C_DEV" value="i2c2" description="I2C device to use for Lidar-lite"/>
+    <configure name="LIDAR_LITE_I2C_ADDR" value="0x62" description="slave address of Lidar-lite"/>
+  </doc>
+  <header>
+    <file name="lidar_lite.h"/>
+  </header>
+  <init fun="lidar_lite_init()"/>
+  <periodic fun="lidar_lite_downlink()" freq="10"/> <!-- for debug -->
```
gautierhattenberger: I suggest to add `autorun="FALSE"` so that debug message sending is only activated on demand

sw/airborne/modules/lidar/lidar_lite.c
```diff
@@ -0,0 +1,117 @@
...
+      /*
+      lidar.distance_raw = (uint16_t)((lidar.trans.buf[0] << 8) + lidar.trans.buf[1]);
+      lidar.distance = update_median_filter(&lidar_filter, (int32_t)lidar.distance_raw);
+      */
+      //lidar.distance_raw = (uint32_t)((lidar.trans.buf[0] << 8) | lidar.trans.buf[1]);
+      lidar.distance_raw = update_median_filter(&lidar_filter, (uint32_t)((lidar.trans.buf[0] << 8) | lidar.trans.buf[1]));
+      lidar.distance = ((float)lidar.distance_raw)/100.0;
+
+      // send message
+      AbiSendMsgAGL(AGL_LIDAR_LITE_ID, lidar.distance);
```
gautierhattenberger: I suggest to do something like https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/range_finder/teraranger_one.c#L127 in case the lidar is not used for AGL measurement (can still be activated by default)

conf/modules/lidar_lite.xml
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="lidar_lite" dir="lidar">
```
gautierhattenberger: The teraranger range finder have be placed in a folder `range_finder`. There is also a `sonar` folder. We should probably decide if we agree on a common folder for this type of sensors.

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -111,6 +111,10 @@
 #define AGL_TERARANGER_ONE_ID 5
 #endif
 
+#ifndef AGL_LIDAR_LITE_ID
+#define AGL_LIDAR_LITE_ID 5
```
gautierhattenberger: This should be ID 6

conf/modules/lidar_lite.xml
podhrmic: IMHO it makes sense to group the sensors based on their categories - i.e. lidars will be in `lidar` folder, sonars in `sonar` folder. If terraranger is a lidar then probably `lidars` (or `lidar`) folder would be logical?

sw/airborne/subsystems/abi_sender_ids.h
podhrmic: Good call. Fixed.

sw/airborne/modules/lidar/lidar_lite.c
podhrmic: Hmm, I am not a big fan of even more ifdef directives. I ll add a variable to switch it on/off so it can be controlled remotely (and not always enabled).

conf/modules/lidar_lite.xml
podhrmic: Good point, done. How would you activate it then?

conf/modules/lidar_lite.xml
gautierhattenberger: The teraranger is not using a laser source, but still based on light (http://www.teraranger.com/technology/). Can we still call that a lidar ?

conf/modules/lidar_lite.xml
gautierhattenberger: a setting will appear in the "modules" settings tab in the GCS (the last tab in settings).

sw/airborne/modules/lidar/lidar_lite.c
gautierhattenberger: that works as well with a variable

podhrmic:
Previously it failed because of wrong pprzlink commit, now it should be fixed.

conf/modules/lidar_lite.xml
podhrmic: Hmm, I guess there will never be a good category for all sensors:-/ Personally I would be OK with leaving terranger in a separate folder, since it is rather unique. Technically it is not a lidar since it doesn't use laser (https://en.wikipedia.org/wiki/Lidar). What is your preference?

podhrmic:
@gautierhattenberger does it look good as is?

conf/modules/lidar_lite.xml
gautierhattenberger: Keep things like they are for now.

OpenUAS:
A little closer to "the other" opensource AP naming maybe welcomed. IMHO it should not be the technology used but the function of a sensor, so rangefinder as base, where we can have lidar, sonar, IR and whatnot based sensors.

PR #1933, opened 2016-11-03
[no_wind_upload] 

dewagter:
- initial solution was to not send wind unless the (first) airframe firmware was fixedwing
 - alternative could be to check if AT LEAST 1 firmware is fixedwing
 - alternative could be to add a flag to link/server

please vote

### DIFF
--- a/sw/ground_segment/cockpit/live.ml
+++ b/sw/ground_segment/cockpit/live.ml
@@ -457,6 +457,10 @@ let create_ac = fun ?(confirm_kill=true) alert (geomap:G.widget) (acs_notebook:G
   let via_http = Str.string_match (Str.regexp "http") af_url 0 in
   let af_xml = ExtXml.parse_file ~noprovedtd:via_http af_file in
 
+  (** Get firmware name *)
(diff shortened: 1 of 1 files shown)


flixr:
Checking available firmwares in the airframe file is simply not robust... and you can already turn of sending of the WIND_INFO message in the misc page of the GCS notebook.
If you want to always disable it for a specific airframe, I suggest to add a define to the GCS section in the airframe file and parse it from the GCS (like e.g. the AC_ICON).

dewagter:
if the GCS can do this in another way, I propose to drop this PR.

gautierhattenberger:
I will add an option to disable it at startup anyway, I think it is missing.

dewagter:
that is great, thanks a lot

-Christophe

flixr:
You mean to disable it for all aircrafts?

Am 04.11.2016 11:22 vorm. schrieb "Christophe De Wagter" <
notifications@github.com>:

> that is great, thanks a lot
>
> -Christophe
>
> On Fri, Nov 4, 2016 at 11:17 AM, Gautier Hattenberger <

PR #1935, opened 2016-11-03
[nav] goto waypoint with heading towards the waypoint

dewagter:
set heading towards waypoint while flying a line in rotorcraft & call failsafe from flightplan

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
@@ -202,8 +202,8 @@ void guidance_hybrid_airspeed_to_attitude(struct Int32Eulers *ypr_sp)
 
     // 2) calculate roll/pitch commands
     struct Int32Vect2 hover_sp;
-    if (norm_sp_airspeed > (4 <<
-                            8)) { //if the setpoint is beyond 4m/s but the ref is not, the norm of the hover sp will stay at 4m/s
+    //if the setpoint is beyond 4m/s but the ref is not, the norm of the hover sp will stay at 4m/s
+    if (norm_sp_airspeed > (4 << 8)) {
       hover_sp.x = (guidance_hybrid_airspeed_sp.x << 8) / norm_sp_airspeed * 4;
       hover_sp.y = (guidance_hybrid_airspeed_sp.y << 8) / norm_sp_airspeed * 4;
     } else {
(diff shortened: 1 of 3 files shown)


sw/airborne/firmwares/rotorcraft/navigation.h
```diff
@@ -212,6 +222,9 @@ bool nav_check_wp_time(struct EnuCoor_i *wp, uint16_t stay_time);
   }
 
 #define NavSetManual nav_set_manual
+#define NavSetFailsafe { \
```
flixr: Would be better as `#define NavSetFailsafe nav_set_failsafe`
Or even better we start to get rid of the macros and use the functions directly...

sw/airborne/firmwares/rotorcraft/navigation.h
```diff
@@ -128,6 +130,14 @@ extern bool nav_set_heading_current(void);
     dist2_to_wp = get_dist2_to_waypoint(_wp); \
   }
 
+#define NavGotoWaypointHeading(_wp) { \
```
flixr: There is no good reason why this should be a macro instead of a static inline function...

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -569,13 +569,24 @@ bool nav_set_heading_towards_waypoint(uint8_t wp)
...
+}
+
 /** Set heading to the current yaw angle */
 bool nav_set_heading_current(void)
 {
   nav_heading = stateGetNedToBodyEulers_i()->psi;
   return false;
 }
 
+bool nav_set_failsafe(void) {
```
flixr: If the return value is always false anyway, I would be in favor of making it a void function and requiring people to use `call_once` instead of `call`.

sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
```diff
@@ -202,8 +202,8 @@ void guidance_hybrid_airspeed_to_attitude(struct Int32Eulers *ypr_sp)
 
     // 2) calculate roll/pitch commands
     struct Int32Vect2 hover_sp;
-    if (norm_sp_airspeed > (4 <<
-                            8)) { //if the setpoint is beyond 4m/s but the ref is not, the norm of the hover sp will stay at 4m/s
+    //if the setpoint is beyond 4m/s but the ref is not, the norm of the hover sp will stay at 4m/s
+    if (norm_sp_airspeed > (4 << 8)) {
```
gautierhattenberger: If this bit shift is a POS_FLOAT_OF_BFP, the macro should be used, same for lines below.

sw/airborne/firmwares/rotorcraft/navigation.h
podhrmic: Yes please.

flixr:
Merging as it is now...
All the nav macros should be cleaned up in one go...

PR #1939, opened 2016-11-04
[speech] Speech RC-Lost

dewagter:
is it useful for anyone else to have speech on RC-Lost?

### DIFF
--- a/sw/ground_segment/cockpit/live.ml
+++ b/sw/ground_segment/cockpit/live.ml
@@ -883,6 +883,10 @@ let get_fbw_msg = fun alarm _sender vs ->
   let mode = PprzLink.string_assoc "rc_mode" vs in
   if mode = "FAILSAFE" then begin
     log_and_say alarm ac.ac_name (sprintf "%s, mayday, AP Failure. Switch to manual." ac.ac_speech_name)
+  end;
+  let rc = PprzLink.string_assoc "rc_status" vs in
+  if (rc = "LOST") || (rc = "REALLY_LOST") then begin
+    log_and_say alarm ac.ac_name (sprintf "%s, RC lost." ac.ac_speech_name)
   end
 
 let get_telemetry_status = fun alarm _sender vs ->


sw/ground_segment/cockpit/live.ml
```diff
@@ -883,6 +883,10 @@ let get_fbw_msg = fun alarm _sender vs ->
   let mode = PprzLink.string_assoc "rc_mode" vs in
   if mode = "FAILSAFE" then begin
     log_and_say alarm ac.ac_name (sprintf "%s, mayday, AP Failure. Switch to manual." ac.ac_speech_name)
+  end;
+  let rc = PprzLink.string_assoc "rc_status" vs in
```
gautierhattenberger: this value is already in `status` variable a few lines above

sw/ground_segment/cockpit/live.ml
```diff
@@ -883,6 +883,10 @@ let get_fbw_msg = fun alarm _sender vs ->
...
+  if (rc = "LOST") || (rc = "REALLY_LOST") then begin
+    log_and_say alarm ac.ac_name (sprintf "%s, RC lost." ac.ac_speech_name)
```
dewagter: actually, i do not know for sure, because we completely removed the warning about failsafe in our branch

flixr:
any updates here?

podhrmic:
@dewagter any updates here?

sw/ground_segment/cockpit/live.ml
OpenUAS: If a start is without RC ( so status RC=NONE) we should make sure we get no message. For the rest it is a very good addition IMHO. Extra parameter in GSC to disable it welcomed...

OpenUAS:
Propose to honor the Pull, and if it really distracts or annoys to much, a new enhancement issue can be made to add a parameter to GCS to enable it, e.g. -userclossmessage (Wih a shorter better naming ofcourse ;)

dewagter:
This is still desired, as it can improve safety.

kirkscheper:
@dewagter there are still some open issues with this PR, see the commets from @gautierhattenberger  above.

dewagter:
I think most speech phrases should be "event" based instead of periodic. The only exception might be the battery, but even there an event "15% left" "12% left" "8% left" is more meaningful than a periodic "battery low". I will see when I can find time to dive back into OCAML ...

PR #1940, opened 2016-11-04
[generator] Qt project generator with VPath fix

kevindehecker:
Replaces #1533. Description over there is still accurate. Two additions:
1. The user needs to manually set PAPARAZZI_HOME, PAPARAZZI_SRC and TARGET in the build environment after opening .creator file. 
2. The user needs to manually set the build and clean steps. Disable the check at "all" and put in "Make Arguments":   -C /home/houjebek/paparazzi -f Makefile.ac AIRCRAFT=<name> <target>.compile. For example:  -C /home/houjebek/paparazzi -f Makefile.ac AIRCRAFT=Iris ap.compile

This needs to happen only once, after which QT will create a .creator.user file

### DIFF
--- a/Makefile.ac
+++ b/Makefile.ac
@@ -41,6 +41,12 @@ override TARGET=ap
 endif
 endif
 
+# Enable QT
+PAPARAZZI_QT_GEN?=
+ifneq (,$(findstring qt,$(MAKECMDGOALS)))
+PAPARAZZI_QT_GEN=1
+endif
+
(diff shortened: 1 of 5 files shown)


flixr:
Nice.
Can you please fix the indentation of the python files to 4 spaces?
Otherwise it looks good to me (although I haven't tested it).

kevindehecker:
Check

PR #1946, opened 2016-11-12
Vectornav as AHRS

podhrmic:
Use Vectornav as either INS (as before) or AHRS (for example for indoor applications)


podhrmic:
@gautierhattenberger can you merge this pls? Apparently I don't have the rights to do so:-/

PR #1949, opened 2016-11-16
[imu] add delay to start configuring mpu9250

gautierhattenberger:
According to the datasheet, the max startup time is 100 ms. If configuration starts too early, it might not be correct.

### DIFF
--- a/sw/airborne/peripherals/mpu9250.c
+++ b/sw/airborne/peripherals/mpu9250.c
@@ -28,9 +28,6 @@
  */
 
 #include "peripherals/mpu9250.h"
-#include "mcu_periph/sys_time.h"
-
-#define MPU9250_REG_STARTUP_TIME 100 // max registery read/write startup time in ms (from datasheet)
 
 void mpu9250_set_default_config(struct Mpu9250Config *c)
 {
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
do you have some code for that already ?

dewagter:
@fvantienen what happened to the "keep retrying" code for the IMU?

gautierhattenberger:
We have 3 solutions for this:
- keep the proposed solution that is only an open-loop protection wait long enough, but don't check sanity after (which works if your sensor is plugged)
- implement the WHO_AM_I check as it is done in the mpu60x0_spi code (which will keep trying even without sensor)
- do both (which doesn't make sense if we keep the current implementation as is does the WHO_AM_I test before the timer check)

Cast your vote! (I'm lazy so I vote merge this until we need something better)

fvantienen:
I vote for option 2 as that one is consistent with the mpu6050. @dewagter that was the mpu6050.
But if no one has time to implement this I'might OK with merging this.

gautierhattenberger:
The same with the WHO_AM_I style, seems to work as well

flixr:
LGTM

PR #1950, opened 2016-11-16
Px4 imu fixes

kevindehecker:
The lsm303d accelero was misbehaving in pprz. Rewrote the driver and cleaned up deprecated stuff.

Test flown with Iris, which now works well in att mode.


podhrmic:
LGTM. @gautierhattenberger any comments?

gautierhattenberger:
except for the strange configuration value you have mentioned above, it looks fine

kevindehecker:
Indeed, should have been deg/s. 120 also seems to works fine with my Iris, so I kept it like that for now. 
The slider range in GCS is often too small (especially for smaller quads), so enlarged the range while we are at it.

PR #1951, opened 2016-11-16
Kill switch now also works for fbw mode

kevindehecker:
I had a few close calls with the kill switch not working (with my Iris+ with Pixhawk and the 3dr frysky remote ) when it is in fbw mode, so...

Tested and works with the Iris+

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/main_fbw.c
+++ b/sw/airborne/firmwares/rotorcraft/main_fbw.c
@@ -244,7 +244,17 @@ static void fbw_on_rc_frame(void)
 {
   /* get autopilot fbw mode as set by RADIO_MODE 3-way switch */
   if (radio_control.values[RADIO_FBW_MODE] < (MIN_PPRZ / 2)) {
-    fbw_mode = FBW_MODE_MANUAL;
+
+#ifdef RADIO_KILL_SWITCH
+    if (radio_control.values[RADIO_KILL] < (MIN_PPRZ / 2)) {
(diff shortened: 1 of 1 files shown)


podhrmic:
@kevindehecker What do you mean by fbw mode? I noticed that no matter what my radio settings are, the MODE0 is always RC_DIRECT, but MODE1 and MODE2 can be changed to whatever I need. Does this take care of that too?

kevindehecker:
No, unfortunately not.

FBW mode (fly by wire) is a feature implemented on the Pixhawk, which (e.g. in case the AP crashes) enables the pilot to always override with RC control. This mode makes sense in case of e.g. fixed wing operation, which often don't necessarily require stabilization to be human-controllable. However, for a quadrotor such as the Iris+ it's useless, unless an IMU is added to the FBW processor such as in the Outback board. (and although the Pixhawk has two IMU's, both are connected to the AP)

The Pixhawk has an auxiliary processor (its got an STM32F1 FBW processor and an STM32F4 AP processor) to take care of the servo commands and RC receiver,  and a main processor which runs the AP and stuff. The RC commands are passed on to the AP by the FBW. And before that happens, the FBW interprets the RC commands in order to not depend on the AP. So, if the RC command says FBW, the AP cannot override.

Currently there is no define to disable RC-Direct ( -->"FBW mode") from an airframe xml. I agree we should add one.

podhrmic:
OK, merging as is and hoping for a way to disable the RC direct in the future PR.

PR #1953, opened 2016-11-21
Added define to disable rc-direct fbw mode manual

kevindehecker:
FBW mode (fly by wire) is a feature implemented on the Pixhawk, which (e.g. in case the AP crashes) enables the pilot to always override with RC control. This mode makes sense in case of e.g. fixed wing operation, which often don't necessarily require stabilization to be human-controllable. However, for a quadrotor such as the Iris+ it's useless, unless an IMU is added to the FBW processor such as in the Outback board. (and although the Pixhawk has two IMU's, both are connected to the AP)

The Pixhawk has an auxiliary processor (its got an STM32F1 FBW processor and an STM32F4 AP processor) to take care of the servo commands and RC receiver, and a main processor which runs the AP and stuff. The RC commands are passed on to the AP by the FBW. And before that happens, the FBW interprets the RC commands in order to not depend on the AP. So, if the RC command says FBW, the AP cannot override.

Currently there is no define to disable RC-Direct ( -->"FBW mode") from an airframe xml. This PR adds one.

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_iris_indi.xml
+++ b/conf/airframes/TUDELFT/tudelft_iris_indi.xml
@@ -52,6 +52,7 @@
       <define name="USE_KILL_SWITCH_FOR_MOTOR_ARMING" value="1" />
       <define name="RADIO_KILL_SWITCH" value="RADIO_KILL" />
     </module>
+    <define name="FBW_ENABLE_MANUAL" value="false"/>
     <!--  <module name="radio_control" type="spektrum">-->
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
Looks okay, except I don't really like the name `FBW_ENABLE_MANUAL`. I can't really find a really better name. Maybe `FBW_MODE_AUTO_ONLY` ?
You should add the documentation with this new setting in the code anyway.

kevindehecker:
Done

PR #1954, opened 2016-11-21
Added 2x3 way mode switch

kevindehecker:
In the FrSky remote that comes with the 3dr Irir+, the mode switch is mixed with the RTL switch (return to launch). The RTL switch negates the value of the mode switch.

This PR enables pprz to distill the mode unregardless of the STL switch position.

If anyone knows a better place for the RADIO_MODE_3x2 define, that would be great. Would be nice to have that in the rc xml instead of the airframe xml?

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_conf.xml
+++ b/conf/airframes/TUDELFT/tudelft_conf.xml
@@ -183,7 +183,7 @@
    telemetry="telemetry/default_rotorcraft_slow.xml"
    flight_plan="flight_plans/TUDELFT/tudelft_delft_basic.xml"
    settings="settings/rotorcraft_basic.xml settings/control/rotorcraft_guidance.xml [settings/control/stabilization_att_int_quat.xml] settings/control/stabilization_indi.xml"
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
I really don't see where this is different from the normal 3 way position. Their is only one channel (`RADIO_MODE`) being used.

kevindehecker:
I'm not sure what you mean, but this is the situation:

**RTL** has two positions (on/off)
**Mode** has three positions (M1/M2/M3)

Here's the output of the int16 mode ppm message:
**RTL	Mode	value**
Off	M1	-9500
Off	M2	-4800
Off	M3	-1850
On	M1	2100
On	M2	4900
On	M3	9600

So, the 3-way mode switch on this RC is incompatible with the normal 3-way in pprz, as the mode switch on this RC is mixed with the RTL switch (making it the equivalent of a 6-way switch). This PR solves the issue (basically by ignoring the RTL switch).

gautierhattenberger:
Ok, that's more clear now. You should then add the above explanation as a comment in the code.

kevindehecker:
Will do.
However, I wonder whether there is a way to have such a define in the radio xml instead of the airframe file?

gautierhattenberger:
I don't really see how you can do that easily from the channel description of the radio file

kevindehecker:
OK, then it'll be an air-frame setting I guess. 

I hope the comment is not too long here.

gautierhattenberger:
a comment is never too long!

PR #1955, opened 2016-11-24
Changed the behaviour of var in command laws

EwoudSmeur:
It used to be int16_t, but that easily overflows, for example in https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/examples/quadshot_asp21_spektrum.xml#L44

servo_value and command_value are already int32_t, so I don't see a reason
to stick to int16_t


flixr:
LGTM

PR #1960, opened 2016-12-07
Test with Matrix C++ library

podhrmic:
I want to test the C++ support for actually including C++ code. Indeed the compilation works fine if I *dont* use any C++ code, but that is not very helpful.

I am using a demo module with code borrowed from PX4/Matrix library https://github.com/PX4/Matrix - eventually I might want to use this code in `sw/ext` but for now lets assume this is OK (just ignore the files)

When I compile this code (AggeAir conf / Iris airfame) it fails at the linker script with:
```
/usr/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/lib/armv7e-m/softfp/libc.a(lib_a-sbrkr.o): In function `_sbrk_r':
sbrkr.c:(.text._sbrk_r+0xc): undefined reference to `_sbrk'
/usr/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/lib/armv7e-m/softfp/libc.a(lib_a-abort.o): In function `abort':
abort.c:(.text.abort+0xa): undefined reference to `_exit'
/usr/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/lib/armv7e-m/softfp/libc.a(lib_a-signalr.o): In function `_kill_r':
signalr.c:(.text._kill_r+0x10): undefined reference to `_kill'
/usr/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/lib/armv7e-m/softfp/libc.a(lib_a-signalr.o): In function `_getpid_r':
signalr.c:(.text._getpid_r+0x0): undefined reference to `_getpid'
collect2: error: ld returned 1 exit status
```

What I gathered from the web, it has something to do with libraries not being in the right order. @flixr you mentioned that g++ somehow reorders libraries, can we tell the linker which order it should use?

For ChibiOS (AggieAir conf/ Minion Lia) this fails with:
```
g++: error: unrecognized command line option '-mthumb'
...
arm-none-eabi-gcc: fatal error: no input files
```

Here is where the cpp files should be included: https://github.com/paparazzi/paparazzi/blob/cpp_support/conf/Makefile.chibios#L207 but I am not sure how to specify them - @gautierhattenberger ? Also I believe that some flags will have to be modified.

### DIFF
--- a/conf/airframes/AGGIEAIR/aggieair_conf.xml
+++ b/conf/airframes/AGGIEAIR/aggieair_conf.xml
@@ -7,7 +7,7 @@
    telemetry="telemetry/AGGIEAIR/aggieair_rotorcraft.xml"
    flight_plan="flight_plans/rotorcraft_basic_geofence.xml"
    settings="settings/rotorcraft_basic.xml settings/control/rotorcraft_guidance.xml settings/nps.xml"
-   settings_modules="modules/gps.xml"
+   settings_modules="modules/lidar_lite.xml modules/gps.xml"
    gui_color="#ffff954c0000"
   />
   <aircraft
@@ -29,7 +29,7 @@
    telemetry="telemetry/AGGIEAIR/aggieair_fixedwing.xml"
    flight_plan="flight_plans/AGGIEAIR/BasicTuning_Launcher.xml"
    settings="settings/fixedwing_basic.xml settings/nps.xml settings/control/ctl_basic.xml"
-   settings_modules="modules/gps.xml modules/nav_survey_poly_osam.xml modules/nav_skid_landing.xml"
+   settings_modules="modules/nav_skid_landing.xml modules/nav_survey_poly_osam.xml modules/gps.xml"
    gui_color="#00009e93ffff"
   />
 </conf>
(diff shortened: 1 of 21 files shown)


fvantienen:
These errors means you have system calls in the included code which are partly implemented in your libc.  This means you either need to implement the above functions or remove all system calls. Example system calls are 'printf', 'open', 'close' etc.

podhrmic:
Thanks @fvantienen - that might be a problem though because the matrix lib uses a couple of standard libraries, and probably won't work without them. I suppose it works on Pixhawk because they have Nuttx. Maybe ChibiOS would solve this issue?

flixr:
As @fvantienen already said, this seems to be mostly (only?) related to the system calls which our libc doesn't provide...
If you want to use a C++ linear algebra library, I would suggest [Eigen](http://eigen.tuxfamily.org)

noether:
I would also suggest to use Eigen. However, you will also need to implement some "dummy" syscalls, such as "abort" if you compile Eigen with some DEBUG flags or with the optimization flag -O0.

podhrmic:
Closing for now, as this should be addressed properly in some future PR.

podhrmic:
I ll keep the branch for now as a reference though.

PR #1961, opened 2016-12-08
localize global wayoints when init from flightplan

EwoudSmeur:
This makes sure that when the ltpdef is valid, the global waypoints will
be initialized. This is needed, because when waypoints_init is called, the ltpdef is not valid yet and nothing is initialized.

### DIFF
--- a/sw/airborne/subsystems/ins/ins_int.c
+++ b/sw/airborne/subsystems/ins/ins_int.c
@@ -53,6 +53,7 @@
 
 #include "math/pprz_geodetic_int.h"
 #include "math/pprz_isa.h"
+#include "subsystems/navigation/waypoints.h"
 
 
 #if USE_SONAR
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
This is an issue that we found with @dewagter during IMAV2016 but forgot to open an issue for this.
The actual problem is that when moving the INS from subsystems to modules, the init order have changed, it was before `autopilot_init` and now it is after. This `autopilot_init` is also initializing the navigation and the waypoints. This results in the described behavior (global WP are wrong without flight plan init) as the `localize_all_waypoint` needs an initialized INS to work.
So, your patch is indeed working, but only for this particular INS and it doesn't solve the underlying problem.
The solution to solve this is to move `autopilot_init` after `modules_init` after checking that there is no side effect. And at some point, the conversion of the navigation to modules should also solves the problem (if modules are loaded in the correct order: airframe file order for now, based on dependency someday).

EwoudSmeur:
Ok, thanks for the info Gautier! So shall I start testing if moving autopilot_init gives problems?

gautierhattenberger:
would be very nice if you do so !

flixr:
@gautierhattenberger how about just calling `waypoints_init()` again after `modules_init()` as a stopgap measure?

flixr:
Actually `waypoints_localize_all()` should be enough...

gautierhattenberger:
sure, that's a nice option for now

PR #1965, opened 2016-12-14
[ext] update to latest libopencm3

flixr:
update is needed to have basic support for STM32F41x

Libopencm3 does not yet have clock setup for the 100MHz, but running at 84MHz seems to work fine...


flixr:
Based on the changes in libopencm3, I don't expect any regressions...
Seems the only significant changes to what we already use were the ADC refactoring, which has been addressed...

@gautierhattenberger should we simply merge this or do you want to do more extensive tests first?

gautierhattenberger:
did you test it on a really board ?

flixr:
My Lisa/M and Lisa/MX work fine on my desk, but I haven't done any more tests than that...

gautierhattenberger:
tested on elle0 as well and working

PR #1966, opened 2016-12-14
[rotorcraft] make sure global waypoints are localized

flixr:
Stopgap measure to fix #1961
This is actually only needed if USE_INS_NAV_INIT is used (the default).


flixr:
@gautierhattenberger can we merge this then?

PR #1967, opened 2016-12-14
[navigation][flight plan] mandate call_once for functions that should only run once

flixr:
The `call` primitive of the flight plan can be used to repeatedly call a function until it returns false.
In most cases the function is only supposed to be called only once though...

This has led to a state where a lot of functions that are supposed to be called from the flight plan have an additional `return false` to make them `call`-able.

From just looking at the flight plan it is hard to determine if a function will indeed be only called once or is actually called each time the flight plan navigation is run until it is "done" (returns false).
To make this distinction more explicit and hopefully more clear (and the code cleaner), one should use `call` only when really needed and `call_once` otherwise.

This PR removes the "dummy" `return false` from a lot of functions, hence mandating that `call_once` is used for them. Eventually we shouldn't have any cases like that anymore...
All flight plans were updated accordingly.

The most common case where `call` still has to be used are navigation routines a la `nav_x_run`.

This should be "backwards safe" (not backwards compatible) in the sense that if you try use `call` for a function that needs to be called with `call_once` (since it now returns void), you will get an "invalid use of void expression" error:
```
In file included from firmwares/rotorcraft/navigation.c:40:0:
/home/flixr/code/paparazzi/var/aircrafts/Quad_LisaMX/ap/generated/flight_plan.h: In function 'auto_nav':
/home/flixr/code/paparazzi/var/aircrafts/Quad_LisaMX/ap/generated/flight_plan.h:129:9: error: invalid use of void expression
         if (! (NavKillThrottle())) {
         ^
```

### DIFF
--- a/sw/airborne/modules/nav/nav_survey_poly_rotorcraft.c
+++ b/sw/airborne/modules/nav/nav_survey_poly_rotorcraft.c
@@ -61,7 +61,7 @@ uint8_t Poly_Size = POLYSURVEY_DEFAULT_SIZE;
 float Poly_Distance = POLYSURVEY_DEFAULT_DISTANCE;
 
 
-bool nav_survey_poly_setup_towards(uint8_t FirstWP, uint8_t Size, float Sweep, int SecondWP)
+void nav_survey_poly_setup_towards(uint8_t FirstWP, uint8_t Size, float Sweep, int SecondWP)
 {
   float dx = waypoints[SecondWP].enu_f.x - waypoints[FirstWP].enu_f.x;
   float dy = waypoints[SecondWP].enu_f.y - waypoints[FirstWP].enu_f.y;
@@ -72,7 +72,7 @@ bool nav_survey_poly_setup_towards(uint8_t FirstWP, uint8_t Size, float Sweep, i
   //if values passed, use it.
   if (Size == 0) {Size = Poly_Size;}
   if (Sweep == 0) {Sweep = Poly_Distance;}
-  return nav_survey_poly_setup(FirstWP, Size, Sweep, ang);
+  nav_survey_poly_setup(FirstWP, Size, Sweep, ang);
 }
 
 struct Point2D {float x; float y;};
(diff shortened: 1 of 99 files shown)


sw/airborne/modules/nav/nav_survey_poly_rotorcraft.c
```diff
@@ -72,7 +72,6 @@ bool nav_survey_poly_setup_towards(uint8_t FirstWP, uint8_t Size, float Sweep, i
   //if values passed, use it.
   if (Size == 0) {Size = Poly_Size;}
   if (Sweep == 0) {Sweep = Poly_Distance;}
-  return nav_survey_poly_setup(FirstWP, Size, Sweep, ang);
```
gautierhattenberger: Shouldn't this function still be called before the return ?

flixr: ups, yes... good catch!

flixr: fixed..

flixr:
What do you guys think about this?

gautierhattenberger:
On the general idea, I'm fine with these changes

sw/airborne/modules/nav/nav_survey_poly_rotorcraft.c
```diff
@@ -155,7 +155,7 @@ bool nav_survey_poly_setup(uint8_t EntryWP, uint8_t Size, float sw, float Orient
   CSurveyStatus = Init;
 
   if (Size == 0) {
-    return true;
+    return;
```
flixr: I updated the run function to return without doing anything if the size is zero...

OpenUAS: The OK as for rotorcraft, quick glance fixedwing look OK, Thx. Note that The call_once (or call for that matter ) was discussed quite some time ago. For now to go forward i think it is fine. In future I rather see a nice non tech flightplan descriptive language and users writing flightplans should just be able to write e.g simply <SetGroundReferenceHere"/>. But that is another discussion. Yes I know, one could write a wrapper for nice flighplan language encapsulating current implementation.

BTW as for SurveySize returning if 0, well if I did not arrive at my survey area one could wish to set it as moving twarth gathering data or let it tepend on fuel reserve. now it cannot move toward But if ever needed it that way myself, I'll file a new issue ;)

BTW not testflown and simmed yet, will do, with finger on TX Mode switch ;) when it lnds in master and hopefully not -20 deg C.

PR #1968, opened 2016-12-14
[boards] add support for openpilot revo nano board

flixr:
- While the STM32F411 can run at up to 100MHz, this is not supported out-of-the-box with libopencm3 yet.
  So we use the 84MHz clock setup from the STM32F401.
- ADC pins not set up yet


gautierhattenberger:
Looks fine so far

flixr:
merging as it seems to work so far (except for missing ADC setup, additional PRs for that welcome :-)

PR #1969, opened 2016-12-14
[datalink] maximum MSG_SIZE is 256

flixr:
This cropped up some time ago when someone wanted to send a large array...

I don't see any reason to limit the maximum message size to 128 byte on the airborne side, since the protocol can handle up to 256 bytes....


flixr:
At least the airframes using the cc3d and naze32 still build fine...

PR #1972, opened 2016-12-17
only run adaptive throttle estimation if desired zd was updated

### LINKED ISSUE #1857: Adapted Throttle causes drop in guidance height

flixr:
Only run adaptive throttle estimation if we are in flight and
the desired vertical velocity (zd) was updated (i.e. we ran hover_loop before).
This means that the estimation is not updated when using direct throttle commands.

This should fix #1857

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
@@ -103,7 +103,7 @@ int32_t guidance_v_delta_t;
 
 float guidance_v_nominal_throttle;
 bool guidance_v_adapt_throttle_enabled;
-
+static bool desired_zd_updated;
 
 #define GUIDANCE_V_GUIDED_MODE_ZHOLD      0
 #define GUIDANCE_V_GUIDED_MODE_CLIMB      1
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
The nice thing about running adaptation in manual throttle is that it was easier to see if it works and if the convergence rate of the filter is correct. Should we add (yet another) flag to enable it in manual for debugging ?

flixr:
I don't see how it could ever properly work in manual mode... since you pass the "desired" vertical velocity to `gv_adapt_run`, but this is not set (and can't be known) in manual mode. It just has the last value or zero, but that will cause a wrong adaption if you are actually not trying to hover...

gautierhattenberger:
That was the idea, hover in manual to check that estimation is not doing something stupid before switching to altitude control where estimation and actual control are interacting each other. But it is not a big deal, you can merge like this if you want.

flixr:
> That was the idea, hover in manual to check that estimation is not doing something stupid

But if you don't hover and instead really fly around, it definitely will do something stupid, namely not properly estimate your hover throttle...

Issue #1973, opened 2016-12-19
Incorrect variable type in ABI messages


ETitus:
The AbiSendMsgOPTICAL in opticflow_module.c sends float opticflow_result.noise_measurement, however the ABI definition of AbiSendMsgOPTICAL_FLOW handles a uint8_t instead. 

This results in loss of information when logging this quality. I expected values between 0 and 1, but only received 0s and 1s.

I came across this issue when improving the OF algorithms.

flixr:
closed with #1974

PR #1974, opened 2016-12-19
fix loss of information due to quality variable type declaration in OF ABI

ETitus:
To fix issue #1973, I changed the abi.xml file to treat quality as a float instead of a uint8_t instead.

This currently seems the best option. 
Alternatively I could address the FIXME in the AbiSendMsgOPTICAL in opticflow_module.c:
"// FIXME, scale to some quality measure 0-255"

However, if we change the float noise_measurement in the struct opticflow_result_t to uint8_t noise_measurement to optimize for speed, while multiplying the relevant variables by 255, other code, i.e. the filtering in the edgeflow algorithm would break as float noise_measurement is used in calculations.

**NB** I also updated the type declaration in optical_flow_landing.c to prevent compile errors.

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -61,7 +61,7 @@
       <field name="flow_y" type="int16_t">Flow in y direction from the camera (in subpixels)</field>
       <field name="flow_der_x" type="int16_t">The derotated flow calculation in the x direction (in subpixels)</field>
       <field name="flow_der_y" type="int16_t">The derotated flow calculation in the y direction (in subpixels)</field>
-      <field name="quality" type="uint8_t"/>
+      <field name="quality" type="float"/>
       <field name="size_divergence" type="float">Divergence as determined with the size method (in 1/seconds)</field>
       <field name="dist" type="float">Distance to ground</field>
     </message>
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
The choice of using uint8 was most likely done to match mavlink quality factor for optic flow. I don't mind changing to float as long as all documentation and comments are updated accordingly.
@fvantienen or @guidoAI ?

flixr:
LGTM, but please also keep in mind that anywhere this noise is actually used in a filter it should be "converted" into a proper variance, meaning it doesn't have a value between 0-1 anymore...
A unit-less noise is not that helpful for most things in the end... you somehow need to scale that to a usable variance which is not unit-less anymore.

flixr:
Also there are other places where this needs to be updated accordingly, e.g. optical_flow_landing.c:vertical_ctrl_optical_flow_cb

ETitus:
@flixr The proposed solution only changes the abi message so it contains a float value between 0-1 instead of (int) 0 OR 1. I didn't scale it to 0-255 exactly because filters use the value as 0-1. Or is the scaling solution better and should that be implemented and changed in the filters too? @knmcguire?

As far as I know it is not used in other places, other than the ABI.xml and in optical_flow_landing.c:vertical_ctrl_optical_flow_cb(which I updated in the pull request too).

flixr:
Functionally there is no difference whether you scale 0-1 or 0-255, it is essentially still unit-less...
So it makes sense to use float here...

But in a Kalman filter the noise you use for a measurement is a variance and not 0-1, so if a filter uses this opticflow quality, it needs to convert it to an actual measurement noise variance...

Ah, sorry... somehow missed that you already updated the optical_flow_landing code.

flixr:
Can you also remove/or update the now outdated comment in opticflow_module.c?

PR #1977, opened 2016-12-29
[python] simple live mag plotter/calib

flixr:
Reads IMU_MAG_RAW messages from ivy and does an "online" min/max calibration and final calibration upon CTRL-C.
If the plot option is given, it plots the raw mag data and when the plot window is closed, calibration is performed


flixr:
Not the most sophisticated tool, but should do the job to calibrate the mag without having to start/stop the server and run the calibration on the log...

PR #1978, opened 2017-01-07
Control modules

gautierhattenberger:
Convert control, guidance and navigation subsystems to modules.

In most cases, the proper documentation and settings are included in the modules.
Fixedwing, rotorcraft and test config are updated. Other conf files are still using the old settings files but since they are not deleted (yet), they can still be used.

Some parts were automatically included (guidance and nav for rotorcraft) so as a temporary hack, it is autoloaded from stabilization modules. This can be improved in the future with dependency and autopilot generation systems.


conf/modules/stabilization_adaptive_fw.xml
```diff
@@ -0,0 +1,77 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="stabilization_adaptive" dir="stabilization">
+  <doc>
+    <description>
+      Adaptive attitude and lateral (heading) control for fixedwing aircraft.
+      Mostly based on PID and optional reference generators and feedforward gains.
+      Adaptive parameters should not be used for normal operation at the moment.
```
flixr: What do you mean by this?

gautierhattenberger: There is a kind of low pass filter to adapt the feedforward gains in flight. That's why it was called 'adaptive' in the first place. But it wasn't really working when I did this years ago and never touched it again...
I can remove the line if you think it is too much ambiguous.

gautierhattenberger:
I wasn't sure about removing the settings files as it will break personal confs as well. When keeping both, people will see new settings and that will "tell" them that they should update. We can remove them later before the release for instance.
I agree the other confs in the repo could be updated.

PR #1980, opened 2017-01-12
[modules] convert radio_control subsystems to modules

gautierhattenberger:
- calling init functions will be done later, it needs to be attached to the FBW part
- add 'cond' attribute option to more nodes in module makefile section


gautierhattenberger:
My idea for this was to first print the `configure` with a `value` (from modules and airframe file), then the board makefile, then the `configure` with `default` attribute.

flixr:
Jep, that would do the job :-)

PR #1981, opened 2017-01-12
[modules] convert FDM subsystems to modules

gautierhattenberger:


conf/modules/fdm_crrcsim.xml
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="fdm_crrcsim" dir="fdm">
+  <doc>
+    <description>
+      CRRCSIM FDM backend for NPS simulator
+
+      Only for fixed-wing.
+      CRRCSIM is running as an external program using the MNAV inputdev.
```
flixr: Can you add some extra information on what CRRCSIM is and how to start that.
Link to a wiki page would also suffice.

conf/modules/fdm_jsbsim.xml
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="fdm_jsbsim" dir="fdm">
+  <doc>
+    <description>
+      JSBSim FDM backend for NPS simulator
+
+      Use the paparazzi-jsbsim package if it is installed, otherwise look for JSBsim under /opt/jsbsim
```
flixr: Link to JSBSim and NPS wiki page would also be nice here...

conf/modules/fdm_jsbsim.xml
```diff
@@ -0,0 +1,32 @@
...
+    </description>
+    <configure name="JSBSIM_ROOT" value="/opt/jsbsim" description="set root directory for JSBSim library"/>
+    <configure name="JSBSIM_PKG" value="yes|no" description="enable or disable using pkg-config to get library flags"/>
+  </doc>
+  <header/>
+  <makefile target="nps">
+    <configure name="JSBSIM_ROOT" default="/opt/jsbsim"/>
+    <configure name="JSBSIM_INC" value="$(JSBSIM_ROOT)/include/JSBSim"/>
+    <configure name="JSBSIM_LIB" value="$(JSBSIM_ROOT)/lib"/>
+    <configure name="JSBSIM_PKG" value="$(shell pkg-config JSBSim --exists && echo 'yes')"/>
```
flixr: I think this should use `default` instead of `value`

conf/modules/fdm_jsbsim.xml
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="fdm_jsbsim" dir="fdm">
+  <doc>
+    <description>
+      JSBSim FDM backend for NPS simulator
+
+      Use the paparazzi-jsbsim package if it is installed, otherwise look for JSBsim under /opt/jsbsim
+    </description>
+    <configure name="JSBSIM_ROOT" value="/opt/jsbsim" description="set root directory for JSBSim library"/>
```
flixr: description should make it more clear that this is only used if `JSBSIM_PKG` is `no`

conf/modules/fdm_jsbsim.xml
```diff
@@ -0,0 +1,32 @@
...
+    <configure name="JSBSIM_PKG" value="yes|no" description="enable or disable using pkg-config to get library flags"/>
```
flixr: maybe we can actually rename this to something more meaningful like `FIND_JSBSIM_VIA_PKG_CONFIG`, which would default to yes...?

PR #1985, opened 2017-01-15
Actuators modules

gautierhattenberger:
Convert actuators subsystems to modules.

Some drivers (4015, 4017, pwm) makefiles are kept so that autoload for some boards is still working.


gautierhattenberger:
I'm planning to make board (+ firmware and target ?) modules as well. They would be loaded based on the selected target and would depend on the necessary modules (actuators, baro, ....).

PR #1990, opened 2017-01-19
[modules] simulator modules (SIM and NPS)

gautierhattenberger:
Preparing simulator modules. The existing makefiles are kept for now.


flixr:
What is the reason/benefit of keeping the makefiles?

gautierhattenberger:
I needs the dependency system to be used properly. I have started working on this with a colleague, but it will take some time. I just want to prepare the modules for a later use. As soon as it is working, I'll remove the old makefiles.

PR #1992, opened 2017-01-27
[board] add support for Parrot Disco plane

gautierhattenberger:
- mostly based on Bebop2
- add PWM output and SBUS input on Linux (based on ArduPilot work)
- some changes to the Linux serial port lib to handle arbitrary baudrate
  (this code could be improved and factorized in the future)
- all sensors are working, the video system is not tested

### DIFF
--- a/sw/airborne/arch/linux/mcu_periph/pwm_sysfs.c
+++ b/sw/airborne/arch/linux/mcu_periph/pwm_sysfs.c
@@ -47,7 +47,7 @@
 // Write in file
 static int write_file(const char *path, const char *fmt, ...);
 // Read file content
-//static int read_file(const char *path, const char *fmt, ...);
+static int read_file(const char *path, const char *fmt, ...) __attribute__((unused));
 
 int pwm_sysfs_init(struct PWM_Sysfs *pwm, char *base_path,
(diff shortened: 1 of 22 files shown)


fvantienen:
Looks good to me

sw/airborne/arch/linux/mcu_periph/pwm_sysfs.c
```diff
@@ -0,0 +1,179 @@
...
+  va_end(args);
+
+  if (ret < 1) {
+    return -errno_bkp;
+  }
+
+  return ret;
+}
+
+/*
```
flixr: whats up with this commented out code?

gautierhattenberger: original code had getter functions for the different parameters (by reading values in the system files). I commented this part to remove an unused warning and forgot it. If you think it is not useful for future improvements (checking status) I can remove it.

flixr: Then I would rather use a pragma to silence the warning and keep the code there...

gautierhattenberger:
done

PR #1993, opened 2017-01-30
Fix for buidling indi_rate

EwoudSmeur:
With this commit, you can only use indi rate, when you are using
stabilization indi or indi_simple.

This should solve the build problems.


flixr:
I guess it would be good then to add it as a dependency:
`<depends>stabilization_indi|stabilization_indi_simple</depends>`

EwoudSmeur:
Thanks Felix, good idea!

flixr:
Thx!

PR #1994, opened 2017-01-30
script for offline estimation of INDI parameters

EwoudSmeur:
For indi_simple, with an example file logger


gautierhattenberger:
I think the name of the file `estimation_script.m` is not reflecting the purpose of the script.

In a more general way, the file logger should not hard-code the data to write.

EwoudSmeur:
Good points, thanks!

PR #1997, opened 2017-01-31
Fix compile error when not using intermcu with px4flash module

kevindehecker:
If anyone knows a better solution...?

Basically I need to undefine the PX4IO stuff in case of a pixracer.

### DIFF
--- a/conf/modules/px4_flash.xml
+++ b/conf/modules/px4_flash.xml
@@ -1,10 +1,9 @@
 <!DOCTYPE module SYSTEM "module.dtd">
-
 <module name="px4_flash">
   <doc>
     <description>
-      Flash pixhawk via PX4 bootloader.
-      Flashes the PX4IO (STM32F1) through the PX4 bootloader, or resets the PX4FMU (STM32F4) to be flashed directly by the F4' PX4 bootloader.
+      Flash Pixhawk via PX4 bootloader.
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
You can use the `cond` attribute to disable some parts if needed. See https://github.com/paparazzi/paparazzi/blob/master/conf/modules/actuators_spektrum.xml#L26 for example.

kevindehecker:
Almost, I'm attempting to use `cond="ifdef INTER_MCU_AP"`, however when compiling for the Pixhawk (e.g. Iris) which has INTER_MCU_AP in ap.config defined, it doesn't work. 

Apparently this define is not known in the xml. What do I need to include to get access to it? In the spektrum example the define is configured from the airframe xml, but in that case we can better just revert commit f8b4b175d099985496a983dacd67b38452aaee53 from @podhrmic and configure the uart itself from the airframe xml such as what was done in the earlier tmp-fix commit e56c96c64e79d026038121865152e88c9458ac2e.

gautierhattenberger:
With this `ifdef` you must refer to a makefile variable (e.g. coming from a `configure` and not a compilation flag (coming from a `define`). You can have a look at the generated makefile in `var/aircrafts/<AC>/Makefile.ac`.

kevindehecker:
Yes, thanks figured it out. Ready for merge! (it fixes the Splash airframe compile problem in master)

PR #1998, opened 2017-02-01
Generic rotation matrix for mounted stereocamera

knmcguire:
This pull request adds a rotation matrix to the stereocam module group. The stereocam in combination with the lisa-s, the new lisa MXs, and the ardrone are usually mounted on there in different direction (facing backwards, forwards and downwards). Before, we had to constantly double check the direction of the velocities to transform that manually to body fixed coordinates, which was very prone to mistakes. Same goes with the derotation of optical flow on the stereocam. 

This addition should make it easier and more generic for platforms with the mounted stereocamera.


sw/airborne/modules/stereocam/stereocam2state/stereocam2state.c
```diff
@@ -56,48 +72,70 @@ void stereocam_to_state(void)
...
 
-  // Derotate velocity and transform from frame to body coordinates
-  // TODO: send resolution directly from stereocam
+//Rotate veloci back to quad's frame
+  struct FloatVect3 quad_body_vel;
+  struct FloatRMat stereocam_to_body;
 
-  float vel_body_x = - vel_x;
-  float vel_body_y = vel_y;
+  float_rmat_inv(&stereocam_to_body, &body_to_stereocam);
```
kirkscheper: stereocam_to_body desn't seem to be used.

sw/airborne/modules/stereocam/stereocam2state/stereocam2state.c
```diff
@@ -56,48 +72,70 @@ void stereocam_to_state(void)
...
+                             );
+
+  // Reusing the OPTIC_FLOW_EST telemetry messages, with some values replaced by 0
   uint16_t dummy_uint16 = 0;
   int16_t dummy_int16 = 0;
   float dummy_float = 0;
 
-  DOWNLINK_SEND_OPTIC_FLOW_EST(DefaultChannel, DefaultDevice, &fps, &dummy_uint16, &dummy_uint16, &flow_x, &flow_y, &dummy_int16, &dummy_int16,
-		  &vel_x, &vel_y,&dummy_float, &dummy_float, &dummy_float);
+  DOWNLINK_SEND_OPTIC_FLOW_EST(DefaultChannel, DefaultDevice, &fps, &dummy_uint16, &dummy_uint16, &flow_x, &flow_y,
```
kirkscheper: looks like the formatting messed up here.

sw/airborne/modules/stereocam/stereocam.c
```diff
@@ -76,6 +111,9 @@ extern void stereocam_disparity_to_meters(uint8_t *disparity, float *distancesMe
 
 extern void stereocam_start(void)
 {
+  struct FloatEulers euler = {RadOfDeg(STEREO_BODY_TO_STEREO_PHI), RadOfDeg(STEREO_BODY_TO_STEREO_THETA), RadOfDeg(STEREO_BODY_TO_STEREO_PSI)};
```
flixr: For defines representing angles it is actually preferred if they are in radians. You can then easily specify them in degrees in your airframe file: `<define name="STEREO_BODY_TO_STEREO_PHI" value="90" unit="deg"/>`

knmcguire: Just to clarify: If I define it with _unit="deg"_ in the airframe file, I do not have to use the _RadOfDeg_ anymore in stereocam.c anymore?

flixr: Yes, it will read the value from the xml file as deg, but make the define in rad, see https://wiki.paparazziuav.org/wiki/Units

knmcguire: Alright clear, this should be implemented in my latest commits

PR #2000, opened 2017-02-07
Improved ChibiOS support

podhrmic:
- [x] minor updates in the ChibiOS makefile
- [x] add support for Lia_1.1_chibios
- [x] add support for Lisa M chibios
- [ ] add support for Pixhawk chibios
- [x] add support for VMS ECU chibios (documentation on the wiki TBD)


podhrmic:
Indeed not complete yet, but wanted to get the changes out already since I changed the makefile - does it work on Apogee @gautierhattenberger ?

gautierhattenberger:
It is working for Apogee. I just renamed a variable with a more explicit name.

gautierhattenberger:
@podhrmic Do you still have things to do for li[s]a and px4 boards ?

podhrmic:
Yes, I am on travels so I haven't had a chance to test it yet:( Hopefully by end of the week.

podhrmic:
FYI here is about the bootloader and ChibiOS: http://www.chibios.com/forum/viewtopic.php?f=3&t=417&start=10

flixr:
It would be good to not duplicate the default options like leds, uarts, actuators... maybe we should put them in `<board>_defaults.makefile` files and include them in the locm3 and chibios versions?

podhrmic:
I was thinking about that - but then I was like well, maybe the defaults are different for each flavor of the board? Lisa 2.0 vs Lisa 2.1 vs Lia? If we are reasonably sure that the defaults are the same then I would be happy merging them into one.

podhrmic:
@gautierhattenberger can we mark this as "initial support" and merge as is? The PX4 compiles and blinks a LED, nothing else yet (have to properly configure the IO in board.h), Lia/Lisa seems to work fine.
The reason for the merge is that I won't get to finish and test the boards properly until early March (have to attend to other things now) and it would be nice to have it in master before some additional refactoring happens.

conf/Makefile.chibios
```diff
@@ -39,7 +60,6 @@ Q=@
 RTOS_DEBUG ?= 0
 
 ifeq (,$(findstring $(RTOS_DEBUG),0 FALSE))
-$(info DEBUGGING ON)
```
gautierhattenberger: Why did you remove this info. I find it useful. Is it possible to at least keep the `DEBUGGING ON` ?

conf/flash_modes.xml
```diff
@@ -47,6 +47,7 @@
       <board name="li[s]?a_mx_.*"/>
       <board name="px4fmu_.*"/>
       <board name="elle*"/>
+      <board name="lia_.*"/>
```
gautierhattenberger: lia board should be handled by the general lisa regexp above. If it is not the case, better fix it.

gautierhattenberger:
If you can address the two comments, I'll merge this. FYI we are about to add the support for the STM32F7 based on ChibiOS for our new board. We wanted to fly with it a bit before, but the weather is terrible these days.

conf/Makefile.chibios
podhrmic: I found it annoying - since most of the time I was compiling I had the debugging off. Maybe have it display the message only if the debugging is ON? And make it more descriptive - such as " ChibiOS debug mode is ON" - sounds good?

conf/flash_modes.xml
podhrmic: that's what I thought, but frankly I didn't spend much time figuring it out - adding lia explicitly made the trick

podhrmic:
@gautierhattenberger I ll clean it up a little more and also take care of @flixr 's comment, hopefully today. Excited for F7 based autopilot - do you have more info about it?

gautierhattenberger:
See https://github.com/enacuavlab/paparazzi/tree/stm32f7 for the code. The board itself is a rather big one dedicated to scientific research and designed for easier integration.

conf/Makefile.chibios
podhrmic: Done - prints a message only when the DEbugging is ON

conf/flash_modes.xml
podhrmic: Confirmed - the regular expression `li[s]?a_[ms]?_.*` doesn't cover Lia boards. I don't think you can easily make a general enough regexpr, so I am happy with adding Lia board separately.

flixr: it should cover `lia_m_*` and `lia_s_*` boards... which one does it not cover?
But maybe it is easier adding them separately anyway...

PR #2001, opened 2017-02-07
Module with servo tester function

EwoudSmeur:
useful for identifying the response of a servo

### DIFF
--- a/conf/modules/servo_tester.xml
+++ b/conf/modules/servo_tester.xml
@@ -1,0 +1,36 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="servo_tester" dir="servo_tester">
+  <doc>
+    <description>
+      Servo tester
+      Has a setting to change the input to the servo in millisecond units.
+
+      You have to give the servo you want to test the name SERVO_TEST, and don't use
+      it in the command laws.
+
+      Additionally, you can make the servo do a series of step maneuvers, with increasing
+      amplitude. Tip: Open up the servo and measure the voltage on the potmeter with
+      a high speed recorder device (like a saleae Logic analyzer with analog inputs). This 
+      will let you model the response of the servo.
+    </description>
+    <define name="TIME_PER_DEFLECTION" value="0.8" description="Time between servo steps for the 'servo_run' feature"/>
+  </doc>
+  <settings>
+    <dl_settings NAME="Servo test">
+      <dl_settings NAME="Servo tester">
+        <dl_setting var="servo_test_val" min="1000" step="1" max="2000" module="modules/servo_tester/servo_tester" shortname="servo_input"/>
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/servo_tester/servo_tester.h
```diff
@@ -0,0 +1,39 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/**
+ * @file modules/ctrl/gain_scheduling.h
```
flixr: please fix doxygen

sw/airborne/modules/servo_tester/servo_tester.c
```diff
@@ -0,0 +1,76 @@
...
+    }
+  } else if(do_servo_run == true) {
+    /*Reset de servo run*/
+    do_servo_run = false;
+    servo_run_counter = 0;
+    servo_test_val = servo_mean;
+  }
+
+  Bound(servo_test_val, SERVO_SERVO_TEST_MIN, SERVO_SERVO_TEST_MAX);
+  Set_SERVO_TEST_Servo(servo_test_val); \
```
flixr: superfluous `\`

conf/modules/servo_tester.xml
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="servo_tester" dir="servo_tester">
+  <doc>
+    <description>
+      Servo tester
+      Has a setting to change the input to the servo in PPRZ units
```
flixr: Please add more description on how to use it, which servo etc..

sw/airborne/modules/servo_tester/servo_tester.c
```diff
@@ -0,0 +1,76 @@
...
+
+void servo_tester_periodic() {
+
+  /*Go up and down with increasing amplitude*/
+  if(do_servo_run && (servo_run_counter<8)) {
+    servo_test_val = servo_mean + servo_run_counter * servo_run_step * ((servo_run_counter % 2)*2-1);
+    servo_run_timer += 1;
+
+    /*Give 400 loop cycles time to reach the setpoint*/
+    if(servo_run_timer > 400) {
```
flixr: Would be more generic and portable to not hardcode the number of periodic cycles here.
E.g. use a time in sec and multiply it with `SERVO_TESTER_PERIODIC_FREQ` which is generated from the freq specified in the module xml.
You should also take out the freq attribute there, so it can easily be used on e.g. fixedwing airframes with lower periodic freq.

EwoudSmeur:
Thanks for your review Felix!

PR #2003, opened 2017-02-09
hybrid vehicles: gradually transition back from forward flight

EwoudSmeur:


podhrmic:
@EwoudSmeur can you provide more description? I.e. what is this good for and how to use it?

EwoudSmeur:
Ah yes of course. Before, if you wanted to go back from forward flight to hover, the pitch angle setpoint would jump back to hover. Some vehicles can handle this, but possibly not all. With this commit when you switch back the setpoint will gradually come up again, just like in the transition to forward flight.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -610,10 +608,16 @@ void guidance_h_from_nav(bool in_flight)
...
-  //Add 0.00625%
-  transition_percentage += 1 << (INT32_PERCENTAGE_FRAC - 4);
+  if(to_forward) {
+    //Add 0.00625%
+    transition_percentage += 1 << (INT32_PERCENTAGE_FRAC - 4);
+  }
+  else {
+    //Subtract 0.00625%
+    transition_percentage -= 1 << (INT32_PERCENTAGE_FRAC - 4);
+}
```
flixr: please fix indentation...

PR #2004, opened 2017-02-09
[indi] support rate mode

EwoudSmeur:
Also prevent unused parameter warning for the thrust callback


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -75,6 +75,10 @@ bool indi_use_adaptive = true;
 bool indi_use_adaptive = false;
 #endif
 
+#ifndef STABILIZATION_INDI_MAX_RATE
+#define STABILIZATION_INDI_MAX_RATE 6.0
```
flixr: doxygen comment for this would be nice, especially also documenting the unit this should have

PR #2005, opened 2017-02-09
[INDI] support for servos

EwoudSmeur:
Now you can add dynamics for servos next to the first order dynamics of
motors. Also, you can give a rate limit in PPRZ/loop units


OpenUAS:
Great!

Issue #1585, opened 2016-03-30
support multiple IMUs running simultaneously


flixr:
Currently there is no proper way to support multiple IMU drivers running simultaneously.
This would be useful for cases where you e.g. want to compare the data from different IMUs, maybe switch to one with less vibrations, switch between external and internal mag, etc...

The IMU drivers already publish the sensor data via ABI using their own ID and the subscribers (e.g. AHRS) can select which one to use.

The main remaining problems are the single global `imu` struct and single IMU calibration.
- [ ] get rid of global `imu` struct
  - each driver keeps it's own copy of the scaled measurements and publishes those via ABI
- [ ] each driver uses their own scale functions or at least their own IMU_x_SENS/NEUTRAL defines
- [ ] send telemetry IMU messages without using the global `imu` struct
  - means adding an IMU/component id to the messages to distinguish?
  - means a lot of duplicated send message code?
  - or subscribe to one specific (e.g. primary or currently active) IMU in `imu.c` and publish those?
- [ ] `BODY_TO_IMU` for each IMU, just like the sensitivity/neutrals
  - howto use `imu_SetBodyToImuCurrent` for each IMU separately?
- [ ] run ahrs_aligner for multiple IMUs separately?
##

gautierhattenberger:
done with #2890

PR #2008, opened 2017-02-12
MPU based IMUs: default scaling based on selected range

flixr:
Set default gyro/accel sensitivity based on selected range.


flixr:
This is not finished yet... and I'm unsure which approach is nicer:
- have one common default file like the `imu_mpu60x0_defaults.h`
- set the defaults in the header file for each IMU "implementation" (like e.g. for apogee and px4fmu)

The difference being that if we use a common defaults file, the range defines need to have the same name and it might not be consistent with the defines for the other IMU parameters anymore... but we have more duplication.

gautierhattenberger:
I would be more in favor of each IMU having its own definitions. If we finally manage to not rely anymore on the global imu struct (among other things) it will be easier to run several IMU at the same time and then it can be annoying to configure them with the same variables.

flixr:
Ok, done for MPU60X0...
When using multiple IMUs, the issue is actually not the global IMU struct anymore, but rather the scaling defines (if the IMU driver doesn't have it's own scaling implementation and uses the one from imu.c)...

flixr:
Ok, didn't remember correctly... the global IMU struct is still an issue, see #1585

flixr:
Anyway, this looks to merge for me (once it was tested).

EwoudSmeur:
Works on apogee with 16g and 2000deg/s scaling!

PR #2009, opened 2017-02-14
Autopilot refactor

gautierhattenberger:
With this, fixedwing and rotorcraft are mostly using the same interface for the autopilot. Some specific code and messages handling are still firmware dependent.
A large part of the autopilot logic of the fixedwing is moved from main_ap to autopilot_static.
More getter/setter functions are provided.
The rest of the systems, the simulators and the configurations are also updated. Unfortunately it does have a small impact on airframe configurations and flight plans when variables where directly used.
~~The functionality of the strip to go back to nav mode (when clicking on mode status) is disabled for now as FW and rotorcraft nav have different index.~~


OpenUAS:
WOW thx, yeah, merge FW and rotorcraft the sooner the better. Will see how I can quickly fix just move airframes... or @gautierhattenberger you can do it you want pull request honored  now

sw/airborne/firmwares/rotorcraft/main_ap.c
```diff
@@ -305,60 +305,58 @@ void failsafe_check(void)
...
 #if !USE_GENERATED_AUTOPILOT
   if (radio_control.status == RC_REALLY_LOST &&
-      autopilot_mode != AP_MODE_KILL &&
-      autopilot_mode != AP_MODE_HOME &&
-      autopilot_mode != AP_MODE_FAILSAFE &&
-      autopilot_mode != AP_MODE_NAV &&
-      autopilot_mode != AP_MODE_MODULE &&
-      autopilot_mode != AP_MODE_FLIP &&
-      autopilot_mode != AP_MODE_GUIDED) {
+      autopilot_get_mode() != AP_MODE_KILL &&
```
OpenUAS: Is there a chance "autopilot_get_mode()" is different between calls and would "autopilot.mode" as a value maybe be better here

conf/modules/actuators_dummy.xml
```diff
@@ -4,9 +4,12 @@
   <doc>
     <description>
       Dummy actuators driver
+      This module prevents autoloading of actuators by fixedwing firmware
     </description>
   </doc>
   <header/>
-  <makefile/>
+  <makefile>
+    <configure name="ACTUATORS"/>
```
flixr: Maybe we should rather use `<configure name="ACTUATORS" value="none"/>`?
That might be more clear and also consistent with the configure option we have for LEDs....

sw/airborne/firmwares/rotorcraft/main_ap.c
flixr: I don't understand your question...

gautierhattenberger: @OpenUAS it will have exactly the same behavior than before, it is just a simple getter function that doesn't change the variable. The only possible inconsistency is if you start changing the mode without proper protection in multi-threaded environment or in interrupts. It is not the case at the moment (and hopefully it will never be the case!).

conf/modules/actuators_dummy.xml
gautierhattenberger: Is there a nice way to test that `ACTUATORS` is either null, undefined or equal to `none` on this line: https://github.com/paparazzi/paparazzi/blob/master/conf/firmwares/fixedwing.makefile#L33 ?

gautierhattenberger: hum... I guess defaulting to `none` just before should be enough

sw/airborne/firmwares/rotorcraft/main_ap.c
OpenUAS: @gautierhattenberger, yes, that was indeed my worry, mode change with unintended effect. But since you also testfly a lot we both will discover and inconsistency soon ;-) . Thanks for the code integration, really appreciated :1st_place_medal:

gautierhattenberger:
tested and working on fixedwing (Apogee) and rotorcraft (elle0), using the static autopilot.

flixr:
What do we do about the functionality to get out of home mode by clicking on the strip?

gautierhattenberger:
With this the firmware should be automatically detected and if not, the button is not connected.

PR #2011, opened 2017-02-14
define an airspeed to use in coordinated turn calculation

EwoudSmeur:


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.c
```diff
@@ -41,6 +41,10 @@
 #define STABILIZATION_ATTITUDE_DEADBAND_E 0
 #endif
 
+#ifndef COORDINATED_TURN_AIRSPEED
```
flixr: A doxygen comment explaining this define would be good

EwoudSmeur:
I actually wanted to do that, but it must have slipped my mind

Issue #2014, opened 2017-02-17
[Mission] Replace next element


Fabien-B:
Replace next element without affecting the current one.

podhrmic:
@Fabien-B can you elaborate please?

gautierhattenberger:
closed in favor of the PR #2015

PR #2015, opened 2017-02-17
[mission] Adds insertion mode.

Fabien-B:
Adds ReplaceNexts insertion mode

Fixes issue #2014

### DIFF
--- a/sw/airborne/modules/mission/mission_common.c
+++ b/sw/airborne/modules/mission/mission_common.c
@@ -74,9 +74,9 @@ bool mission_insert(enum MissionInsertMode insert, struct _mission_element *elem
       mission.insert_idx = 1;
       break;
     case ReplaceNexts:
-        tmp = (mission.current_idx + 1) % MISSION_ELEMENT_NB;
-        mission.elements[tmp] = *element;
-        mission.insert_idx = (mission.current_idx + 2) % MISSION_ELEMENT_NB;
+      tmp = (mission.current_idx + 1) % MISSION_ELEMENT_NB;
+      mission.elements[tmp] = *element;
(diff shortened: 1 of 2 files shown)


Fabien-B:
It allows to replace the next mission element without affecting the current one.

podhrmic:
@Fabien-B can you elaborate please? Also, you can just push the changes into the current pull request (in this case #2014 instead of opening a new one)

Fabien-B:
This insertion mode could be used in a software made for controlling the drone in mission mode (a "dynamic planner" software).
It could be useful to cancel some missions send to the drone but not yet running, without affecting the current behaviour of the drone.

Lets say the drone is currently following a path between 3 waypoints WP1, WP2 and WP3.
The user of the dynamic planner already planned next missions blocks : lets say a circle around WP2 and a GoTo WP4.
While the drone is following the path, the operator change its mind, and wants to replace the next element by an other, for example a survey, or wants to remove the GoTo WP4.

Sending a mission block with REPLACE_ALL insertion mode will leads the drone to an uncertain trajectory, depending on where the drone currently is.

The REPLACE_NEXTS insertion mode could then be used.

sw/airborne/modules/mission/mission_common.c
```diff
@@ -73,6 +73,10 @@ bool mission_insert(enum MissionInsertMode insert, struct _mission_element *elem
       mission.current_idx = 0;
       mission.insert_idx = 1;
       break;
+    case ReplaceNexts:
+        tmp = (mission.current_idx + 1) % MISSION_ELEMENT_NB;
```
gautierhattenberger: can you follow the 2 spaces indentation

Fabien-B:
Sorry, identation should be ok now.

PR #2016, opened 2017-02-17
Robird

kevindehecker:
This PR adds Robird, which is a flappy wing vehicle resembling a falcon with a Pixhawk autopilot. 

1. It has some R7008SB SBUS2 receiver , which is different than some other SBUS2 receivers, which is different then SBUS receivers. To boring to go into, but apparantely PX4 has a better implementation so I borrowed some info there. This works fine for this equipment, but would be interesting to test on current sbus stuff.
2. The bird can glide pretty well, but only if the wing is locked in a certain position. I created a module, which reads a wing position sensor, which can lock the wing exactly right. However, this is done by setting the thrust to a fixed value just before attempting to lock it (there are some speed inertia dynamics to account for), and then after a rotation sets the thrust to 0 (which actively breaks the motor). I could not think of a nice way to do everything from the module code, so  I needed to change the autopilot a bit. It's working fine, but if anyone can think of something better...?

Its not been flight tested yet as that is scheduled for over two weeks, so I would like to add this to pprz already now.


conf/radios/T14SG_SBUS.xml
```diff
@@ -0,0 +1,60 @@
...
+-- function: logical command
+-- average: (boolean) channel filtered through several frames (for discrete commands)
+-- min: minimum pulse length (micro-seconds)
+-- max: maximum pulse length (micro-seconds)
+-- neutral: neutral pulse length (micro-seconds)
+-- Note: a command may be reversed by exchanging min and max values
+-->
+
+<!DOCTYPE radio SYSTEM "radio.dtd">
+<radio name="Futaba T10CG with SBUS" data_min="900" data_max="2100" sync_min ="5000" sync_max ="15000" pulse_type="POSTIVE">
```
gautierhattenberger: please set correct radio name

sw/airborne/boards/px4fmu_2.4.h
```diff
@@ -184,12 +184,34 @@
 /* allow to define ADC_CHANNEL_VSUPPLY in the airframe file*/
 #ifndef ADC_CHANNEL_VSUPPLY
 #define ADC_CHANNEL_VSUPPLY ADC_2
+#define DefaultVoltageOfAdc(adc) (0.00975*adc)
 #endif
 
-#if USE_ADC_2
-#define DefaultVoltageOfAdc(adc) (0.00975*adc)
-#else
-#define DefaultVoltageOfAdc(adc) (0.0021*adc) // scale internal vdd to 5V
```
gautierhattenberger: what happen to this case ?

sw/airborne/firmwares/rotorcraft/autopilot.c
```diff
@@ -74,6 +74,7 @@ bool   autopilot_power_switch;
 
 bool   autopilot_ground_detected;
 bool   autopilot_detect_ground_once;
+int lock_wings;
```
gautierhattenberger: I don't think it is the best position for this

sw/airborne/firmwares/rotorcraft/autopilot_static.c
```diff
@@ -173,8 +173,14 @@ void autopilot_static_periodic(void)
     SetCommands(commands_failsafe);
   } else {
     guidance_v_run(autopilot_in_flight);
-    guidance_h_run(autopilot_in_flight);
+    guidance_h_run(autopilot_in_flight);
```
gautierhattenberger: please remove trailing white spaces here and there

sw/airborne/firmwares/rotorcraft/autopilot_static.c
```diff
@@ -173,8 +173,14 @@ void autopilot_static_periodic(void)
     SetCommands(commands_failsafe);
   } else {
     guidance_v_run(autopilot_in_flight);
-    guidance_h_run(autopilot_in_flight);
+    guidance_h_run(autopilot_in_flight);    
     SetRotorcraftCommands(stabilization_cmd, autopilot_in_flight, autopilot_motors_on);
+
+    if (lock_wings == 1 && autopilot_motors_on && commands[COMMAND_THRUST] > 0) {
```
gautierhattenberger: I think this should either go above (set `stabilization_cmd` with the proper value in a control loop) or below (as part of the actuators mixing defined in the airframe file) in the autopilot architecture.

sw/airborne/modules/glide_wing_lock/glide_wing_lock.c
```diff
@@ -0,0 +1,101 @@
...
+#define WING_POS_LOCK_SWITCH RADIO_AUX2
+#endif
+
+void glide_wing_lock_init(void) {
+  adc_buf_channel(ADC_CHANNEL_MOTORSENSOR, &adcbuf, 1);
+}
+
+void glide_wing_lock_event() {    
+  static int lockstate = 0;
+  if (radio_control.values[RADIO_AUX2] > (MIN_PPRZ / 2)) { // check glide switch
```
gautierhattenberger: I suggest to make this configurable if someone doesn't have an AUX2 channel on his radio

sw/airborne/subsystems/radio_control/sbus_common.c
```diff
@@ -54,6 +54,124 @@
 #endif
 
 
+
+/*
+ * S.bus decoder matrix. (extracted from PX4)
```
gautierhattenberger: Can you provide instead some pointer on the SBUS2 protocol ? It looks hardly possible that this changes won't affect the support of SBUS1.
As a side not, an other pull request is also changing some elements of the decoding function (#2018), but only for efficiency, no extra protocol support.

kevindehecker: I tried to implement [this](https://www.google.nl/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&ved=0ahUKEwjq2r_JnKHSAhWKfhoKHQQ9Aq8QFggjMAE&url=http%3A%2F%2Fforum.autoquad.org%2Fdownload%2Ffile.php%3Fid%3D2319%26sid%3Db9bf8e54c8231cfb51691db4a3733673&usg=AFQjCNG4SJJZK_DtNhTvczfZfYZVOi5eKg&sig2=g37S5wm92t7mXeJK3YVMUg), but then I found the working implementation of PX4 / Ardupilot. From what I understand on fora it seems to work with all known sbus stuff, and it work for me at least.

The main thing is that the end byte is different between the different sbus receivers. For original sbus it's just 0x0, for later generations it varies (e.g. per 4 blocks). In this PR, the original 0x0 end byte is accepted like normal, so it should just work. I just miss the equipment to test it. 
SBUS2 is extra different in the sense that it provides telemetry, but I have not had the time to implement that yet fully (but it is on the todo list). The foundation is there now.

sw/airborne/firmwares/rotorcraft/autopilot_static.c
kevindehecker: I totally agree this is not the nicest solution. The thing is that setting zero thrust before here also seems to influence the other actuators (like ailerons). I really just want to set main motor to standstill and leave other control as is. Also, I like it for every controller (so, make it independent whether we use INDI or PID or whatever). This seemed to be the only place.

Ideally I would rather do this directly from the module somehow, but I could not find a way to do so. (as the event functions are called at the wrong moment, and there is no obvious way to influence that). This combined with extra AP->FBW complexity made me decide to do it here...

gautierhattenberger: It is true that in your case the AP/FBW split is annoying. Then I'm suggesting to make the `SetRotorcraftCommand` configurable. Currently there is 3 per-defined case, but we could put a `#ifndef SetRotorcraftCommand` around that (and include `modules.h` before) so that you can define your own set command function in a module. What do you think ?

sw/airborne/subsystems/radio_control/sbus_common.c
gautierhattenberger: thanks for the link, I'll have a look with my colleague who wrote our sbus driver.

gautierhattenberger: @kevindehecker could you test your SBUS2 radio with the `sbus2` branch of paparazzi repository (see https://github.com/paparazzi/paparazzi/commit/bba83260fffec4aa7c792ac48227d1366e047988) ? Thanks

kevindehecker: I actually tried what you've done over there before I made this, but it gave me incorrect results... (the received bytes seemed to be constantly changing in loops of 4 if I remember correctly)
Also, the current pprz implementation is quite messy regarding some of the protocol definitions. For example, SBUS_FLAGS_BYTE is set to 22, while it should be 23 (counted from 0). For some reason, pprz does not count the start byte in the package...
That initially really made it difficult to compare and understand what's going on. There is some more stuff like that which I encountered last week but don't remember from the top of my head.

Have you tried this on your original sbus equipment and found it faulty? I can try this again of course, but probably Friday at the earliest.

sw/airborne/firmwares/rotorcraft/autopilot_static.c
kevindehecker: Hmm, yes that sounds interesting...! If I am allowed to include modules here it can be much nicer.

sw/airborne/subsystems/radio_control/sbus_common.c
gautierhattenberger: I have updated the sbus2 branch. It now works for SBUS1, SBUS2 in 14CH mode and SBUS2 in 12CH mode with my R7008SB receiver and T14SG RC.
It is possible that some other end bytes will appear as it seems they are telling what is frame frame coming after the regular RC frame (voltage, gps, alti and so on). At some point we may wonder if we should care about this end byte or not as it is not a checksum and we don't use (for now) the other information.
@kevindehecker if you have the time would be nice if you can test it as well.

kevindehecker: OK I will Friday!

sw/airborne/firmwares/rotorcraft/autopilot_static.c
kevindehecker: Didn't test this yet (Friday)... but something like this I suppose?

I tried to include modules.h in autopilot_utils directly, but that gives all sorts of trouble :(

gautierhattenberger:
@kevindehecker for your implementation of the SetRotorcraftCommands, I suggest this one: 2803da37e758db875d51e51c07384a96db3c8018. With this you can just re-implement the `set_rotorcraft_commands` function, as the default implementation is defined WEAK. With this no need for extra includes of modules.h.

Also don't forget to merge with master and fix conflicts.

sw/airborne/subsystems/radio_control/sbus_common.c
kevindehecker: Works, numbers I get are a bit different. Whatever.

kevindehecker:
All done. Thanks again for the tips @gautierhattenberger !

gautierhattenberger:
closed with d585ee9b08feea229e321906a456743983a50a14, 63388b2d40395acb8dac81b57b3582afe6955101 and 9f74c89f3bfc01bfcba2373219d486143922a0b0

PR #2019, opened 2017-02-20
[autopilot] fix autopilot motor on var for arming on switch & throttle

masierra:
Some variables were left unchanged in in the autopilot refactor.


flixr:
Thanks!

PR #2027, opened 2017-02-25
add temperature control module for Bebop2

martinmm:
It is possible to control the temperature of the MPU6xxx in the Bebop2
through resistors powered by PWM_6. Seems this does not make a big
difference even if flying in cold conditions but does rather improve
things than do harm.


flixr:
Nice!
Naming wise I think it would be nicer to call it bebop_temp_ctrl or parrot_temp_ctrl as it doesn't provide (directly) INS related functionality.

And maybe autoload it for the bebop imu module?

martinmm:
@flixr Not sure about a perfect name, ins_ is misleading about the functionality but parrot_ is maybe a little too broad.

flixr:
I would have suggested bebop_temp_ctrl, but this might also be applicable for the Parrot disco?

sw/airborne/modules/ins/ins_temp_ctrl.c
```diff
@@ -0,0 +1,154 @@
...
+
+  if (ins_temp_ctrl_ok == 1) {
+    /* minimal PI algo without dt from Ardupilot */
+    error = (float) ((INS_TEMP_TARGET) - temp_current);
+
+    /* Don't accumulate errors if the integrated error is superior
+     * to the max duty cycle(pwm_period)
+     */
+    if ((fabsf(sum_error) * INS_TEMP_CTRL_KI < INS_TEMP_CTRL_DUTY_MAX)) {
+        sum_error = sum_error + error;
```
flixr: plz fix indentation

sw/airborne/modules/ins/ins_temp_ctrl.c
```diff
@@ -0,0 +1,154 @@
...
+
+    if (output > INS_TEMP_CTRL_DUTY_MAX) {
+      output = INS_TEMP_CTRL_DUTY_MAX;
+    } else if (output < 0) {
+      output = 0;
+    }
+
+     if (dprintf(pwm_heat_duty_fd, "%u", output) < 0)
+       /* could not set duty cycle */
+       {}
```
flixr: Would it be useful to print a warning on the console in this case?

martinmm:
> Would it be useful to print a warning on the console in this case?

just a printf?

martinmm:
@flixr it also works for the Disco, yes. Needs to be tested for that. But not for all Parrot units and I do not really like the brand name in there.

flixr:
yeah, just printf could be useful...? Similar to when the video device can't be initialized..

Then maybe call it sensor_temp_ctrl or imu_temp_ctrl as it really just heats the IMU sensor and doesn't have anything to do with an INS...

PR #2028, opened 2017-02-27
Wind estimation

gautierhattenberger:
Add a wind estimation algorithm, based on Unscented Kalman Filter for real time estimation of all 3 components. The computation is done in a ChibiOS thread so it runs in parallel of the autopilot. The algorithm itself is generated from a Matlab/Simulink model.

This module can also serves as an example of integration of generated code and ChibiOS threads with mutex and signals synchronization.

Some extra tools and modification:
- angle of attack and sideslip update in NPS
- second angle of attack (e.g. sideslip) in aoa_pwm driver
- Meso-NH reader to feed NPS with realistic meteorological (wind) data

The algorithm itself is still an ongoing research topic and might be updated in the future. It shouldn't be used for "normal" operations.


sw/simulator/mesonh/mesonh.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+#nps_udp_socket = None
+#nps_udp_dest = None
+
+start_time = time.time()
+
+
+def get_wind(east, north, up):
+	t = time.time() - start_time
```
flixr: please use 4 spaces for indentation

sw/simulator/mesonh/mesonh.py
```diff
@@ -0,0 +1,203 @@
...
+                print(s)
+            ivy_callback.last_print = time.time()
+            ivy_callback.acid_print.clear()
+
+
+ivy_callback.last_print = time.time()
+ivy_callback.acid_print = {}
+
+
+#def send_wind_nps_udp(acid, wnorth, weast, wup):
```
flixr: what's up with this commented out code?

gautierhattenberger: Not sure anymore, this is a student work from last year. I thought this was the working version, but maybe not. Should I remove it or keep it as a reference until I fix it or find a better version (some colleagues from LAAS are also working on this) ?

podhrmic:
@gautierhattenberger I am glad you opened up this - we are also using simulink generated code for an attitude estimation algorithm (PR to follow) and I have a few related questions:

1. do you know what is the licence of the generated files? There is nothing explicitly stated, but I am not 100% sure that it can be GPL by default (although I haven't investigated it)
2. when making modules like this, do we still want to make them arch-independent or not? If not, there should be checks to avoid compilation for non-target arch, like:
```
#ifndef CHIBIOS
#error "Only for chibios
#endif
```
so it is more descriptive
3. the generated code is indeed big - doing this on larger scale (multiple modules) could inflate the repository - should we worry about that?

gautierhattenberger:
1. You're right it is not clear. It seems that you can choose the license of the original Simulink model as you like (BTW, our model is here: https://github.com/enacuavlab/UKF_Wind_Estimation), and you can even release the generated code in open-source, as long as you are using your toolbox license for your own work. It means that you are not suppose to use you Matlab license "on demand" for other people. If someone else wants to change something he either need to buy an Embedded Coder toolbox license and generate its code from the model or change the generated code by hand (which he probably won't do...).
2. I would keep it here, it is easier to find. At some point we can have some API to hide the RTOS functions (see pprz_mutex.h file). I'll update that later. In the meantime, I'll add the compilation error.
3. Doing it by hand would also inflate the repository, the difference is that hand made code is easier to factorize. I wouldn't worry to much for now. I have idea to allow loading of modules from external sources (but not much time to work on it) so we could have extra repositories for exotic modules.

podhrmic:
OK, LGTM. You already mentioned that the generated code is hard to maintain (change) so I would add a link to the original simulink files.

PR #2030, opened 2017-03-03
added hott sumd radio control input

mlaiacker:
Added Graupner HOTT SUMD serial rc control protocol. 
I used the sbus implementation as base for the hott sumd input. 
Tested with lisa m 2.0 uart1 and Graupner GR-12L with 9 and 12 channels. 
Should work with 2 to 32 channels.
and example airframe config.

### DIFF
--- a/conf/airframes/max/microjet_lisa_m.xml
+++ b/conf/airframes/max/microjet_lisa_m.xml
@@ -1,0 +1,222 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<!-- Microjet Multiplex (http://www.multiplex-rc.de/)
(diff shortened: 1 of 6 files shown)


flixr:
Thanks :-)
HOTT doesn't have inverted polarity like SBUS, right? At least that part is commented out and it seems like you can remove all the polarity stuff...

Can you please also fix the code style? Two spaces for indentation, no trailing whitespaces, ... See http://docs.paparazziuav.org/latest/stylec.html
You can also run the `fix_code_style.sh` script over the new files...

mlaiacker:
code is now also flight tested. 
fix_code_style did not work: "syntax error line 10 [:-q: unary operator expected".
fixed some style by hand.

flixr:
LGTM.

Very strange that the fix_code_style.sh script doesn't work for you, what system are you using? What is your /bin/bash?

PR #2032, opened 2017-03-05
[imu] Vectornav as IMU (gyro and accel data only)

podhrmic:
Allows using Vectornav as IMU - useful if you want to test AHRS algorithms with calibrated gyro/accel data provided by vectornav.

### DIFF
--- a/sw/airborne/modules/imu/imu_vectornav.c
+++ b/sw/airborne/modules/imu/imu_vectornav.c
@@ -1,0 +1,162 @@
+/*
+ * Copyright (C) 2016 Michal Podhradsky, michal.podhradsky@aggiemail.usu.edu
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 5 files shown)


sw/airborne/modules/imu/imu_vectornav.c
```diff
@@ -0,0 +1,162 @@
...
+ * triggers warning in case the IMU stops sending data
+ * and performs initial configuration if needed
+ */
+void imu_vectornav_periodic(void)
+{
+  static uint16_t last_cnt = 0;
+  static uint16_t sec_cnt = 0;
+
+  sec_cnt = imu_vn.vn_packet.counter -  last_cnt;
+  imu_vn.vn_freq = sec_cnt; // update frequency counter
```
gautierhattenberger: is it a frequency or a period ?

sw/airborne/modules/imu/imu_vectornav.c
```diff
@@ -0,0 +1,162 @@
...
+
+  // we want at least 75% of periodic frequency to be able to control the airfcraft
+  if (imu_vn.vn_freq >= (PERIODIC_FREQUENCY*0.75)) {
+    imu_vn.vn_status = VNNotTracking;
```
gautierhattenberger: this status is not really used, what is the purpose of it ?

podhrmic: Good point, the idea was to somehow tell the user that the IMU stopped working, but I am not sure what would be the best way to do it?

sw/airborne/modules/imu/imu_vectornav.c
podhrmic: Frequency - it counts how many packets you received since last call of `send_vn_info` (ideally once a second).

podhrmic:
Merging as is, the IMU status check will be added later (and hopefully in a more consistent manner).

PR #2033, opened 2017-03-07
simulation of bebop1

EwoudSmeur:
I made a simulation file for the bebop1, that takes into account actuator dynamics. It also adds a a spinup torque for the rotors. I think the inertia should be roughly ok. Since I am using the actuators directly, bypassing the motor_mixing, I had to change the nps a bit.

I looked into simulating an electric engine with a propeller, but there are many parameters which I would have to reverse engineer, so it did not seem worthwhile. And then I am still not sure if I can get out of JSBsim the same behaviour as what I get from test flights, given the architecture of JSBsim.


podhrmic:
LGTM

PR #1902, opened 2016-10-12
[rtcm3-rtk] RTK using ublox M8P-2

dewagter:
- [x] depends on https://github.com/paparazzi/pprzlink/pull/35

PR #2034, opened 2017-03-08
[rtcm] RTK with UBlox8 modules

gautierhattenberger:
This PR comes as a replacement on #1902
It is the same code with the NMEA mask fix and a fragmentation of packets to fit in RF modems if necessary.


gautierhattenberger:
@dewagter or any one from TUDelft, are you fine with these modifications ? (splitting the frames so it can work by default on XBee as well, configurable for bigger frames)

flixr:
Basically LGTM, but the entries in `.gitignore`, `.packages`, `control_panel_xml.example` and `default_rotorcraft.xml` are missing compared to #1902

gautierhattenberger:
sorry I forgot them

PR #2035, opened 2017-03-09
Must turn on kill switch once before kill switch can be turned off 

kevindehecker:
I was hunting for another bug,  but then I came across this:

Currently in the AP when arming through the kill switch, the kill switch must be turned on before switching  it off will arm the engines. (so if the rc was left on with kill off and throttle up when the AP is powered up, it will not attack you)

In the FBW side this is not the case. Not completely sure if we want it, but it would be more consistent I guess.


podhrmic:
LGTM

PR #2042, opened 2017-03-20
[extra_dl] Extend module functionality

podhrmic:
Summary of the commit below:
comment

### DIFF
--- a/conf/modules/extra_dl.xml
+++ b/conf/modules/extra_dl.xml
@@ -11,7 +11,24 @@
   </header>
   <init fun="extra_pprz_dl_init()"/>
   <periodic fun="extra_pprz_dl_periodic()" freq="TELEMETRY_FREQUENCY" autorun="TRUE"/>
-  <event fun="ExtraDatalinkEvent()"/>
+
+  <!--
+  The event function listens on the extra telemetry port (from the payload computer for example)
+  and parses the received "datalink" messages, such as MOVE_WP.
+  We need this function to be able to listen to the extra telemetry port  
+
+  As a result all messages normally handled by the autopilot (such as MOVE_WP) are processed.
+  -->
+  <event fun="extra_pprz_dl_event()"/>
+
+  <!-- 
+  The "datalink" tag appends the specified "message" at the end of dl_parse_msg() function.
+  In case their is match in the message ID, "fun" is called on the message.
+
+  This automatically checks for the specified messages in regular telemetry (from GCS)
+  as well as on the extra telemetry (from the payload computer).
+  -->
+  <datalink message="PAYLOAD_COMMAND" fun="extra_pprz_dl_parse_payload_cmd()"/>
   <makefile target="ap">
     <configure name="EXTRA_DL_PORT" default="uart1" case="upper|lower"/>
     <define name="EXTRA_DOWNLINK_DEVICE" value="$(EXTRA_DL_PORT_LOWER)"/>
--- a/conf/modules/extra_dl.xml
+++ b/conf/modules/extra_dl.xml
@@ -21,14 +21,6 @@
   -->
   <event fun="extra_pprz_dl_event()"/>
 
-  <!-- 
-  The "datalink" tag appends the specified "message" at the end of dl_parse_msg() function.
-  In case their is match in the message ID, "fun" is called on the message.
-
-  This automatically checks for the specified messages in regular telemetry (from GCS)
-  as well as on the extra telemetry (from the payload computer).
-  -->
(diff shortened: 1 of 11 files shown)


conf/modules/extra_dl.xml
```diff
@@ -11,7 +11,24 @@
...
+  <event fun="extra_pprz_dl_event()"/>
+
+  <!-- 
+  The "datalink" tag appends the specified "message" at the end of dl_parse_msg() function.
+  In case their is match in the message ID, "fun" is called on the message.
+
+  This automatically checks for the specified messages in regular telemetry (from GCS)
+  as well as on the extra telemetry (from the payload computer).
+  -->
+  <datalink message="PAYLOAD_COMMAND" fun="extra_pprz_dl_parse_payload_cmd()"/>
```
gautierhattenberger: I don't think this datalink event should be part of the module. `extra_dl` is only here to provide the IO stream for an extra telemetry/datalink channel. It is up to the module controlling your payload to implement this function. Also, another payload might want to reuse the same message but in a different way.

sw/airborne/modules/datalink/extra_pprz_dl.c
```diff
@@ -47,13 +47,25 @@
 #include "subsystems/datalink/telemetry.h"
 
 
+
+bool extra_dl_msg_available;
+uint8_t extra_dl_buffer[MSG_SIZE]  __attribute__((aligned));
```
gautierhattenberger: What it the reason to have a dedicated buffer here ?

sw/airborne/modules/datalink/extra_pprz_dl.h
```diff
@@ -41,11 +41,20 @@
...
 extern struct pprz_transport extra_pprz_tp;
 
-/* Datalink Event */
-#define ExtraDatalinkEvent() {                            \
-    pprz_check_and_parse(&EXTRA_DOWNLINK_DEVICE.device, &extra_pprz_tp, dl_buffer, &dl_msg_available); \
-    DlCheckAndParse(&EXTRA_DOWNLINK_DEVICE.device, &extra_pprz_tp.trans_tx, dl_buffer); \
-  }
+
+// example of checking for correct PAYLOAD_COMMAND message
+#define EXPECTED_PAYLOAD_LENGTH 8
```
gautierhattenberger: again, these things are clearly too specific to be part of this generic module

sw/airborne/modules/nav/nav_geofence.h
```diff
@@ -30,8 +30,7 @@
  * home_mode_max_alt is (optionally) defined in the flight plan
  * GEOFENCE_DATALINK_LOST_TIME is defined in the airframe config file
  */
-
-#include <stdbool.h>
+#include "subsystems/datalink/datalink.h"
```
gautierhattenberger: I would place this line inside the `#ifdef GEOFENCE_DATALINK_LOST_TIME` so that it is only included if really needed

conf/modules/extra_dl.xml
podhrmic: I see - I ll move it to a dedicated module.

sw/airborne/modules/datalink/extra_pprz_dl.c
podhrmic: This way we have one buffer for *regular* datalink messages, and one for the *extra* datalink messages. If we don't use the extra buffer, the two streams would have to share the same buffer. Does it make sense?

sw/airborne/modules/datalink/extra_pprz_dl.h
podhrmic: Noted, will move this.

sw/airborne/modules/nav/nav_geofence.h
podhrmic: Good point, will move this.

sw/airborne/modules/datalink/extra_pprz_dl.c
gautierhattenberger: Actually, the `dl_buffer` is just a temporary buffer that receives data from transport buffers (one per stream) and calls datalink functions and callback. This is done in sequence so unless the different streams are coming from different threads (and we would have other issues here probably) it is not really needed to have a different buffer. On the other hand, if we don't have memory issues, it shouldn't be a problem to have a second one.

podhrmic: OK, lets just keep the extra buffer then.

podhrmic:
@gautierhattenberger pls see the changes.

PR #1836, opened 2016-08-07
bebop2 no toiletboil add +50 to x mag

bartremes:

PR #2043, opened 2017-03-20
[MAG_CALIB_UKF] Online magnetomer calibration using unscented Kalman filter

wv-tud:
Magnetometer performance on the bebop2 has been below optimal. Several fixes have provided increased usability (#1836 #1864) however they require calibration and regular re-calibration of the magnetometer and as far as I heard do not allow for long duration flights due to magnetometer drift (@OpenUAS @bartremes)

In this module an unscented kalman filter is implemented based on [TRICAL](http://github.com/sfwa/TRICAL) to calibrate the magnetometer online. This should prevent magnetometer drift during flight. 

I have been testing this non-flying on several bebop2's and it seems to provide accurate calibration. However due to weather conditions I haven't been able to do any NAV flight tests. 

It would be awesome if anybody else with access to either bebop1 or bebop2 could verify this.

**Current limitations:**

- Currently this is only tested on the bebop2 since it is the only platform I am currently able to test.
- I have not flown in NAV mode yet due to windy conditions in the Netherlands. During RC ATT flight the bebop does provide the correct heading
- The UKF should also be able to correct for motor current disturbances so mag_current_coef shouldn't be necessary (this hasn't been thoroughly investigated yet!)


EwoudSmeur:
Nice Wilco, I am curious to see how it performs!

I wonder if the disturbances of motor currents are not fluctuating too much to estimate online...

Would it be possible to put the function call in subsystems/imu.c, such that we don't need to update all specific IMUs?

gautierhattenberger:
Actually, I think it is better to keep it as a module, but just bind to mag data (like filters) so you are independent of the IMU implementation

wv-tud:
@EwoudSmeur in the imu.c file there seem to be no functions which get called after every magnetometer reading except for the weak function imu_scale_mag() which would still be implementation dependant I figured

With regards to the current compensation that will definately have to be tested more extensively

wv-tud:
@gautierhattenberger Is is possible to bind the module to the mag data but ensure that the calibration runs before the AHRS?

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,206 @@
...
+        /** Update magnetometer UKF **/
+        meas[0] = MAG_FLOAT_OF_BFP( _imu->mag.x );
+        meas[1] = MAG_FLOAT_OF_BFP( _imu->mag.y );
+        meas[2] = MAG_FLOAT_OF_BFP( _imu->mag.z );
+        TRICAL_estimate_update(&mag_calib, meas, expected_mag_field);
+        TRICAL_measurement_calibrate(&mag_calib, meas, calib_meas);
+        /** Save calibrated result **/
+        _imu->mag.x = (int32_t) MAG_BFP_OF_REAL( calib_meas[0] );
+        _imu->mag.y = (int32_t) MAG_BFP_OF_REAL( calib_meas[1] );
+        _imu->mag.z = (int32_t) MAG_BFP_OF_REAL( calib_meas[2] );
```
gautierhattenberger: I think the "proper" way to deal with this is to let this module bind to the "real" magnetometer, run this function on mag event, send a new ABI message (it will have a different id) and let the user tell that his filter should listen to this second message instead of the default value (broadcast). It is a bit annoying as the user needs to specify the correct ID for the correct filter (something like `<configure name="AHRS_ICQ_MAG_ID" value="MAG_UKF_CALIB_ID"\>`, but it will guarantee the correct sequence and complete independence of the modules.

sw/airborne/subsystems/imu.c
```diff
@@ -123,7 +123,7 @@ void imu_init(void)
 #if defined IMU_MAG_X_NEUTRAL && defined IMU_MAG_Y_NEUTRAL && defined IMU_MAG_Z_NEUTRAL
   VECT3_ASSIGN(imu.mag_neutral,   IMU_MAG_X_NEUTRAL,   IMU_MAG_Y_NEUTRAL,   IMU_MAG_Z_NEUTRAL);
 #else
-#if USE_MAGNETOMETER
+#if USE_MAGNETOMETER && (!defined UKF_CALIBRATE_FIELD_SENSOR || !UKF_CALIBRATE_FIELD_SENSOR)
```
gautierhattenberger: better find a more generic name, or just remove the `UKF_`

sw/airborne/subsystems/imu/imu_bebop.c
```diff
@@ -87,10 +93,13 @@ void imu_bebop_init(void)
   //the magnetometer of the bebop2 is located on the gps board,
   //which is under a slight angle
   struct FloatEulers imu_to_mag_eulers =
-          {0.0, RadOfDeg(8.5), 0.0};
+  { 0.0, RadOfDeg( 8.5 ), M_PI };
```
gautierhattenberger: why changing this ?

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,206 @@
...
+#include "generated/airframe.h"
+#include "subsystems/gps.h"
+
+
+#define PRINT(string,...) fprintf(stderr, "[CALIB_UKF->%s()] " string,__FUNCTION__ , ##__VA_ARGS__)
+
+#if !defined MAG_CALIB_UKF_GEO_MAG_TIMEOUT
+#define MAG_CALIB_UKF_GEO_MAG_TIMEOUT 0
+#else
+PRINT_CONFIG_VAR(MAG_CALIB_UKF_GEO_MAG_TIMEOUT)
```
wv-tud: I thought #ifndef usage was discouraged - ([PPRZ C Style guide](http://docs.paparazziuav.org/latest/stylec.html)):
"If you really need to test whether a symbol is defined or not, test it with the defined construct."

sw/airborne/subsystems/imu/imu_bebop.c
wv-tud: This way all relevant magnetometer rotations are grouped in a single FloatEulers structure, in stead of the 180 degree rotation being done on Line 150

sw/airborne/modules/calibration/mag_calib_ukf.c
gautierhattenberger: This is the case when you want to check for TRUE or FALSE (and assume FALSE if not defined). In this case you really want to check if it is defined or not, so `ifndef` is fine.

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,206 @@
...
+bool   update_geo_mag_field = true;
+struct FloatVect3 H = { .x = MAG_CALIB_UKF_NORM, .y = 0.0f, .z =  0.0f};
+#endif
+
+#if USE_MAGNETOMETER && MAG_CALIB_UKF_HOTSTART
+static FILE* fp;
+char hotstart_file_name[512];
+#endif
+
+void mag_calib_ukf_init(struct Imu *_imu) {
```
wv-tud: Would a warning  to remove calibration from airframe file when `IMU_MAG_X_NEUTRAL`, `IMU_MAG_Y_NEUTRAL` or `IMU_MAG_Z_NEUTRAL` is defined suffice?

I don't think there is any real disadvantage to defining them, except that they are not necessary, since the UKF will correct for any wrongly entered NEUTRAL value.
Changing these values in your airframe file between flights however will upset the hot-start functionality.

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: Definately a cleaner solution :+1:

sw/airborne/modules/calibration/mag_calib_ukf.c
gautierhattenberger: Then I guess they can be even used as a valid initial value for aircraft that doesn't support the hot start functionality (non-linux boards). So no warning at all as long as online calibration is activated.

conf/modules/mag_calib_ukf.xml
```diff
@@ -0,0 +1,29 @@
...
+  </doc>
+  <header>
+    <file name="mag_calib_ukf.h"/>
+  </header>
+  <periodic fun="mag_calib_hotstart_write()" freq="0.25"/>
+  <makefile>
+    <file name="mag_calib_ukf.c"/>
+    <flag name="CFLAGS" value="I$(PAPARAZZI_SRC)/sw/ext/TRICAL/include"/>
+    <flag name="CFLAGS" value="I$(PAPARAZZI_SRC)/sw/ext/TRICAL/src"/>
+    <file name="../../../ext/TRICAL/src/TRICAL.c"/>
```
wv-tud: It seems to append the name argument to the current directory. Using $(PAPARAZZI_SRC) will result in: 
```
make[1]: *** No rule to make target '/home/fred/paparazzi/var/aircrafts/bebop2/ap/modules/calibration//home/fred/paparazzi/sw/ext/TRICAL/src/TRICAL.c', needed by '/home/fred/paparazzi/var/aircrafts/bebop2/ap/modules/calibration//home/fred/paparazzi/sw/ext/TRICAL/src/TRICAL.o'.  Stop.
```

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: And additionally, what would be the best way to make sure the calibrated values are stored in the imu struct? Since that is the origin of the periodic telemetry:
```
static void send_mag_scaled(struct transport_tx *trans, struct link_device *dev)
{
  pprz_msg_send_IMU_MAG_SCALED(trans, dev, AC_ID,
                               &imu.mag.x, &imu.mag.y, &imu.mag.z);
}
```

gautierhattenberger: The last to bind is the first to be called. So in your case, the ahrs is most likely initialized before the mag_calib module. Since this is defined by the loading order in the airframe file, you can't tell for sure that the sequence is correct (but result should be deterministic). Forcing the bind to correct source should guarantee that the order is always correct.
Now, if you change the content of imu struct instead of working on a local copy, it is the global imu struct that should be modified in the current implementation. But again, no guarantee, a particular implementation could send something else.
So you should make a copy of the received mag data, apply your filter and send back this value (and eventually update by hand the global imu struct if you want to see the result in the telemetry).

conf/modules/mag_calib_ukf.xml
gautierhattenberger: strange, it is working like this for `gps_piksi` or `tlsf` modules

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: Ah ok, so it probably works since subsystems are initialized before modules.

However if I send another abi message after processing the one from the IMU won't the AHRS see duplicate magnetometer messages if one of the following defines isn't correctly alterred:
`AHRS_FC_MAG_ID`, `AHRS_ICE_MAG_ID`, `AHRS_MLKF_MAG_ID`, `AHRS_FINV_MAG_ID`,  `AHRS_DCM_MAG_ID`, `AHRS_ICQ_MAG_ID`, `INS_FINV_MAG_ID`

I've tried to resolve this by hardcoding the defines for all current AHRS systems in  my header file:

```
#ifdef AHRS_FLOAT_CMPL_WRAPPER_H
#undef AHRS_FC_MAG_ID
#define AHRS_FC_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef AHRS_INT_CMPL_EULER_WRAPPER_H
#undef  AHRS_ICE_MAG_ID
#define AHRS_ICE_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef AHRS_FLOAT_MLKF_WRAPPER_H
#undef  AHRS_MLKF_MAG_ID
#define AHRS_MLKF_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef AHRS_FLOAT_INVARIANT_WRAPPER_H
#undef  AHRS_FINV_MAG_ID
#define AHRS_FINV_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef AHRS_FLOAT_DCM_WRAPPER_H
#undef  AHRS_DCM_MAG_ID
#define AHRS_DCM_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef AHRS_INT_CMPL_QUAT_WRAPPER_H
#undef  AHRS_ICQ_MAG_ID
#define AHRS_ICQ_MAG_ID MAG_CALIB_UKF_ID
#endif

#ifdef INS_FLOAT_INVARIANT_WRAPPER_H
#undef  INS_FINV_MAG_ID
#define INS_FINV_MAG_ID MAG_CALIB_UKF_ID
#endif
```

is this acceptable?

conf/modules/mag_calib_ukf.xml
wv-tud: Ah ok now I see, the `<file>` attribute has a `dir` tag which should be used

sw/airborne/modules/calibration/mag_calib_ukf.c
gautierhattenberger: This will only work if the the related filter is including this before the binding. I would start with a warning at compile time for now, to remind the user to set the value correctly. An other option less intrusive for filters is to check if one of this define have been actually defined to `MAG_CALIB_UKF_ID`.

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,200 @@
...
+      H.y                     = (float) geo_mag.vect.y / n;
+      H.z                     = (float) geo_mag.vect.z / n;
+      mag_field_from_geo_mag  = time(0);
+      update_geo_mag_field    = false;
+      VERBOSE_PRINT("Updating local magnetic field from geo_mag module (Hx: %4.2f, Hy: %4.2f, Hz: %4.2f)\n", H.x, H.y, H.z);
+    }
+  }
+  if (MAG_CALIB_UKF_GEO_MAG_TIMEOUT && GpsFixValid() && difftime(time(0), mag_field_from_geo_mag) >= MAG_CALIB_UKF_GEO_MAG_TIMEOUT) {
+    geo_mag.ready           = false;
+    geo_mag.calc_once       = true;   ///< Geo_mag will not re-update the calculation when the throttle is on so this is neccesary
```
wv-tud: Why does `geo_mag` module check if the motors are running? 
Are the computation so intensive that they cannot be run during flight?

wv-tud: @gautierhattenberger Should the `geo_mag` timed updates be transferred to the `geo_mag` module instead of being triggered here? If not should I make the updates a periodical function in this module instead of checking for the time here?

flixr:
Nice!

Can you please also add TRICAL in `sw/ext/Makefile` analogous to e.g. chibios?
Then the submodule will be automatically initialized and updated on make...

flixr:
Does TRICAL only run on Linux or would it also run on bare metal?
If it only runs on Linux, this should be mentioned in the module doc and a check added in the C file.
If it also runs on bare metal, then the HOTSTART file writing needs to be disabled or implemented to use something like persistent settings....

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+
+TRICAL_instance_t mag_calib;
+static abi_event mag_ev;
+
+time_t mag_field_from_geo_mag;
+bool   update_geo_mag_field = true;
+struct FloatVect3 H = { .x = MAG_CALIB_UKF_NORM, .y = 0.0f, .z =  0.0f};
+
+#if MAG_CALIB_UKF_HOTSTART
+static FILE *fp;
```
flixr: no reason to make those global variables...

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+char hotstart_file_name[512];
+#endif
+
+void mag_calib_ukf_init(void)
+{
+#if USE_MAGNETOMETER
```
flixr: I don't think we need to protect this with USE_MAGNETOMETER.....

wv-tud:
I'm not quite sure how to asses that.
So far as I see it uses `memset`, `memcpy`, `assert`, `math`, `string` floating points en a pragma I haven't managed to fix except silence `MUST_ITERATE( X, X )`

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: I'm not quite sure how to asses that.
So far as I see it uses `memset`, `memcpy`, `assert`, `math`, `string`, floating points en a pragma I haven't managed to fix except silence `MUST_ITERATE( X, X )`

wv-tud:
I'm not quite sure how to asses that.
So far as I see it uses `memset`, `memcpy`, `assert`, `math`, `string`, floating points en a pragma I haven't managed to fix except silence `MUST_ITERATE( X, X )`

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: This way I thought there is no downside to keeping it in my optitrack airframe files

sw/airborne/modules/calibration/mag_calib_ukf.c
flixr: Yes, on the Cortex-M based boards it takes too long, so it can't run in flight as it would delay the control loops...

flixr:
At a glance it looks like it doesn't need an OS... but you are using OS only syscalls... e.g. for time plz use our sys_time...

Regarding the geo_mag integration: it would be better to not directly include the geo_mag header and use the struct directly, but bind to the GEO_MAG ABI message and use that value to update the local magnetic field.

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+  if (MAG_CALIB_UKF_GEO_MAG_TIMEOUT && GpsFixValid() && difftime(time(0), mag_field_from_geo_mag) >= MAG_CALIB_UKF_GEO_MAG_TIMEOUT) {
+    geo_mag.ready           = false;
+    geo_mag.calc_once       = true;   ///< Geo_mag will not re-update the calculation when the throttle is on so this is neccesary
+    update_geo_mag_field    = true;
+  }
+  /** Update magnetometer UKF and calibrate measurement **/
+  if (mag->x != 0 || mag->y != 0 || mag->z != 0) {
+    /** Rotate the local magnetic field by our current attitude **/
+    struct FloatQuat *body_quat = stateGetNedToBodyQuat_f();
+    struct FloatRMat body_rmat;
```
flixr: there is no reason to first create a rotation matrix from the quaterion, you can directly use `float_quat_vmult`

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+    /** Save calibrated result **/
+    mag->x = (int32_t) MAG_BFP_OF_REAL(calibrated_measurement[0]);
+    mag->y = (int32_t) MAG_BFP_OF_REAL(calibrated_measurement[1]);
+    mag->z = (int32_t) MAG_BFP_OF_REAL(calibrated_measurement[2]);
+    VERBOSE_PRINT("magnetometer measurement (x: %4.2f  y: %4.2f  z: %4.2f) norm: %4.2f\n", measurement[0], measurement[1], measurement[2], hypot(hypot(measurement[0], measurement[1]), measurement[2]));
+    VERBOSE_PRINT("magnetometer bias_f      (x: %4.2f  y: %4.2f  z: %4.2f)\n", mag_calib.state[0], mag_calib.state[1],  mag_calib.state[2]);
+    VERBOSE_PRINT("expected measurement     (x: %4.2f  y: %4.2f  z: %4.2f) norm: %4.2f\n", expected_mag_field[0], expected_mag_field[1], expected_mag_field[2], hypot(hypot(expected_mag_field[0], expected_mag_field[1]), expected_mag_field[2]));
+    VERBOSE_PRINT("calibrated   measurement (x: %4.2f  y: %4.2f  z: %4.2f) norm: %4.2f\n\n", calibrated_measurement[0], calibrated_measurement[1], calibrated_measurement[2], hypot(hypot(calibrated_measurement[0], calibrated_measurement[1]), calibrated_measurement[2]));
+  }
+  AbiSendMsgIMU_MAG_INT32(MAG_CALIB_UKF_ID, get_sys_time_usec(), mag);
```
flixr: it would be better to resend the message with the original time stamp instead of the time when the computation is done

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+#endif
+PRINT_CONFIG_VAR(MAG_CALIB_UKF_NORM)
+
+#if !defined MAG_CALIB_UKF_NOISE_RMS
+#define MAG_CALIB_UKF_NOISE_RMS 2e-1f
+#endif
+PRINT_CONFIG_VAR(MAG_CALIB_UKF_NOISE_RMS)
+
+#if !defined MAG_CALIB_UKF_HOTSTART
+#define MAG_CALIB_UKF_HOTSTART TRUE
```
gautierhattenberger: default to FALSE as it is not working in all cases ?

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,199 @@
...
+  TRICAL_norm_set(&mag_calib, MAG_CALIB_UKF_NORM);
+  TRICAL_noise_set(&mag_calib, MAG_CALIB_UKF_NOISE_RMS);
+
+  H.x = AHRS_H_X;
+  H.y = AHRS_H_Y;
+  H.z = AHRS_H_Z;
+  VERBOSE_PRINT("Local magnetic field loaded from airframe file (Hx: %4.2f, Hy: %4.2f, Hz: %4.2f)\n", H.x, H.y, H.z);
+
+  mag_calib_hotstart_read();
+  AbiBindMsgIMU_MAG_INT32(IMU_BOARD_ID, &mag_ev, mag_calib_ukf_run);
```
flixr: would be good to make the IMU_ID here configurable... and then check in the callback that you don't receive your own republished message (i.e. sender_id != MAG_CALIB_UKF_ID)

gautierhattenberger: The ID should be something like `MAG_CALIB_UKF_MAG_ID` with a default to:
```
#ifndef MAG_CALIB_UKF_MAG_ID
#define MAG_CALIB_UKF_MAG_ID ABI_BROADCAST
#endif
```
This way it can be used with and external mag or binded to a specific one when multiple sensors are present.

sw/airborne/modules/calibration/mag_calib_ukf.c
flixr: I'm not a fan of this... it would make sense to not add this module if you don't use the mag...
and even if you have it added it will only result in some extra computation, but not be used.

sw/airborne/modules/calibration/mag_calib_ukf.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @file "modules/calibration/mag_calib_ukf.h"
+ * @author w.vlenterie
+ * Calibrate magnetometer using UKF
+ */
+
+#ifndef MAG_CALIB_UKF_H
+#define MAG_CALIB_UKF_H
+
+#ifdef AHRS_FLOAT_CMPL_WRAPPER_H
```
flixr: this is brittle and error prone... and probably wont even work as intended in a lot of cases where file is never included...

wv-tud: would you know a better way to solve it? Letting the user define the ` MAG_CALIB_UKF_ID` inside the AHRS section of the airframe file also seems a little odd/unnecessary.

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: It does work with pre-defined mag field parameters, as it is initialized as `AHRS_H_*`. 

However since the UKF is correcting based on this field it should be corrected when large distance is covered. To solve this I included a timeout (disabled by default) which can be set to update the geo_mag module.

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: removed it

wv-tud:
@flixr How would I go and implement OS independent persistent settings? I'd like to prevent the user from having to copy settings to their airframe file for the obvious inconvenience (it's 12 floats).

sw/airborne/modules/calibration/mag_calib_ukf.h
flixr: Right now I can't think of a nice solution that works robustly...

gautierhattenberger:
Persistent setting API is not really nice for this at the moment. I would suggest to let this part aside for now. A static initial calibration would already help to have a faster convergence as this filter would only correct small variations.

flixr:
At the moment this only compiles on Linux anyway... making it work on bare metal could be a future improvement.
But until then you should add a check in the the C file if it is compiled on Linux (and note it in the module doc).
(Ab)using persistent settings should work, but maybe not really that nice as @gautierhattenberger said...

wv-tud:
The hotstart/persistent settings help in two major ways. 

1. As @gautierhattenberger mentions it helps the filter converge faster. From no initial conditions especially a typical magnetometer calibration motion (aka rotating around all body axes) helps it converge fast.
2. It seems to help the AHRS set up correctly, otherwise the heading only slowly drifts to its correct heading whilst the filter has already converged. Maybe this can be fixed by resetting the heading after the filter has converged?

sw/airborne/modules/calibration/mag_calib_ukf.c
gautierhattenberger: if `geo_mag` needs update because the distance to init point is significant, it should be handled by `geo_mag` module and propagated to filters through ABI message.

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,205 @@
...
+  AbiBindMsgGEO_MAG(ABI_BROADCAST, &h_ev, mag_calib_update_field);    ///< GEO_MAG_SENDER_ID is defined in geo_mag.c so unknown
+}
+
+void mag_calib_ukf_run(uint8_t sender_id, uint32_t stamp, struct Int32Vect3 *mag)
+{
+  float measurement[3] = {0.0, 0.0, 0.0}, calibrated_measurement[3] = {0.0, 0.0, 0.0};
+  if (sender_id != MAG_CALIB_UKF_ID) {
+    /** Update geo_mag based on MAG_CALIB_UKF_GEO_MAG_TIMEOUT (0 = no periodic updates) **/
+    if (MAG_CALIB_UKF_GEO_MAG_TIMEOUT && GpsFixValid() && (get_sys_time_msec() - timestamp_geo_mag) >= 1000 * MAG_CALIB_UKF_GEO_MAG_TIMEOUT) {
```
flixr: I would remove the whole geo_mag recalculation stuff... (re)triggering this based on time doesn't make that much sense anyway, as it only changes if you have a significantly different position.
And it creates a dependency on geo_mag...

wv-tud:
@flixr What part (except for hotstart which can be disabled) doesn't on other platforms? And how do I implement such a check?

sw/airborne/modules/calibration/mag_calib_ukf.c
wv-tud: Would this also not matter for fixedwings which could of course travel much further?

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,205 @@
...
+      float_quat_vmult(&expected_measurement, body_quat, &H);
+      float expected_mag_field[3] = { expected_measurement.x, expected_measurement.y, expected_measurement.z };
+      /** Update magnetometer UKF **/
+      measurement[0] = MAG_FLOAT_OF_BFP(mag->x);
+      measurement[1] = MAG_FLOAT_OF_BFP(mag->y);
+      measurement[2] = MAG_FLOAT_OF_BFP(mag->z);
+      TRICAL_estimate_update(&mag_calib, measurement, expected_mag_field);
+      TRICAL_measurement_calibrate(&mag_calib, measurement, calibrated_measurement);
+      /** Save calibrated result **/
+      mag->x = (int32_t) MAG_BFP_OF_REAL(calibrated_measurement[0]);
```
flixr: You shouldn't modify the mag measurement that was passed to the callback... make your own copy and then send that via ABI.

wv-tud: this allows the scaled magnetometer message to show the calibrated values

flixr: not necessarily... depending on the imu/mag implementation this does not point to the mag in the global imu struct, but the local mag values of the mag driver...

wv-tud: Could I then include imu.h and modify the imu.mag values directly? And should i use a static struct for the copy?

flixr: That would be ok for now (as long as we still have the global struct).
For the copy simply make a global struct for the mag values.

gautierhattenberger:
@wv-tud Please have a look at https://github.com/paparazzi/paparazzi/commit/7fdd0c95ef287668596b79a8636e6ea515ba82e5
I made a few changes, including compilation on non-Linux autopilot (was failing because of `assert` in TRICAL lib, disabled with `NDEBUG` option) and removing dependency to geo_mag and gps (if geo_mag needs update based on time or distance, it should be able to handle this itself).
The code is running on a STM32F4 board, and it seems it does work, but the convergence is rather slow, so better start from a valid static calibration. Not tested in flight, but if you and @flixr agree with my changes, I think it can go to master like that.

wv-tud:
I'll have a look tonight thanks. Were you assessing convergence speed from the mag_scaled or psi message? Because after the magnetometer is properly calibrated it can take quite a while for psi to also show the correct value (not sure how to fix this)

gautierhattenberger:
I looked at the mag_scaled min/max values

wv-tud:
All changes look OK. Does 'NDEBUG' also remote the `-Wshadow` warning for `jn` in `3dmath.h`?

What timespan should I be thinking about when you mention rather slow? And is this using a static UAV or are you rotating/moving by hand? Decreasing the noise RMS could lead to faster convergence, perhaps the default is too high (since I couldn't find it in the bebops magnetometer datasheet).. 
Thanks for testing!

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,279 @@
...
+static char hotstart_file_name[512];
+#endif
+
+void mag_calib_ukf_init(void)
+{
+  TRICAL_init(&mag_calib);
+  TRICAL_norm_set(&mag_calib, MAG_CALIB_UKF_NORM);
+  TRICAL_noise_set(&mag_calib, MAG_CALIB_UKF_NOISE_RMS);
+  mag_calib_hotstart_read();
+#if MAG_CALIB_UKF_USE_INITIAL_STATE
```
gautierhattenberger: why not checking directly if `MAG_CALIB_UKF_INITIAL_STATE` is defined ?

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,279 @@
...
+  float initial_state[12] = MAG_CALIB_UKF_INITIAL_STATE;
+  mag_calib.state[0] = initial_state[0];
```
gautierhattenberger: better use `memcpy` here

sw/airborne/modules/calibration/mag_calib_ukf.c
```diff
@@ -0,0 +1,279 @@
...
+}
+
+void mag_calib_send_state(void)
+{
+  // TODO: find a better way to relay the filter state
+  char data[256];
+  uint8_t i;
+  for (i = 0; i < 12; i++) {
+    snprintf(data, 256, "%s, mag_calib.state[%d]: %0.6f", AIRFRAME_NAME, i, mag_calib.state[i]);
+    DOWNLINK_SEND_INFO_MSG(DefaultChannel, DefaultDevice, strlen(data), data);
```
gautierhattenberger: There is a `PAYLOAD_FLOAT` message to send an array of float

conf/modules/mag_calib_ukf.xml
```diff
@@ -0,0 +1,46 @@
...
+      
+      For more information see TRICAL project page (https://www.github.com/sfwa/TRICAL).
+    </description>
+    <define name="[AHRS/INS]_XXX_MAG_ID" value="MAG_CALIB_UKF_ID" description="Select correct input mag for your estimation filter (AHRS or INS), replace XXX by your filter name as defined in sw/airborne/subssytems/abi_sender_ids.h"/>
+    <section name="MAG_CALIB_UKF" prefix="MAG_CALIB_UKF_">
+      <define name="NORM" value="1.0f" description="Measurement norm of magnetometer"/>
+      <define name="NOISE_RMS" value="2e-1f" description="Noise RMS of magnetometer"/>
+      <define name="ABI_BIND_ID" value="ABI_BROADCAST" description="Bind to a specific source for input data, broadcast ID by default"/>
+      <define name="INITIAL_STATE" value="{0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0}" type="float[12]" description="Initial state to be used by filter (suggested for non-Linux-based boards). You can get the state via settings."/>
```
gautierhattenberger: correct format for this is
```
<define name="INITIAL_STATE" value="0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0" type="float[]" />
```

PR #2045, opened 2017-03-21
Drawing circles in the GCS if the user is employing the Guidance Vector Field

noether:
Hello,

now if the tracked ellipse in the guidance vector field is a circle, then the mode is changed to HORIZONTAL_MODE_CIRCLE. By doing this, the GCS will draw the circumference on the screen once a CIRCLE msg is received by the server at ground.

### DIFF
--- a/sw/airborne/modules/guidance/gvf/gvf.c
+++ b/sw/airborne/modules/guidance/gvf/gvf.c
@@ -58,7 +58,7 @@ static void send_gvf(struct transport_tx *trans, struct link_device *dev)
       plen = 6;
       break;
     default:
-      plen = 0;
+      plen = 1;
       break;
   }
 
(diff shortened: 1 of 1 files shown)


sw/airborne/modules/guidance/gvf/gvf.c
```diff
@@ -237,6 +238,9 @@ bool gvf_ellipse(uint8_t wp, float a, float b, float alpha)
     gvf_trajectory.p[3] = 60;
   }
 
+  if (gvf_trajectory.p[2] == gvf_trajectory.p[3])
+    horizontal_mode = HORIZONTAL_MODE_CIRCLE;
```
gautierhattenberger: this mode should be set to HORIZONTAL_MODE_WAYPOINT if not a circle nor a segment

noether: haha, I have pushed at the same time you made the comment :P

PR #2046, opened 2017-03-22
Small update to the circular formation script

noether:
1. Sigmas are asked in degrees now (instead of radians).
2. The array of floats from the GVF msg is parsed "differently".
3. The error signals are printed properly.

I have also added an example (to be described in the wiki :P)


sw/ground_segment/python/gvf/gvfFormation.py
```diff
@@ -111,9 +111,9 @@ def main():
         interface.shutdown()
         return
 
-    B = np.loadtxt(sys.argv[1])
-    desired_sigmas = np.loadtxt(sys.argv[2])
-    ids = np.loadtxt(sys.argv[3])
+    B = np.loadtxt('./formation/' + sys.argv[1])
```
gautierhattenberger: not sure it is interesting to force a local path to `formation` folder, you may want to have your input files in different directories or even pass an absolute path

noether: yepe, I do not think is a good idea either.

noether: I will commit the change once I have edited the wiki with the example. Maybe after explaining the example I will come with more small changes.

gautierhattenberger: I suggest to add
```
  <modules>
    <module name="gvf_module"/>
  </modules>
```
to your demo flight plan then!

noether:
I have updated the wiki with the committed example

https://wiki.paparazziuav.org/wiki/Module/guidance_vector_field#Circular_formations

PR #2047, opened 2017-03-29
Full reatured fixedwing example for Pixracer AP board

OpenUAS:
Still needs quite some testflights. Weather permitting tests this weekend. If it works as expected a pull request honoring after that.


podhrmic:
@OpenUAS any news here? Has it been tested and flown?

OpenUAS:
Yes, test flown Auto2 on "classic" control, but not tuned ETECS yet. The next week test and merge. ad to still get a good V (I Works) reading on analog, but sensor was not working.
Will then  update AC with Dynamic mag compensation and tuned and flown on ETECS. Maybe even a small video if time permits.

OpenUAS:
100% auto test-flight performed. Could use more testing... as always ;-)

gautierhattenberger:
right, conf and control_panel shouldn't be under revision control. It seems this is a bit messy with a lot of unwanted changes.

PR #2048, opened 2017-03-30
Edge flow fixes

ETitus:
Each commit contains different fixes for different problems.
Most fixes are clear by the commit message, if not please ask why I did certain things.

Note that for the RESOLUTION_FACTOR in opticflow_calculator.c I'm quite sure it is implemented now as it was intended. However I am not fully sure it is sufficient, as information is lost earlier regardless due to the use of integers instead of floats.

Apart from those I have a few remarks/questions with respect to things that might improve speed a bit, or remain a possible problem. However I have not fixed them as I am not sure if my proposed solution is valid/matters.

1)
I believe line 201 in edge_flow.c is useless, given line 181:
https://github.com/ETitus/paparazzi/blob/EdgeFlow_fixes/sw/airborne/modules/computer_vision/lib/vision/edge_flow.c#L201
Removing line 201 might result in a slightly faster run?

2)
looking at line 124 and 143 in edge_flow.c, the for loop goes over c=-1, c=0 and c=1. But the loop does nothing useful per definition for c=0, because Sobel[c + 1] * something will always be 0 as Sobel[0+1]=Sobel[1]=0.
I am not sure how much, or actually little computation time this will save, but having the for loop only loop c=-1 and c=1 could prevent useless computations.
https://github.com/ETitus/paparazzi/blob/EdgeFlow_fixes/sw/airborne/modules/computer_vision/lib/vision/edge_flow.c#L124

3)
If one is to switch between LK and EF frequently the if statement in line 451 in opticflow_calculator.c is run multiple times, mallocing space, that is not being freed somewhere.
This can either be moved to the init function, and set to 0 when the if in 451 is run.
https://github.com/ETitus/paparazzi/blob/EdgeFlow_fixes/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c#L451

The same goes for the if in line 240, where 2 images are created and not freed upon switching away.

https://github.com/ETitus/paparazzi/blob/EdgeFlow_fixes/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c#L240


ETitus:
Commit 2ac81ac contains proposed speed up fixes 1) and 2) as described in previous comment

flixr:
@dewagter @OpenUAS @knmcguire @fvantienen @rmeertens @guidoAI can you guys please look at this or find someone to review/test this?

knmcguire:
@ETitus did you try this on a ARdrone 2.0? Did you fly and hover with it?

ETitus:
No I only did handheld tests to verify the memory didn't leak/indices kept within the arrays and switching back and forth between EF and LK functioned properly.

It would be interesting to fly in the Zoo on Optitrack and compare truth to old/new EF/LK though.

knmcguire:
It would be good to also try it out in flight. I don't think that comparing it with the real velocity would be necessary (as it is still in construction). If it is able to hover on it, and no problems exist while switching between LK and EF than it would be good. Do you have time to do such a test (and let me know so I can watch)?

knmcguire:
Btw. I probably will test it also on a Bebop 1 just to be sure. Maybe this afternoon

ETitus:
Today @knmcguire and I flew the Ardrone, after making a INDI conf which is added in the latest commit.
The performance looked similar to the current master. 
@knmcguire will try to fly this on a Bebop before the pull request can be granted.

knmcguire:
Seems to be working fine. This pull request is ready to merge

PR #2051, opened 2017-04-09
guidance_h heading setpoint in float

EwoudSmeur:
heading setpoint in float for convenience and robustness
I also changed the rotorcraft_fp and hover_loop messages to send the float value (see pprzlink pr)

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
@@ -173,9 +173,9 @@ void guidance_h_init(void)
 
   INT_VECT2_ZERO(guidance_h.sp.pos);
   INT_VECT2_ZERO(guidance_h_trim_att_integrator);
-  INT_EULERS_ZERO(guidance_h.rc_sp);
-  guidance_h.sp.heading = 0;
-  guidance_h.sp.heading_rate = 0;
+  FLOAT_EULERS_ZERO(guidance_h.rc_sp);
+  guidance_h.sp.heading = 0.0;
+  guidance_h.sp.heading_rate = 0.0;
   guidance_h.gains.p = GUIDANCE_H_PGAIN;
   guidance_h.gains.i = GUIDANCE_H_IGAIN;
   guidance_h.gains.d = GUIDANCE_H_DGAIN;
@@ -318,7 +318,7 @@ void guidance_h_read_rc(bool  in_flight)
       stabilization_attitude_read_rc(in_flight, FALSE, FALSE);
       break;
     case GUIDANCE_H_MODE_HOVER:
-      stabilization_attitude_read_rc_setpoint_eulers(&guidance_h.rc_sp, in_flight, FALSE, FALSE);
+      stabilization_attitude_read_rc_setpoint_eulers_f(&guidance_h.rc_sp, in_flight, FALSE, FALSE );
 #if GUIDANCE_H_USE_SPEED_REF
       read_rc_setpoint_speed_i(&guidance_h.sp.speed, in_flight);
       /* enable x,y velocity setpoints */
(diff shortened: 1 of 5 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -555,6 +555,7 @@ void guidance_h_nav_enter(void)
   reset_guidance_reference_from_current_position();
 
   nav_heading = stateGetNedToBodyEulers_i()->psi;
+  guidance_h.sp.heading = stateGetNedToBodyEulers_f()->psi;
```
flixr: This is an additional change and not only a result of changing it to float... what is the reason behind it?

sw/ground_segment/tmtc/rotorcraft_server.ml
```diff
@@ -174,7 +174,7 @@ let log_and_parse = fun ac_name (a:Aircraft.aircraft) msg values ->
             and desired_alt = foi32value "carrot_up" /. pos_frac in
             a.desired_pos <- Aircraft.add_pos_to_nav_ref nav_ref ~z:desired_alt (desired_east, desired_north);
             a.desired_altitude <- desired_alt +. (hmsl_of_ref nav_ref a.d_hmsl);
-            a.desired_course   <- foi32value "carrot_psi" /. angle_frac
+            a.desired_course   <- fvalue "carrot_psi" /. angle_frac
```
flixr: As mentioned in https://github.com/paparazzi/pprzlink/pull/49 I don't think it is needed to change this message...

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
EwoudSmeur: Ah yes, so for this commit this is not essential. The reason for this is that I would like to have the option to not use nav_heading, as there are vehicles for which this is not a free variable (hybrids for instance). Then guidance_h.sp.heading needs to be reset on enter, as it will not be set from the (reset) nav_heading.
I thought for now this would not hurt.

sw/ground_segment/tmtc/rotorcraft_server.ml
EwoudSmeur: So that means we will never change this message? Don't you get md5 complaints anyway if you try to listen to an AC you did not upload?
It is a really small thing, but I rather make it as clean as possible now we are at it.

flixr: Of course not....
But the message are the most important/prominent external interface with a simple API built around it in pprzlink. And at least the important core messages shouldn't be changed light heartedly. Anyone building something on pprzlink (like e.g. FlyingRobotCommander) will have to update it as well if a message that is used changes...

So IMHO "core" messages should not be changed if it doesn't provide a clear advantage or fix something...

Regarding the MD5sum: you will only get the complaint if you rebuilt your aircraft with the new version but didn't upload it... if you only update paparazzi itself (with the changed message) but not rebuild your airframe there is no complaint...

So in my view the "clean" way is to not introduce unnecessary changes to the "external API" (i.e. messages) when there is no additional benefit apart from "internal" refactoring.

EwoudSmeur: Ah yes I had not thought of external applications. I will remove the message change.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
EwoudSmeur: So do you think this is fine or shall I remove it?

flixr: I think it is fine (although I haven't tested it), maybe just mention it in the PR description...

flixr:
@gautierhattenberger this looks good to me, maybe you can have a quick look as well?

podhrmic:
LGTM. Has it been tested @EwoudSmeur ?

podhrmic:
Merging as is since @flixr approved it.

PR #2052, opened 2017-04-10
Poly survey via Guidance Vector Field addition

noether:
Some updates regarding the Guidance Vector Field (I will update the wiki accordingly).

- The gains 'ke' and 'kd' are not "global" anymore and each trajectory has its own set of gains. If the user runs a mission with different trajectories, then there is not need to change the gains on the fly anymore, e.g. poly survey (see later). The user can also tune the gains separately and save them in its airframe.

- More primitives regarding ellipses, sin and lines have been added. In particular, now there is 'gvf_segment' where the aircraft tracks a segment defined by two points.

- The set NAV now has 'nav_survey_polygon_gvf_run()', which does exactly the same as 'nav_survey_polygon_run()', but employing the GVF instead of the Carrot. One of the advantages is that during the segments, the aircraft really tracks the line and it is not going to a waypoint (Carrot).


noether:
mmm I do not understand/find which tests my code failed, any help? :P

flixr:
The problem is that you added gvf functions (specifically `gvf_set_direction`) to nav_survey_polygon and now that depends on the gvf module... and since the gvf module is of course not added (and shouldn't be) to each airframe using survey_polygon, it fails for those...

noether:
ahh I see the problem then. Thanks Felix.

Then it is not only gvf_set_direction, but gvf_circle and gvf_segment, i.e., all the functions that substitute the carrot guidance by the gvf_guidance.

Mmm, what would you suggest? For example, to move this navigation function to the gvf module? or to
put the new two functions in the nav_survey_poly between #IFDEF GVF? (maybe this is dirty?) another ideas?

OpenUAS:
Not really the fault of @noether , well maybe add, #ifdef USE_AD0 around the  static const uint32_t ADC_AD0CR_SEL_HW_SCAN = 0  and rest etc. etc. maybe. A warning about an unused var breaks the build, very strict policy indeed ;-)

flixr:
@OpenUAS the build doesn't break because there are unused vars...
Is because of the additions to the nav_survey_polygon module that now effectively requires the gvf module to be loaded as we...
@noether if you can, it would be nicer to put those functions in the gvf module... as far as I can see there is no reason to have them in the nav_survey_polygon module....

noether:
yes, I think it will be the best solution, I will put these navigation functions in the gvf module.

Thanks for the input guys :P

noether:
do not accept this pull yet. I had to synchronize my repo in order to have it ready in another computer on Monday :P.

noether:
Hello, 

if you do not have any further comments or requests, the pull request is ready to be merged.

conf/flight_plans/demo_gvf.xml
```diff
@@ -1,21 +1,35 @@
-<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
```
flixr: that doesn't look correct, the `flight_plan.dtd` is in the same directory as the flightplan, so it was correct before

conf/modules/nav_survey_polygon.xml
```diff
@@ -14,12 +14,12 @@
...
         - min_rad       minimal radius when navigating
         - altitude      the altitude that must be reached before the flyover starts
-      2. Run the survey with nav_survey_polygon_run()
+      2. Run the survey with nav_survey_polygon_run() for employing the carrot guidance algorithm
 <!--
 Block example:
 @verbatim
 <block name="Poly Survey" strip_button="Poly Survey">
-  <call fun="nav_survey_polygon_setup(first_wp, size, angle, sweep_width, shot_dist, min_rad, altitude)"/>
+  <call once fun="nav_survey_polygon_setup(first_wp, size, angle, sweep_width, shot_dist, min_rad, altitude)"/>
```
flixr: should be `call_once` with an underscore...

sw/airborne/modules/guidance/gvf/gvf.h
```diff
@@ -53,6 +53,14 @@ typedef struct {
   float p[16];
 } gvf_tra;
 
+typedef struct {
+  int seg;
```
flixr: can you please add some inline doxygen comments on what these variables mean and what unit they have?

sw/airborne/modules/guidance/gvf/nav/nav_survey_polygon_gvf.c
```diff
@@ -0,0 +1,326 @@
...
+  divider = (gvf_survey.sweep_vec.y * gvf_survey.dir_vec.x) - (gvf_survey.sweep_vec.x * gvf_survey.dir_vec.y);
+
+  //calculate the leftmost point if one sees the dir vec as going "up" and the sweep vec as going right
+  if (divider < 0.0) {
+    for (i = 1; i < gvf_survey.poly_count; i++)
+      if ((gvf_survey.dir_vec.x * (waypoints[gvf_survey.poly_first + i].y - small.y)) + (gvf_survey.dir_vec.y *
+          (small.x - waypoints[gvf_survey.poly_first + i].x)) > 0.0) {
+        VECT2_COPY(small, waypoints[gvf_survey.poly_first + i]);
+      }
+  } else
```
flixr: please add brackets to the else clause (or run the fix_code_style script over the file)

sw/airborne/modules/guidance/gvf/trajectories/gvf_line.c
```diff
@@ -30,11 +30,28 @@
...
 #include "gvf_line.h"
 
-#ifndef GVF_LINE_ALPHA
-#define GVF_LINE_ALPHA 0
+#ifndef GVF_LINE_KE
+#define GVF_LINE_KE 1
 #endif
 
-gvf_li_par gvf_line_par = {GVF_LINE_ALPHA};
+#ifndef GVF_LINE_KN
```
flixr: can you please add a doxygen description to these defines?

sw/airborne/modules/guidance/gvf/trajectories/gvf_line.h
```diff
@@ -32,10 +32,18 @@
 #include "modules/guidance/gvf/gvf.h"
 
 typedef struct {
-  float alpha;
+  float ke;
+  float kn;
+  float heading;
 } gvf_li_par;
 
+typedef struct {
```
flixr: same: please add doxygen inline comments for the struct members variables

sw/ground_segment/python/gvf/gvfFormation.py
```diff
@@ -53,13 +53,16 @@ def message_recv(ac_id, msg):
...
-    no_telemetry = 0
     for ac in list_aircraft:
-        if ac.a == -999 or ac.XY[0] == -999:
-            print "Waiting for telemetry of aircraft ", ac.id
-            no_telemetry = 1
-
-    if no_telemetry:
+        waiting_for_msgs = 0
+        if ac.a == -999:
+            print "Waiting for GVF msg of aircraft ", ac.id
```
flixr: please add `from __future__ import print_function` at the beginning and use the print function, so this also works with python3

noether:
thanks a lot Felix for your time! I will take a look at your comments tomorrow.

noether:
done, I have added also extra doxygen comments. Let me know whether they are properly presented or not, since it is the first time for me.

sw/ground_segment/python/gvf/gvfFormation.py
```diff
@@ -57,10 +57,10 @@ def formation(B, ds, radius, k):
     for ac in list_aircraft:
         waiting_for_msgs = 0
         if ac.a == -999:
-            print "Waiting for GVF msg of aircraft ", ac.id
+            print ("Waiting for GVF msg of aircraft ", ac.id)
```
flixr: in python you usually don't put a space between the function name and the opening bracket...

sw/airborne/modules/guidance/gvf/gvf.h
```diff
@@ -32,6 +32,13 @@
 
 #include "std.h"
 
+/** \typedef gvf_conf
```
flixr: I think this would work, although I'm not sure about the `\param` keyword... also we usually use the `@` javadoc notation, see http://docs.paparazziuav.org/latest/styledoxygen.html

You could write this in an easier way though:
```
/**
 * Parameters for the GVF
 */
typedef struct {
   float ke; ///< Gain defining how agressive is the vector field
   float kn; ///< Gain for making converge the vehile to the vector field
   float error; ///< Error signal (unit ?)
   int8_t s; ///< Right or left direction (and how is left/right encoded?)
} gvf_con;
```

Also please add the units of the variables in the description.

sw/ground_segment/python/gvf/gvfFormation.py
noether: you are right yep

sw/airborne/modules/guidance/gvf/gvf.h
noether: cool, thanks for the link. I will do it as you suggest then, probably on Sunday or Monday.

By the way, I have written the units where it was possible. Here in particular, the error signal does not have any units that make sense. In fact, they are just level sets https://en.wikipedia.org/wiki/Level_set .

The direction also does not have any preference, it is just 1 or -1 and depending on how the trajectory is defined and what are the parameters, could be left or right. I will also add a miniexplanation in the comment about this.

podhrmic:
@noether any updates here?

noether:
Not really, 

the updates are the ones described at the beginning. The next 'big' update will be the integration of the gvf  in the GCS, i.e., to draw the trajectories and the predicted track to be followed by the aircraft. However,  I have to learn first a bit of ocaml :P.

conf/modules/gvf_module.xml
```diff
@@ -7,65 +7,54 @@ For more details we refer to https://wiki.paparazziuav.org/wiki/Module/guidance_
     </description>
   </doc>
```
gautierhattenberger: add description of the parameters

conf/modules/gvf_module.xml
```diff
@@ -7,65 +7,54 @@ For more details we refer to https://wiki.paparazziuav.org/wiki/Module/guidance_
...
-          <dl_setting MAX="90" MIN="-90" STEP="1" VAR="gvf_ellipse_par.alpha" shortname="alpha" param="GVF_ELLIPSE_ALPHA"/>
+          <dl_setting MAX="5" MIN="0.0" STEP="0.01" VAR="gvf_ellipse_par.ke" shortname="ell_ke" param="GVF_ELLIPSE_KE"/>
+          <dl_setting MAX="5" MIN="0.0" STEP="0.01" VAR="gvf_ellipse_par.kn" shortname="ell_kn" param="GVF_ELLIPSE_KN"/>
+          <dl_setting MAX="150" MIN="0.0" STEP="10" VAR="gvf_ellipse_par.a" shortname="ell_a" param="GVF_ELLIPSE_A"/>
+          <dl_setting MAX="150" MIN="0.0" STEP="10" VAR="gvf_ellipse_par.b" shortname="ell_b" param="GVF_ELLIPSE_B"/>
+          <dl_setting MAX="90" MIN="-90" STEP="1" VAR="gvf_ellipse_par.alpha" shortname="ell_alpha" param="GVF_ELLIPSE_ALPHA"/>
         </dl_settings>
           <dl_settings NAME="Line">
-          <dl_setting MAX="180" MIN="-180" STEP="1" VAR="gvf_line_par.alpha" shortname="alpha" param="GVF_LINE_ALPHA"/>
+		  <dl_setting MAX="5" MIN="0.0" STEP="0.01" VAR="gvf_line_par.ke" shortname="line_ke" param="GVF_LINE_KE"/>
```
gautierhattenberger: check indent in this file

sw/airborne/modules/guidance/gvf/gvf.h
```diff
@@ -76,23 +99,28 @@ struct gvf_Hess {
...
-extern void gvf_set_gains(float ke, float kd);
 extern void gvf_set_direction(int8_t s);
 
 // Straigh line
-void gvf_line(float x, float y, float alpha);
+extern bool gvf_line_XY_heading(float x, float y, float heading);
+extern bool gvf_line_XY1_XY2(float x1, float y1, float x2, float y2);
 extern bool gvf_line_wp1_wp2(uint8_t wp1, uint8_t wp2);
-extern bool gvf_line_wp_heading(uint8_t wp, float alpha);
+int out_of_segment_area(float x1, float y1, float x2, float y2, float d1, float d2);
```
gautierhattenberger: this could be static only in the C file

sw/airborne/modules/guidance/gvf/gvf.c
```diff
@@ -196,38 +227,104 @@ bool gvf_line_wp1_wp2(uint8_t wp1, uint8_t wp2)
...
+  float y2 = waypoints[wp2].y;
+
+  int s = out_of_segment_area(x1, y1, x2, y2, d1, d2);
+  if (s != 0) {
+    gvf_control.s = s;
+  }
+
+  gvf_line_XY1_XY2(x1, y1, x2, y2);
+  return true;
+}
```
noether: For the segment functions, I have renamed them to segment_loop. For the next update I will add the 'segment function' so it will return false once the segment is finished (I need to think how to implement the logic for this new function).

noether:
It seems that the conflict with pprzlink has been solved by merging this branch with the latest paparazzi/master.

noether:
I do not understand why I still have 'conflicts' with pprzlink mmmm. I have synchronized this branch with the latest pprz/master and made a 'make clean all' (so I know it is pointing to the correct commit of pprzlink).

Last time I have solved it with a 'big rebase', which was an ugly solution.... any ideas?

podhrmic:
@noether sorry I had to make some small changes in pprzlink

noether:
@podhrmic  haha, no worries man :P. Next time I know how to fix such an issue without too much mess .

podhrmic:
@noether apparently there are still some requested changes from @flixr regarding the docs and the code indentation. Could you please address those so I can merge this?

noether:
@podhrmic 

I think I have already addressed the points of @flixr. Please, let me know whether this is not the case and I will add the requested docs.

By the way, last week a friend at ENAC has found a small bug in this pull request. It is not important but it can be annoying when one uses the poly_survey with the GVF. I will fix it this week and then update this pull request.

If you are ok with this upcoming update, then we can merge it :P.

noether:
Oki, I think this is the last time I add new features in this pull request.

I have added new primitives for following straight lines (endless lines, segments and 'segment loops'). The GVF wiki has been updated accordingly with the new available functions for the user.

Let me know whether the style or other things are ok for you, or whether something needs to be better documented in the code.

Issue #2040, opened 2017-03-19
Pixhawk fwb upload unreliable


podhrmic:
on Ubuntu 16.04 with the latest master, I am unable to flash the FBW through AP (using the `px4flash` module). It either receives ERROR from the AP, or freezes completely (have to restart the pprzcenter).

Tried on several Pixhawks, the same results.

When flashing the FWB with a JTAG (Black Magic probe), everything indeed works fine - so I am wondering if there is something with the `px4flash` module?

kevindehecker:
I'm having the same problem. As a work around; if you time it just right you can still upload to FBW. For me it is just after the second slow blink of the AP sys time led, but it depends how much you still need to build and how fast your laptop is.
If it didn't work make sure to kill the upload script process manually before attempting again (e.g. killall python ), because pressing stop from pprz center won't do it.

I'll try to look into this probably in two weeks or so.

podhrmic:
Cool, thanks @kevindehecker

podhrmic:
@kevindehecker any updates here?

kevindehecker:
I gave it a try, tried a few things,, but could not fix the problem. Then 2 weeks holiday :) Hopefully in the next few days I'll have time again.

OpenUAS:
Having Pixhawk, Pixracer and various PX clones, and not using JTAG on them we know annoyance: Timing the USB plug-in just in time etc.etc. As developers, well.. we manage; but it is for users quite an issue we experience every time. The first experience with PPRZ should be smooth(er). If one cannot upload without to much hassle, it is quite frustrating and users tend to look elsewhere for an open-source AP solution. So a little higher status label ;)

podhrmic:
Seems to be fixed with #2057 so closing.

PR #2057, opened 2017-04-25
Fix px4 fbw flash problem, related to #2040

kevindehecker:
There were some timing issues that disturbed communication when flashing the fbw side of px4.

### DIFF
--- a/sw/airborne/modules/px4_flash/px4_flash.c
+++ b/sw/airborne/modules/px4_flash/px4_flash.c
@@ -72,7 +72,7 @@ void px4flash_init(void)
   setToBootloaderMode = false;
 #ifdef INTER_MCU_AP
   px4ioRebootTimeout = false;
-  px4iobl_tid = sys_time_register_timer(15.0, NULL); //20 (fbw pprz bl timeout)-5 (px4 fmu bl timeout)
+  px4iobl_tid = sys_time_register_timer(12.0, NULL); //20 (fbw pprz bl timeout)-5 (px4 fmu bl timeout) - 3 (uncertainty and random delays)
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
It looks fine, would be nice if someone else can try to confirm that it is working better. @OpenUAS ?

podhrmic:
I ll test it today and let you know.

agressiva:
Tested here
working much better.

I uploaded de fbw code 10 times and get only one error.

podhrmic:
Works great here too. Weird that it was such a small issue, but this definitely improved the uploading experience. Thanks @kevindehecker

kevindehecker:
For future reference, some details. It's a combination of factors really.

The main problem was that the FBW side set the baud rate to slow 10s before the AP side, so they could not communicate properly but the AP side did not know.  The total flash window is also about 10s....

So, if fbw upload was attempted, you really had to time it perfectly just after the AP side booted but before the FBW side was set to slow baud yet (approx 0.5s window). If you missed it, the AP side could respond to the upload script with "FBWOKOK"(while actually things were not OK) and enabled the pass-through communication to the fbw side. The px4 upload script was then started, but could not get the correct responses anymore from the fbw side and starts an endless loop of retries. 

The second problem is that the output of the px4 upload python script is only printed when the script exits, which never happened in this case, So in pprz Center it seemed like it just froze.

Then the third problem is that when manually stopping make, the px4 python script is not killed with it automatically (needs to be killed manually). So, when another attempt is made to upload, the previous script is still trying simultaneously. Creating the total mass that we observed.

This PR fixed the first main problem, but the second and third are still there. (But they are not problems now)

PR #2059, opened 2017-05-04
OpenOCD moved the location of scripts. This patch fixes the defaults.

esden:
This is not tested as I do not have a setup to test this at the moment. @agressiva pointed out that this is an issue. Maybe he can confirm that this patch solves his issue.

### DIFF
--- a/conf/Makefile.chibios
+++ b/conf/Makefile.chibios
@@ -343,8 +343,8 @@ GDB = $(shell which arm-none-eabi-gdb)
 
 
 # Settings for OOCD
-$(TARGET).OOCD_INTERFACE=flossjtag
-OOCD_INTERFACE=flossjtag
+$(TARGET).OOCD_INTERFACE=ftdi/flossjtag
+OOCD_INTERFACE=ftdi/flossjtag
 OOCD = $(shell which openocd)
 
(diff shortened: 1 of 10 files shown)


agressiva:
it worked.
tested with lisa_m 1.0 and floss_jtag

flixr:
Hm... if they changed it, we would need to set this dependent on the OpenOCD version...

esden:
I think they changed it already several versions ago. As so few people seem to be using openocd programming in general, adding additional logic to choose the script location seems superflous to me. Especially because it only applies to the interfaces that use an FTDI chip and we use interfaces that are not FTDI based for some of the boards.

flixr:
At least we should find out since which version this is the case and on which distris the older one is still in use...

esden:
Just checked the interface/ftdi script subdirectory and scripts have been added in OpenOCD V0.6.0 in 2012. I think it is safe to consider OpenOCD V0.5.0 from 2011 as obsolete. I would hope that 5 years is enough time for everyone to update their packages. :D

podhrmic:
@flixr shall we merge as is?

flixr:
Merging as Ubuntu trusty already has openocd 0.7.0

PR #2060, opened 2017-05-06
Fixed copilot module

podhrmic:
Now it works both with rotorcraft and fixedwing firmwares.

The problem is that the waypoints are handled differently in each, and Copilot has to process the WP_MOVED message.

I wanted to take care of this at the makefile level, rather than having a bunch of `ifdefs` but I needed to pass the information about which firmware is being used (hence added the `FIRMWARE` variable in the makefile). It would be nice if this was eventually handled by the generators and I didn't have to use the `<raw>` section in the module xml file, but it works fine for now.


podhrmic:
Merging as I need it for other work and the changes don't affect anything besides the copilot module. I am happy to revisit this later.

conf/modules/copilot.xml
```diff
@@ -35,7 +35,27 @@
   <datalink message="COPILOT_STATUS_DL" fun="copilot_parse_copilot_status_dl(buf)"/>
   <datalink message="MOVE_WP" fun="copilot_parse_move_wp_dl(buf)"/>
   <makefile target="ap">
-    <file name="copilot.c"/>
+    <file name="copilot_common.c"/>
+      <raw>
+        # attach the correct file based on firmware
+        RESULT=0
+
+        ifneq (,$(findstring ROTORCRAFT,$($(TARGET).CFLAGS)))
```
gautierhattenberger: there is a `firmware` attribute for the `makefile` node in order to do this

podhrmic: I noticed it too late, fixed it with https://github.com/paparazzi/paparazzi/commit/e30a03cc0073b55a5a2ff59fe4ec298c810d22d0

PR #2061, opened 2017-05-06
Added test for aggieair config

podhrmic:
Added a test for aggie_air conf. It really helps me with development, and it does't hurt anyone.

Indeed, if each developer has their own conf test it would be probably too much, but I don't think that is likely to happen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -308,6 +308,10 @@ test: test_math test_examples
 test_coverity: all
 	CONF_XML=conf/conf_tests_coverity.xml prove tests/aircrafts/
 
+# test AggieAir conf
+test_aggieair: all
+	CONF_XML=conf/airframes/AGGIEAIR/aggieair_conf.xml prove tests/aircrafts/
+
 # compiles all aircrafts in conf_tests.xml
 test_examples: all
 	CONF_XML=conf/conf_tests.xml prove tests/aircrafts/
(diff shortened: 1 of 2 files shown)


podhrmic:
@flixr are you cool with this?

gautierhattenberger:
I don't mind adding this but what is the reason ? you want to run this on a CI server ?

flixr:
I also don't really mind one more test target, but I want to avoid the situation where we start adding this for every conf...
For my conf I simply made an alias in my .zshrc (for you probably .bashrc) to avoid having to type so much...

podhrmic:
@flixr would you mind sharing an example?

flixr:
Added to my `.zshrc`:
```
alias pprz_flixr='CONF_XML=conf/airframes/FLIXR/flixr_conf.xml prove tests/aircrafts/'
```
I think for bash the alias definition is the same....

podhrmic:
And then how do you use it? `make pprz_flixr` doesn't' seem to work for me:-/

flixr:
just `pprz_flixr`... it's a stupid name, I should have better called it something like `test_flixr` or so...
You type in a shell alias as if it was a command, and the shell will simply replace it by what you set the alias to...

flixr:
So we can close this then?

podhrmic:
Actually I would prefer to merge this - you need to set up your PPRZ_HOME and PPRZ_SRC to get it working, and I have multiple clones of pprz on my computer, so it would be harder to set it up. Is that cool?

flixr:
Fair enough...

PR #2062, opened 2017-05-06
First shot towards a separate HITL target

podhrmic:
@gautierhattenberger I am trying to add HITL as a separate target so it is easier to use.

The idea is that `hitl` is de facto `nps` target, except it has a different name (and uses a few different files). So I want to have `nps_common.makefile` and then `nps.makefile` and `hitl.makefile` (see below).

Then I have to have a toplevel `Makefile.hitl` that only calls `Makefile.nps` and that should be it - right?

I can'get it to compile, I suspect there is some change with generators that I have to add the target to?


gautierhattenberger:
at least you need to add `hitl` to the list of default possible targets in sw/lib/ocaml/env.ml

podhrmic:
@gautierhattenberger thanks for the pointer. I changed it, but I am still not seeing `hitl.makefile` being called - if I run HITL target I see:
```
...
make[1]: Leaving directory '/home/podhrad/Documents/Paparazzi/paparazzi'
make[1]: Entering directory '/home/podhrad/Documents/Paparazzi/paparazzi'
>>>Calling nps.makefile
>>>Calling nps_common.makefile
GENERATE /home/podhrad/Documents/Paparazzi/paparazzi/var/aircrafts/Minion_RP3/hitl/generated/airframe.h from airframes/AGGIEAIR/aggieair_minion_rp3_lia.xml
...
```

even though the target is `hitl` , not `nps`

Then indeed the compilation fails.

Any ideas? I added the `hitl` target to `env.ml`

gautierhattenberger:
correct makefile selection is coming from this line: https://github.com/paparazzi/paparazzi/blob/master/conf/firmwares/subsystems/shared/nps.makefile#L17

podhrmic:
I specify it in `hitl.makefile` (see https://github.com/paparazzi/paparazzi/blob/hitl_updates/conf/firmwares/subsystems/shared/hitl.makefile#L13 ) but the problem is that `hitl.makefile` is never called in the first place.

podhrmic:
`$(info >>>Calling hilt.makefile)` never shows up in the console, instead I can see `https://github.com/paparazzi/paparazzi/blob/hitl_updates/conf/firmwares/subsystems/shared/nps.makefile#L9` even when I call `hitl` target

podhrmic:
@gautierhattenberger do you have an idea why compiling `hitl` target would still call `conf/firmwares/subsystems/shared/nps.makefile` first?

flixr:
@podhrmic nps.makefile is always added/called from rotorcraft.makefile...

flixr:
Use the `--debug` option of make to find what makefiles are parsed, etc...
e.g. ` make --debug=v AIRCRAFT=Ark_Hexa_1.8 hitl.compile`

podhrmic:
Thanks for help, I figured it out. Have to do some cleanup still, but it works now.

podhrmic:
@flixr @gautierhattenberger would you be OK with merging this as-is, so I can update the wiki page about HITL and have a separete `hitl` target which makes the whole setup more user-friendly?

It needs some polishing, but it doesn't break anything and works just fine for what it is - unfortunately I won't have time to sit down and properly finish it for a couple of weeks, and I would like to avoid merge conflicts (plus I would like to have the wiki page updated).

flixr:
Can you at least take out the "calling xx makefile" infos?

podhrmic:
OK, never mind - probably better to merge it once it is clean.

PR #2066, opened 2017-05-23
Succesfully tested Disco at v1.3

kevindehecker:


AlessioDurante14:
Does it work also with version 1.4?
Is there an update which allows to work with Parrot Disco at v1.4?
Thank you very much

PR #2070, opened 2017-06-01
Added a Region of Interest feature to the fast9 corner detector and fixed a reallocation bug

manoskyr:
- Bugfix : If the reallocation of the `ret_corners` array caused it to be moved in memory, the calling function would lose track of the new corners because the pointer was only passed by value into the `fast9_detect()`. Now it is passed by reference to avoid this.

- Minor change : Local variable `corner_cnt` of `fast9_detect()` changed to `uint16_t` instead of `uint32_t`.

- _RoI_ feature not used in optic flow calculations for now, but will be when feature management is implemented.

### DIFF
--- a/sw/airborne/modules/computer_vision/lib/vision/fast_rosten.c
+++ b/sw/airborne/modules/computer_vision/lib/vision/fast_rosten.c
@@ -46,12 +46,12 @@ static void fast_make_offsets(int32_t *pixel, uint16_t row_stride, uint8_t pixel
  * @param[in] y_padding The padding in the y direction to not scan for corners
  * @param[in] *num_corners reference to the amount of corners found, set by this function
  * @param[in] *ret_corners_length the length of the array *ret_corners.
- * @param[in] *ret_corners array which contains the corners that were detected.
+ * @param[in] **ret_corners pointer to the array which contains the corners that were detected.
  * @param[in] *roi array of format [x0 y0 x1 y1] describing the region of interest in the image where the corners will be detected. If null, the whole image is used.
 */
-void fast9_detect(struct image_t *img, uint8_t threshold, uint16_t min_dist, uint16_t x_padding, uint16_t y_padding, uint16_t *num_corners, uint16_t *ret_corners_length,struct point_t *ret_corners, uint16_t *roi) {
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/computer_vision/lib/vision/fast_rosten.c
```diff
@@ -46,30 +46,45 @@ static void fast_make_offsets(int32_t *pixel, uint16_t row_stride, uint8_t pixel
...
+	  y_end =img->h - 3 - y_padding;
+  }
+  else{
+	  x_start = roi[0] > 0 ? roi[0] : 3 + x_padding;
+	  y_start = roi[1] > 0 ? roi[1] : 3 + y_padding;
+	  x_end = roi[2] < (img->w - 3 - x_padding) ? roi[2] : img->w - 3 - x_padding;
+	  y_end = roi[3] < (img->h - 3 - y_padding) ? roi[3] : img->h - 3 - y_padding;
+
+  }
+
```
kirkscheper: The spacing here is incorrect, run fix_code_style.sh on your edited files.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
```diff
@@ -279,10 +279,12 @@ void calc_fast9_lukas_kanade(struct opticflow_t *opticflow, struct opticflow_sta
   // FAST corner detection
   // TODO: There is something wrong with fast9_detect destabilizing FPS. This problem is reduced with putting min_distance
   // to 0 (see defines), however a more permanent solution should be considered
+  // last parameter (for ROI detection) set to NULL because feature management is not implemented yet.
   fast9_detect(img, opticflow->fast9_threshold, opticflow->fast9_min_distance,
                opticflow->fast9_padding, opticflow->fast9_padding, &result->corner_cnt,
                &opticflow->fast9_rsize,
-               opticflow->fast9_ret_corners);
+               &opticflow->fast9_ret_corners,
+			   0);
```
kirkscheper: Should use NULL for pointers instead of 0, just like you have it in the comments. It is to reinforce that is a pointer input.

sw/airborne/modules/computer_vision/lib/vision/fast_rosten.c
manoskyr: Did not know about that, sorry.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
manoskyr: Yeap, you are write, missed that one. Changes commited.

PR #2075, opened 2017-06-14
WLS control allocation

EwoudSmeur:
This is a pull request for a control allocation scheme, made by Daan Höppener, a student of mine. It allows you to specify priorities for different control objectives through a weighting in the cost function. This is why it has weighted least squares (WLS) in the name. Prioritizing parts of the control objective (pitch and roll over yaw for instance) is important for stable flight when there are saturations. Normally, the optimal control input can be found with the pseudo-inverse. However, when there are saturations, WLS-control allocation will satisfy the important parts of the control objective at the cost of the unimportant parts.

The optimal solution for this cost function is found using the active set method, which requires solving multiple least squares problems. For this we are using a (trimmed down) library from John Burkardt (http://people.sc.fsu.edu/~jburkardt/cpp_src/r8lib/r8lib.html). I wasn't sure what header we need for these files.

Probably I can still improve some stuff here and there, but I wanted to get the ball rolling with this pr...

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_bebop_indi_actuators.xml
+++ b/conf/airframes/TUDELFT/tudelft_bebop_indi_actuators.xml
@@ -100,7 +100,7 @@
     <!-- setpoint limits for attitude stabilization rc flight -->
     <define name="SP_MAX_PHI" value="45" unit="deg"/>
     <define name="SP_MAX_THETA" value="45" unit="deg"/>
-    <define name="SP_MAX_R" value="120" unit="deg/s"/>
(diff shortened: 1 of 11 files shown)


podhrmic:
Exciting! For the header - the library is GPL, so perhaps copying their licence info there and a link to the source would be good?

EwoudSmeur:
Hi @OpenUAS 
Thanks for merging, but I think there has not been the chance for a good review. Not many people look into this code on a regular basis, so these pull requests are one of the few occasions for me to get feedback, and for others to object to anything I am doing.
I am thinking we should maybe revert and give the review another shot...

PR #2083, opened 2017-06-22
[pprzlink] allow to compile paparazzi with PPRZLINK v2

gautierhattenberger:
To activate, build project with correct option:
`PPRZLINK_LIB_VERSION=2.0 make`
(note that `make clean` is usualy required before that)

At the moment, only a few messages in datalink.c are using explicitly the new format. More examples will come later, in particular regarding inter UAV communications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,11 @@ LOGALIZER=sw/logalizer
 SUBDIRS = $(PPRZCENTER) $(MISC) $(LOGALIZER) sw/tools
 
 #
+# Communication protocol version
+#
+PPRZLINK_LIB_VERSION ?= 1.0
+
+#
 # xml files used as input for header generation
 #
 CUSTOM_MESSAGES_XML = $(CONF)/messages.xml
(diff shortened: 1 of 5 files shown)


podhrmic:
@gautierhattenberger  I ll look at it next week.

dewagter:
Dear @gautierhattenberger,

Thank you for the nice work. This vehicle to vehicle communication + more message ID's is an important step.

I could not review all the code changes in detail.

It does look like this brings us a lot closer to the mavlink protocol and I was wondering: if PPRZ1 and PPRZ2 protocol are not compatible anyway due to extra header bytes, could we not also add the counter, and swap the header byte so PPRZ2 would be parsable by mavlink decoders? From there on the step to harmonizing messages (content) would be smaller.

gautierhattenberger:
@dewagter Our first try on this new protocol was to be as close as possible to mavlink, but in the end, it wasn't really convincing:
- we have the class ID that they don't have at all and that is very important to extend the number of messages and avoid ambiguity
- having the receiver ID in the header (so for all messages) is really interesting I think (for multi-uav communications) instead of adding it as a field to some messages (`ac_id` fields could be removed in the future)
- we are closer to mavlink1 but they are now working on mavlink2 which is much more complicated with functionality that we don't handle at the moment
- I don't think mixing messages definitions and protocols is very safe. I think we should focus on a better support of multiple message systems rather than trying to merge them (I personally don't like much mavlink messages, and they certainly don't care about integrating our messages in their standard set)
- I don't want to add more than 2 bytes to keep it small
- I placed the sender and receiver ID at the beginning as this part could in the future be integrated in the transport layer as a 'service' (like usual network stuff, actually xbee and udp transport already hold this information)

To sum up, I think we should focus on improving our own system without adding to much external constraints, and improve the support of other communication systems and middlewares separately.

dewagter:
@gautierhattenberger 

Thank you for this information. It is great that you looked into this.

I do see the need for more messages and the use of target id (which is already in the xbee_api but not in others and not used air-to-air yet as far as I know). Also keeping it small is nice.

Seeing some drone hacking stuff going on lately, maybe it's also nice to have a low-profile protocol.

So this seems to be the best option.

PR #2084, opened 2017-06-26
[module] add support for Decawave DW1000 modules

gautierhattenberger:
Implement trilateration algorithm for positionning using the
UltraWideBand modules DW1000 from Decawave.
The interface with the modules is done with an arduino using the low
level library https://github.com/thotro/arduino-dw1000 with some
modifications.
The data extraction and trilateration are working, calibration and real
flight have not been done at the moment.


gautierhattenberger:
The modified arduino library: https://github.com/enacuavlab/arduino-dw1000/tree/pprz

PR #2087, opened 2017-07-09
upgrades to optical flow landing module

kirkscheper:
This cleans up the optical flow module a bit and adds more options to be set from the airframe file. Also moves the computation of the covariance and mean to the statistics file in the math library. I also cleaned up a warning in the optical flow calculator with the shadow variable name 'state'. @guidoAI can you take a look at this?

@gautierhattenberger I made a stat.c file and added it to the rotorcraft makefile. Is this alright? I thought it is better if it compiled than in the header...

### DIFF
--- a/conf/firmwares/rotorcraft.makefile
+++ b/conf/firmwares/rotorcraft.makefile
@@ -80,7 +80,7 @@ endif
 # Math functions
 #
 ifneq ($(TARGET), fbw)
-$(TARGET).srcs += math/pprz_geodetic_int.c math/pprz_geodetic_float.c math/pprz_geodetic_double.c math/pprz_trig_int.c math/pprz_orientation_conversion.c math/pprz_algebra_int.c math/pprz_algebra_float.c math/pprz_algebra_double.c
+$(TARGET).srcs += math/pprz_geodetic_int.c math/pprz_geodetic_float.c math/pprz_geodetic_double.c math/pprz_trig_int.c math/pprz_orientation_conversion.c math/pprz_algebra_int.c math/pprz_algebra_float.c math/pprz_algebra_double.c math/pprz_stat.c
 
 $(TARGET).srcs += subsystems/settings.c
 $(TARGET).srcs += $(SRC_ARCH)/subsystems/settings_arch.c
--- a/sw/airborne/math/pprz_stat.c
+++ b/sw/airborne/math/pprz_stat.c
@@ -41,7 +41,8 @@ int32_t mean_i(int32_t *array, uint32_t n_elements)
 {
   // determine the mean for the vector:
   float sum = 0.;
-  for (uint32_t i = 0; i < n_elements; i++) {
+  uint32_t i;
+  for (i = 0; i < n_elements; i++) {
     sum += (float)array[i];
   }
 
@@ -78,7 +79,8 @@ int32_t covariance_i(int32_t *array1, int32_t *array2, uint32_t n_elements)
   float sumX = 0., sumY = 0., sumXY = 0;
 
   // Determine the covariance:
-  for (uint32_t i = 0; i < n_elements; i++) {
+  uint32_t i;
+  for (i = 0; i < n_elements; i++) {
     sumX += (float)array1[i];
     sumY += (float)array2[i];
     sumXY += (float)(array1[i]) * (float)(array2[i]);
@@ -101,7 +103,8 @@ float mean_f(float *array, uint32_t n_elements)
 {
   // determine the mean for the vector:
   float sum = 0.;
-  for (uint32_t i = 0; i < n_elements; i++) {
+  uint32_t i = 0;
+  for (i = 0; i < n_elements; i++) {
     sum += array[i];
   }
 
@@ -137,7 +140,8 @@ float covariance_f(float *arr1, float *arr2, uint32_t n_elements)
   float sumX = 0., sumY = 0., sumXY = 0;
 
   // Determine the covariance:
-  for (uint32_t i = 0; i < n_elements; i++) {
+  uint32_t i;
+  for (i = 0; i < n_elements; i++) {
     sumX += arr1[i];
     sumY += arr2[i];
     sumXY += arr1[i] * arr2[i];
--- a/sw/airborne/math/pprz_stat.c
+++ b/sw/airborne/math/pprz_stat.c
@@ -1,0 +1,147 @@
+/*
+ * Copyright (C) 2017 Kirk Scheper
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file pprz_array_float.c
+ * @brief Array funcitons.
+ *
+ */
+
+#include "pprz_stat.h"
+
+/*********
+ * Integer implementations
+ *********/
+
+/**
+ * Get the mean value of an array
+ * This is implemented using floats to handle scaling of all variables
+ * @param[out] mean The mean value
+ * @param[in] *array The array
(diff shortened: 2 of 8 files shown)


conf/firmwares/rotorcraft.makefile
```diff
@@ -80,7 +80,7 @@ endif
 # Math functions
 #
 ifneq ($(TARGET), fbw)
-$(TARGET).srcs += math/pprz_geodetic_int.c math/pprz_geodetic_float.c math/pprz_geodetic_double.c math/pprz_trig_int.c math/pprz_orientation_conversion.c math/pprz_algebra_int.c math/pprz_algebra_float.c math/pprz_algebra_double.c
+$(TARGET).srcs += math/pprz_geodetic_int.c math/pprz_geodetic_float.c math/pprz_geodetic_double.c math/pprz_trig_int.c math/pprz_orientation_conversion.c math/pprz_algebra_int.c math/pprz_algebra_float.c math/pprz_algebra_double.c math/pprz_stat.c
```
gautierhattenberger: I suggest to also add this to `firmwares/subsystems/fixedwing/autopilot.makefile` so that both firmwares have the same "math" lib available.

sw/airborne/math/pprz_stat.c
```diff
@@ -0,0 +1,151 @@
...
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file pprz_array_float.c
+ * @brief Array funcitons.
+ *
+ */
+
+#include "pprz_stat.h"
```
gautierhattenberger: use "full" path `math/pprz_stat.h`

sw/airborne/math/pprz_stat.c
```diff
@@ -0,0 +1,151 @@
...
+ * Get the mean value of an array
+ * This is implemented using floats to handle scaling of all variables
+ * @param[out] mean The mean value
+ * @param[in] *array The array
+ * @param[in] n Number of elements in the array
+ */
+int32_t mean_i(int32_t *array, uint32_t n_elements)
+{
+  // determine the mean for the vector:
+  float sum = 0.;
```
gautierhattenberger: even better, initialize with `0.f`, same for other functions

sw/airborne/math/pprz_stat.h
```diff
@@ -33,47 +33,13 @@ extern "C" {
...
-  float me = 0.;
-  float see = 0.;
-  for (int i = 0; i < nb; i++) {
-    me += array[i];
-    see += array[i] * array[i];
-  }
-  me /= nb;
-  return (see / nb - me * me);
-}
+extern int32_t mean_i(int32_t *array, uint32_t n_elements);
```
gautierhattenberger: I would keep the doxygen comments for the header as well, although I don't know how doxygen is extracting the information

gautierhattenberger:
@kirkscheper do you need more review from @guidoAI or can I merge this ?

kirkscheper:
I think I will wait till @guidoAI has time to review this as it is his code. Although this is flight tested, I want to make sure that what I changed is still what he intended the module to do.

sw/airborne/modules/ctrl/optical_flow_landing.c
```diff
@@ -41,127 +41,151 @@
...
+
+// struct containing most relevant parameters
+struct OpticalFlowLanding of_landing_ctrl;
 
 // sending the divergence message to the ground station:
 static void send_divergence(struct transport_tx *trans, struct link_device *dev)
 {
   pprz_msg_send_DIVERGENCE(trans, dev, AC_ID,
-                           &divergence, &divergence_vision_dt, &normalized_thrust,
+                           &(of_landing_ctrl.divergence), &divergence_vision_dt, &of_landing_ctrl.nominal_thrust,
```
guidoAI: Nominal thrust was supposed to be a constant, while normalized thrust would be between 0 (minimum thrust available to the controller) and 1 (maximum thrust allowed) - which is hence varying during control. I would prefer normalized thrust here. 

divergence is low-passed. Is of_landing_ctrl.divergence as well? We don't want copies of the divergence from the vision, we want the divergence that is used by control.

sw/airborne/modules/ctrl/optical_flow_landing.c
```diff
@@ -171,434 +195,326 @@ void vertical_ctrl_module_run(bool in_flight);
...
       }
-
-      // low-pass filter the divergence:
-      divergence = divergence * of_landing_ctrl.lp_factor + (new_divergence * (1.0f - of_landing_ctrl.lp_factor));
-      previous_message_nr = vision_message_nr;
-      dt = 0.0f;
-
-    } else {
-      // after re-entering the module, the divergence should be equal to the set point:
-      if (ind_hist <= 1) {
```
guidoAI: Was this moved or removed?

kirkscheper: Was removed. 

It was not clear to me why the divergence should be equal to the set point for one time step when re-entering the module, it would seem that setting the divergence to the set point then back to the real divergence (which would be close to 0 at the beginning of the landing) would not be good for the covariance estimation. 

Advise if that was not the idea.

sw/airborne/modules/ctrl/optical_flow_landing.c
kirkscheper: I just replaced the divergence with the of_landing.divergence so it should be the same. of_landing.divergence is used in the PID control function so I guess it is what you want right?

kirkscheper:
@guidoAI Due to the hardcoded corrections from the optical flow module on the bebop I can't test this module on a vehicle that is not a Bebop. I added some todo's to the code to make this module less dependent on the platform used (the best would be to have a helper function in the optical flow module to convert the output flow to scaled flow, in time and with this strange 1.28 factor). 

I did test fly a slightly altered version of this from kirkscheper:event_based_flow

ETitus:
Hi @kirkscheper, @guidoAI 

I might be wrong, but I was looking at the code and spotted something that might be incorrect. It involves cov_array_filled. 

Basically I think you need to wait for the window+delay instead of the only the window. While I think it shouldn't make a large difference in performance(the covariance is wrong for only 15 delay steps after starting the module,), it might make a difference if the module is started when the drone is not hovering stable.
The following example is with the standard window size of 30, and delay steps of 15:

cov_array_filled is set to true as soon as the ind_hist reaches 29, if I understand it correctly.

During the first 30 steps the divergence_history[ind_hist] will fill from index 0-29
Same goes for the past_divergence_history. However this will be filled using divergence_history[ind_past]
And ind_past will go from 15-29 0-14 (-15 till 0, but + windows size to avoid negatives)

Basically this will fill past_divergence_history half with zeros and the second half with actual values from divergence_history.
 
Now if I understand correctly, it wouldn't make sense to compute anything with this past_divergence_history array until it is also properly filled(see bold for change)

if (!cov_array_filled && ind_hist + 1 == of_landing_ctrl.window_size **+ of_landing_ctrl.delay_steps**) {
    cov_array_filled = true;
}

Titus

kirkscheper:
@ETitus  Thanks for the review. Nice find with this. I didn't use the second covariance method so never noticed this. Your recomened fix however wouldn't work as the value of ind_hist is wrapped on window_size so the logic statement you recommend would never evaluate true.

I have pushed an update which should fix the problem. Basically, if using the second covariance method you fill the array twice before using the arrays. Let me know if you have any future problems. Also let me know if you flight test this so that we can confirm everything is working as intended and we can merge this.

ETitus:
@kirkscheper, ah I didn't notice the wrapping of ind_hist, that is new too, and better indeed!
Your fix is quite elegant, as it makes sure that aslong as you are using the 1 array method, the past index isn't even used :+1: The only time that would give trouble is if someone would switch mid flight(and thus not reinitializing when reentering), though I assume that won't happen.

I am currently also working on improving my 3 axis version of OF hover and I will also try to fly this and cross check :)

A side note, I did find an (critical) error in the EF algorithm, that prevents EF divergence being used for my/this/any code exchangeable with LK. I haven't tried to tune the required values yet, but I will also try to do that for this module. Meanwhile, what would you say is the right approach, it is broader than this code, but creating a vision pull request while yours is still pending isn't the best approach I'd say.

Shall I update this pull request instead? Do I pull request your branch, or how do I add to this pull request?

kirkscheper:
I didn't consider the option of changing in flight... too many options here. How about this?

This commit shouldn't affect PR for the edgeflow so you can make your pull request now. Alternatively, if you test fly this code I can merge it asap so you can do your stuff.

guidoAI:
By the way, I tried to test the module on Friday. By the time I had a Bebop 2 working (long story), I could just test long enough to see that size_divergence was sometimes a NaN... which seems like a problem that I resolved long ago. I had no time to get into the details... but will do when back from holiday.

kirkscheper:
OK, seems like the nans issue is originated in the Lucas-Kanade optical flow and not here so I will merge this now.

PR #2088, opened 2017-07-16
tools/airframe_editor PEP8 and Python 3

guillaume-florent:
Hi. While reading the some Python code of the project, I thought I might contribute to PEP8 compliance and Python 3 compatibility. Willing to do more of it as I learn paparazzi and maybe enhance the Python coding standards document. Really hoping I can help on this great project !!

### DIFF
--- a/sw/tools/airframe_editor/airframe_editor.py
+++ b/sw/tools/airframe_editor/airframe_editor.py
@@ -1,30 +1,189 @@
 #!/usr/bin/env python
+# coding: utf-8
+
+r"""Airframe editor"""
 
 from __future__ import print_function
+
+from os.path import join
+
+# Own Modules
+import gui_dialogs
+import xml_airframe
+import paparazzi
 
 import pygtk
 import gtk
 pygtk.require('2.0')
 
-from os import path
-
-# Owm Modules
-import gui_dialogs
-import xml_airframe
-import paparazzi
-
 
 # Airframe File
-airframe_file = path.join(paparazzi.airframes_dir, "examples/quadrotor_lisa_m_2_pwm_spektrum.xml")
+airframe_file = join(paparazzi.airframes_dir,
+                     "examples/quadrotor_lisa_m_2_pwm_spektrum.xml")
 
 
-class AirframeEditor:
+class AirframeEditor(object):
+    def __init__(self):
+        self.window = gtk.Window(gtk.WINDOW_TOPLEVEL)
+        self.window.set_title("Paparazzi Airframe File Editor")
+
+        self.my_vbox = gtk.VBox()
+
+        # MenuBar
+        mb = gtk.MenuBar()
+
+        # File
+        filemenu = gtk.Menu()
+
+        # File Title
+        filem = gtk.MenuItem("File")
+        filem.set_submenu(filemenu)
+
+        openm = gtk.MenuItem("Open")
+        openm.connect("activate", self.open)
+        filemenu.append(openm)
+
+        exitm = gtk.MenuItem("Exit")
+        exitm.connect("activate", gtk.main_quit)
+        filemenu.append(exitm)
+
+        mb.append(filem)
+
+        # Help
+        helpmenu = gtk.Menu()
+
+        # Help Title
+        helpm = gtk.MenuItem("Help")
+        helpm.set_submenu(helpmenu)
+
+        aboutm = gtk.MenuItem("About")
+        aboutm.connect("activate", self.about)
+        helpmenu.append(aboutm)
+
+        mb.append(helpm)
+
+        self.my_vbox.pack_start(mb, False)
+
+        # #### Buttons
+        self.btnExit = gtk.Button("Exit")
+        self.btnExit.connect("clicked", self.destroy)
+        self.btnExit.set_tooltip_text("Close application")
+
+        self.btnOpen = gtk.Button("Open")
+        self.btnOpen.connect("clicked", self.open)
+
+        self.btnRun = gtk.Button("Reorganize XML")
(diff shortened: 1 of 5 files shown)


guillaume-florent:
Added some numpy style docstrings. C
I hope this is okay as I could not find a clearly defined standard for the docstrings.
Same question regarding unit tests as I could not locate them for Python. I would recommend pytest but would like to have your thoughts first.

gautierhattenberger:
Hello. Thank you for improving our code and making it compatible for Python3. It looks fine to me but I'm not an expert in Python's coding standards, maybe @flixr can have a comment on this ?

BTW, if you are looking for some improvements to do related to Python code in Paparazzi, we can give some pointers on on-going/unfinished part of the project.

guillaume-florent:
Sure, if you can give me the pointers to the python-related on-going/unfinished part of the project I can work on these parts in priority. Cheers.

flixr:
pytest is a good choice!
Regarding the docstring style: we don't have any rule there (yet), but we should probably add that to http://docs.paparazziuav.org/latest/stylepython.html

Personally I prefer the reST or google docstring style, as they are more compact... IMHO the google style is still more compact (at least in terms of vertical space) than the numpy style, while still being more readable than the reST style.
We can set sphinx to parse google style (or numpy) via an extension...

sw/tools/airframe_editor/airframe_editor.py
```diff
@@ -1,37 +1,204 @@
...
-from __future__ import print_function
+r"""Airframe editor"""
 
-import pygtk
-import gtk
-pygtk.require('2.0')
+from __future__ import print_function
 
-from os import path
+from os.path import join
```
flixr: Personally I prefer to import not only the `join` but rather `path` and use `path.join` as for me this makes it more clear in the code what is joined (a lot of packages/classes have join methods).
But this is my personal preference and not a strict rule or anything...

sw/tools/airframe_editor/xml_airframe.py
```diff
@@ -76,13 +92,18 @@ def reorganize_airframe_xml(airframe_xml):
     find_and_add_sections_with_name(airframe_xml, airframe, "XSENS")
 
     find_or_add_group(airframe_xml, airframe, "GAINS")
+
     # Fixedwing
-    find_and_add_sections_with_name(airframe_xml, airframe, "HORIZONTAL CONTROL")
+    find_and_add_sections_with_name(airframe_xml, airframe,
```
flixr: IMHO we don't need to enforce a 80char limit... 100 or even 120 is just as fine nowadays and IMHO makes it more readable in cases like this.

guillaume-florent: I understand and I had the same point of view and researched where the 80 chars limit came from. Apparently, it comes from visualizing 'side by side' several files in editor like emacs. But in many other IDEs, 100 or 120 is fine. Maybe 80 should be condidered as a very soft rule that can be broken if readability is better.

guillaume-florent: If that is okay with you, I will update the CONTRIBUTING.md file with elements from our discussion for Python and from info gleaned on the wiki (mostly for C). About OCaml (currently learning it), I found some standards (e.g. https://caml.inria.fr/resources/doc/guides/guidelines.en.html) but I have no idea which one is preferred on the project.

guillaume-florent: Gautier mentionned some Python parts of Paparazzi that need more attention. Do you know which ones?

gautierhattenberger: One of my student started a new design (in Python) for the Paparazzi Center two years ago. You can have a look at this page: http://wiki.paparazziuav.org/wiki/Paparazzi_Center/Evolutions
The code is here now: https://github.com/paparazzi/paparazzi/tree/master/sw/supervision/python
Several things could be done, starting by fixing the paths (it doesn't work if you don't start the program from its own directory). I also wanted to improve the design a bit, by hiding some panels when they are not needed/wanted for instance.

guillaume-florent: I have already had at look at that (to understand how the Paparazzi Center works) since I currently better understand the logic in Python than in OCaml ;-) I will have more time in September to spend on it and I am definitely interested in trying to improve it. Can you be a bit more specific on which panels to show/hide and when? Cheers.

gautierhattenberger: I have opened a new issue regarding discussion around the paparazzi center: https://github.com/paparazzi/paparazzi/issues/2093

sw/tools/airframe_editor/paparazzi.py
```diff
@@ -1,81 +1,156 @@
...
+if sys.version_info[0] == 2:
+    PY2 = True
+else:
+    PY2 = False
+
+if PY2 is True:
+    from commands import getstatusoutput
+else:
+    from subprocess import getstatusoutput
+
```
tcunis: The if-structure and the boolean variable here are pretty redundant. Please use

```python
    PY2 = (sys.version_info[0] == 2)
    if PY2:
        #etc.
```
or, if `PY2` is never used again (as it seems to be here) consider directly inserting the condition expression into the last if-else-structure.

sw/tools/airframe_editor/airframe_editor.py
tcunis: I agree with @flixr's argument. On the other hand, the former gives the entire `path` object/structure, when we just need the `join` function of it. As a compromise, what about

```python
from os.path import join as pathjoin
```

gautierhattenberger:
closing for timeout and repository from user have been removed anyway

PR #2094, opened 2017-07-26
Updating gvfFormation.py

noether:
The script now uses the correct library from /var/python (now it can work with pprlink v2.0 for example).

It seems that the 'array msgs' do not need to be split anymore (otherwise, with .split it crashes).

### DIFF
--- a/sw/ground_segment/python/gvf/gvfFormation.py
+++ b/sw/ground_segment/python/gvf/gvfFormation.py
@@ -8,7 +8,7 @@ import numpy as np
 import sys
 from os import path, getenv
 PPRZ_HOME = getenv("PAPARAZZI_HOME", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
-PPRZ_HOME = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
+PPRZ_SRC = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
 sys.path.append(PPRZ_SRC + "/sw/lib/python")
 sys.path.append(PPRZ_HOME + "/var/lib/python")
 from pprzlink.ivy import IvyMessagesInterface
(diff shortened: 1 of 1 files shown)


sw/ground_segment/python/gvf/gvfFormation.py
```diff
@@ -7,9 +7,9 @@
 import numpy as np
 import sys
 from os import path, getenv
-PPRZ_SRC = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
-sys.path.append(PPRZ_SRC + "/sw/lib/python")
-sys.path.append(PPRZ_SRC + "/sw/ext/pprzlink/lib/v1.0/python")
+PPRZ_HOME = getenv("PAPARAZZI_HOME", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
+sys.path.append(PPRZ_HOME + "/sw/lib/python")
+sys.path.append(PPRZ_HOME + "/var/lib/python")
```
gautierhattenberger: Actually, only this one should use PPRZ_HOME.
Things in conf/ and var/ should use PAPARAZZI_HOME, everything else (sw, doc, data, ...) should use PAPARAZZI_SRC. That is for the day we fix the binary debian packaging for paparazzi...

noether: yep, I was wondering that before committing. Now it is clearer :P.

sw/ground_segment/python/gvf/gvfFormation.py
```diff
@@ -7,9 +7,10 @@
 import numpy as np
 import sys
 from os import path, getenv
+PPRZ_HOME = getenv("PAPARAZZI_HOME", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
 PPRZ_SRC = getenv("PAPARAZZI_SRC", path.normpath(path.join(path.dirname(path.abspath(__file__)), '../../../../')))
 sys.path.append(PPRZ_SRC + "/sw/lib/python")
-sys.path.append(PPRZ_SRC + "/sw/ext/pprzlink/lib/v1.0/python")
+sys.path.append(PPRZ_HOME + "/var/lib/python")
```
flixr: should use `path.join(PPRZ_HOME, "/var/lib/python")` instead of string addition

noether: oki. There will be soon a new update. I will include your suggestion by then :P.

PR #2095, opened 2017-07-27
Fix INDI rpy setpoint function

EwoudSmeur:
The function quat_from_rpy_cmd_i does not seem to act as desired.
The quat_int controller uses simple conversion from eulers, which works
intuitively on a test case, so this one is preferred.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -306,9 +306,9 @@ void stabilization_indi_set_failsafe_setpoint(void)
 void stabilization_indi_set_rpy_setpoint_i(struct Int32Eulers *rpy)
 {
   // stab_att_sp_euler.psi still used in ref..
-  memcpy(&stab_att_sp_euler, rpy, sizeof(struct Int32Eulers));
+  stab_att_sp_euler = *rpy;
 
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
Where are the commands coming from ? Flight plan or radio control ? It is fine to change the function as long as the inputs are really euler angles, which is not always the case.
For the `quat_from_rpy_cmd_i`, is it failing at specific angles like close to 90° ?

EwoudSmeur:
In fact it is not often used, but for instance in guidance_opticflow_hover.c. Also one of our students (@tomvand) was using it in a guidance_module thing. When using it there, the drone was actually flying in the wrong direction when this function was called instead of the euler to quat one. 

I am not 100% sure when rpy is 'meant' to be an orientation vector and how that beats Euler angles. But I think more often than not, rpy is just meant as Euler angles...

ETitus:
I encountered similar problems in module modes. Temporarily solving it by making sure I would always flight with yaw 0, so the problem would not occur. I'll check if it solves my problem too :)

ETitus:
I can confirm that solved my problem too. It seems the yaw wasn't taken into account when giving pitch and roll commands. Especially visible at 180 degree angles, as that turns + into - and the other way around.

EwoudSmeur:
If there are no objections I'll go ahead an merge this...

OpenUAS:
Please do, appriciated

PR #2098, opened 2017-08-05
fix SRTM download

flixr:
the SRTM download page now redirects to https
so enable support for TLS

needs the Debian/Ubuntu package libocamlnet-ssl-ocaml-dev

### DIFF
--- a/sw/lib/ocaml/META.pprz.template
+++ b/sw/lib/ocaml/META.pprz.template
@@ -1,5 +1,5 @@
 description = "Paparazzi UAS package"
-requires = "unix,str,pprzlink,xml-light,lablgtk2,glibivy,netclient"
+requires = "unix,str,pprzlink,xml-light,lablgtk2,glibivy,netclient,nettls-gnutls"
 version = "1.0"
 directory = ""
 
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
I guess this is failing if you don't install the extra ssl library. Should we wait a bit that people update their system for the new paparazzi-dev or do we merge this now ?

flixr:
I would say we merge it and announce it on the mailing list.

flixr:
Also this currently only works with ocaml > 4, tls is not loaded for older version (as on trusty).

gautierhattenberger:
tested and working, thanks Felix

kevindehecker:
So, I got an error after a clean make:

<....>
ocamlfind: Package `nettls-gnutls' not found
Makefile:167: recipe for target 'compat.cmi' failed
make[1]: *** [compat.cmi] Error 2
Makefile:148: recipe for target 'libpprz' failed
make: *** [libpprz] Error 2

apt-get installing libocamlnet-ssl-ocaml-dev fixed it.

flixr:
@kevindehecker jep, as the description says....
Also if you would have updated paparazzi-dev to the latest version it would have pulled this dependency in already.... or are you not using that meta package?

kevindehecker:
I pulled master, then I got this error I posted above. I needed to manually apt-get install to fix it. Tried to do a make clean && make first, but no luck.

I thought I'd post the exact error here anyway so you at least trace the problem back here if other people have similar issues.

podhrmic:
@kevindehecker I manually installed the package with `sudo apt install  libocamlnet-ssl-ocaml-dev` and it fixed the problem without reinstalling the dependencies.

PR #2101, opened 2017-08-16
Flight plan functions for autopilot mode guided

knmcguire:
This pull request has a new paparazzi module, which includes some basic functions that can be used from the flight plan directly when the autopilot is in guided mode. This enables the possibility to also use the flight plan when no perfect 3D fix is known (like GPS or OptiTrack), since most are focused position waypoints. Since indoor autonomous flight is becoming more common application, this module can be very helpful as it turns the flight plan into a state machine, which can reactively respond to its environment. I have been using this principle for my flights for a while now so therefore I want to merge it into paparazzi master branch.

This pull request includes:
-  flight_plan_guided module
- example airframe with the module included (for gazebo simulator)
-  example flight plan which uses the module (I added it to the conf_example.xml, so you can try it out yourselves)
-  adjusted the color filter to send an ABI message if an color (obstacle) is detected. It serves as an example how this detection can be used in the flight plan.

I hope that you also find this module useful and please comments if there are some issues or if it has to be adjusted for more general use within paparazzi


sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,427 @@
...
+ * @param[in] fail_altitude, wait for altitude in [m] (positive z-axis is up)
+ * @param[in] speed, speed in [m/s] (positive z-axis is up)
+ * @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+bool WaitUntilSpeedOrAltitude(float speed, float fail_altitude)
+{
+  if (autopilot.mode != AP_MODE_GUIDED) { return true; }
+
+  if (stateGetPositionEnu_f()->z > -fail_altitude) { return false; }
+  if (stateGetSpeedEnu_f()->z < -speed) { return true; }
```
kirkscheper: Same strange negation here. Also why fail_altitude and not just altitude like you have speed.

if (stateGetPositionEnu_f()->z < altitude) { return true; }
if (stateGetSpeedEnu_f()->z < speed) { return true; }

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,427 @@
...
+
+/********************************************//**
+ *  Timer functions
+ ***********************************************/
+
+/**
+ * ResetSpecialTimer: Reset a special timer, that works outside of the block and stage time
+ * @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+float specialtimer = 0;
```
kirkscheper: specialtimer is a bit special name, maybe nav_timer? 


As it is currently implemented, this timer does not run continuously but only when waitUintilTimer or  WaitUntilTimerOrAltitude is called. If you areally want a timer that is always running you should add it to common_flight_plan.h and navigation.c but only reset it with your ResetNavTimer.

conf/airframes/examples/ardrone2_FPguided_gazebo.xml
```diff
@@ -0,0 +1,212 @@
...
+    <module name="actuators" type="ardrone2"/>
+    <module name="imu" type="ardrone2"/>
+    <!-- gps: "ublox" or change to "sirf" for usage with parrot flight recorder -->
+    <module name="gps" type="ublox"/>
+    <module name="stabilization" type="int_quat"/>
+    <module name="ahrs" type="int_cmpl_quat"/>
+    <module name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
```
gautierhattenberger: Can you the modules from this `modules` node inside the `firmware` node ? I recommend to stop using it as it is planed to be removed some day.

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,427 @@
...
+ * WaitUntilTimerOrAltitude: Blocks while waiting until the timer reaches the specified amount of seconds or an altitude has been reached
+ * @param[in] sec, wait for amount of seconds [sec]
+ * @param[in] fail_altitude, wait for altitude in [m] (positive z-axis is up)
+ * @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+bool WaitUntilTimerOrAltitude(float sec, float fail_altitude)
+{
+  if (autopilot.mode != AP_MODE_GUIDED) { return true; }
+
+  if (stateGetPositionEnu_f()->z > fail_altitude) { return false; }
```
kirkscheper: if (stateGetPositionEnu_f()->z < altitude) { return true; }

conf/flight_plans/rotorcraft_guided_flightplan.xml
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
+
+<flight_plan alt="1.0" ground_alt="0" lat0="51.990634" lon0="4.376789" max_dist_from_home="20" name="Rotorcraft Optitrack (Delft)" security_height="0.3">
+  <header>
+	  #include "autopilot.h"
+    #include "guidance/guidance_h.h"
+    #include "guidance/guidance_v.h"
+    #include "modules/flight_plan_guided/flight_plan_guided.h"
```
gautierhattenberger: At least this one shouldn't be needed as modules header are automatically available for flight plans. For the other ones, if they are required by your new module, they should be included in `flight_plan_guided.h`.
As a side note, you can include modules from the flight plan itself. I means that you don't need a customized airframe that is matching a particular flight plan.

conf/modules/flight_plan_guided.xml
```diff
@@ -0,0 +1,30 @@
...
+
+<module name="GUIDED_FlightPlan" dir="flight_plan_guided">
+  <doc>
+    <description>GUIDED_FlightPlan</description>
+  </doc>
+
+  <settings>
+    <dl_settings>
+      <dl_settings NAME="guided_flight_plan">
+        <dl_setting var="nom_flight_alt" min="0.2" step="0.1" max="3." module="flight_plan_guided/flight_plan_guided" shortname="nom_flight_alt" param="NOM_FLIGHT_ALT"/>
```
gautierhattenberger: I guess `nom` stands for `nominal` ? I suggest using the full name to make it clear. If it is too long for the GCS settings tab, use the `shortname` attribute.

conf/modules/flight_plan_guided.xml
```diff
@@ -0,0 +1,30 @@
...
+    </dl_settings>
+  </settings>
+
+  <header>
+    <file name="flight_plan_guided.h"/>
+  </header>
+
+  <init fun="flight_plan_guided_init()"/>
+
+  <makefile target="ap">
```
gautierhattenberger: you can use `target="ap|nps"` here

sw/airborne/modules/computer_vision/colorfilter.c
```diff
@@ -58,6 +68,17 @@ struct image_t *colorfilter_func(struct image_t *img)
                                        color_cr_min, color_cr_max
                                       );
 
+
+  if (COLORFILTER_SEND_OBSTACLE) {
+    if (color_count > 20)
+      AbiSendMsgOBSTACLE_DETECTION(ABI_BROADCAST, 1,
```
gautierhattenberger: You are not supposed to send a message with  `ABI_BROADCAST` id, this is only for receiving. You should add the ID of your module in `subsystems/abi_sender_ids.h` (choose a "category" matching your module or start a new one) and use it here.

sw/airborne/modules/computer_vision/colorfilter.c
```diff
@@ -58,6 +68,17 @@ struct image_t *colorfilter_func(struct image_t *img)
...
+                                   0);
+    else
+      AbiSendMsgOBSTACLE_DETECTION(ABI_BROADCAST, 10,
```
gautierhattenberger: same here

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,427 @@
...
+ * @author K. N. McGuire 2017
+ *
+ */
+
+#include "modules/flight_plan_guided/flight_plan_guided.h"
+#include "subsystems/ins.h"
+#include "firmwares/rotorcraft/navigation.h"
+#include "firmwares/rotorcraft/guidance/guidance_h.h"
+#include "firmwares/rotorcraft/guidance/guidance_v.h"
+#include "modules/sonar/sonar_bebop.h"
```
gautierhattenberger: This board dependent you should use that directly. Either use the state interface or the ABI messages.

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,427 @@
...
+ * ChangeVerticalMode: Change vertical mode in autopilot guided mode
+ * @param[in] mode, number of the specific horizontal mode that you want to change to
+ * @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+bool ChangeVerticalMode(uint8_t mode)
+{
+  guidance_v_mode_changed(mode);
+  return false;
+}
+
```
gautierhattenberger: I agree you shouldn't use mode change function from guidance layer directly

knmcguire:
@gautierhattenberger Could you specify the comment about mixing with image processing, why it is not a good idea? 

Right now, it only receives a ABI message from the color filter module with a possible obstacle on the path, since I do want to keep the image processing itself separate, however I do want to use the knowledge from it, so the MAV react on it in real-time. Besides, the information does not have to come from an image processing module but can also come from a range sensor for instance.  It is the major reason why I use the  flightplan_guided module in the first place, because I wanted a module where it all comes together (external sensor information about the environment -> behavior).

kirkscheper:
You only use distance_obstacle in the flight plan so don't need it in the flight_plan_guided. You can get the value of the sensor in the flightplan directly. Just write that value in the range sensor to a parameter defined as extern in the header and include the header in the flight plan. That way you don't mix up functionality.

gautierhattenberger:
As @kirkscheper pointed out, if you mix functionalities (providing guided mode control in FP, avoiding obstacle based on vision), you (or someone else) won't be able to reuse your code easily for an other purpose.

knmcguire:
Alright, I see your points. 

I did wanted to avoid using external variables, but since the flightplan can not read specified abi messages (yet), I will use them for this occasion.

Thanks for your feedback. I will go through all your comments and update the pull request in the coming days

knmcguire:
Actually, the ability to read ABI messages from the flight plan sounds really nice. That really keeps all the modules separate, but the flight plan can be a sort of state machine where it all comes together. Maybe for a future pull request?

gautierhattenberger:
This is indeed useful but not so easy and will need some work in the flight plan generator to create automatic bindings. As an example, here is a module I made for reading sonar data from flight plans: http://docs.paparazziuav.org/latest/module__agl_dist.html

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
knmcguire: Actually, this suggestion will cause the function to constantly return true if it has not passed the altitude (so the function will keep on running), which means it will not hit the timer check in the mean time. That is not what I want with this function

knmcguire:
Alright, I tried to implement all your feedback and have added some improvements to the code as well.

@gautierhattenberger  I managed to read the abi message completely from the flight plan now (check commit 7d0d9bb). It is an bit unconventional way I supposed and I do get "unused variables" warning, but it works :) Could you take a look at it and see if this is okay, or that there is an better way to do this?

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+// start and stop modules
+#include "generated/modules.h"
+
+float nominal_alt; // nominal flight altitude
+
+//TODO: Find equivelant function in math (couldn't find it yet...)
+
+/* Wraps an angle in radians between -PI and +PI */
+float wrapToPi_guided(float ang)
```
flixr: This should be removed or rather a float_course_normalize function added to math/pprz_algebra_float.h and maybe a macro analogous to INT32_COURSE_NORMALIZE.

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+
+/********************************************//**
+ *  Start functions for guided flight plan
+ ***********************************************/
+
+/**
+ * KillEngines: Kill engines from flight plan directly
+ * @param[return] Falseboolean for flight plan to only run once
+ */
+uint8_t KillEngines(void)
```
flixr: As @gautierhattenberger already pointed out, wrapper functions like this should be removed and the actual functions called directly using `call_once` instead of `call`...

kirkscheper:
@flixr "The flight plan is only meant to be executed if you are in NAV mode, but now you want to change the mode to GUIDED but still follow the commands from the flight plan..." 

Exactly the functionality we want. We would like to be able to follow a fixed set of autonomous commands (state machine) without using waypoints or having a position estimate. Since the flight plan is run regardless of the mode the vehicle is in it seems like a good solution for us. If you have a better or different solution please let us know. 

Of course when the flight plan was designed, probably there was no plan of flying without waypoints or position estimate but our goals have evolved and the code has to evolve with it. I do however emphatically agree that not useful wrappers are useful and we should try to reduce that as much as possible.

I guess an alternative would be flying in module mode and implementing the state machine there but seems redundant when the flight plan is already running and has similar functionality and would otherwise not be used (since we are not in NAV).

Like @knmcguire said we have been using the flight plan in some form or another to execute autonomous flight in gps denied environments since IMAV last year and only expect to expand that capability so it is important to decide on the framework for that development soon. One of the main reasons we didn't try merging this earlier is that @dewagter was not a fan of the implementation, however over the past few months there has been no work done in this direction so we are trying to emphasise a need for this type of functionality in the future. 

So again, any insight or ideas to (robustly, elegantly and "maintainably") achieve this functionality are welcome.

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/**
+ * SpeedRotateToHeading: Will slowly turn to a given heading with a fixed speed
+ * @param[in] rotation_rate, speed to slowly rotate the heading(clockwise is positive)
+ * @param[in] heading, here the MAV will stop spinning [rad] (clockwise is positive)
+ *  @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+bool SpeedRotateToHeading(float rotation_rate, float heading)
+{
```
kirkscheper: This is also a blocking function, move it below?

conf/flight_plans/rotorcraft_guided_flightplan.xml
```diff
@@ -0,0 +1,105 @@
...
+    </block>
+
+    <block name="Start Engine">
+      <call fun="StartEngines()"/>
+      <while cond="1"/>
+    </block>
+    
+    <block name="Take off">
+      <call fun="Climb(0.2)"/>
+      <call fun="WaitUntilAltitude(nominal_alt)"/>
```
kirkscheper: I still don't get why you use this, how often do you need to change the take off altitude without reflashing the controller?

conf/airframes/examples/ardrone2_FPguided_gazebo.xml
```diff
@@ -0,0 +1,210 @@
...
+      <define name="COLORFILTER_SEND_OBSTACLE" value="TRUE"/>
+    </module>
+
+    <module name="video_rtp_stream">
+      <define name="VIEWVIDEO_CAMERA" value="front_camera"/>
+      <define name="VIEWVIDEO_DOWNSIZE_FACTOR" value="4"/>
+      <define name="VIEWVIDEO_QUALITY_FACTOR" value="60"/>
+    </module>
+
+    <module name="flight_plan_guided"/>
```
gautierhattenberger: I think this should be loaded from the flight plan, so people who want to use it won't forget to add it.

conf/flight_plans/rotorcraft_guided_flightplan.xml
```diff
@@ -0,0 +1,105 @@
...
+PRINT_CONFIG_VAR(FP_GUIDED_OBSTACLE_DETECTION_ID)
+#endif
+
+static abi_event obstacle_detection_ev;
+static void obstacle_detection_cb(uint8_t sender_id, float distance_obstacle, float heading_obstacle);
+float distance_obstacle_FP;
+static void obstacle_detection_cb(uint8_t UNUSED(sender_id), float distance_obstacle, float UNUSED(heading_obstacle))
+{
+  distance_obstacle_FP = distance_obstacle;
+}
```
gautierhattenberger: By doing it like this, it is creating static variables everywhere the flight_plan.h header is included. It may work, but I don't think it is what we want. You could at least do something like this:
```
#ifdef NAV_C
abi_event obstacle_detection_ev;
float distance_obstacle_FP = 0.f;
void obstacle_detection_cb(uint8_t UNUSED(sender_id), float distance_obstacle, float UNUSED(heading_obstacle))
{
  distance_obstacle_FP = distance_obstacle;
}
#else
extern abi_event obstacle_detection_ev;
extern float distance_obstacle_FP;
extern void obstacle_detection_cb(uint8_t sender_id, float distance_obstacle, float heading_obstacle);
#endif
```
I guess the proper solution is to generate this code automatically from a new dedication xml description in the flight plan. I'll think about it...

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * WaitUntilAltitude: Blocks while waiting until a certain altitude is reached
+ * @param[in] altitude in [m] (positive z-axis is up)
+ * @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ */
+bool WaitUntilGround(void)
+{
+  if (autopilot.mode != AP_MODE_GUIDED) { return true; }
+  if (stateGetPositionEnu_f()->z > 0.1) { return true; }
```
gautierhattenberger: And what if ground is not the `z=0` plane ? Why not have a generic `WaitUntilAlt` that looks for a range [alt-epsilon, a+epsilon]. Or if you really need, and extra option to tell if you wait until below, over or equal to a given altitude.

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * WaitforHeading: Blocks untill wanted heading is achieved
+ * @param[in] heading, waits untill MAV reached certain heading [rad] (clockwise is positive)
+ *  @param[return] Boolean for flight plan to keep running (true) or to just run once (false)
+ *
+ * TODO: Wrap the angle around -pi and pi
+ */
+bool WaitforHeading(float heading)
+{
+  if (fabs(wrapToPi_guided(heading) - stateGetNedToBodyEulers_f()->psi) < 0.1) {
```
gautierhattenberger: This test could be implement in the `NavCourseCloseTo` that is currently a void macro in `firmwares/rotorcraft/navigation.h`

sw/tools/generators/gen_flight_plan.ml
```diff
@@ -940,6 +940,7 @@ let () =
       check_altitude (float_of_string alt) xml;
       check_altitude_srtm (float_of_string alt) xml !fp_wgs84;
 
+      Xml2h.define "NOMINAL_ALT" (sprintf "%.0f /* nominal altitude of the flight plan */" (float_of_string alt));
```
gautierhattenberger: I would prefer `NAV_ALT` or `NAV_DEFAULT_ALT` for consistency with other defines

gautierhattenberger:
The guided mode was made for off-board control at the guidance level while skipping the flight plan logic. But the control loops in NAV and GUIDED should be the same. So if you need to mix the two, it means that the NAV API is not flexible enough to do it alone. @knmcguire, did you identify the missing control or blocking point for using pure NAV in indoor environment ? Is it the "speed" control ? The fact that GPS is probably reporting NO_FIX ? Something else ?

sw/airborne/modules/flight_plan_guided/flight_plan_guided.c
```diff
@@ -0,0 +1,352 @@
...
+  return true;
+}
+
+
+/********************************************//**
+ *  Timer functions
+ ***********************************************/
+
+/**
+ * ResetSpecialTimer: Reset a special timer, that works outside of the block and stage time
```
knmcguire: I didn't change it actually, that is what I do want, that it only runs if I tell it to. But if you do not see the immediate use for it, I will change it according to the suggestion

kirkscheper: If you want it like this just change the comment to not say it "works outside of the block and stage time".

conf/airframes/examples/ardrone2_FPguided_gazebo.xml
knmcguire: Ah I see. Is it possible to load modules in the flightplan directly? I didn't know this was possible. Do I add it within <header> ?

conf/flight_plans/rotorcraft_guided_flightplan.xml
knmcguire: Ah yes, this is an better option. I did get many unused variable warnings so this will fix the problem. I do agree about that the proper solution is to automatically generate it, so I hope that will be possible

conf/airframes/examples/ardrone2_FPguided_gazebo.xml
gautierhattenberger: see https://github.com/paparazzi/paparazzi/blob/master/conf/flight_plans/demo_gvf.xml#L35 for example

knmcguire:
@gautierhattenberger what I found lacking to the current NAV_mode was that it is too centered on waypoint navigation with an global coordinate system. For indoor navigation I assume no localization system is available, therefore I prefer more "egocentric" way of navigation, like: move 1 meter forward from the current position in body fixed coordinates, instead of moving waypoint. I think that is a lot more intuitive for the task.  Right now all the functions existing to do this, is all only available for guided mode.

About the speed control, that is not necessarily the purpose, but since we can get velocity from a bottom camera, it is the type of control I use the most. But if we integrate it for odometry, position control can be used as well. I just want the MAV to control his state relative from his current position, in its body fixed coordinates.

gautierhattenberger:
I looked again at the GUIDED mode API, and it seems that for most (all ?) of your wrapper functions to move the MAV, you could use directly the body-relative functions from https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/autopilot_guided.h and in worst case the very generic `autopilot_guided_update`.

I also look at the real differences between NAV and GUIDED at the guidance level, at they are more or less the same except that the guidance setpoint is coming from the `navigation_carrot` for NAV while in GUIDED it is set separately by calling the proper functions.
It means that the kind of wrapper that would be useful to fly in NAV in your case is something that set the `navigation_carrot`, `nav_heading` and `nav_altitude` (that are global) from body-frame setpoints. It is actually done this way by GUIDED like here : https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/autopilot_guided.c#L58

As a side, the `NO_GPS_NEEDED_FOR_NAV` might help as well.

flixr:
While technically you can call any function you like in the flight plan, IMHO you have to be careful which ones you use.
In nearly all modes the flight plan is executed "alongside" so that it will e.g. transition to geo init, standby, etc. while you are flying in "manual" and sort of do the right thing if you switch to NAV later.
While most things that are currently called from the flight plan won't directly call control loops and instead just set some variables (like @gautierhattenberger already pointed out), it probably won't work as intended if they actually set commands directly in the end...

While you have "protected" your wrapper functions to only call the functions if you are in GUIDED mode, this somewhat goes against the (initial) intention.
Also meaning that if you decide to use guided functions in the flight plan, you can't (or rather shouldn't) send any guided mode commands via telemetry (or a companion computer) anymore. Because if you do that you will suddenly have two conflicting sets of instructions....

So from a higher level perspective it seems to me that we should probably be rather looking at ways to achieve your goals (relative commands, etc. from the flight plan) without using guided mode.
Also as already pointed out, it is not necessary to have a global position to use the flight plan (although this can certainly be improved).

knmcguire:
Thank you  for comments. 

I did not know that the autopilot_guided file existed at the moment when I made the pull request. It seems like those functions look very similar to the functions at the end of guidance_h.c and guidance_v.c, which is also a bit conflicting... I definitely do not want to make a third of this kind to paparazzi because that will only create more confusion.

Tomorrow I will try to make a flight plan with functions you propose and see if it can work out for my purposes. Then I will decide if I will keep this pull request or make a new one.  In the mean time I am still very interested in trying to find a way to read abi messages from the flight plan, since that is an important aspect that I need for my flights.

gautierhattenberger:
@knmcguire see https://github.com/paparazzi/paparazzi/pull/2106
It should compile, but I didn't really properly tested it.

knmcguire:
ah thanks! I will test it out and see if it works
________________________________

knmcguire:
I have now removed the flight_plan_guided module and tried to build up the flight plan with existing functions only.

 I tried to do this with the existing autopilot_guided functions, however failed to achieve the wanted functionality... I will need to add a lot more flags to the autopilot_guided_update function in order to achieve those combinations I want (for instance x, y in body velocity and z only in NED position). The separate guidance functions provide more flexibility for now. If you have concerns in terms of safety, I could try again but that will  have to be another time.

gautierhattenberger:
Are you sure that the update function is not doing the job ? For your example, if you set bits 1, 3 and 5 (or 7) to 1, you should have vx and vy in body frame, z as NED altitude (and yaw as angle offset (or rate), so 0 if you don't want to change).
Actually it is an exact body frame, but more a decoupling horizontal/vertical, so z is always in NED and the rotation to body frame only consider the yaw angle, not roll and pitch.

knmcguire:
I tried did try the update function, but setting the bits directly from the flight_plan does not seem to work... Very likely I am doing something wrong there, because I do not correctly understand the bitsetting procedure

But then again,  to have       <call_once fun="SetBit(flags,0)"/> for every single functionality does not sound like an elegant solution either.

gautierhattenberger:
That's not quite the correct way to set the flag. For instance, in my example bits 1, 3, 5 should be set, so the first parameter for the update function is: `0b00101010 = 0x2A = 42` (and it is not even joke... ).
Actually we should add some defines so you could pass `GUIDED_FLAG_XY_REL | GUIDED_FLAG_YAW_OFFSET | ...`.

hooperfly:
@gautierhattenberger thanks for the adding the `defines` and thanks for the Douglas Adams reference. :)

knmcguire:
ah yes you are totally right, I misunderstood its purpose. I thought I had to make a new number per situation but this makes more sense ( 42 of course   hahaha ) . The function autopilot_guided_update is working for me now. However, autopilot_guided.h does need to be included to the headers of the flight plan apparently, is that what we want as well?

Yes, adding the defines is a good idea. So if I understand correctly, if I want the combination I want, I can do an bit OR operation on the defines FLAG_XY_BODY | FLAG_YAW_OFFSET | FLAG_XY_VEL and use that as the first argument of autpilot_guided_update. I can work with that.

So about another thing, I am wondering what to do with this pull request, because I reverted quite a lot of my changes already and the 18 commits is a bit too much.  I could close this and make another one with just a flight_plan, using only the autopilot_guided functions and defines, after pull request #2106 has been accepted?

conf/flight_plans/rotorcraft_guided_flightplan.xml
```diff
@@ -0,0 +1,120 @@
+<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
+
+<flight_plan alt="1.0" ground_alt="0" lat0="51.990634" lon0="4.376789" max_dist_from_home="20" name="Rotorcraft Optitrack (Delft)" security_height="0.3">
+  <header>
+
+#include "autopilot.h"
```
gautierhattenberger: This include shouldn't be needed anymore, I made it available by default as it is often used.
Could you also fix the indentation of the rest of the file ?

sw/airborne/firmwares/rotorcraft/autopilot_guided.h
```diff
@@ -96,9 +96,17 @@ extern void autopilot_guided_update(uint8_t flags, float x, float y, float z, fl
 #define GUIDED_FLAG_Z_OFFSET    (1<<2)
 #define GUIDED_FLAG_YAW_OFFSET  (1<<3)
 #define GUIDED_FLAG_XY_VEL      (1<<5)
-#define GUIDED_FLAG_2_VEL       (1<<6)
+#define GUIDED_FLAG_Z_VEL       (1<<6)
```
gautierhattenberger: thank you, those two are really too close on french keyboards...

knmcguire:
Latest update where I used only the autopilot_guided_update() functions to achieve the "state machine" I want. I implemented the latest commit with the abi_messages too and tested it out, and it seems to all work. If nothing is in the way (or no abi message is send), The MAV is supposed to complete the square but in this case it turns a bit more
screenshot from 2017-08-23 16-10-22

I first wanted to add the combined flags in the flight plan as a <variable>. However, I get an error if I add one and do not want to give it setting values... bit odd. I now added it to the autopilot_guided.h, since I figured somebody might want to use it as well.

sw/airborne/firmwares/rotorcraft/autopilot_guided.h
```diff
@@ -96,9 +96,17 @@ extern void autopilot_guided_update(uint8_t flags, float x, float y, float z, fl
...
 #define GUIDED_FLAG_YAW_RATE    (1<<7)
 
 
+// Useful Combination of the flags fir the autopilot_guided_update
```
gautierhattenberger: I'm not so found of these defines, as everybody might want its own combination in the end. And even if it is shorter, it is not always easier to read.
If you want them for your flight plan, you can have these lines in the `header` section.

Issue #172, opened 2012-03-30
improve cam handling


flixr:
The cam_point module
- sends the CAM message, with camera orientation (roll,pitch) and target xy coordinates.
- server calculates Lat/long from the CAM message and sends CAM_STATUS ground message
- you can enable the cam footprint in the GCS (reads CAM_STATUS)

The difference of using SHOW_CAM_COORDINATES and the CAM_POINT message is that you already get the coordinates in lat/lon via telemetry, but apparently can't show the cam footprint in the gcs...

Currently missing:
- the CAM message is only read for fixedwings (server_fw.ml)
- only handes cameras that can move in roll/pich axes, but not ones that can move around yaw (look left/right).
- cam_aperture (also aspect ratio) is hardcoded in live.ml... this should really be settable via option.
##

rijeshaugustine:
I've been looking into this.

I have one other quam with the current implementation.
At ground level the cam footprint still shows a large square footprint, this seems to indicate that height is not being taken into account. 

I reworked the module so that it would read a Horizontal angle of view and a vertical angle of view from the airframe file. 

I was unclear as to the purpose of all the processing that was happening in server.ml. I fly without a gimbal so only have to worry about the angles of the airplane.
Currently I have the CAM_STATUS message being used to trigger the cam update but don't actually read the contents of the message. 
Finally I used a polygon shape rather than a rectangle to plot the footprint. 
https://github.com/rijeshaugustine/paparazzi/commit/159134aa9bb998a92cbc3525eaa4d76e200f29e1

This commit removes the functionality of camera roll/pitch. If camera roll/pitch were sent in CAM_STATUS (rather than lat and long) this could easily be incorporated into the current math.
I have simplified the code significantly. Am I missing something by doing so? Essentially I am using the field of view and angles of the airplane to find the centre point of each line segment of the footprint and then getting the 4 corners. 

I've also had a look at getting camera yaw incorporated. I was thinking there could be a quaternion for the camera angles and a quaternion for the plane angles. Multiply them together and convert back to angles. I'm not sure if this will work though due to the order the rotations happen. I would imagine that both yaw rotations would have to be dealt with first and then pitch and roll can be dealt with.

rijesha:
This issue is almost solved, just need a little bit more information.
The cam message only being read by fixedwings has been fixed already.
Pull Request #2103 has the relevant GCS code for cam_aperture and a fully gimbaled camera.

[...]

PR #2103, opened 2017-08-16
fix hardcoded cam footprint field of view and AGL

rijesha:
The cam footprint angle of view can now be adjusted in the airframe.xml file. The old cam footprint math used height in MSL not AGL. In addition the footprint did not adjust properly when the aircraft pitched or rolled.


rijesha:
I mention of few of these fixes in a comment for issue #172

sw/ground_segment/cockpit/live.ml
```diff
@@ -326,19 +333,6 @@ let mark = fun (geomap:G.widget) ac_id track plugin_frame ->
             let png = sprintf "%s/var/logs/%s" Env.paparazzi_home name in
             GdkPixbuf.save png "png" dest;
             incr i;
-
-      (* Computing the footprint: front_left and back_right *)
```
gautierhattenberger: this part is actually not related to the camera footprint but to the "mark" button and screenshot tool. See http://wiki.paparazziuav.org/wiki/GCS#Actions
please add back this code.

sw/ground_segment/cockpit/live.ml
```diff
@@ -1521,6 +1532,45 @@ let get_shapes = fun (geomap:G.widget)_sender vs ->
 let listen_shapes = fun (geomap:G.widget) ->
   safe_bind "SHAPE" (get_shapes geomap)
 
+let get_cam_status = fun (geomap:MapCanvas.widget) _sender vs ->
```
gautierhattenberger: The point the camera is pointing at (taking into account the bank angles and the ground alt) is already computed in the server program: https://github.com/paparazzi/paparazzi/blob/master/sw/ground_segment/tmtc/server.ml#L165
Your current position and this cam point should be enough to compute your polygon without extra information (maybe AGL).
Also be careful that at large bank angle, the projection of the footprint may send some corner very "far". I don't know how GTK canvas are dealing with this.

sw/ground_segment/cockpit/live.mli
```diff
@@ -54,6 +55,12 @@ type aircraft = private {
     mutable in_kill_mode : bool;
     mutable speed : float;
     mutable alt : float;
+    mutable agl : float;
```
gautierhattenberger: I suspect that in fact only `agl` is needed for your needs

sw/lib/ocaml/mapTrack.ml
```diff
@@ -276,50 +270,16 @@ object (self)
     zone#set [`X1 x1; `Y1 y1; `X2 x2; `Y2 y2; `OUTLINE_COLOR "#ffc0c0"; `WIDTH_PIXELS 2]
 
     (** moves the rectangle representing the field covered by the camera *)
-  method move_cam = fun cam_wgs84 mission_target_wgs84 ->
+  method move_cam = fun positions mission_target_wgs84 ->
     match last, cam_on with
         Some last_ac, true ->
-          let (cam_xw, cam_yw) = geomap#world_of cam_wgs84
```
gautierhattenberger: the math that you have in your `get_cam_status` function could go here as well

sw/lib/ocaml/mapTrack.ml
```diff
@@ -276,50 +270,16 @@ object (self)
...
-          let cam_field_half_width = ( tan (cam_half_aperture) ) *. oblic_distance in
-          ac_cam_cover#set [`X1 (-. cam_field_half_width);
-                            `Y1 (-. cam_field_half_height_1);
-                            `X2 (cam_field_half_width);
-                            `Y2(cam_field_half_height_2);
+          let points = geomap#convert_positions_to_points positions in
+          ac_cam_cover#set [`POINTS points;
                             `OUTLINE_COLOR color];
-          cam#affine_absolute (affine_pos_and_angle 1.0 cam_xw cam_yw cam_heading);
+          (*cam#affine_absolute points; *)
```
gautierhattenberger: remove if not needed anymore

sw/ground_segment/cockpit/live.ml
rijesha: Isn't there some ambiguity if the camera is on a gimbal? It is unknown if the camera has any rotation in the direction it is pointing. Would it be better to move all the calculations into the server program and modify the cam status message to transmit the 4 corners of footprint. My current code also does not take into account any gimbal effects that the original may have, because the gimbal information is only being read on the server.

rijesha: If a point can't be plotted, the GCS will stop updating the footprint and retain the last plotted footprint. Once the aircraft has leveled off and a footprint can be plotted, then the footprint continues to update. I have performed multiple flights with this code without any GCS error messages or crashing.

sw/ground_segment/cockpit/live.ml
rijesha: oops my bad.

sw/ground_segment/cockpit/live.ml
rijesha: Even without a gimbal just a camera point is  not enough. An aircraft traveling north at a 30deg pitch, will point a camera is the same direction as an aircraft traveling east with a 30deg roll.

gautierhattenberger: The gimbal information (pan and tilt angle) are currently taken from the telemetry CAM message (if sent by the drone, assuming 0 by default). With this, the `server` agent is computing a target point. Isn't this enough with altitude above ground to compute the footprint ? I don't really like the idea of sending the coordinates of the corners of the footprint. I prefer either a target point or absolute angles.

rijesha: The Horizontal field of view and the vertical field of view of the camera may not be the same. We need one program to read the fields from airframe.xml. If the camera is pointed straight down the target point will be directly under the aircraft no matter what angle the camera is panned at. Due to pan angle the footprint may be a wide rectangle, a skinny rectangle, or a diamond shape. This test case can then be applied to a variety of tilt angle. Am I missing something with the target point? The pieces of information we need to fully calculate a footprint are ac.pitch, ac.roll, ac.yaw, ac.agl, ac.pos, cam.pan, cam.tilt, cam.hfv, and cam.vhf. IMHO the server should send cam.pan and cam.tilt messages to the GCS, GCS will then compute the footprint (having all the information), GCS will then send the the information to mapTrack.

rijesha:
Im going to close this pull request and make the suggested changes and open it again

gautierhattenberger:
@rijesha sorry I couldn't help much, I'm really busy these days

rijesha:
@gautierhattenberger I have moved all of the code to the server e3abd262c6c09277d0a561bba9e3605655abd99c. I have also used quaternions to handle a gimbal. Commit a6126a0f685be8d66014ef5abdab8da8f3710abe is how the cam footprint code would look like it it were in the GCS. I decided to put it in the server so that the cam footprint information is broadcasted out for other programs to easily use. I still need to clean up some of the code, but I figured I should get your opinion on whether the code is better in the server or the gcs before proceeding.

sw/ground_segment/tmtc/server.ml
```diff
@@ -167,25 +168,58 @@ let send_cam_status = fun a ->
...
-                        "cam_long", PprzLink.Float ((Rad>>Deg)wgs84.posn_long);
+                        "cam_lat_tr", PprzLink.Float ((Rad>>Deg)geo_1.posn_lat);
+                        "cam_lon_tr", PprzLink.Float ((Rad>>Deg)geo_1.posn_long);
+                        "cam_lat_tl", PprzLink.Float ((Rad>>Deg)geo_2.posn_lat);
+                        "cam_lon_tl", PprzLink.Float ((Rad>>Deg)geo_2.posn_long);
+                        "cam_lat_bl", PprzLink.Float ((Rad>>Deg)geo_3.posn_lat);
+                        "cam_lon_bl", PprzLink.Float ((Rad>>Deg)geo_3.posn_long);
+                        "cam_lat_br", PprzLink.Float ((Rad>>Deg)geo_4.posn_lat);
+                        "cam_lon_br", PprzLink.Float ((Rad>>Deg)geo_4.posn_long);
                         "cam_target_lat", PprzLink.Float ((Rad>>Deg)twgs84.posn_lat);
```
rijesha: @gautierhattenberger  Is there a better way of doing this? My OCaml is not good enough to figure out how to make arrays.

sw/ground_segment/tmtc/server.ml
```diff
@@ -167,25 +168,58 @@ let send_cam_status = fun a ->
...
+  
+          let gimRot = quaternion_from_angle a.cam.phi a.cam.theta 0.0
+          and acRot = quaternion_from_angle a.roll a.pitch (-.a.heading) in
+  
+          let tr_rotated = multiply_quaternion acRot (multiply_quaternion gimRot tr)
+          and tl_rotated = multiply_quaternion acRot (multiply_quaternion gimRot tl) 
+          and br_rotated = multiply_quaternion acRot (multiply_quaternion gimRot br) 
+          and bl_rotated = multiply_quaternion acRot (multiply_quaternion gimRot bl) in
+          
+          let absF (f:float) = if f > 0.0 then f else (f *. -1.0) in
```
gautierhattenberger: it already exists with `abs_float`

sw/ground_segment/tmtc/server.ml
gautierhattenberger: you can send comma-separated values like `DL_VALUES` (https://github.com/paparazzi/paparazzi/blob/master/sw/ground_segment/tmtc/server.ml#L204 and https://github.com/paparazzi/paparazzi/blob/master/sw/ground_segment/cockpit/live.ml#L956)

sw/ground_segment/cockpit/live.ml
```diff
@@ -406,6 +414,16 @@ let get_speech_name = fun af_xml def_name ->
     fvalue "SPEECH_NAME" default_speech_name
   with _ -> default_speech_name
 
+let get_cam_aov = fun af_xml ->
```
gautierhattenberger: Do you really need the cam field of view in the GCS since you are doing the computation in server agent ? It doesn't seemed to be used here anymore. Some question for roll, pitch, etc

rijesha: Thanks for the review! I will clean up all of my code and removed any extra variables/functions I no longer need.

sw/ground_segment/cockpit/live.ml
```diff
@@ -1260,10 +1260,20 @@ let listen_flight_params = fun geomap auto_center_new_ac alert alt_graph ->
...
-    ac.track#move_cam cam_wgs84 target_wgs84
+    let lats_str = PprzLink.string_assoc "lats" vs in
+    let longs_str = PprzLink.string_assoc "longs" vs in
+    
+    let lats = List.map float_of_string (Str.split list_separator lats_str) in
+    let longs = List.map float_of_string (Str.split list_separator longs_str) in
+
+    let target_wgs84 = { posn_lat = (Deg>>Rad)(a "cam_target_lat"); posn_long = (Deg>>Rad)(a "cam_target_long") } in
+    
+    let geo_1 = { posn_lat = (Deg>>Rad)(List.nth lats 0); posn_long = (Deg>>Rad)(List.nth longs 0) }
```
rijesha: @gautierhattenberger  In the example you gave me for DL_Settings Arrays were used. I don't think there should be too much performance degradation keeping it in lists. Please correct me if I am wrong. Do i actually need to put exception handling in for List.nth? Will safe_bind catching it and notifying the user be sufficient?

gautierhattenberger: Considering the sizes involved, list or arrays should not make a big difference.
The `safe_bind` should catch exceptions and print them.

rijesha:
Ok. I think this is now ready to be pushed. Pprzlink PR https://github.com/paparazzi/pprzlink/pull/59

gautierhattenberger:
please update for the correct PPRZLINK version

gautierhattenberger:
you should also rebase on top of master to solve the conflicts, otherwise I can't merge it

rijesha:
Oops I didn't rebase. I just merged master in. It should still work though?

PR #2105, opened 2017-08-21
Air to air gps

gautierhattenberger:
This is used by the traffic info module to have position of surrounding aircraft at higher rate and lower latency.
Based on PPRZLINK v2.

### DIFF
--- a/sw/airborne/modules/multi/traffic_info.c
+++ b/sw/airborne/modules/multi/traffic_info.c
@@ -79,7 +79,11 @@ bool parse_acinfo_dl(void)
   uint8_t msg_id = IdOfPprzMsg(dl_buffer);
 
   /* handle telemetry message */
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/multi/traffic_info.c
```diff
@@ -79,7 +79,11 @@ bool parse_acinfo_dl(void)
   uint8_t msg_id = IdOfPprzMsg(dl_buffer);
 
   /* handle telemetry message */
+#if PPRZLINK_DEFAULT_VER == 2
```
kirkscheper: Looks good, only I just now noticed this PPRZLINK_DEFAULT_VER in the pprzlink repo. For my own info, is there a reason it is 'default', is there a way to have multiple versions simultaneously where one is default and another isn't? Why not just pprzlink_version?

PR #2106, opened 2017-08-21
[flight_plan] add binding to ABI messages from flight plan

gautierhattenberger:
This feature allows to create an automatic binding to a ABI message
while specifying the list of variable names ('_' are ignored).
A local (static) variable is updated to the latest value each time a new
message is received and it can be read in any flight plan instruction.

Example of usage:

 <variables>
  <abi_binding name="VELOCITY_ESTIMATE" vars="_, vx, vy, vz, _"/>
 </variables>

please not the in this example, the first and the last variable are
ignored but the number of variable must match the number of fields in
ABI messages description (conf/abi.xml).

An extra attribute 'id' allows to set a specific sender ID (by default
ABI_BROADCAST).


knmcguire:
Eventhough I can not review this pull request properly, I did test out this functionality quickly in the flightplan in gazebo, with the AGL abi-message:
 _<variables>
  <abi_binding name="AGL" vars="height"/>
 </variables>_
 I created a block with an downlink  message just to see if the variable actually updated:
    _<block name="read abi">
      <call_once fun="DOWNLINK_SEND_ESTIMATOR(DefaultChannel, DefaultDevice,&height, &height)"/>
      <deroute block="read abi"/>
    </block>_
And the value was exactly the same (with a slight delay) as the sonar_distance variable from SONAR telemetry message. So it seems to do the trick.
screenshot from 2017-08-21 22-17-09

PR #2107, opened 2017-08-24
Change stereocamera message parser to pprzlink

kirkscheper:
This replaces the message parser for the stereo camera from ITU-R BT.656 to pprzlink. With this I am also squeezing the functionality of stereocam2state and state2stereocam into the stereocam file. This should help with maintenance of docs and code. I updated the airframes that used the stereocam and added an airframe of my own to ensure that this module is included in tests in the future.

This does break the droplet module however. This implementation of the droplet will be replaced by a better one in the coming weeks.


conf/flight_plans/flight_plan.dtd
```diff
@@ -14,8 +14,9 @@
 <!ELEMENT corner EMPTY>
 <!ELEMENT kml EMPTY>
 
-<!ELEMENT variables (variable*)>
+<!ELEMENT variables (variable|abi_binding)*>
```
gautierhattenberger: Looks like this pull request is not correctly rebased on top of master

kirkscheper: yea, doesn't look right at all.

conf/airframes/KS/ks_bebop2_stereo.xml
```diff
@@ -0,0 +1,233 @@
...
+      <define name="AHRS_HEADING_UPDATE_GPS_MIN_SPEED" value="0"/>
+    </module>
+    <module name="ins" type="extended"/>
+    <!-- Disable use of GPS: -->
+    <!--    <define name="INS_INT_GPS_ID" value="ABI_DISABLE"/>
+    </module> -->
+    
+  </firmware>
+
+  <modules main_freq="512">
```
kirkscheper: Will do. However I used an airframe from the TUDELFT as a template for this, maybe you should just change all the airframes on master to be only modules if that is where we are going. Might be nice to do that before v5.12.

gautierhattenberger: This will be indeed a big help, although we won't enforce this for v5.12 release but hopefully the next one.

gautierhattenberger:
sorry, you need to fix a few conflicts here after #1909

Issue #2068, opened 2017-05-30
RTP streaming problem on ARDrone2


tomvand:
Using the master branch of Paparazzi, RTP video streaming does not seem to work on the ARDrone2.
I am using the examples/ardrone2.xml airframe which contains the video_thread module and video_rtp_stream. When I open the .sdp file with VLC to view this stream, I only receive a single frame before the video stops. VLC's media statistics show that only one video block was decoded. I added a printf to rtp_packet_send(...) and verified that the drone *is* sending new packets.

This problem did not occur with v5.10 of Paparazzi. Overwriting modules/computer_vision/viewvideo.c, /lib/encoding/rtp.c and .h with their v5.10 versions also fixes the problem.

I'm using Ubuntu 16.04 with VLC 2.2.2.

OpenUAS:
Yes, now looking into it...

ETitus:
I've had the same problem as well at some point when I pulled master. Before that I was on v5.10. 

I assumed it was an error at my end though, as **closing VLC and reopening the .sdp file actually shows you a new single frame**. However, now it seems a true error.

OpenUAS:
@ETitus Thanks, yes it is a real issue on ARDrone2, will test on BB1 and BB2 asap. Assigne issue to self for the time being of retesting. It would be great if @wilcoschoneveld could look into it.

dewagter:
RTP is super dependent on time. Real-time Transport, in fact, has
everything to do with time. The default behavior for a viewer is to buffer
until it is sure it can show the stream without hesitations. So for a few
seconds, the maximal download delay is analyzed and the stream is shown
with this delay (so the buffer never depletes and the viewer can show the
steam without interruption).

see:
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/lib/encoding/rtp.c#L133

Most likely something with frame time here...



-Christophe

dewagter:
Actually: I think I got it:

https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/lib/encoding/rtp.c#L110

the seconds must be multiplied by 1000000 to get microseconds. otherwise
you never reach 1 second.

Titus, Tom and/or Erik, can you try this?

Previously it had the time move slower than real-time so the video was
always outdated and that caused most viewers to show it with less delay.
Maybe this can still be an option with a define for instance:

[...]

PR #2110, opened 2017-09-07
[RTP] Revert to protocol specification: namely linearly increasing co…

dewagter:
…unter with 90kHz rate for FMT26. See #2068

### DIFF
--- a/sw/airborne/modules/computer_vision/lib/encoding/rtp.c
+++ b/sw/airborne/modules/computer_vision/lib/encoding/rtp.c
@@ -31,11 +31,39 @@
 
 #include "rtp.h"
 
-static void rtp_packet_send(struct UdpSocket *udp, uint8_t *Jpeg, int JpegLen, uint32_t m_SequenceNumber,
+static void rtp_packet_send(struct UdpSocket *udp, uint8_t *Jpeg, int JpegLen, uint16_t m_SequenceNumber,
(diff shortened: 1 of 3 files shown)


dewagter:
not tested yet

dewagter:
This pull request reverts the change of putting the image time in milliseconds of the day multiplied by 9/100 in the RTP stream as this is not acceptable for the protocol. The RTP protocol requires a linearly increasing time stamp and it is less important that it has the correct seconds but more important that it is constant.

Changes:

 - rename: frame_number to packet number (there is no frame number, only frame time in RTP)
 - change packet_number to protocol 16bit
 - create a frame_time_counter with 90kHz resolution
 - use the VIDEO_RATE to increment it with fixed intervals
 - update all doxygen and add references to protocol docs

This code was tested on Ardrone2:

 - before this commit, I also only get the first frame and then nothing anymore
 - with this update, I get minimal delay (<0.5sec) streaming video

OpenUAS:
Grr have missed the request, shoul refresh my browser window more often ; )

gautierhattenberger:
is this closed with #2068 ?

dewagter:
Yes, the patch was pushed to master separately: https://github.com/paparazzi/paparazzi/commit/57772d6a8a1ddbc17432318c94f7221c83943d97

PR #2111, opened 2017-09-11
Distributed circular formation

noether:
I think it is ready for being merged in master. It is a nice example for the new Paparazzi Link 2.0.

Let me know whether I have to do something else (btw, the paparazziproxy.py is also included... maybe Gautier has other plans for this).


gautierhattenberger:
I think it is not properly rebased/merged with latest master

noether:
Oki

Shall I do a rebase or pull from the current master?

gautierhattenberger:
I will squash everything into a single commit in the end, so I guess it doesn't really matter.

conf/airframes/TUDELFT/tudelft_conf.xml
```diff
@@ -123,6 +123,7 @@
   <aircraft
    name="CX10"
    ac_id="18"
+   ac_id="15"
```
kirkscheper: this is still not properly rebased on master

sw/airborne/modules/dcf/dcf.c
```diff
@@ -0,0 +1,178 @@
...
+        tableNei[i][3] = dcf_control.timeout;
+      } else {
+        tableNei[i][3] = (uint16_t)timeout;
+
+        float e = dcf_control.theta - (tableNei[i][1] + (tableNei[i][2])) * M_PI / 1800.0;
+        if (e > M_PI) {
+          e -= 2 * M_PI;
+        } else if (e <= -M_PI) {
+          e += 2 * M_PI;
+        }
```
kirkscheper: it is better to use a while loop to wrap the angle. You should use the macro NormRadAngle(e) (from include/std.h) here instead.

sw/airborne/modules/dcf/dcf.c
```diff
@@ -0,0 +1,178 @@
...
+#define DCF_TIMEOUT 1500
+#endif
+
+dcf_con dcf_control = {DCF_GAIN_K, DCF_RADIUS, DCF_TIMEOUT, 0};
+
+/*! Default number of neighbors per aircraft */
+#ifndef DCF_MAX_NEIGHBORS
+#define DCF_MAX_NEIGHBORS 4
+#endif
+int16_t tableNei[DCF_MAX_NEIGHBORS][4];
```
kirkscheper: Can you make this an array of structs with elements containing nei_id, theta, desired_sigma and the last_theta (merge the last_theta and tableNei). It will make the use of the elements much more intuitive.

Additionally, consider using more descriptive variable names? Theta and sigma are not very intuitive without any explanation. Another one last_theta -> msg_time

It also doesn't look like the tableNei[i][3] is used anywhere, why do you store the timeout?

sw/airborne/modules/dcf/dcf.h
```diff
@@ -0,0 +1,52 @@
...
+#define DCF_H
+
+#include "std.h"
+
+typedef struct {
+  float k;
+  float radius;
+  uint16_t timeout;
+  float theta;
+} dcf_con;
```
kirkscheper: We don't typedef structs in paparazzi. define the strtuct then reference it with struct dcf_con. This helps to reinforce that these variables are structs.

sw/ground_segment/python/udp_link/pprzlink_proxy.py
```diff
@@ -0,0 +1,99 @@
...
+
+# GCS ID, OUT port, IN port
+gcs_conf = [0, 4243, 4242]
+
+# list ID, OUT port, IN port
+mavs_conf = [
+        [102, 4244, 4245],
+        [103, 4246, 4247],
+        [104, 4248, 4249]
+        ]
```
kirkscheper: One would expect these to be optional program inputs (which could default to these values if not defined). Same with the DEFAULT_ADDRESS and the gcs_conf.

sw/airborne/modules/dcf/dcf.c
```diff
@@ -0,0 +1,178 @@
...
+        error_sigma[i] = (uint16_t)(e * 1800.0 / M_PI);
+      }
+    }
+  }
+
+  u *= -gvf_control.s*dcf_control.k;
+
+  gvf_ellipse_XY(xc, yc, dcf_control.radius + u, dcf_control.radius + u, 0);
+
+  if ((now - last_transmision > 200) && (autopilot_get_mode() == AP_MODE_AUTO2)) {
```
kirkscheper: This 200ms seems like it should be a variable that can be set in the airframe like DCF_TIMEOUT.

conf/modules/distributed_circular_formation.xml
```diff
@@ -0,0 +1,42 @@
...
+  </doc>
+
+  <settings name="DCF">
+    <dl_settings>
+      <dl_settings NAME="DCF">
+        <dl_settings NAME="Control">
+          <dl_setting MAX="20" MIN="0" STEP="0.2" VAR="dcf_control.k" shortname = "Gain" param="DCF_GAIN_K"/>
+          <dl_setting MAX="200" MIN="0" STEP="1" VAR="dcf_control.radius" shortname = "Radius" param="DCF_RADIUS"/>
+          <dl_setting MAX="5000" MIN="0" STEP="1" VAR="dcf_control.timeout" shortname = "Timeout" param="DCF_TIMEOUT"/>
+        </dl_settings>
```
kirkscheper: These variables should also be described above in the description as they can also be set in airframe files.

conf/modules/distributed_circular_formation.xml
```diff
@@ -0,0 +1,42 @@
...
+  <header>
+    <file name="dcf.h"/>
+  </header>
+
+  <init fun="dcf_init()"/>
+  <!--periodic fun="send_theta_to_nei()" freq="5."/-->
+
+  <datalink message="DCF_REG_TABLE" fun="parseRegTable()"/>
+  <datalink message="DCF_CLEAN_TABLE" fun="parseCleanTable()"/>
+  <datalink message="DCF_THETA" fun="parseThetaTable()"/>
```
kirkscheper: is it really nessesary to have two messages for reg_table and clean_table. Can't the clean table just be a variable in the reg_table which if set requests that the table be reset? I am just thinking that if we have many messages just containing one variable that the messages list will fill up pretty quick.

sw/airborne/modules/dcf/dcf.c
```diff
@@ -0,0 +1,178 @@
...
+  struct pprzlink_msg msg;
+
+  for (int i = 0; i < DCF_MAX_NEIGHBORS; i++)
+    if (tableNei[i][0] != -1) {
+      msg.trans = &(DefaultChannel).trans_tx;
+      msg.dev = &(DefaultDevice).device;
+      msg.sender_id = AC_ID;
+      msg.receiver_id = tableNei[i][0];
+      msg.component_id = 0;
+      pprzlink_msg_send_DCF_THETA(&msg, &(dcf_control.theta));
```
kirkscheper: If you are sending the same data to all the aircraft, why send it multiple time? Why not just send it once with the msg.receiver_id as a broadcast id like 255 or 0. Actually, you don't check the id in the parseThetaTable atm so it seems like the messages will just be received multiple times. 

I am not totally aware how the new messages are handled, @gautierhattenberger does the pprzlink decoder only propagate messages that are designated to you or do you need to check that yourself?

conf/modules/distributed_circular_formation.xml
kirkscheper: Additionally, you can combine the REG_TABLE and the THETA message together looking at the sender_id of the message, ground station messages are id 0, other aircraft would be non zero.

conf/airframes/TUDELFT/tudelft_conf.xml
noether: True, 

does smb know whether it should be 18 or 15?

conf/modules/distributed_circular_formation.xml
noether: They are described in detail the wiki:

https://wiki.paparazziuav.org/wiki/Module/guidance_vector_field#Fully_distributed_circular_formations

A reference in the head of the file will be enough?

conf/modules/distributed_circular_formation.xml
noether: DCF_THETA should be a single message since it is the only thing that need to be exchange by the aircraft while flying. Better to keep it as simple/minimum as possible.

I agree that I need to think better how to design better the set of messages/payload. The first step was to make the algorithm work. Since I believe that Gautier is thinking about how to deal with or organize the air-to-air messages (in this case they are not really 'datalink'), maybe we can leave  this suggestion for later.

sw/airborne/modules/dcf/dcf.c
noether: I agree that it would be more intuitive. However, I believe it will make the code more cumbersome in many places. What about a comment on top of this line describing each element? (they are also described in the wiki).

Sigma and theta are described in the wiki and in the corresponding papers. I will write a comment in the code and put the reference to the wiki.

I will do the same with last_theta (last time that the aircraft received a theta).

The timeout is sent in a telemetry message. An operator in the ground can monitor the frequency of lost messages and decide to have a longer timeout (if a neighbor is in 'timeout' then it left the formation).

sw/airborne/modules/dcf/dcf.c
noether: The while loop is not necessary here, since by construction it is only required one iteration. Nevertheless, I will employ the suggested function. I suspected it existed :P .

sw/airborne/modules/dcf/dcf.c
noether: Agree. In fact, it defines the frequency for the air-to-air message.

sw/airborne/modules/dcf/dcf.c
noether: It is not to all the aircraft, but to a subset of them (the neighbors of the aircraft). You can have ten aircraft flying in total but a particular aircraft only needs to interact with a subset of them. The algorithm is distributed in the sense that not everybody needs to share information with everybody.

Yes, in fact for XBee radios, the filtering of the message happens at the radio chip, since the ID of the aircraft is the same as the ID for the radio.

sw/airborne/modules/dcf/dcf.h
noether: Is this really a strict rule or a suggestion? mmm I was doing the same multiple times with the GVF algorithm (for each trajectory and for the control variables). I mean, I can stop this practice and change it here, but for the GVF it will be quite some work...

sw/ground_segment/python/udp_link/pprzlink_proxy.py
noether: Totally agree. In fact, Gautier has plans for this file. 

This was a quick patch for making the simulation work with air-to-air messages and it should not be here indeed.

noether:
Thanks @kirkscheper for the suggestions and checking the code!

conf/airframes/TUDELFT/tudelft_conf.xml
noether: True.

Any idea whether it was 15 or 18?

sw/airborne/modules/dcf/dcf.c
```diff
@@ -0,0 +1,178 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ */
+
+#include <math.h>
+#include <std.h>
+#include <stdio.h>
```
gautierhattenberger: don't forget to remove this one if no debug print

noether:
We still have to test it in simulation once the "new" Pprzlink_proxy is included in master.

sw/airborne/modules/dcf/dcf.h
```diff
@@ -0,0 +1,63 @@
...
+
+extern struct dcf_con dcf_control;
+
+struct dcf_tab {
+  int16_t tableNei[DCF_MAX_NEIGHBORS][4];
+  int16_t error_sigma[DCF_MAX_NEIGHBORS];
+  uint32_t last_theta[DCF_MAX_NEIGHBORS];
+};
+
+extern struct dcf_tab dcf_tables;
```
gautierhattenberger: wouldn't it be cleaner (or more complicated) to have a table of structure rather than a structure of tables ?

sw/ground_segment/python/gvf/dcfInitTables.py
```diff
@@ -0,0 +1,81 @@
...
+time.sleep(2)
+
+if len(list_ids) == 2:
+    B.shape = (2,1)
+
+for count, column in enumerate(B.T):
+    index = np.nonzero(column)
+    i = index[0]
+
+    msg_clean_a = PprzMessage("datalink", "DCF_REG_TABLE")
```
gautierhattenberger: it is probably possible to just make a message, send it, change the id and send it again
same thing below to init add new elements to the table

noether: Yepe, it is possible :P.  But I wanted to do it like this because for a pair of neighbors msg_a for one, msg_b for the other one. Easy to debug and write when I wrote the script for the first time.

noether:
oh, now I realize about the post address in the licence and to eliminate this TUDELFT xml file...

gautierhattenberger:
merged and closed with https://github.com/paparazzi/paparazzi/commit/b93251561050efa4ad5aa01ed42a660fc5c6f444

PR #2113, opened 2017-09-28
F3 and X-Vert support

kevindehecker:
Adds support for:
1. STM32F373x
2. The Horizon Hobbies Eflite X-Vert VTOL (based on the STM32F373xC)
3. STM32F303x (preliminary)
4. The F3 Discovery board (based on the STM32F303)

All are based on Chibios. In order to fit the RAM memory footprint on the 32K RAM of the STM32F37, the thread working area sizes have been reduced. It all seems to work without fault, so I don't think it is necessary to increase the size for larger chips such as the F7. 
Another thing is that the PPRZ_TRIG_CONST needed to be defined as CONST. This placed the sin integer lookup table on flash. The other defines such as PPRZ_TRIG_INT_COMPR_FLASH do not seem to have any effect. Of course, the new define PPRZ_TRIG_INT_USE_FLOAT also works.

The x-vert has been flight tested. It can hover manually. Further steps (autonomous, forward flight) are planned. A wiki page about the x-vert is in the works.

Known problems:
The X-vert ESCs provide motor feedback, but this is not yet read out
The X-vert may have a current sensor, but it is not yet read out
The X-vert board has an CYRF chip for radio control, but this is not supported yet in this PR.
The Chibios stm32F37 implementation has a problem with the hard FPU. When enabled, totally weird results happen.


podhrmic:
I ll look over it over the weekend.

conf/boards/xvert_1.0.makefile
```diff
@@ -0,0 +1,71 @@
...
+
+ARCH=chibios
+$(TARGET).ARCHDIR = $(ARCH)
+
+RTOS=chibios
+
+## FPU on F3
+
+##TODO: there is some bug when using the hard FPU unit!!
+USE_FPU=no
```
podhrmic: can you file a bug report at ChibiOS forum about it? Giovanni is typically fairly responsive.

sw/airborne/arch/chibios/mcu_periph/adc_arch.c
```diff
@@ -344,25 +349,45 @@ void adc_init(void)
...
   adc_watchdog.channel = 0;
-  adc_watchdog.vmin = (1<<12)-1; // max adc
+  adc_watchdog.vmin = (1 << 12) - 1; // max adc
 #endif
 
   adcgrpcfg.circular = TRUE;
   adcgrpcfg.num_channels = ADC_NUM_CHANNELS;
   adcgrpcfg.end_cb = adc1callback;
   adcgrpcfg.error_cb = adcerrorcallback;
+#if defined(__STM32F373xC_H)
```
podhrmic: it would be nice to have a better way of specifying the driver than adding ifdefs, perhaps separate files for each chibios arch - but that is indeed out of scope of this PR

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
```diff
@@ -86,38 +86,38 @@ static void handle_i2c_thd(struct i2c_periph *p)
...
-    memcpy(i->dma_buf, (void*)t->buf, (size_t)(t->len_w));
+    memcpy(i->dma_buf, (void *)t->buf, (size_t)(t->len_w));
     status = i2cMasterTransmitTimeout(
-        (I2CDriver*)p->reg_addr,
-        (i2caddr_t)((t->slave_addr)>>1),
-        (uint8_t*)i->dma_buf, (size_t)(t->len_w),
-        (uint8_t*)i->dma_buf, (size_t)(t->len_r),
-        tmo);
-    memcpy((void*)t->buf, i->dma_buf, (size_t)(t->len_r));
+               (I2CDriver *)p->reg_addr,
```
podhrmic: stylecheck pls?

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
```diff
@@ -198,7 +198,7 @@ static struct i2c_init i2c1_init_s = {
 struct i2c_errors i2c1_errors;
 // Thread
 static __attribute__((noreturn)) void thd_i2c1(void *arg);
-static THD_WORKING_AREA(wa_thd_i2c1, 1024);
+static THD_WORKING_AREA(wa_thd_i2c1, 128);
```
podhrmic: I would be worried that this might overflow for larger i2c buffers - could you add `ifndef` guard and specify this number in the board file to override the default?

sw/airborne/arch/chibios/mcu_periph/spi_arch.c
```diff
@@ -347,15 +346,15 @@ static __attribute__((noreturn)) void thd_spi1(void *arg)
   }
 }
 
-static THD_WORKING_AREA(wa_thd_spi1, 1024);
+static THD_WORKING_AREA(wa_thd_spi1, 256);
```
podhrmic: again - would be better to have this board-dependent with larger defaults.

sw/airborne/arch/chibios/mcu_periph/uart_arch.c
```diff
@@ -128,7 +128,7 @@ static __attribute__((noreturn)) void thd_uart1_rx(void *arg)
   }
 }
 
-static THD_WORKING_AREA(wa_thd_uart1_rx, 1024);
+static THD_WORKING_AREA(wa_thd_uart1_rx, 256);
```
podhrmic: dtto - board specific defines for the thread sizes if needed to be smaller than defaults

sw/airborne/firmwares/rotorcraft/main_chibios.c
```diff
@@ -42,7 +42,7 @@
  * Also include FBW on single MCU
  */
 static void thd_ap(void *arg);
-static THD_WORKING_AREA(wa_thd_ap, 8192);
+THD_WORKING_AREA(wa_thd_ap, 1024);
```
podhrmic: again - depending on the autopilot configuration it might not fit in 1k, should be board dependent

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
```diff
@@ -208,324 +208,325 @@ void ahrs_fc_update_accel(struct FloatVect3 *accel, float dt)
...
-     * <0.66G = 0, 1G = 1.0, >1.33G = 0
-     */
-
-    const float g_meas_norm = float_vect3_norm(&filtered_gravity_measurement) / 9.81;
-    ahrs_fc.weight = 1.0 - ahrs_fc.gravity_heuristic_factor * fabs(1.0 - g_meas_norm) / 10.0;
-    Bound(ahrs_fc.weight, 0.15, 1.0);
-  } else {
-    ahrs_fc.weight = 1.0;
-  }
+    static struct FloatVect3 filtered_gravity_measurement = {0., 0., 0.};
```
podhrmic: isn't pprz style with 2 spaces? Pls run the stylecheck script on this.

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
```diff
@@ -208,324 +208,325 @@ void ahrs_fc_update_accel(struct FloatVect3 *accel, float dt)
...
-  const float mag_rate_update_gain = 2.5;
-  RATES_ADD_SCALED_VECT(ahrs_fc.rate_correction, r2, (mag_rate_update_gain * res_norm));
-  const float mag_bias_update_gain = -2.5e-4;
-  RATES_ADD_SCALED_VECT(ahrs_fc.gyro_bias, r2, (mag_bias_update_gain * res_norm));
+    /* project mag on local tangeant plane */
+    struct FloatEulers ltp_to_imu_euler;
+    float_eulers_of_rmat(&ltp_to_imu_euler, &ahrs_fc.ltp_to_imu_rmat);
 
-}
+    const float cphi   = cosf(ltp_to_imu_euler.phi);
```
podhrmic: what is the reason for changes here (they aren't mentioned in the PR description)?

sw/tools/qtc.creator.user_template
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE QtCreatorProject>
-<!-- Written by QtCreator 4.1.0, 2017-01-23T23:23:44. -->
+<!-- Written by QtCreator 4.4.0, 2017-09-08T21:31:25. -->
```
podhrmic: this file should probably be a part of .gitignore

sw/tools/find_vpaths.py
```diff
@@ -3,55 +3,71 @@
 #It finds the correct VPATH to each source file from make, and then feeds all info to gcc -MM.
 import sys
 import os
-from os import path, getenv
+from os import path, getenv, environ
```
podhrmic: how are these changes related to PR?

sw/tools/qt_project.py
```diff
@@ -96,8 +96,8 @@
 # Parse the header files
 headers_all =  ""
 line = "1"
-while line:
-    line=ap_srcs_list.readline()
+for line in ap_srcs_list:
+    #line=ap_srcs_list.readline()
```
podhrmic: remove the comment?

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
kevindehecker: Actually, all these changes were done by fix_code_style

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
kevindehecker: Weird, all this is done by fix_code_style. I only did the two minor changes in commit 8cdc6d7057618a0a5bf4735bcc7a63e40ce927e4

sw/tools/find_vpaths.py
kevindehecker: In order to get a good qt project with a chibios based board, some changes were needed. Since they are in a clear seperate commit I did not bother creating a separate PR

sw/tools/qtc.creator.user_template
kevindehecker: Actually, it should not. The template is used to generate an accepted qtc project.user. Unfortunately it currently means that the template has to be updated when qtc changes. Looking for a better solution, but nothing as of yet.
(other than that, I can recommend qt over eclipse. It only shows you the files that are actually used)

conf/boards/xvert_1.0.makefile
kevindehecker: It's probably something with wrong settings for the minimal float number. Was planning to look into it a bit further myself later.

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
kevindehecker: Fix_code_style is actually what caused this...

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
kevindehecker: I don't think that's how it works. The data goes in different buffers. E.g. the ublox I2C gps (used in this PR) actually is a rather large buffer user compared to al other sensors, and it does not seem to be a problem. Also, the chibios examples themselves use this number. I think they were oversized in pprz. Of course could still add the ifdef overrides if still required, but think these numbers here should be the default.

I'm not sure though what happens when the buffer is to small, I may try that when I get the chance, 
It would also be nice if someone with e.g. a Chimera can test this.

(same for all the other THD_WORKING_AREA's)

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
kevindehecker: If I run it again, it doesn't change it. It does give some weird error warning though:
~/paparazzi$./fix_code_style.sh sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
Using astyle version 2.05.1
(standard_in) 1: syntax error
./fix_code_style.sh: line 18: [: -eq: unary operator expected
Unchanged  sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c

podhrmic: Don't know why the script does the warning, maybe @flixr knows more?

However, the file broke the style check - compare lines 206 `const struct FloatVect3 vel_tangential_body =` before and after. Clearly the correct style is before applying the style check. 

Pls revert and check your changes manually. 

Also there is a typo at line 212 - the end bracket should be `}` instead of `)`

sw/airborne/subsystems/ahrs/ahrs_float_cmpl.c
podhrmic: Ok makes sense

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
podhrmic: OK, sounds reasonable. I think ChibiStudio plugin for Eclipse can tell you the used memory for each thread.

podhrmic:
@kevindehecker the only remaining thing is the code style for ahrs.c file (see my comment above), the rest looks good (at least for the time being).

kevindehecker:
Done!

podhrmic:
LGTM. @gautierhattenberger or @flixr do you want to add anything or good to merge?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -340,9 +340,17 @@ static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct I
   indi.u_in.r = indi.u[2].o[0] + indi.du.r;
 
   //bound the total control input
+#ifdef INDI_FULL_AUTHORITY
```
flixr: This new option should be documented in the module xml....

flixr: Also please use `#if FLAG` instead of `#ifdef FLAG` according to http://docs.paparazziuav.org/latest/stylec.html#stylecpp

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
gautierhattenberger: i2c buffers are global variables so not counted for the stack size. Only automatic variables (more or less the local variables) created by the thread will use the stack memory, a few pointers in the case of this driver. So I guess 128 is fine.

sw/airborne/firmwares/rotorcraft/main_chibios.c
gautierhattenberger: Here, I don't agree to reduce the stack size. We are running some estimation filters that requires some memory. So I suggest to keep the 8K by default and make it configurable.

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
gautierhattenberger: I wrote this part of the code. The code_style script should be correct, it is probably my VIM settings for aligning multi-line parameters that are a bit different.

PR #2114, opened 2017-10-10
Adding rangesensor module

knmcguire:
This pull request includes several items relating to adding range sensors to MAVs:
- Compatibility to read out the  new ToF laser ring through UART (designed by @dewagter)
- Adding laser range sensors to the ardrone 2.0 gazebo model (Bottom laser range sensor is send through like a sort of sonar)
- Made an module that creates a velocity forcefield based on those sensors, which can be used within guided flight.

Please let me know if you have questions or comments about this pull request


gautierhattenberger:
First, a general question: it seems that the range sensor data are not much configurable (up, down, right, left, front, back) in the module itself and in the ABI message. Would it be possible to have an array to store the measurements and some parameters to specify the range or directions ? This way, we could use several types of laser range finder with the same API.
What do you think ?

knmcguire:
Indeed, that what I first had intended, but it was a bit difficult to select the sensors without knowing the exact name, that's why I named it explicitly bottom front etc. But you are right it is not really generic.

I could have another look at it, and then I can add them to the abi message as in one array for the values, and 3 arrays for the Euler angles in which they are oriented. However, then I do have to make the range sensor forcefield module as generic as well, so that will require a bit more coding :) but probably will be for the best.

knmcguire:
I tried to make it more generic, and was partly successful. There are still 2 major errors that I couldn't solve, one gazebo related and one paparazzi related:

- Gazebo: Apparently I can not read out the pose of the sensor and retrieve the orientation from that (I just get zeros). I believe this is a bug in gazebo since I can't do this for the cameras as well, so I'll report it to them. In the mean time, I have a little hack, where I write the orientation in the sensor name in a fixed manner, so I can retrieve it from the string. This is just a very temporary solution.
- Paparazzi: If I add an array to the message of the abi message like this:  "<field name="distance" type="int16_t[]" unit="mm"/>" if I remake paparazzi, it compiles with no problem, but if I compile the NPS, I get a whole list of errors of the abi message... So this is probably a bug.

kirkscheper:
I will wait to give detailed feedback till the abi stuff is fixed but here are some high level comments. 

Try to generalize your work, Consider your naming, e.g "RANGE_FORCEFIELD" could just be FORCEFIELD as the signal could be generated by a different sensor process. another example is "range_finders.left" this could be more generic and descriptive e.g. obstacle_dist.left. The naming of range_sensor is also quite generic, I guess in most of your uses in this PR range sensors means a ring of laser range finders. Try to be specific when that is useful (eg reading the data from the sensor ring) and generic when that is useful (e.g. using the data that is transferred after the abi).

The modules section of the airframe is no longer used, all modules should be in the firmware section.

Try to keep the list of commits short and descriptive. A commit should contain a set of code changes that are dependent on each other and contain a single idea. The idea of a commit is that the user can checkout any commit and the code should compile. Commits like "[gazebo] added range modules, but gazebo is unable to find it! maybe add range sensors back to the ardrone model directly??" or "[range_module] latest changes based on the IMAV2017" are not vary useful. Try squashing some commits to just the commits that are useful or into one commit.

You sometimes use if, else if, else statements where the else is empty. I don't think that is particularly useful, just don't use an else if it should be empty.

You made changes to "mag_pitot_uart.c" which shouldn't have anything to do with this PR. It looks like you are running to the mag_pitot to parse the intermcu msgs but that is not the idea. You should make a range_sensor module read the intermcu msgs (indeed similar to way that mag_pitot or stereocam does it but in a separate module). @gautierhattenberger I can imagine situations where we may have multiple members on the same intermcu bus, we should think of a way of handling this, possibly having a module read the bus and have modules subscribe to the messages similar to the datalink messages.

It also looks like the module is not included in the airframe so isn't used by your example. On that point I also don't really like the idea of having gazebo specific airframes as the idea is that you should switch the target from ap to nps with the same airframe not make unique airframes for each target.

You quite often cast a float to a float (in range_sensor_forcefield.c) this is not good practice.

The list of "IDs of Range sensor related variables" should all start the same to reinforce the fact that they are in the same group. Additionally, there is already an AGL group. More conceptually, what is the difference between a range sensor is quite a generic term, for e.g what would be the difference between AGL_SONAR_NPS and AGL_RANGE_SENSORS. Maybe sometimes you should be more specific to the sensor type (part number). The idea of these ids is to allow modules reading the messages to discriminate the senders of the data and choose the one that it wants, in that case the grouping should not be based on the sensor type but on the data type, what is in the message. You see this with the grouping of GPS or IMU or AGL ect. Careful here.

knmcguire:
Hey Kirk, Thanks for your extensive feedback!

About the range forcefield module, after the abi message issue has been fixed, it will be completely changed to make it more generic, to add an forcefield in any orientation and translate that to the body fixed velocities. It might be a too big of a project for this pull request so I might remove it for now.

About the commit messages. It might be a personal thing, but I generally like commits with tiny changes with small changes or fix, instead of the onces with a huge diff in the code (since that can be seen already in the total diff). This specifies the thought process and it is easy to monitor the development, so I do not mind the long commit list at all. Usually I always build before I commit anyway (except for the last one but that is a special issue). But like I said, that is very personal preference and I do not find that much documentation on what the preferred strategy is. On documenting, yes it could be clearer, it is sometimes out of laziness too, especially since pull request usually get squashed on the paparazzi... 

About the mag_pitot_uart.c, this was partially code that I received from @dewagter. This pull request was more about the range forcefield module, which is used by both the NPS (by gazebo) as the AP (through intermcu), however now the focus is more on gazebo because reading out the rangesensors is a bit more work than with mcu_pitot_uart.

The range module I switched off for now, was to compile the gazebo nps just by itself, without conflicting ABI messages, to see what the error messages for that one is. 


Anyway, I will wait for the abi issue to be fixed first. In the mean time, I will be reinstalling my computer so I won't do any new commits for the next few days.

knmcguire:
I made a lot of changes to make this code more generic:

- Followed the advice of Gautier and now each ray sensor's pose it read out by NPS. However, I talked to @kirkscheper and we do agree that the orientation should be defined from the airframe file (as this is different for every vehicle). So the pose from gazebo is a check if those match
- Following up on the top, I also added the orientation define from the airframe to the mag_pitot_uart module 
- THe range forcefield model is now generic, so it will create a forcefield velocity in the oposite direction of the given orientation of the range sensor.

So I did test this out in gazebo already (the guided flightplan already uses it), however I will try to test out the forcefield on the actual platform with a laser range array on it. In the mean time, the latest changes can be reviewed and commented on. I hope it is going into the right direction

conf/airframes/TUDELFT/tudelft_origami_lisamxs_wifi_indi_stereoboard.xml
```diff
@@ -0,0 +1,320 @@
...
+   <!--module name="gps" type="ubx_ucenter"/-->
+   <module name="send_imu_mag_current"/>
+<!--Stereo camera modules: Reintergrate this with the newest pprzlink!!>
+   <module name="stereocam">
+     <define name= "STEREO_BODY_TO_STEREO_PHI" value="90" unit="deg"/>
+     <define name= "STEREO_BODY_TO_STEREO_THETA" value="0" unit="deg"/>
+     <define name= "STEREO_BODY_TO_STEREO_PSI" value="90" unit="deg"/>
+     <define name="STEREO_UART" value="UART1"/>
+   </module-->
+   <module name="mag_pitot_uart.xml">
```
kirkscheper: You shouldn't use the mag_pitot_uart module to read the messages but add a module specially to do that.

sw/airborne/modules/range_forcefield/range_forcefield.c
```diff
@@ -0,0 +1,137 @@
...
+ * @param[in] min_vel_command_lc, minimum velocity for the forcefield generation [m/s]
+ * @param[in] max_vel_command_lc, maximum velocity for the forcefield generation [m/s]
+ * @param[out] vel_body_x, adjusted body velocity in x direction [m/s]
+ * @param[out] vel_body_y, adjusted body velocity in y direction [m/s]
+ * */
+
+void range_sensor_single_velocity_force_field(struct FloatVect3 *vel_avoid_body, float range, struct FloatEulers *body_to_sensors_eulers,
+    float avoid_inner_border, float avoid_outer_border, float min_vel_command_lc, float max_vel_command_lc)
+{
+  static const float max_sensor_range = 2;
```
kirkscheper: This depends on the sensors used and should then be a variable that can be overwritten from the airframe.

sw/airborne/modules/range_forcefield/range_forcefield.c
```diff
@@ -0,0 +1,137 @@
...
+  float difference_inner_outer = avoid_outer_border - avoid_inner_border;
```
kirkscheper: This i only used in the else, compute it there. Also, since you only use this computation once I wouldn't recommend assigning it to a variable #efficiencyefficiencyefficiency

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -83,6 +85,35 @@ static struct gazebocam_t gazebo_cams[VIDEO_THREAD_MAX_CAMERAS] =
 { { NULL, 0 } };
 #endif
 
+#ifdef NPS_SIMULATE_RANGE_SENSORS
+#include "subsystems/abi.h"
```
kirkscheper: this include should be inside of  extern "C" {} like on line 46. You can just add it here around the includes.

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -83,6 +85,35 @@ static struct gazebocam_t gazebo_cams[VIDEO_THREAD_MAX_CAMERAS] =
...
+static void gazebo_init_range_sensors(void);
+static void gazebo_read_range_sensors(void);
+#ifndef NPS_GAZEBO_RANGE_MAX_SENSORS
+#define NPS_GAZEBO_RANGE_MAX_SENSORS 255
+#endif
```
kirkscheper: This seems unnecessary as the number of range sensors is defined in the model and is fixed.

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -83,6 +85,35 @@ static struct gazebocam_t gazebo_cams[VIDEO_THREAD_MAX_CAMERAS] =
...
+
+gazebo::sensors::RaySensorPtr RaySensorPtr_array[NPS_GAZEBO_RANGE_MAX_SENSORS];
+
+
+#ifndef NPS_GAZEBO_RANGE_AMOUNT_SENSORS
+#define NPS_GAZEBO_RANGE_AMOUNT_SENSORS 0
```
kirkscheper: NUMBER instead of amount

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -83,6 +85,35 @@ static struct gazebocam_t gazebo_cams[VIDEO_THREAD_MAX_CAMERAS] =
...
+#endif
+
+
+#ifndef NPS_GAZEBO_RANGE_ORIENTATION
+#define NPS_GAZEBO_RANGE_ORIENTATION 0, 0, 0
+#endif
```
kirkscheper: This should not be used but rather use the REMOTE_GROUND_ORIENTATIONS, this whay simulation and reality are the same.

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -585,4 +624,130 @@ static void read_image(
...
+        range_sensors_uint16[index_of_raysensor] = UINT16_MAX;
+      }
+      for (int n = 0; n < 3; n++) {
+        range_sensors_orientation_int32[index_of_raysensor * 3 + n] =
+          ANGLE_BFP_OF_REAL(range_orientation[index_of_raysensor * 3 + n]);
+      }
+      ray_sensor_count_selected++;
+    }
+
+#ifdef NPS_GAZEBO_RANGE_ORIENTATION_AGL
```
kirkscheper: again, this should be the same define that is used in flight

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -585,4 +624,130 @@ static void read_image(
...
+    /* One of the range sensors can be selected to act like a sonar. It will check which of
+     * the ray sensors is looking downwards (or any orientation you specify), within 5 degrees
+     * of variation, and select that as the sonar.
+     */
+    if (RaySensorPtr_array[i]->Range(0) != 0 && !isinf(RaySensorPtr_array[i]->Range(0)) &&
+        RadOfDeg(5) > fabs((float)pose_sensor.rot.GetRoll() - range_orientation_agl[0]) &&
+        RadOfDeg(5) > fabs((float)(-1 * pose_sensor.rot.GetPitch()) - range_orientation_agl[1]) &&
+        RadOfDeg(5) > fabs((float)(-1 * pose_sensor.rot.GetYaw()) - range_orientation_agl[2])) {
+      range_sensor_down = RaySensorPtr_array[i]->Range(0);
```
kirkscheper: alg instead of range_sensor_down. Also agl should be from the angle looking down not from the angle_orientation_agl matrix.

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -585,4 +624,130 @@ static void read_image(
...
+  if (ray_sensor_count_selected != NPS_GAZEBO_RANGE_AMOUNT_SENSORS)
+    cout << "ERROR: you have defined " << NPS_GAZEBO_RANGE_AMOUNT_SENSORS << " sensors in your airframe file, but only "
+         << (int)ray_sensor_count_selected << " sensors have been found in the gazebo simulator, "
+         "with the same orientation as in the airframe file " << endl;
+
+  // SEND ABI MESSAGES
+  // Standard range sensor message
+  AbiSendMsgRANGE_SENSORS_ARRAY(RANGE_SENSOR_ARRAY_RAY_SENSOR_GAZEBO_ID, ray_sensor_count_selected, range_sensors_uint16,  range_sensors_orientation_int32);
+  // Down range sensor as "Sonar"
+  if (range_sensor_down != 0) {
```
kirkscheper: range_sensor_down is a float, not a good inequality

conf/airframes/examples/ardrone2_rangesensors_gazebo.xml
```diff
@@ -0,0 +1,222 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
```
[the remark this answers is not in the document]
knmcguire: I do agree with this feedback, however it seems that in reality this is not a simple... Not only do I have to change the dynamics of the ardrone, I also have to change the values of imu, motor_mixing, servos (since the pocketdrone is in a plus configuration). 

The airframe will get very cluttered if I separate all of this in in the <target> under <firmware> tag, I wanted to do this in the <sections> or <servos> instead but the <target> tag does not exist there. Any ideas on how to make this more elegant?

sw/simulator/nps/nps_fdm_gazebo.cpp
knmcguire: changing the name to alg I agree. I would like to keep the range_orientation_agl since it seems less hardcoded and clean, so at least it is clear what the downward angle is, which is defined on top of the this file.

sw/simulator/nps/nps_fdm_gazebo.cpp
knmcguire: That is true, but the number of sensors can not be read out until gazebo is initialized.  So this value is just to be save that it will not overflow (this amount is not the same as the number of sensors that will be selected to read out the value).

kirkscheper: Now you have the variable defined as REMOTE_GROUND_AMOUNT_SENSORS, this is available when you compile papaarazzi in generated/airframes.h. You can just use that

knmcguire:
Alright, the next big update to this pull request. The biggest changes:

- Made a separate module to read out the laser ring (called laser_range_array)
- In gazebo, the orientations are just read out in the init function, improves efficiency

sw/airborne/modules/range_finder/laser_range_array.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef LASER_RANGE_ARRAY_H
+#define LASER_RANGE_ARRAY_H
+
+#include "std.h"
+#include "generated/airframe.h"
+#include "pprzlink/pprz_transport.h"
+
+/* Main magneto pitot strcuture */
```
gautierhattenberger: wrong comment

gautierhattenberger:
See my last (small) comment and the conflicting files. After this it is I think good to merge.

kirkscheper:
Closed in preference of #2158

PR #2117, opened 2017-10-14
Fix bugs of quat_from_earth_cmd_f()

sulijune:
[rotorcraft] stabilization quat_from_earth_cmd_f() :  fix b_x and b calculation
- b_x is rotation matrix first row in NED
- dot(b,n) = 0

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.c
@@ -76,12 +76,12 @@ void quat_from_earth_cmd_f(struct FloatQuat *quat, struct FloatVect2 *cmd, float
   /* as rotation matrix */
   struct FloatRMat R_rp;
   float_rmat_of_quat(&R_rp, &q_rp);
-  /* body x-axis (before heading command) is first column */
+  /* body x-axis (before heading command) is first row */
   struct FloatVect3 b_x;
-  VECT3_ASSIGN(b_x, R_rp.m[0], R_rp.m[3], R_rp.m[6]);
-  /* body z-axis (thrust vect) is last column */
+  VECT3_ASSIGN(b_x, R_rp.m[0], R_rp.m[1], R_rp.m[2]);
+  /* body z-axis (thrust vect) is last row */
   struct FloatVect3 thrust_vect;
-  VECT3_ASSIGN(thrust_vect, R_rp.m[2], R_rp.m[5], R_rp.m[8]);
+  VECT3_ASSIGN(thrust_vect, R_rp.m[6], R_rp.m[7], R_rp.m[8]);
 
   /// @todo optimize yaw angle calculation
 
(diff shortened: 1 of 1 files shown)


sulijune:
Fix bugs of quat_from_earth_cmd_f()

sulijune:
Fix bugs of quat_from_earth_cmd_f()

sulijune:
When I test quat_from_earth_cmd_f() , I transform `quat` into Euler angles, psi is not equate to `heading`. I think the `b_x` and `b` may be wrong.
I calculate these vectors in the NED Coordinate System.

kirkscheper:
@EwoudSmeur can you give a once over?

EwoudSmeur:
I think you are right, and the Euler psi should correspond to het heading, at least if the pitch angle does not approach 90 degrees. Maybe you can give more explanation on why you state that:
    b_x is rotation matrix first row in NED
    dot(b,n) = 0
Like it is explained in the code, it seems to me that b_x is the aircraft body axis in the NED frame, so then it should be the column. Maybe @flixr can say something about it?

sulijune:
# the test code:
paparazzi/sw/Makefile:
```Makefile
Q=@
BUILDDIR= .

CC= gcc
CFLAGS= -fpic
INCLUDES= -Iairborne/firmwares/rotorcraft/stabilization -Iairborne -Iinclude

SRC = ntest_quat.c
SRC += $(wildcard airborne/math/*.c)
SRC += airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.c

OBJ= $(SRC:.c=.o)

LIBNAME=libpprzmath

all:$(OBJ)
	$(Q)$(CC) $(OBJ) -lm

shared_lib: $(OBJ)
	@echo CREATE shared $(LIBNAME)
	$(Q)$(CC) -shared -o $(BUILDDIR)/$(LIBNAME).so $(OBJ) -lm

%.o: %.c
	@echo $(SRC)
	$(CC) -c $< $(CFLAGS) $(INCLUDES) -o $@

clean:
	$(Q)rm -f $(BUILDDIR)/*.o $(BUILDDIR)/$(LIBNAME).so

.PHONY: all clean shared_lib install_shared_lib
```

paparazzi/sw/ntest_quat.c:
```C
#include <stdio.h>
#include "stabilization_attitude_quat_transformations.h"

int main(int argc, char *argv[])
{
	struct FloatQuat quat;
	struct FloatVect2 cmd = {
		-(M_PI)/3,(M_PI)/3
	};
	float heading = 1.0;
	quat_from_earth_cmd_f(&quat, &cmd, heading);
	  printf("heading = %f\n", heading);
	  printf("Psi     = %f\n", atan2f((quat.qx*quat.qy+quat.qi*quat.qz)*2,(1-2*(quat.qy*quat.qy+quat.qz*quat.qz))));
	return 0;
}
```

I think  `heading` = `Psi`.

sulijune:
equations
picture

flixr:
I'll take a closer look at this asap... but that unfortunately now means not before the 16th...

OpenUAS:
@flixr Would like to test, a closer look really appreciated

kirkscheper:
@flixr do you have time to look at this? Otherwise I think we will just merge cause its seems fine.

gautierhattenberger:
@OpenUAS v5.14 is already released, so either make a 5.16 or a 5.14_backport milestone, but it is confusing to use something already done

OpenUAS:
:tada: so now can re-testfly in a spring sun day, (... or snowy one ;))

PR #2119, opened 2017-10-15
[ins] updated hff to accept velocity as only input and add bias

kirkscheper:
This adds a bias estimate to the hff filter and allows the filter to run with only velocity inputs as well as position estimates. This also adds the possibility to have a local position estimate added to the ins. 

This has been tested with a bebop 2 and a pocket drone. It is important to define the noise characteristics to suit the drone being used but once done the performance is quite good.

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_ardrone2_opticflow_indi.xml
+++ b/conf/airframes/TUDELFT/tudelft_ardrone2_opticflow_indi.xml
@@ -11,7 +11,7 @@
 
     <define name="USE_SONAR" value="TRUE"/>
 
-	<!-- Subsystem section -->
+    <!-- Subsystem section -->
     <module name="telemetry" type="transparent_udp"/>
     <module name="radio_control" type="datalink"/>
     <module name="motor_mixing"/>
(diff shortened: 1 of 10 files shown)


kirkscheper:
I just included the messges.xml here (and in the other PR) for the review (and so others can easily test). It will not be included in the final version.

conf/airframes/TUDELFT/tudelft_ardrone2_opticflow_indi.xml
```diff
@@ -24,9 +24,7 @@
       <define name="AHRS_USE_GPS_HEADING" value="FALSE"/>
     </module>
     <module name="ins" type="extended"/>
-  </firmware>
-
-  <modules main_freq="512">
+
```
guidoAI: Kirk Kirk Kirk

guidoAI: Please make it hff_extended

kirkscheper:
I did a flight test with this today in combination with the updated opticalflow and was able to fly pretty accurate square patter using body relative position offsets in guided mode. It all seems to work pretty well!

gautierhattenberger:
you can then remove the messages.xml file and update for the correct version of PPRZLINK, then it is good to merge

PR #2120, opened 2017-10-15
[sim] updates to gazebo so simulate pressure and acceleration

kirkscheper:
Added some additional warnings to the simulator to better resemble the flight code and addressed the new warnings.
Added the capability to turn off the GPS, useful for testing indoor environments.
Added some additional debugging capabilities.

I came across a strange warning that I didn't see before:
cc1plus: warning: command line option '-std=gnu99' is valid for C/ObjC but not for C++

I can't find where the cflags are being added to the CXXFLAGS but this warning is also in the current master. Any ideas?

### DIFF
--- a/conf/Makefile.nps
+++ b/conf/Makefile.nps
@@ -32,7 +32,14 @@ DEBUG_FLAGS ?= -ggdb3
 # Launch with "make Q=''" to get full command display
 Q=@
 
-CFLAGS  = -W -Wall
+# flags for warnings (C and C++)
+WARN_FLAGS += -W -Wall -Wextra
+WARN_FLAGS += -Wunused -Wcast-align -Wpointer-arith -Wswitch-default -Wmissing-declarations
+#WARN_FLAGS += -Wcast-qual
+#WARN_FLAGS += -Wredundant-decls
+#WARN_FLAGS += -Wshadow
+
+CFLAGS  = $(WARN_FLAGS)
 CFLAGS += $(INCLUDES)
 CFLAGS += $($(TARGET).CFLAGS)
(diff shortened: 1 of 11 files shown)


kirkscheper:
The warnings seem to be a result of #2113 @kevindehecker can you take a look at it?

gautierhattenberger:
This might come from the find_vpath.py script, as it is using either gcc or g++ for all source file depending if it founds at least one c++ file or not. Compilation flags are not differentiated anymore.
Even if it is not the main issue, it should be solved I think

kevindehecker:
The error is a caused because apparently nps is compiled as as g++ project. But for c++ the flag -std=gnu99 is not valid. So, not completely sure if this flag is still required for anything but I think it can be removed. If you remove the line 
`CFLAGS += -std=gnu99`
from conf/Makefile.nps it compiles fine for me.

Otherwise we need to set up separated cxxflags and cflags and feed them accordingly to the c/c++ compiler.

gautierhattenberger:
@kevindehecker @kirkscheper  I confirm that the issue is coming from the `find_vpath.py` script that compiles all the files with either gcc or g++ without distinction. I'm not really sure why this stuff is used for, so it would be nice to find a solution. I will comment the faulty lines in a few days if nobody have a better idea to handle this.

kevindehecker:
So, removing CFLAGS += -std=gnu99 from makefile.nps is not an option then?

kevindehecker:
Line 172 of what file?

kevindehecker:
Ah, in makefile.ac... I see it now. Wait I'll fix that

PR #2128, opened 2017-10-20
Fix incorrect in-flight acceleration measurements in Gazebo

tomvand:
This PR fixes a bug in the accelerations measured by gazebo.

Gazebo's GetWorldLinearAccel() did not seem to take the ardrone's velocity_decay (drag) into account, instead it measured the drone's acceleration *without drag* which resulted in incorrect measurements
in-flight (the drone's acceleration in world frame did not go to zero when the drone reached its steady-state velocity).

Fixed by manually deriving the drone's world velocity over time, as in [Gazebo's own IMU implementation](https://bitbucket.org/osrf/gazebo/src/e26144434b932b4b6a760ddaa19cfcf9f1734748/gazebo/sensors/ImuSensor.cc?at=default&fileviewer=file-view-default#ImuSensor.cc-370).

Created an issue for Gazebo [here](https://bitbucket.org/osrf/gazebo/issues/2363/linear-velocity_decay-not-included-in-link).

### DIFF
--- a/sw/simulator/nps/nps_fdm_gazebo.cpp
+++ b/sw/simulator/nps/nps_fdm_gazebo.cpp
@@ -316,10 +316,12 @@ static void init_gazebo(void)
  */
 static void gazebo_read(void)
 {
+  static gazebo::math::Vector3 vel_prev;
+  static double time_prev;
+
   gazebo::physics::WorldPtr world = model->GetWorld();
   gazebo::math::Pose pose = model->GetWorldPose(); // In LOCAL xyz frame
   gazebo::math::Vector3 vel = model->GetWorldLinearVel();
-  gazebo::math::Vector3 accel = model->GetWorldLinearAccel();
   gazebo::math::Vector3 ang_vel = model->GetWorldAngularVel();
   gazebo::common::SphericalCoordinatesPtr sphere =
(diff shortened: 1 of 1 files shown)


sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -329,6 +331,15 @@ static void gazebo_read(void)
   /* Fill FDM struct */
   fdm.time = world->GetSimTime().Double();
 
+  // Find world acceleration by differentiating velocity
+  // model->GetWorldLinearAccel() does not seem to take the velocity_decay into account!
+  // Derivation of the velocity also follows the IMU implementation of Gazebo itself:
+  // https://bitbucket.org/osrf/gazebo/src/e26144434b932b4b6a760ddaa19cfcf9f1734748/gazebo/sensors/ImuSensor.cc?at=default&fileviewer=file-view-default#ImuSensor.cc-370
+  float dt = fdm.time - time_prev;
```
kirkscheper: Shouldn't this be a double?  fdm.time and time_prev are both doubles. gazebo::math::Vector3 also contain doubles.

tomvand: Yes, fixed

PR #2132, opened 2017-10-23
Drag-based velocity estimator

tomvand:
This PR adds a new module 'dragspeed' that uses the accelerometer to estimate the drag acting on the quadrotor, which is then transformed back into its velocity. This allows the velocity to be estimated using the IMU *without integration*. For usage instructions see sw/airborne/modules/dragspeed/readme.pdf.

**This PR requires the DRAGSPEED message in pprzlink, see [the corresponding PR](https://github.com/paparazzi/pprzlink/pull/62).**

Example result:
total_filter_0_8


sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	float vx = -(ACCEL_FLOAT_OF_BFP(accel->x) - dragspeed.zero.x)
+			/ dragspeed.coeff.x;
+	float vy = -(ACCEL_FLOAT_OF_BFP(accel->y) - dragspeed.zero.y)
+			/ dragspeed.coeff.y;
+	// Simple low-pass filter
+	dragspeed.vel.x += (1 - dragspeed.filter) * (vx - dragspeed.vel.x);
+	dragspeed.vel.y += (1 - dragspeed.filter) * (vy - dragspeed.vel.y);
+	// Send as ABI VELOCITY_ESTIMATE message
+	// Note: set VEL_DRAGSPEED_ID to ABI_DISABLE to disable
+#if VEL_DRAGSPEED_ID
```
kirkscheper: I think the ABI_DISABLE is for the receiving side of the msg. If the reciever has abi_disable then the message isn't sent to the receiving callback (like what you did with DRAGSPEED_ACCEL_ID). The sender id should be fixed beforehand in abi_sender_ids.h

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	dragspeed.filter = DRAGSPEED_FILTER;
+	// Register callbacks
+	register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_DRAGSPEED,
+			send_dragspeed);
+	AbiBindMsgIMU_ACCEL_INT32(DRAGSPEED_ACCEL_ID, &accel_ev, accel_cb);
+}
+
+bool dragspeed_calibrate_coeff(void) {
+	dragspeed.do_calibrate_coeff = 1;
+	return FALSE;
```
kirkscheper: Run fix_code_style. You can also add the pprz code style to your IDE so this doesn't happen in the future. There are project guidelines for eclipse, sublime and qt in the head folder (or the tools sub-folder). If you use a different IDE you can refer to the code style here http://docs.paparazziuav.org/latest/stylec.html

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	calibrate_zero(accel);
+}
+
+/**
+ * Calibrate drag coefficient by comparing accelerometer measurements to INS
+ * velocities.
+ * Should be performed with VEL_DRAGSPEED_ID set to ABI_DISABLE, but for safety
+ * the ABI messages are also disabled automatically when calibration is active.
+ *
+ * This routine assumes that the accelerometers have been zeroed beforehand.
```
kirkscheper: Why assume, if this important for the operation of the calibration routine you could add a flag to the dragspeed struct to include a flag that indicates if the accel has been zero'ed?

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	if (!dragspeed.do_calibrate_coeff) {
+		return;
+	}
+
+	// Calculate INS velocity in body frame
+	struct FloatEulers *att = stateGetNedToBodyEulers_f();
+	struct EnuCoor_f *vel_ins = stateGetSpeedEnu_f();
+	struct FloatVect2 vel_ins_body = {
+			cos(att->psi) * vel_ins->y + sin(att->psi) * vel_ins->x,
+			-sin(att->psi) * vel_ins->y + cos(att->psi) * vel_ins->x
```
kirkscheper: Can you use the float implementations of sin and cos (sinf and cosf).

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	};
+	// Calibrate coefficient when velocity is sufficiently high
+	if (abs(vel_ins_body.x) > 0.5) {
```
kirkscheper: abs is for int, I guess you want fabsf.

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+	}
+	do_calibrate_prev = dragspeed.do_calibrate_zero;
+	// Return when calibration is not active
+	if (!dragspeed.do_calibrate_zero) {
+		return;
+	}
+
+	// Average accelerometer readings when velocity is sufficiently low
+	struct EnuCoor_f *vel_ins = stateGetSpeedEnu_f();
+	float ins_speed = sqrt(vel_ins->x * vel_ins->x + vel_ins->y * vel_ins->y);
```
kirkscheper: sqrtf

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+ * The zero-velocity readings can change between flights, e.g. because of small
+ * differences in battery or outer hull positions.
+ */
+static void calibrate_zero(struct Int32Vect3 *accel) {
+	// Reset when new calibration is started
+	static int do_calibrate_prev = 0;
+	static struct FloatVect2 zero;
+	static int num_samples = 0;
+	if (dragspeed.do_calibrate_zero && !do_calibrate_prev) {
+		zero.x = 0;
```
kirkscheper: 0.f

Small thing but it helps to reinforce that this variable is a float, as opposed to the num_samples.

sw/airborne/modules/dragspeed/dragspeed.h
```diff
@@ -0,0 +1,53 @@
...
+	// Estimated velocity
+	struct FloatVect2 vel;
+	// Low-pass filter
+	float filter;
+	// Drag coefficient calibration
+	struct FloatVect2 coeff;
+	int do_calibrate_coeff;
+	// Zero calibration
+	struct FloatVect2 zero;
+	int do_calibrate_zero;
```
kirkscheper: The way you use this, I would expect that a boolean is better than an int here and for do_calibrate_coeff.

Also, why call them do_... Doesn't really add much. Just calibrate_zero is already enough IMHO

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+			/ dragspeed.coeff.y;
+	// Simple low-pass filter
+	dragspeed.vel.x += (1 - dragspeed.filter) * (vx - dragspeed.vel.x);
+	dragspeed.vel.y += (1 - dragspeed.filter) * (vy - dragspeed.vel.y);
+	// Send as ABI VELOCITY_ESTIMATE message
+	// Note: set VEL_DRAGSPEED_ID to ABI_DISABLE to disable
+#if VEL_DRAGSPEED_ID
+	if (!dragspeed.do_calibrate_coeff && !dragspeed.do_calibrate_zero) {
+		AbiSendMsgVELOCITY_ESTIMATE(VEL_DRAGSPEED_ID, stamp,
+				dragspeed.vel.x, dragspeed.vel.y, 0, DRAGSPEED_R);
```
kirkscheper: This will also send the vertical velocity estimate through to the ins. Maybe we should add a flag to define which axis' to update. If you need help with this pass by my office.

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+#define DRAGSPEED_COEFF_X 1.0 /// Drag coefficient (mu/m) of the linear drag model along x axis, where m*a = v*mu
+#endif
+
+#ifndef DRAGSPEED_COEFF_Y
+#define DRAGSPEED_COEFF_Y 1.0 /// Drag coefficient (mu/m) of the linear drag model along y axis, where m*a = v*mu
+#endif
+
+#ifndef DRAGSPEED_ZERO_X
+#define DRAGSPEED_ZERO_X 0.0 /// Accelerometer reading (x axis) when stationary [m/s^2]
+#endif
```
kirkscheper: DRAGSPEED_ZERO_X and Y are not used atm.

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,228 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/dragspeed/dragspeed.c"
+ * @author Tom van Dijk
+ * This module estimates the velocity of rotorcraft by measuring the drag force using the accelerometer.
```
kirkscheper: I know you have the accompanying pdf but can you add a little more description here, like what are the assumptions you make that you can accurately estimate the drag (no wind, constant vehicle geometry, ect)?

sw/airborne/modules/dragspeed/dragspeed.c
tomvand: As a workaround until the axis flag is added, I tried setting the z velocity estimate to stateGetSpeedNed_f()->z as this should be closer to the drone's true vertical velocity than zero. Tested this with the INS set to 'extended' and USE_GPS_ALT defined, but then it resulted in large altitude deviations. These did not seem to occur when the z velocity was set to zero. (Nor with -stateGetSpeedNed_f()->z, but these velocity estimates are expressed in the front-right-down body frame, right? I could not find any documentation on the axis system used in this message). I also found that existing code (modules/optical_flow/px4flow.c) also sets the vertical velocity estimate to zero. What would you advise me to do here?

tomvand:
Fixed all of the issues except for the vertical velocity estimate, see #2144. Until this issue is fixed, this module will result in incorrect vertical behaviour of the drone.

gautierhattenberger:
Please fix conflicts on `ardrone.world` file (probably needs a rebase or merge with master) and set the correct PPRZLINK version (you message is now merged)

tomvand:
Merged latest master which removes ardrone.world, updated pprzlink to its latest master.

conf/modules/dragspeed.xml
```diff
@@ -0,0 +1,50 @@
...
+      This module makes the following assumptions:
+      - The drag acting on the quadrotor grows linearly with velocity.
+      - The size and weight of the drone are constant (changes in size and weight
+        require re-calibration).
+      - Pitch and roll angles are small (less than ~20 deg).
+      - No wind (when used as ground speed).
+    </description>
+    <define name="DRAGSPEED_SEND_ABI_MESSAGE" description="Send VELOCITY_ESTIMATE ABI messages" value="TRUE|FALSE (default: TRUE)"/>
+    <define name="DRAGSPEED_ACCEL_ID" description="Accelerometer ABI ID to subscribe to (default: ABI_BROADCAST)"/>
+    <define name="DRAGSPEED_COEFF" description="Drag coefficient (mu/m), where Drag coefficient (mu/m) of the linear drag model, where m*a = v*mu"/>
```
kirkscheper: I guess this one should be actually DRAGSPEED_COEFF_X and DRAGSPEED_COEFF_Y

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,262 @@
...
+    struct Int32Vect3 *accel)
+{
+  // Estimate current velocity
+  float vx = -(ACCEL_FLOAT_OF_BFP(accel->x) - dragspeed.zero.x)
+      / dragspeed.coeff.x;
+  float vy = -(ACCEL_FLOAT_OF_BFP(accel->y) - dragspeed.zero.y)
+      / dragspeed.coeff.y;
+  // Simple low-pass filter
+  dragspeed.vel.x += (1 - dragspeed.filter) * (vx - dragspeed.vel.x);
+  dragspeed.vel.y += (1 - dragspeed.filter) * (vy - dragspeed.vel.y);
```
kirkscheper: Not something you need to change but something you should keep in mind it that implementing the low pass filter like this assumes that you have a fixed update of the acceleros. This assumption is probably invariant while flying but is not invariant across vehicles e.g. bebop has an update rate of 500Hz but ardrone has 200Hz so this filter will be different on those vehicle for the same value of dragspeed.filter.

Like I said this is not something that you have to fix but should keep in mind.

sw/airborne/modules/dragspeed/dragspeed.c
```diff
@@ -0,0 +1,262 @@
...
+    num_samples_y = 0;
+  }
+  calibrate_prev = dragspeed.calibrate_coeff;
+  // Return when calibration is not active
+  if (!dragspeed.calibrate_coeff) {
+    return;
+  }
+  // Also return if zero calibration has not been performed yet
+  if (!dragspeed.zero_calibrated) {
+    fprintf(stderr,
```
kirkscheper: Will this compile if not using a linux board?

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -278,6 +278,13 @@
 #endif
 
 /*
+ * IDs of VELOCITY estimates (message 12)
+ */
+#ifndef VEL_DRAGSPEED_ID
+#define VEL_DRAGSPEED_ID 1
```
kirkscheper: I was quite surprised about this but then I realized that the other velocity ids where in the wrong place. 

Note to self. We need to update this for the opticflow_module, stereocamera and the px4flow which are all wrong.

kirkscheper:
@tomvand if you update the description and the fprintf I think this can be merged.

sw/airborne/modules/dragspeed/dragspeed.c
tomvand: Nope, added it to the hexa_lisaL airframe which resulted in compilation errors. Is there a define that I can use to recognize all linux boards or all boards that support printf? I think the warning will be helpful when diagnosing problems with calibration. If not, I will remove the printf.

conf/modules/dragspeed.xml
tomvand: Fixed

kirkscheper:
I just noticed the readme.pdf, its really handy to have the documentation with the code but it would be a little more handy if it was an md file. This way it will be displayed on the github page You can post the images in this thread and refer to the images in the readme.md. For eg. 

`alt text`

results in

alt text

This will have the added benefit of adding a file history as it will be simple text.

I can image that one downside is that on the user code base, the images won't be displayed outside of a browser.

@dewagter I know you have opinions on the documentation, what do you think about this.

tomvand:
Is it also possible to add images to the description in the module XML? It would have the same benefits and problems as the .md file (i.e. simple text but no images when opened outside the browser), but would at least follow the standard set by the other modules and show up correctly at http://docs.paparazziuav.org/latest/ .

kirkscheper:
Good point.

Do you know if it is possible to include another file from within the module xml? I can imagine that it would be nice to have the documentation here with the code but included with the xml.

tomvand:
Looked it up, Doxygen appears to have commands for this: [includedoc](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdincludedoc) and [image](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdimage). I assume that the xml parser just copies the description text, so I would expect the commands to work. Never tried it, though...

dewagter:
Do you know sphinx? It allows inline docs to be parsed, combined with
doxigen if needed and pushed to github pages.

For an example see:

https://github.com/tudelft/tudelft_vision

tomvand:
Just tried the image command in the XML description, seems to work fine:
image
with the image placed in `doc/images/airborne/modules/dragspeed/`.

@kirkscheper Haven't tried the includedoc command yet. However, when including the text it would need to be placed in the module folder *or* the module xml, not both at once (that would result in duplicate text that would both need to be maintained). Although I think that including the documentation inside the module folder would be neat, currently all other modules have their documentation in their XML file so this is probably where users will look first. For that reason, I think it would be better to move the readme inside the XML file, instead of the other way round. Do you agree?

@dewagter Is that already included in paparazzi's documentation system? Otherwise it might be a bit overkill as I'm only trying to document my module.

tomvand:
Moved all documentation to the module XML and removed the readme.pdf since it is now redundant. Also fixed the compilation error on non-linux boards by surrounding the printf statement with `#ifdef __linux__`.

sw/airborne/modules/dragspeed/dragspeed.c
gautierhattenberger: Unfortunately we don't have a standard for that. The proper solution is probably to provide a debug API for Paparazzi with debug levels and proper output depending of the hardware. But that is for an other PR.

PR #2135, opened 2017-10-24
[nps] load aircraft into gazebo at run time + actuator handling

kirkscheper:
This enables gazebo to grab the quadrotor configuration from the airframe file and load up the correct model. This also ensures that the same yaw configuration is used for the simulator as is defined in the motor_mixing to reduce possible user error. Now the only thing is that the order of the actuator names still has to be correct.

Secretly, this also refractors how GPS is ignored in nps. As well as replaces course with heading when GPS_DATALINK is used so that nps correctly copies reality when using natnet/optitrack.

@knmcguire and @tomvand can you take a look here and let me know if there are any issues.

### DIFF
--- a/conf/airframes/KS/ks_bebop2_stereo.xml
+++ b/conf/airframes/KS/ks_bebop2_stereo.xml
@@ -3,7 +3,14 @@
 <airframe name="bebop2_opticflow">
 
   <firmware name="rotorcraft">
-    <target name="ap" board="bebop2"/>
+    <target name="ap" board="bebop2">
+      <module name="stabilization" type="indi_simple"/>
+    </target>
+    <target name="nps" board="pc">
+      <module name="fdm" type="gazebo"/>
+      <module name="stabilization" type="int_quat"/>
+      <configure name="STEREO_PATH" value="/home/kirk/mavlab/stereoboard"/>
+    </target>
 
     <module name="telemetry" type="transparent_udp"/>
     <module name="radio_control" type="datalink"/>
@@ -11,19 +18,15 @@
     <module name="actuators" type="bebop"/>
     <module name="imu" type="bebop"/>
     <module name="gps" type="datalink"/>
-    <module name="stabilization" type="indi_simple"/>
     <module name="ahrs" type="int_cmpl_quat">
       <configure name="USE_MAGNETOMETER" value="FALSE"/>
       <define name="AHRS_USE_GPS_HEADING" value="TRUE"/>
       <define name="AHRS_HEADING_UPDATE_GPS_MIN_SPEED" value="0"/>
     </module>
-    <module name="ins" type="extended"/>
+    <module name="ins" type="hff_extended"/>
 
     <module name="stereocam">
       <configure name="STEREO_UART" value="UART2"/>
-      <define name="STEREO_BODY_TO_STEREO_PHI" value="90."/>
-      <define name="STEREO_BODY_TO_STEREO_THETA" value="0."/>
-      <define name="STEREO_BODY_TO_STEREO_PSI" value="90."/>
     </module>
     
   </firmware>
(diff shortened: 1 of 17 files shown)


conf/airframes/KS/ks_bebop2_stereo.xml
```diff
@@ -200,6 +203,14 @@
     <define name="ACTUATOR_NAMES" value="nw_motor, ne_motor, se_motor, sw_motor" type="string[]"/>
     <define name="JSBSIM_MODEL" value="simple_x_quad_ccw" type="string"/>
     <define name="SENSORS_PARAMS" value="nps_sensors_params_default.h" type="string"/>
+
+    <define name="GAZEBO_AC_NAME" value="simple_x_quad" type="string"/>    
+    <define name="ACTUATOR_THRUSTS" value="1.55, 1.55, 1.55, 1.55" type="double[]"/>
+    <define name="ACTUATOR_TORQUES" value="0.155, 0.155, 0.155, 0.155" type="double[]"/>
+    
+    <define name="GAZEBO_WORLD" value="ardrone.world" type="string"/>
```
tomvand: Should this be "empty.world"?

tomvand:
Looks good. Found no obvious errors in the changes and the simulation appears to work fine. Spawning the aircraft model specified in the airframe xml is definitely a nice addition!

I could not confirm the working of the datalink GPS heading. As far as I'm aware the AHRS is not fully functional in gazebo atm. (I tried a short test flight without bypassed AHRS but this resulted in uncontrolled behavior both with and without the datalink gps); when the AHRS is bypassed, the GPS heading is overwritten by the ground truth value. I have not yet checked if this would be a quick fix to get a reasonable heading estimate during the accelerometer callback.

Note that this update will require existing gazebo worlds to be modified, the existing drone needs to be removed just as you did for the empty world.

One minor suggestion: could you rename the simple_quad model to simple_plus_quad? Makes the name a bit more descriptive and highlights the difference with the simple_x_quad. (Or is this an existing naming convention from jsbsim?)

kirkscheper:
"I could not confirm the working of the datalink GPS heading. As far as I'm aware the AHRS is not fully functional in gazebo atm. (I tried a short test flight without bypassed AHRS but this resulted in uncontrolled behavior both with and without the datalink gps); when the AHRS is bypassed, the GPS heading is overwritten by the ground truth value. I have not yet checked if this would be a quick fix to get a reasonable heading estimate during the accelerometer callback."

This is quite strange, when I build and run the airframe in my conf it works just fine, the vehicle heading is correct and the ahrs works fine (I don't bypass AHRS or INS).

"One minor suggestion: could you rename the simple_quad model to simple_plus_quad? Makes the name a bit more descriptive and highlights the difference with the simple_x_quad. (Or is this an existing naming convention from jsbsim?)"

I thought the same here, I actually thought of  simple_quad_x and  simple_quad_plus (or  simple_quad_+ if that is allowed) but in the end decided on the same naming as with jsbsim. We can of course also rename those. The question is how strongly do you feel about this?
screenshot from 2017-10-25 18-32-51

tomvand:
Have you tried changing the nav_heading in the settings? That caused the drone to become unstable around the yaw axis when I tried it.

I have no strong feelings about the simple quad naming. If it's too much work to rename the jsbsim one (not sure how often its referenced in all the airframe files) it's fine to leave this as it is.

kirkscheper:
Just tried the nav_heading step, no issues!

tomvand:
Ok, then there might be something wrong on my end; I can recheck it later this week. If it works fine for you then I have no further objections against merging.

-------- Oorspronkelijk bericht --------
Van: Kirk Scheper
Datum:25-10-2017 18:56 (GMT+01:00)
Aan: paparazzi/paparazzi
Cc: Tom van Dijk ,Mention
Onderwerp: Re: [paparazzi/paparazzi] [nps] load aircraft into gazebo at run time + actuator handling (#2135)


Just tried the nav_heading step, no issues!

PR #2137, opened 2017-10-24
INDI settings and fan demo

EwoudSmeur:
I added two settings for guidance_indi, and noticed that all attitude indi settings were 'persistent'. I heard from Kirk and Matej that they had trouble with persistent stuff in the past, so I removed that.

Also added an airframe and a flight plan for a 'fan demo', where a bebop flies in front of a fan to show off the INDI guidance. Test flown and works

### DIFF
--- a/conf/airframes/TUDELFT/tudelft_fan_demo.xml
+++ b/conf/airframes/TUDELFT/tudelft_fan_demo.xml
@@ -1,0 +1,192 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="bebop_indi">
+
+  <firmware name="rotorcraft">
+    <target name="ap" board="bebop"/>
+
+    <target name="nps" board="pc">
+      <module name="fdm" type="jsbsim"/>
+      <module name="udp"/>
+    </target>
+
+    <!--define name="USE_SONAR" value="TRUE"/-->
+
+	<!-- Subsystem section -->
+    <module name="telemetry" type="transparent_udp"/>
+    <module name="radio_control" type="datalink"/>
+    <module name="actuators" type="bebop"/>
+    <module name="imu" type="bebop"/>
+    <module name="gps" type="datalink"/>
+    <module name="stabilization" type="indi">
+      <define name="INDI_RPM_FEEDBACK" value="TRUE"/>
+    </module>
+    <module name="stabilization" type="rate_indi"/>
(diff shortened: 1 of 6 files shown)


conf/airframes/TUDELFT/tudelft_fan_demo.xml
```diff
@@ -0,0 +1,192 @@
...
+    <module name="guidance" type="indi">
+      <define name="GUIDANCE_INDI_SPECIFIC_FORCE_GAIN" value="-500.0"/>
+      <define name="GUIDANCE_INDI_THRUST_DYNAMICS" value="0.1"/>
+      <define name="GUIDANCE_INDI_RC_DEBUG" value="FALSE"/>
+      <define name="GUIDANCE_INDI_POS_GAIN" value="0.7"/>
+      <define name="GUIDANCE_INDI_SPEED_GAIN" value="1.5"/>
+    </module>
+  </firmware>
+
+  <modules main_freq="512">
```
kirkscheper: We don't use a separate modules part any more, can you move these to the firmware section?

conf/airframes/TUDELFT/tudelft_fan_demo.xml
```diff
@@ -0,0 +1,192 @@
...
+      <configure name="USE_MAGNETOMETER" value="FALSE"/>
+      <define name="AHRS_USE_GPS_HEADING" value="TRUE"/>
+    </module>
+    <module name="ins" type="extended"/>
+    <module name="guidance" type="indi">
+      <define name="GUIDANCE_INDI_SPECIFIC_FORCE_GAIN" value="-500.0"/>
+      <define name="GUIDANCE_INDI_THRUST_DYNAMICS" value="0.1"/>
+      <define name="GUIDANCE_INDI_RC_DEBUG" value="FALSE"/>
+      <define name="GUIDANCE_INDI_POS_GAIN" value="0.7"/>
+      <define name="GUIDANCE_INDI_SPEED_GAIN" value="1.5"/>
```
kirkscheper: Move these to a GUIDANCE_INDI section? 

The difference is that variables defined here are included in the make flags and should be reserved for defines that are used to enable/disable parts of the code from being compiled. If these are just settings, it is better to place them in a <section>, these are included in generated/airframe.h. This distinction helps to reinforce how these parameters will be used in the code.

EwoudSmeur:
Thanks @kirkscheper

PR #2142, opened 2017-10-30
math functions and high pass filter

EwoudSmeur:
The high pass filter uses doubles to avoid numerical errors.

This is some of the code I use for the Cyclone, planning to merge all of it piece by piece like this.

The 2D matrix stuff I actually don't use for the Cyclone at this point, but I thought it could be useful in the future.

### DIFF
--- a/sw/airborne/math/pprz_algebra_float.c
+++ b/sw/airborne/math/pprz_algebra_float.c
@@ -707,11 +707,8 @@ void float_mat2_mult(struct FloatVect2 *vect_out, float mat[4], struct FloatVect
  * 4x4 Matrix inverse.
  * obtained from: http://rodolphe-vaillant.fr/?e=7
  */
-float float_mat_minor_4d(float m[16], int r0, int r1, int r2, int c0, int c1, int c2);
-void float_mat_adjoint_4d(float adjOut[16], float m[16]);
-float float_mat_det_4d(float m[16]);
 
-float float_mat_minor_4d(float m[16], int r0, int r1, int r2, int c0, int c1, int c2)
+static float float_mat_minor_4d(float m[16], int r0, int r1, int r2, int c0, int c1, int c2)
 {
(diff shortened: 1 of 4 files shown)


sw/airborne/math/pprz_algebra_float.c
```diff
@@ -651,10 +755,11 @@ float float_mat_det_4d(float m[16])
...
 void float_mat_inv_4d(float invOut[16], float mat_in[16])
 {
-    float_mat_adjoint_4d(mat_in, invOut);
+  float_mat_adjoint_4d(mat_in, invOut);
 
-    float inv_det = 1.0f / float_mat_det_4d(mat_in);
-    int i;
-    for(i = 0; i < 16; ++i)
-        invOut[i] = invOut[i] * inv_det;
+  float inv_det = 1.0f / float_mat_det_4d(mat_in);
```
kirkscheper: check if matrix is invertible here before the divide?

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -609,38 +713,38 @@ float float_mat_det_4d(float m[16]);
```
kirkscheper: The lines above this shouldn't be needed, can you remove them plz :) less more is more better...

Also, if you are super generous, can you add function descriptors for some more of these functions (very optional)?

EwoudSmeur:
Thanks Kirk. I copied the check for the 2D case, and if I remember correctly Gautier advised me to use 1e-4 instead of something like FLOAT_EPSILON. I am not sure why, somehow 1e-4 seems a bit arbitrary to me.

kirkscheper:
indeed 1e-4 is arbitrary, I usually use 1e-5 :p 

You can define something in std.h

sw/airborne/math/pprz_algebra_float.c
EwoudSmeur: Without this I get a warning for a missing prototype...

gautierhattenberger:
looks good

PR #2145, opened 2017-10-30
Cyfoam, ctrl effectiveness scheduling, JSBsim

EwoudSmeur:
- Added Cyfoam airframe file
- Added control effectiveness scheduling function
- More advanced scheduling function should be added in the future
- Added simple JSBsim files for Cyclone for a crude simulation

### DIFF
--- a/conf/airframes/ENAC/cyfoam.xml
+++ b/conf/airframes/ENAC/cyfoam.xml
@@ -112,9 +112,6 @@
       <define name="INS_FINV_MAG_ID" value="MAG_HMC58XX_SENDER_ID"/>
     </module>
 
-  </firmware>
-
-  <modules main_freq="512">
     <!--<module name="gps" type="ubx_ucenter"/>-->
     <module name="tlsf"/>
     <module name="pprzlog">
@@ -140,7 +137,8 @@
       <define name="USE_PWM_INPUT1" value="PWM_PULSE_TYPE_ACTIVE_LOW"/>
       <define name="USE_PWM_INPUT2" value="PWM_PULSE_TYPE_ACTIVE_LOW"/>
     </module>
-  </modules>
+
+  </firmware>
 
   <servos driver="Pwm">
     <servo name="ELEVON_LEFT"  no="0" min="1000" neutral="1500" max="2000"/>
@@ -179,13 +177,6 @@
     <define name="BODY_TO_IMU_PHI"   value="180." unit="deg"/>
     <define name="BODY_TO_IMU_THETA" value="90." unit="deg"/>
     <define name="BODY_TO_IMU_PSI"   value="180." unit="deg"/>
-  </section>
-
-  <section name="AUTOPILOT">
-    <define name="MODE_MANUAL"  value="AP_MODE_ATTITUDE_DIRECT"/>
-    <define name="MODE_AUTO1" value="AP_MODE_FORWARD"/>
-    <define name="MODE_AUTO2"  value="AP_MODE_NAV"/>
-    <define name="MODE_STARTUP"  value="AP_MODE_NAV"/>
   </section>
 
   <section name="BAT">
@@ -302,6 +293,13 @@
     <define name="JSBSIM_MODEL" value="cyclone" type="string"/>
     <define name="SENSORS_PARAMS" value="nps_sensors_params_default.h" type="string"/>
     <define name="NO_MOTOR_MIXING" value="TRUE"/>
+  </section>
+
+  <section name="AUTOPILOT">
+    <define name="MODE_MANUAL"  value="AP_MODE_ATTITUDE_DIRECT"/>
+    <define name="MODE_AUTO1" value="AP_MODE_FORWARD"/>
+    <define name="MODE_AUTO2"  value="AP_MODE_NAV"/>
+    <define name="MODE_STARTUP"  value="AP_MODE_NAV"/>
   </section>
 
 </airframe>
--- a/conf/airframes/ENAC/cyfoam.xml
+++ b/conf/airframes/ENAC/cyfoam.xml
@@ -1,0 +1,318 @@
+<!--Foam Cyclone Airframe
+  Chimera AP
+	Xbee API 
+	Ublox M8N
+	SBUS Futaba -->
+
+<airframe name="foam cylone">
+
+  <servos driver="Pwm">
+    <servo name="ELEVON_LEFT"  no="0" min="1000" neutral="1500" max="2000"/>
+    <servo name="ELEVON_RIGHT" no="1" min="1000" neutral="1500" max="2000"/>
+    <servo name="RM" no="2" min="1000" neutral="1100" max="2000"/>
+    <servo name="LM" no="3" min="1000" neutral="1100" max="2000"/>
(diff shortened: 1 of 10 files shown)


conf/modules/tlsf.xml
```diff
@@ -17,6 +17,6 @@
     <define name="MAX_FLI" value="$(TLSF_MAX_FLI)"/>
     <file_arch name="tlsf_malloc_arch.c"/>
     <include name="$(PAPARAZZI_SRC)/sw/ext/tlsf"/>
-    <file name="tlsf.c" dir="$(PAPARAZZI_SRC)/sw/ext/tlsf"/>
+    <file name="tlsf.c" dir="../../tlsf"/>
```
kirkscheper: This doesn't look quite right. As far as I understand it this does "work" but maybe keep the old line commented here for future people (just in case this causes issues in the future).

conf/airframes/ENAC/cyfoam.xml
```diff
@@ -0,0 +1,318 @@
...
+      <!--<define name="AHRS_ICQ_MAG_ID" value="MAG_HMC58XX_SENDER_ID"/>-->
+      <!--[><define name="AHRS_GRAVITY_UPDATE_COORDINATED_TURN" value="TRUE"/><]-->
+      <!--[><define name="AHRS_GPS_SPEED_IN_NEGATIVE_Z_DIRECTION" value="TRUE"/><]-->
+    <!--</module>-->
+
+    <!--<module name="ins"/>-->
+    <module name="ins" type="float_invariant">
+      <define name="INS_PROPAGATE_FREQUENCY" value="500"/>
+      <define name="INS_FINV_MAG_ID" value="MAG_HMC58XX_SENDER_ID"/>
+    </module>
```
kirkscheper: This airframe looks reversed...

kirkscheper: Not gona fix this :(

gautierhattenberger: what is the issue here exactly ?

conf/modules/tlsf.xml
gautierhattenberger: this was working for sure, can it be related to the `vpath` tool (again) ?

conf/airframes/ENAC/cyfoam.xml
kirkscheper: The airframe order is reversed to the conventional order (firmware, control, sections, modes).

conf/modules/tlsf.xml
kirkscheper: it works, it just looks really weird, I would have guessed it would be something like ../ext/tlsf but that didn't work. I am fine with it like this though. Indeed it needs to be changed due to the vpath tool.

EwoudSmeur: I have tried again with the original line, and it does continue compiling, but it gives the following error in the beginning:

../../sw/tools/find_vpaths.py arm-none-eabi-gcc /home/ewoud/paparazzi_check/var/aircrafts/CYFOAM/ap_tmp.list /home/ewoud/paparazzi_check
arm-none-eabi-gcc: error: ERROR:: No such file or directory
arm-none-eabi-gcc: error: found: No such file or directory
arm-none-eabi-gcc: error: more: No such file or directory
arm-none-eabi-gcc: error: than: No such file or directory
arm-none-eabi-gcc: error: once:: No such file or directory
arm-none-eabi-gcc: error: /home/ewoud/paparazzi_check/sw/ext/tlsf/tlsf.c/home/ewoud/paparazzi_check/sw/airborne/arch/chibios/mcu_arch.c: Not a directory

This if statement triggers 34 times for the first srcs[i]:
(line 58)             if os.path.isfile(path.join(vpath, f)) :

I don't really understand what the vpath file does...

EwoudSmeur: Ah this looks like #2182 !
Running it from the command line helped to look at the prints and errors:
python3 sw/tools/find_vpaths.py arm-none-eabi-gcc /home/ewoud/paparazzi_check/var/aircrafts/CYFOAM/ap_tmp.list

podhrmic: `find_vpaths` again. Does it get better with the latest fix in master? (just rebase master on top this PR)

conf/airframes/ENAC/cyfoam.xml
```diff
@@ -0,0 +1,307 @@
...
+
+  </firmware>
+
+  <modules main_freq="512">
+    <!--<module name="gps" type="ubx_ucenter"/>-->
```
kirkscheper: We are phasing out separate modules sections, can you move this to the firmware above (like you already did with your other modules?) additionally, I think the order the modules appear is important for how they are run so consider that, usually, the "subsystems" go first then the more mission specific modules.

sw/airborne/modules/ctrl/ctrl_effectiveness_scheduling.c
```diff
@@ -0,0 +1,79 @@
...
+
+float g1g2_forward[INDI_OUTPUTS][INDI_NUM_ACT] = {FWD_G1_ROLL,
+                                                  FWD_G1_PITCH, FWD_G1_YAW, FWD_G1_THRUST
+                                                 };
+
+float g1g2_hover[INDI_OUTPUTS][INDI_NUM_ACT] = {STABILIZATION_INDI_G1_ROLL,
+                                                STABILIZATION_INDI_G1_PITCH, STABILIZATION_INDI_G1_YAW, STABILIZATION_INDI_G1_THRUST
+                                               };
+
+float g2_both[INDI_NUM_ACT] = STABILIZATION_INDI_G2; //scaled by INDI_G_SCALING
```
kirkscheper: static?

conf/airframes/ENAC/cyfoam.xml
```diff
@@ -0,0 +1,307 @@
...
+    <define name="MAG_X_SENS" value="3.68245686199" integer="16"/>
+    <define name="MAG_Y_SENS" value="3.81802043484" integer="16"/>
+    <define name="MAG_Z_SENS" value="4.01084591146" integer="16"/>
+
+    <define name="BODY_TO_IMU_PHI"   value="180." unit="deg"/>
+    <define name="BODY_TO_IMU_THETA" value="90." unit="deg"/>
+    <define name="BODY_TO_IMU_PSI"   value="180." unit="deg"/>
+  </section>
+
+  <section name="AUTOPILOT">
```
kirkscheper: This is usually at the bottom of the airframe. I think we should discuss what the preferred order is for the future... I think it is beneficial to have a standard but that standard hasn't been clearly defined. @gautierhattenberger is this defined anywhere?

gautierhattenberger: There is no definition of the order of the main nodes in airframe files at the moment. Even the DTD file is not reflecting anything in particular. Also the parser don't really care. We could edit somewhere a "recommended practice" guide. My preferred order is firmware, then "special nodes" (servos, commands, ...) and finally 'sections'.

conf/modules/tlsf.xml
EwoudSmeur: It does get better! Thanks @podhrmic !

gautierhattenberger:
@EwoudSmeur @kirkscheper it looks fine now

PR #2146, opened 2017-10-30
Pullrequest bebop2 tools wifi tools

coppolam:
Added a set of scripts (in sw/tools/parrot/bebop) to allow Bebop 2 drones to connect to a router instead of acting as access points.

### DIFF
--- a/sw/tools/parrot/bebop/wifi_tools/scripts/pprz.conf
+++ b/sw/tools/parrot/bebop/wifi_tools/scripts/pprz.conf
@@ -1,0 +1,8 @@
+# Config file for PPRZ init script
+START_PPRZ=0
+START_TELNET=1
+START_ADB=0
+JOIN_WIFI=1
+WIFI_SSID=huphub
+WIFI_AMODE=none
+WIFI_KEY=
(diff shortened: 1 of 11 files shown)


dewagter:
can you rename the txt to README.md

sw/tools/parrot/bebop/wifi_tools/scripts/pprz.conf
```diff
@@ -0,0 +1,8 @@
+# Config file for PPRZ init script
+START_PPRZ=0
```
kirkscheper: Where is this file used? Why a pprz.conf and a scripts/pprz.conf?

coppolam: Going over the scripts it actually seems that both pprz.conf and scripts/pprz.conf could be superfluous, with pprz_swarmhub.conf being sufficient. I'll double check soon and make a new pull request, either deleting them or making the distinction more clear.

Sai-Jiang:
Hello!

I wonder whether this approach applies to Bebop2 and which firmware version it actually supports.

I have tried several approaches to convert Bebop2 to client but no one succeeds.

Thx very much

PR #2148, opened 2017-10-30
Rename TUDELFT to tudelft, remove tudelft_ prefix

EwoudSmeur:
I know this is a bit painful, but it is also a bit painful every day to type redundant stuff...

Typing capitals is a bit annoying, and since we already have tudelft folders it is redundant to give the files a tudelft_ prefix

Tried to find all places that could be affected and adjust naming accordingly


EwoudSmeur:
merging as it should only affect the tudelft conf

PR #2149, opened 2017-10-30
Gazebo in tudelft conf, with orange avoid example

kevindehecker:
This code is being used in the TUDelft MAVLab course

### DIFF
--- a/conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
+++ b/conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
@@ -3,7 +3,6 @@
 <airframe name="ardrone2">
 
   <firmware name="rotorcraft">
-    <target name="ap" board="ardrone2"/>
 
     <target name="nps" board="pc">
       <module name="fdm" type="gazebo"/>
@@ -22,9 +21,7 @@
     <module name="stabilization" type="int_quat"/>
     <module name="ahrs" type="int_cmpl_quat"/>
     <module name="ins" type="extended"/>
-  </firmware>
 
-  <modules main_freq="512">
     <module name="bat_voltage_ardrone2"/>
     <!-- remove the gps_ubx_ucenter module if you use the sirf gps (flight recorder) -->
     <module name="gps" type="ubx_ucenter"/>
@@ -45,7 +42,7 @@
       <define name="VIEWVIDEO_DOWNSIZE_FACTOR" value="4"/>
       <define name="VIEWVIDEO_QUALITY_FACTOR" value="60"/>
     </module>
-  </modules>
+  </firmware>
 
   <commands>
     <axis name="PITCH" failsafe_value="0"/>
--- a/conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
+++ b/conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
@@ -1,0 +1,213 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="ardrone2">
+
+  <firmware name="rotorcraft">
+    <target name="ap" board="ardrone2"/>
+
+    <target name="nps" board="pc">
+      <module name="fdm" type="gazebo"/>
+    </target>
+
+    <define name="USE_SONAR" value="TRUE"/>
+
+	<!-- Subsystem section -->
+    <module name="telemetry" type="transparent_udp"/>
+    <module name="radio_control" type="datalink"/>
+    <module name="motor_mixing"/>
+    <module name="actuators" type="ardrone2"/>
+    <module name="imu" type="ardrone2"/>
+    <!-- gps: "ublox" or change to "sirf" for usage with parrot flight recorder -->
+    <module name="gps" type="ublox"/>
+    <module name="stabilization" type="int_quat"/>
+    <module name="ahrs" type="int_cmpl_quat"/>
+    <module name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
+    <module name="bat_voltage_ardrone2"/>
+    <!-- remove the gps_ubx_ucenter module if you use the sirf gps (flight recorder) -->
+    <module name="gps" type="ubx_ucenter"/>
+    <module name="send_imu_mag_current"/>
+    <module name="air_data"/>
+    <module name="geo_mag"/>
+    <!--module name="logger_file"/-->
+    <module name="video_thread">
+    </module>
+    <module name="cv_colorfilter">
+      <define name="COLORFILTER_CAMERA" value="front_camera"/>
+    </module>
+
+    <module name="orange_avoider" />
+
+    <module name="video_rtp_stream">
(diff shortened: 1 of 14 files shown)


conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
```diff
@@ -0,0 +1,213 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="ardrone2">
```
kirkscheper: Is the only difference here that gazebo loads the cyberzoo instead of the empty environment? Then I don't think you should add an extra example for this. One example specifically for gazebo is enough.

If you want to add an ardrone that can avoid orange poles that happens to also work in simulation then you can add this as ardrone2_pole_avoider or something like that. I still don't think that this should be in the examples but in the tudelft given that it is a bit specific to the cyberzoo.

conf/airframes/TUDELFT/tudelft_conf.xml
```diff
@@ -430,6 +430,28 @@
    gui_color="green"
   />
   <aircraft
+   name="ardrone2_flightplan_guided_gazebo"
```
kirkscheper: Why add simulation specific vehicles? The idea of the targets is that one airframe can be used for both real flight and simulation.

dewagter: Good point. Only: the orangeavoider runs in a bebop and gazebo does not simulate that yet.

kirkscheper: There are now 3 types of vehicles defined. ARdrone, simple_quad and simple_x_quad. If you would like a bebop specific vehicle you can add it and define the type in the airframe using NPS_GAZEBO_AC_NAME. I think it would be sufficient however just to use the default simple_x_quad  though.

If you are referring to the resolution of the camera, the you would indeed need a new vehicle. Or wait till the end of the week when I implement a way to dynamically add sensors to the models and then add the bebop camera instead of the ardrone camera to the default drone.

One major thing that is not yet implemented that may affect the bebop integration is there are no actuator dynamics implemented to INDI doesn't work. What I have done so far is use INDI in ap and quat_int in nps. I think that this is a fine compromise for now though.

kirkscheper: https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/KS/ks_bebop2_stereo.xml

kirkscheper:
for the spaces in the world file you can use 

sed -i $'s/\t/  /g' *.sdf

This will replace all tabs with spaces (easier to debug in github images, it is also the format that is used when writing the sdf files from the sdf library).

dewagter:
will merge real and sim airframe as soon as https://github.com/paparazzi/paparazzi/pull/2170 is merged

conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
dewagter: This will partially be solved with: https://github.com/paparazzi/paparazzi/pull/2172

But for now we still need a separate airframe file in order to have the orange_avoid module. One solution to this was: https://github.com/dewagter/paparazzi/tree/calibration_xml in which you could split airframe.xml into 2 parts: a common ardrone part and a few specific ones.

kirkscheper: I am not sure why you need a separate airframe. See https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/tudelft/origami_lisamxs_wifi_indi_stereoboard.xml for an example of how to set up a single airframe for ap and nps.

https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/KS/ks_bebop2_stereo.xml shows an example with a bebop.

I also don't see how #2172 solves your issue. That PR only makes it easier to set the default environment from the flightplan. Currently, this can already be done with the NPS_GAZEBO_WORLD define in the airframe file.

All you have to do is correctly define the nps target in https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/tudelft/bebop_orangeavoid_course2017.xml then you can run it in simulation. 

As I mentioned before, the only possible issue is that the image might be a different size than on the real vehicle (the frame rate may also be different). The camera settings defined at the top of that airframe are not considered in nps yet, you will just get a fixed image with dimensions 320x240. This can be changed however, since #2158, sensors can now be added at runtime. So you could potentially make a bebop vehicle instead of the simple_x_quad which doesn't have cameras and add them at runtime with the correct settings. But none of this will be fixed by added a separate airframe.

Maybe if you describe why you need separate files in a bit more detail I can try to change nps to facilitate it.

conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
```diff
@@ -0,0 +1,223 @@
...
+    <module name="actuators" type="ardrone2"/>
+    <module name="imu" type="ardrone2"/>
+    <!-- gps: "ublox" or change to "sirf" for usage with parrot flight recorder -->
+    <module name="gps" type="ublox"/>
+    <module name="stabilization" type="int_quat"/>
+    <module name="ahrs" type="int_cmpl_quat"/>
+    <module name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
```
gautierhattenberger: is it possible to move these modules to the firmware section ?

sw/airborne/modules/orange_avoider/orange_avoider.c
```diff
@@ -30,6 +31,31 @@
 #define VERBOSE_PRINT(...)
 #endif
 
+#ifndef ORANGE_AVOIDER_LUM_MIN
```
gautierhattenberger: these parameters should be described in the documentation of the module

dewagter: sure

conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
dewagter: this is on purpose since that does not work as expected... still investigating. will fix this for all tudelft airframes once solved.

conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
dewagter: @kirkscheper the flying airframe is a bebop1, the simulated one is an ARDrone. Both have not only different board but different pretty much everything. Merging the control gains into 1 file is not working for me. I tried to have gazebo fly with bebop gains but that did not work either.

sw/airborne/modules/orange_avoider/orange_avoider.c
dewagter: fixed

conf/airframes/examples/ardrone2_gazebo_cyberzoo.xml
dewagter: fixed this part

PR #2154, opened 2017-10-31
Module to Read and Broadcast Decawave Serial Communication within Paparazzi

coppolam:
The module works with Decawave DWM1000 modules on Arduino, allowing multiple modules to communicate velocity and height from paparazzi with each-other, on top of measuring inter-antenna range. The link to the library and Arduino build file is provided in the module description. An ABI message with the received information is then broadcasted.


gautierhattenberger:
How is this module different from http://docs.paparazziuav.org/latest/module__dw1000_arduino.html ? Can we merge the two functionality if not exactly the same ?

coppolam:
The module differs in that it is made for anchorless networks where UWB does not function as a trilateriation technology, but as a ranging/communication technology between MAVs. The module enables communication paparazzi on-board variables between MAVs (in this case VX, VY, Z). The agents need to communicate these values in order to achieve anchorless relative localization. From my understanding, this is a different functionality than dw1000_arduino and hence could be kept separate.

conf/abi.xml
```diff
@@ -104,6 +104,14 @@
       <field name="z" type="float" unit="m"/>
       <field name="noise" type="float"/>
     </message>
+    
+    <message name="UWB_COMMUNICATION" id="19">
+      <field name="id" type="uint8_t"/>
+      <field name="range" type="float" unit="m"/>
+      <field name="vx" type="float" unit="m/s^2"/>
```
gautierhattenberger: Is this acceleration or speed ? Unit and name are not matching

conf/modules/decawave_serial_communication.xml
```diff
@@ -0,0 +1,29 @@
...
+  </doc>
+  <header>
+    <file name="decawave_serial_communication.h" />
+  </header>
+  <init fun="decawave_serial_communication_init()" />
+  <periodic fun="decawave_serial_communication_periodic()" freq="30" autorun="TRUE" />
+  <event fun="decawave_serial_communication_event()"/>
+  <makefile target="ap">
+    <file name="decawave_serial_communication.c" />
+    <configure name="SERIAL_UART" default="uart2" case="upper|lower" />
```
gautierhattenberger: `SERIAL_*` is a bit too generic, it should be at least prefixed by something in relation with the module name

sw/airborne/modules/decawave/decawave_serial_communication.c
```diff
@@ -0,0 +1,268 @@
...
+#include "subsystems/abi.h"
+
+PRINT_CONFIG_VAR(SERIAL_UART)
+PRINT_CONFIG_VAR(SERIAL_BAUD)
+
+// define coms link for stereocam
+#define SERIAL_PORT   (&((SERIAL_UART).device))
+struct link_device *xdev = SERIAL_PORT;
+
+#define SerialGetch() SERIAL_PORT ->get_byte(SERIAL_PORT->periph)
```
gautierhattenberger: Not sure these macros are a big help. Sometimes you use `SERIAL_PORT`, sometimes `xdev` (which is not a very explicit name). Would be better to be consistent.

sw/airborne/modules/decawave/decawave_serial_communication.c
```diff
@@ -0,0 +1,268 @@
...
+static uint8_t _dataTotalSend = 0;
+
+static bool _inProgress = false;
+static bool _allReceived = false;
+
+static uint8_t _varByte = 0;
+
+static struct nodeState _states[UWB_SERIAL_COMM_DIST_NUM_NODES];
+float range_float = 0.0;
+
```
gautierhattenberger: In general, you are using here a lot of global variables that could be gathered in a single structure with the buffers, status flags, indexes, uart device pointer, ...

sw/airborne/modules/decawave/decawave_serial_communication.c
```diff
@@ -0,0 +1,268 @@
...
+    if (_varByte == UWB_SERIAL_COMM_SPECIAL_BYTE){
+      i++;
+      _varByte = _varByte + _tempBuffer[i];
+    }
+    if(_dataRecvCount<=UWB_SERIAL_COMM_FLOAT_SIZE){
+      _recvBuffer[_dataRecvCount] = _varByte;
+    }
+    _dataRecvCount++;
+  }
+  if(_dataRecvCount==UWB_SERIAL_COMM_FLOAT_SIZE){
```
gautierhattenberger: when you're done, don't forget to run the `fix_code_style.sh` script on your files

sw/airborne/modules/decawave/decawave_anchorless_communication.c
```diff
@@ -0,0 +1,278 @@
...
+  The example file to flash the Arduino Micro can be found in
+    examples/UWB_localization_v1_0/UWB_localization_v1_0.ino
+ */
+
+#include "decawave_anchorless_communication.h"
+#include "subsystems/datalink/telemetry.h"
+#include "subsystems/radio_control.h"
+#include "state.h"
+#include "pprzlink/messages.h"
+#include "subsystems/datalink/downlink.h"
```
gautierhattenberger: is those includes (downlink.h, messages.h) really needed ?

gautierhattenberger:
Mostly okay. It seems that your files don't have a endline character at the end.
Also, the default baudrate is only 9600 bauds. It looks a bit slow for an application that potentially require a high update rate. Don't you think it could be higher ?

StevenH2812:
@gautierhattenberger It has been tested and working with 9600 baud rate. The lower baud rate ensures that the decawave side will function also with slower microprocessors. For the basic functionality of the module 9600 baud rate is enough since the communication is kept minimal. (6 byte messages, 4 messages back and forth, 20-30 Hz update frequency = 3840-5760 bit/s).  If you have a faster microprocessor then by all means you should increase it to ensure you always have the latest data available, but people can always do so if they do decide to use this module. I noticed with an 8 MHz Atmega328 that 57600 was the limit baud rate, higher baud rates started giving erroneous communication.

That being said, the default can be set higher to say 19200 as well. I think whoever uses this module should tune this value to whatever works well with the microprocessor used on the decawave side anyways, so it's quite arbitrary.

PR #2155, opened 2017-10-31
removed unnecessary pprz.conf files

coppolam:
Removed pprz.conf files from bebop wifi tools as they are not necessary. pprz_swarmhub.conf can be used instead


kirkscheper:
@coppolam is this tested?

coppolam:
Yes, I tried it yesterday on a Bebop 2

PR #2156, opened 2017-10-31
pprz_algebra_float.h new functions

coppolam:
Added a function to invert any matrix and a function to create an identity matrix.

### DIFF
--- a/sw/airborne/math/pprz_algebra_float.c
+++ b/sw/airborne/math/pprz_algebra_float.c
@@ -765,3 +765,56 @@ bool float_mat_inv_4d(float invOut[16], float mat_in[16])
 
   return 0; //success
 }
+
+/** Calculate inverse of any n x n matrix (passed as C array) o = mat^-1
+Algorithm verified with Matlab. 
+Thanks to: https://www.quora.com/How-do-I-make-a-C++-program-to-get-the-inverse-of-a-matrix-100-X-100
+*/
+void float_mat_invert(float **o, float **mat, int n)
+{
+  int i, j, k;
+  float t;
+  float a[n][2*n];
+
+  // Append an identity matrix on the right of the original matrix
+  for(i = 0; i < n; i++) {
+    for(j = 0; j < 2*n; j++) {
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
The new function are considering matrix as 1D arrays, while existing functions are using 2D arrays. I suggest to make them consistent.

kirkscheper:
@gautierhattenberger are there any standards in pprz as to how matricies are defined. Most functions assume a matrix is defined as a double array (a[][]) e.g. float_mat_mul(float **o, float **a, float **b, int m, int n, int l), but some assume a single array (a[]). e.g float_mat_inv_2d(float inv_out[4], float mat_in[4]).

It would appear that a matrix defined by a single array is assumed to be a square matrix, is that correct?

Additionally, what is the policy with inline functions, preferred or not?

gautierhattenberger:
There is no string guidelines at the moment. Double arrays are used for arbitrary size matrices. As you said, it seems that some fixed-size matrix functions are using single array. We could stick to this for now. Which means that in the case of this PR, double arrays are preferred.

Regarding inline or not, I would say that if the function is short enough (3 to 4 lines) it can be inline, and a regular function otherwise. But this is really arbitrary...

coppolam:
I will change them, although for the future I think it might useful to have both versions (double and single). Passing single arrays means that the functions can also be directly used with array operations, so that  functions like float_mat_copy, float_mat_sum could also be used not just for matrix double arrays, but also for arrays.

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -765,3 +765,56 @@ bool float_mat_inv_4d(float invOut[16], float mat_in[16])
...
+
+  /* Cut out the identity, which has now moved to the left side */
+  for(i = 0 ; i < n ; i++ ) {
+    for(j = n; j < 2*n; j++ ) {
+      o[i][j-n] = a[i][j];
+
+    }
+  }
+
+}
```
kirkscheper: file ending

PR #2158, opened 2017-11-02
Add rangesensor module

kirkscheper:
This pull request includes several items relating to adding range sensors to MAVs:

- Compatibility to read out the new ToF laser ring through UART (designed by @dewagter)
- Adding laser range sensor array to gazebo models (Bottom laser range sensor is send through like a sort of sonar)
- Made an module that creates a velocity forcefield based on those sensors, which can be used within guided flight.
- Make SWD_NOPWR default for lisa_mxs boards
- Converted all tabs in the gazebo models to double spaces
- Dynamically add sensors to gazebo vehicles depending on the modules used
- Added simple square environment for collision avoidance navigation testing
- Added some needed math functions

This replaces #2114 due to conflicting files.


gautierhattenberger:
It looks fine for me. It this still incomplete or ready to be merged ?

kirkscheper:
I had an epiphany tonight to improve efficiency so will update tomorrow.

gautierhattenberger:
I'm not completely satisfied with the new `if` statement fo abi bindings in FP as it solves a particular issue. Someone may want a different specific behavior some day and we'll end up with more "options" or attributes. What do you think of the possibility to have a `handler` attribute similar to the settings that allows to specify a custom callback function? So if you don't want the default behavior (update the values with the last message) you can do whatever you want. The callback can be defined in the `header` section of the flight plan or from a module.

kirkscheper:
@knmcguire This is a bit different than the original pull request. I removed the RANGE_SENSOR_ARRAY message and now have the range_forcefield bind to the OBSTACLE_DETECTION message (which now has an elevation field). This way the force field will work with any sensor that can measure distance to an obstacle and not only the range array.

The range_forcefield now also stores the nearest distance experienced in the xyz directions over time so that it could still work if you had multiple sensors pointed in the same direction (or at angles not parallel with the xyz plane. This formulation will also ensure that if you have multiple sensors in the same/similar direction that One important difference is that the resultant velocity command will never exceed the max_vel setting of the range_forcefield. 

The range is also now sent periodically rather than asynchronously.

@knmcguire let me know if this is good or if you had a different plan for this.

@gautierhattenberger I implemented a handler (I hope this is what you meant). The only issue I had was adding an & in the header section. Adding it like & caused the flight_plan.h generator to fail and adding it like &amp; worked fixed the flight_plan.h generator but broke the modules.h generator. So I ended up just using a nested if.

knmcguire:
@kirkscheper  Actually, what you added was exactly what I had in mind for a future pull request, so I am totally okay with it:)

gautierhattenberger:
@kirkscheper please have a look at my proposed changes in paparazzi/pull_request_add_rangesensor_module branch

I made a few changes to the ocaml part to be a bit more "ocaml-ish".

Regarding the issue with `&&` instructions, I don't have a nice solution. You need the escape sequence to have it displayed correctly, but gen_aircraft is parsing the original file (where you may have the escape sequence) while gen_modules is parsing the expanded version in var/aircraft/<AC>/flight_plan.xml which is discarding the escape characters. I cannot do anything about this as it is coming from the ocaml xml lib.
So my answer would be: don't use `&` :(
For simple tests, you can use the `And( , )` macro as suggested in my modified FP.

PR #2160, opened 2017-11-03
Pullrequest pprz math array operations

coppolam:
Added a function to cast a value to an array of integers (so that it may be initialized as desired with one command), and a function to find a specific value within the array.
These functions will be used in later pull requests in order to store aircraft IDs and determine if they are new or not during relative localization and avoidance, but are general enough to be put in pprz_algebra_int.h.

### DIFF
--- a/sw/airborne/math/pprz_algebra_int.h
+++ b/sw/airborne/math/pprz_algebra_int.h
@@ -608,6 +608,14 @@ static inline void int32_vect_zero(int32_t *a, const int n)
   for (i = 0; i < n; i++) { a[i] = 0.; }
 }
 
+/** a = v */
+static inline void int32_vect_value(int32_t *a, const int32_t v, const int n)
+{
+  int i;
+  for(i = 0 ; i < n; i++) { a[i] = v; }
+}
+
+
 /** a = b */
 static inline void int32_vect_copy(int32_t *a, const int32_t *b, const int n)
 {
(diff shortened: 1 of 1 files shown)


sw/airborne/math/pprz_algebra_int.h
```diff
@@ -608,6 +608,14 @@ static inline void int32_vect_zero(int32_t *a, const int n)
   for (i = 0; i < n; i++) { a[i] = 0.; }
 }
 
+/** a = v */
+static inline void int32_vect_value(int32_t *a, const int32_t v, const int n)
```
kirkscheper: Not very descriptive function name. Maybe something like set_value. Also the function description is not accurate (see vect_copy below). I think you mean something like a = v * I(n,1)...

sw/airborne/math/pprz_algebra_int.h
```diff
@@ -657,6 +665,20 @@ static inline void int32_vect_smul(int32_t *o, const int32_t *a, const int32_t s
   for (i = 0; i < n; i++) { o[i] = a[i] * s; }
 }
 
+/** Finds a value s in an array and returns the first location at which it is found, else -1 */
```
gautierhattenberger: I suggest to make the comment more accurate about the return value (true or false) and the value you get from the argument

sw/airborne/math/pprz_algebra_int.h
```diff
@@ -665,7 +665,9 @@ static inline void int32_vect_smul(int32_t *o, const int32_t *a, const int32_t s
   for (i = 0; i < n; i++) { o[i] = a[i] * s; }
 }
 
-/** Finds a value s in an array and returns the first location at which it is found, else -1 */
+/** Find value s in array a. Returns 1 if found, 0 if not found.
```
kirkscheper: hahaha, not to nit pic here but if it is a bool shouldn't the return be true or false?

Issue #2144, opened 2017-10-30
VELOCITY_ESTIMATE ABI message requires vertical velocity


tomvand:
With the current definition of the VELOCITY_ESTIMATE ABI message it is not possible to send only a horizontal velocity estimate, the vertical velocity also has to be included. Setting the vertical velocity estimate to zero or stateGetSpeedNed_f()->z as a workaround results in incorrect behaviour. Can the message be changed to include a flag that indicates which axes are estimated, or separated into a horizontal and vertical estimate?

gautierhattenberger:
A flag can be added similar to https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml#L674 in order to specify if horizontal and/or vertical estimate have been updated

kirkscheper:
closed with #2161

PR #2161, opened 2017-11-03
add flags to velocity and position estimate abi messages

kirkscheper:
This makes it optional to define if you want to update the horizontal and/or vertical parameters of the velocity and position estimates from the abi message. This should address the issue #2144 .

### DIFF
--- a/sw/airborne/modules/guidance_opticflow/guidance_opticflow_hover.c
+++ b/sw/airborne/modules/guidance_opticflow/guidance_opticflow_hover.c
@@ -98,7 +98,7 @@ struct opticflow_stab_t opticflow_stab = {
 
 
 static void stabilization_opticflow_vel_cb(uint8_t sender_id __attribute__((unused)),
-    uint32_t stamp, float vel_x, float vel_y, float vel_z, float noise);
+    uint32_t stamp, float vel_x, float vel_y, float vel_z, float noise_x, float noise_y, float noise_z);
 /**
  * Initialization of horizontal guidance module.
  */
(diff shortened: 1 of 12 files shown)


gautierhattenberger:
In `guidance_opticflow_hover.c` line 100 the declaration of the callback is not updated

kirkscheper:
Wait till #2132 is accepted before merging this. Then I can extend the fixes to that module as well.

tomvand:
Thanks Kirk, really needed this :).

Would it be prudent to add separate bits for the X and Y axes? (In case someone builds a forward-only velocity estimator?)

kirkscheper:
Not really. I can't image a situation where you would only estimate forward velocity with a rotorcraft, you would need both to be able to fly stably. In the odd case that that happens this can always be expanded later.

tomvand:
I agree for rotorcraft, but aren't the ABI messages also used for fixed-wing/hybrid uavs? Expanding later is fine by me btw.

flixr:
If we already add flags, I would really propose to add them for each axis.
Another maybe even better option would be to extend the noise, i.e. create a `noise_x`, `noise_y`, `noise_z`. That way you could even have a different measurement/estimate noise for each axis... and to mark one axis as invalid, just set it to a negative value.

kirkscheper:
I went with @flixr comment. I think this is done now.

tomvand:
Looks good

sw/airborne/modules/guidance_opticflow/guidance_opticflow_hover.c
```diff
@@ -149,28 +149,42 @@ void guidance_h_module_run(bool in_flight)
...
+  if (noise_x >= 0.f)
+  {
+    /* Calculate the error */
+    float err_vx = opticflow_stab.desired_vx - vel_x;
 
-  /* Calculate the integrated errors (TODO: bound??) */
-  opticflow_stab.err_vx_int += err_vx / 512;
-  opticflow_stab.err_vy_int += err_vy / 512;
+    /* Calculate the integrated errors (TODO: bound??) */
+    opticflow_stab.err_vx_int += err_vx / 512;
```
gautierhattenberger: what is this 512 ? If it is the related to the general periodic freq, it shouldn't be hard-coded. Or at least configurable.

kirkscheper: I guess it probably is but I didn't change this, only added a tab.

IMHO this entire module doesn't make much sense. This functionality should be possible with just normal control loops. I also see that this module isn't included in any airframe. I am in favour of deleting this altogether.

gautierhattenberger: I don't mind deleting old code. I guess it was a demo module made by Christophe or Freek. I let you guys decide if it is worth keeping it.

PR #2162, opened 2017-11-04
Added support for encrypted pprzlink

podhrmic:
Paparazzi side, see [#65](https://github.com/paparazzi/pprzlink/pull/65) for details.

The intention was to hide all the crypto from the user, so there is the same Paparazzi code for normal and encrypted version of pprzlink.

## Changes

- bump pprzlink
- added `secure transparent telemetry` module
- I merged and rebased latest master, so it kind of clutters the PR

## For reviewers
- please see comments in the code

**NOTE**: as is it shouldn't be used for any application where security is important (the key exchange part of the crypto algorithm is still missing and we are using fixed keys ATM).

PR #2163, opened 2017-11-04
Added HACL-C module and secure telemetry module

podhrmic:
I decided to separate #2162 into multiple smaller commits. 

This PR adds HACL* C snapshot, so the cryptographic functions can be accessed from within the autopilot. 

No need to review the HACL* C files as they are autogenerated. Eventually I might include just a git submodule, but for now copying the files is fine.


gautierhattenberger:
please don't merge things as important before they are even discussed !

PR #2164, opened 2017-11-04
Added key generator for asymmetric keys

podhrmic:
Generates two header files, `keys_uav.h` and `keys_gcs.h` and puts them into the destination directory. Should be called externally from a makefile.

Makefile command: 
```
$(AIRFRAME_H) : $(CONF)/$(AIRFRAME_XML) $(CONF_XML) $(AIRCRAFT_MD5)
...
  $(Q)cargo run --manifest-path $(PAPARAZZI_HOME)/sw/tools/key_generator/Cargo.toml --release $(AC_GENERATED)
```

Requires Rust and Cargo (get it from https://www.rustup.rs/ )

### DIFF
--- a/sw/tools/key_generator/.gitignore
+++ b/sw/tools/key_generator/.gitignore
@@ -1,0 +1,3 @@
+/target/
+**/*.rs.bk
+Cargo.lock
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
These kind of things should be discussed a bit before merging. Not that I'm against RUST, but for tools it should preferably use Python as we already have dependencies for this.

PR #2165, opened 2017-11-05
Autogenerate communication keys only when needed

podhrmic:
I need to generate keys upon each airframe build for the encrypted link communication.

This change does the job (edit in the Makefile.ac), the keys are generated only when I have the [transparent_secure_telemetry](https://github.com/paparazzi/paparazzi/blob/master/conf/modules/telemetry_transparent_secure.xml) module enabled.

@gautierhattenberger or @flixr - any comments on this solution? I don't like the duplicity in the Makefile, but I don't know how else do the conditional generation.

### DIFF
--- a/Makefile.ac
+++ b/Makefile.ac
@@ -70,6 +70,7 @@ MODULES_H=$(AC_GENERATED)/modules.h
 MODULES_DIR=$(PAPARAZZI_HOME)/conf/modules/
 AUTOPILOT_DIR=$(AC_GENERATED)/
 AIRCRAFT_MD5=$(AIRCRAFT_CONF_DIR)/aircraft.md5
+GENERATE_KEYS ?= 0
 
 UNAME = $(shell uname -s)
 ifeq ("$(UNAME)","Darwin")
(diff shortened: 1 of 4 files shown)


gautierhattenberger:
The main issue is the dependency over an extra language. Since it seems a Python library exists and is packaged for Ubuntu (python-axolotl-curve25519, https://github.com/tgalal/python-axolotl-curve25519), I think it should be used for the key generation.

podhrmic:
I was thinking about that (using an existing Python library), but the intent is to use solely HACL* because it is a formally verified cryptolibrary. Hence (assuming the proofs etc. are correct) it is guaranteed to be correct.

I could write the key generator in C/C++, but I already had Rust bindings for HACL so it was easier this way (plus Rust is a memory safe language which makes the guarantees even stronger). 

Since the key generator is used only for secure comms, it wouldn't interfere with a regular user experience. I know it is another dependency, but I think the added safety is worth the additional overhead.

podhrmic:
Closing for now, as this should be a part of the secure link integration, see #2189 metaissue for details.

PR #2173, opened 2017-11-07
AWB

dewagter:
@wv-tud can you please help merge the useful developments in here


kirkscheper:
I tried to do my best cleaning this up and adding some documentation. This is tested and working. I did notice that the AWB gain scheduling was not very stable so I have deactivated it be default.

gautierhattenberger:
It looks fine.
What about having all the `ae_` and `awb_` variables in a structure (or two) ?
You can also run the style script.

kirkscheper:
I noticed some stability problems and an unrecoverable saturation. All fixed now. I think this is good to merge.

PR #2176, opened 2017-11-08
vision based relative positioning

dewagter:
@wv-tud


kirkscheper:
Its useful if you could add a little description with that is in the pull request or what the added module does for the commit history and pr logs.

Also, don't forget need to run code_style

wackv:
Will do. But won't be for a while due to upcoming flight tests.

wackv:
@benelgiz I haven't gotten around to adding it, I will see if I can do so
asap

sw/airborne/modules/computer_vision/active_random_filter.cpp
```diff
@@ -0,0 +1,1959 @@
...
+ * @file "modules/computer_vision//cv_active_random_filter.c"
+ * @author Wilco Vlenterie (wv-tud)
+ * Active random sampling colour filter
+ */
+
+#include "active_random_filter.h"
+#include <vector>
+#include <ctime>
+
+#define BOARD_CONFIG "boards/bebop.h"       // TODO: WHYYYY?
```
gautierhattenberger: good question, why is this needed ?
if should be something like `#include BOARD_CONFIG` in place of the `#include "boards/bebop.h"` below and removing this line

kirkscheper:
I will update this after #2364

gautierhattenberger:
up?

Issue #2169, opened 2017-11-06
Merge Vision Developments WV


dewagter:
- [ ] codesourcery compiler from https://github.com/Parrot-Developers/toolchains
   - [ ] @fvantienen : https://github.com/tudelft/toolchains
 - [ ] airframes
   - [x] https://github.com/paparazzi/paparazzi/pull/2170 course airframe avoid 
   - [x] https://github.com/paparazzi/paparazzi/pull/2149 orange_avoid in sim: 
   - [ ] https://github.com/paparazzi/paparazzi/pull/2175 course airframe attitude
   - [ ] https://github.com/paparazzi/paparazzi/issues/2171 update move WP to guidance
 - [x] https://github.com/paparazzi/paparazzi/pull/2159 opencv 3.2 from https://github.com/tudelft/opencv_bebop 
 - [ ] https://github.com/paparazzi/paparazzi/pull/2180 ISP + video grabbing update
 - [x] https://github.com/paparazzi/paparazzi/pull/2173 AE_AWB color gains + frame based instead of periodic
 - [ ] H264 hwenc
   - [ ] either https://github.com/tudelft/lib_h264_bebop
   - [ ] or https://github.com/tudelft/tudelft_vision
   - [x] https://github.com/paparazzi/paparazzi/pull/2174 send from direct hardware buffer
   - [ ] https://github.com/paparazzi/paparazzi/issues/2020 update to work with new parrot firmware
 - [ ] TRICAL fix: https://github.com/wv-tud/TRICAL/commit/29ef9c73f728aef04b9d58f7bdafb19fd73b7596
 - [ ] videostab
   - [ ] cv_image_pose
   - [ ] https://github.com/paparazzi/paparazzi/pull/2168 libopengles 
   - [ ] unrotator: https://github.com/wv-tud/paparazzi/blob/master_AS_RTK/sw/airborne/modules/computer_vision/cv_bebop_stabilization.c 
 - [ ] mp4 record
 - [ ] vision based relative positioning
    - [ ] https://github.com/paparazzi/paparazzi/pull/2176 random sampling tracker

PR #2180, opened 2017-11-09
Faster computer vision on bebop using isp

dewagter:
This is the code in use in all the bebop projects where fast good front camera images are important: 

 - drone race
 - course
 - swarming
 - [ ] add an option for the original resolution
 - [ ] merge the VGA resolution option


conf/Makefile.arm-linux-toolchain
```diff
@@ -26,6 +26,8 @@ ifndef HARD_FLOAT
 endif
 endif
 
+PREFIX=/opt/arm-2012.03/bin/arm-none-linux-gnueabi
```
flixr: Will break other linux builds

dewagter: fixed... will be addressed separately: see https://github.com/paparazzi/paparazzi/issues/2169

OpenUAS:
Any open objections or can we merge it? Then I can start retest on BB,BB2 and prelim Disco

kirkscheper:
@OpenUAS This is still marked as incomplete by Christohpe. Also see the check boxes in the original comment.

kirkscheper:
This has been implemented with #2364

PR #2188, opened 2017-11-16
[gps] add support for REMOTE_GPS_LOCAL

gautierhattenberger:
position and velocity is received in local frame
transformation to global frame is handled on airborne side


kirkscheper:
Looks good. I say update the pprzlink sha and merge.

PR #2192, opened 2017-11-27
Update libopencm3 and add Random Number Generator (RNG)

podhrmic:
RNG is a prerequisite for secure pprzlink.
Updated libopencm3 already has drivers for RNG.
ChibiOS would have to be updated too, but the new ChibiOS currently doesn't work with [TLS module](https://github.com/paparazzi/paparazzi/blob/master/conf/modules/tlsf.xml)
LPC21 autopilots don't have the RNG hardware.

Tested on Lisa MX. Most stm32 chips have RNG hardware.

@gautierhattenberger any chance you would want to update TLS to work with the new Chibios?


podhrmic:
The errors are due to #2194 - not because of changes introduced here.

podhrmic:
@gautierhattenberger any comments on this or can I merge?

kirkscheper:
It looks like the commit history is a bit mixed up, your changes seemed to be mixed in with ones that are already upstream. Can you rebase your changes on top of the current master?

gautierhattenberger:
beside the rebase issue, it looks fine

PR #2195, opened 2017-11-28
Draw line fix so that color actually works, plus extra functions for …

guidoAI:
…drawing crosshairs and rectangles.


sw/airborne/modules/computer_vision/lib/vision/image.c
```diff
@@ -578,27 +608,24 @@ void image_show_points(struct image_t *img, struct point_t *points, uint16_t poi
...
+  uint8_t color[4] = {255,255,255,255};
+  int size_crosshair = 5;
+
   // Go through all the points
   for (uint16_t i = 0; i < points_cnt; i++) {
     // Draw a line from the original position with the flow vector
     struct point_t from = {
       vectors[i].pos.x / subpixel_factor,
-      vectors[i].pos.y / subpixel_factor,
-      0,
```
gautierhattenberger: what about the other parameters of the `point_t` structure ? shouldn't they be initialized properly ?
if only x,y are needed in most case, maybe an init macro can be added to images.h to hide the other parameters initialization.

guidoAI: Good remark! point_t got additional properties in a previous pull request - I believe by Manos. I wonder if this is actually the best way to go, as a point as a mathematical concept would not require properties such as the number of times observed. Perhaps a solution would be an additional struct with a point_t as one of its attributes. However, I feel this is beyond the scope of the current pull request.

sw/airborne/modules/computer_vision/lib/vision/image.c
```diff
@@ -578,27 +608,24 @@ void image_show_points(struct image_t *img, struct point_t *points, uint16_t poi
  */
 void image_show_flow(struct image_t *img, struct flow_t *vectors, uint16_t points_cnt, uint8_t subpixel_factor)
 {
+  uint8_t color[4] = {255,255,255,255};
+  int size_crosshair = 5;
```
kirkscheper: In the future, things like this can be declared static so that they are not constantly redeclared. alternatively you can consider using the const descriptor, helps with code clarity.

sw/airborne/modules/computer_vision/lib/vision/image.c
kirkscheper: Can you make an issue about this so that it is addressed in the future?

sw/airborne/modules/computer_vision/lib/vision/image.c
```diff
@@ -725,15 +824,25 @@ void image_draw_line_color(struct image_t *img, struct point_t *from, struct poi
...
+
+      // depending on startx being odd or even, we first have to set U or V
+      if(startx % 2 == 1) {
+          temp_color[0] = color[2];
+          temp_color[2] = color[0];
+      }
+      else {
+          temp_color[0] = color[0];
+          temp_color[2] = color[2];
+      }
```
kirkscheper: Don't forget to run fix code style. You can also change the settings of your editor so you don't have this problem in the future.

sw/airborne/modules/computer_vision/lib/vision/image.c
guidoAI: I made it static now

sw/airborne/modules/computer_vision/lib/vision/image.c
guidoAI: Ran fix code style

sw/airborne/modules/computer_vision/lib/vision/image.c
guidoAI: I don't think it's worth an issue

sw/airborne/modules/computer_vision/lib/vision/image.c
kirkscheper: This was not specific to this line but in general, often you define a variable with a constant value, in that case a static or const (or both) descriptor is interesting. Consider it in the future.

PR #2197, opened 2017-11-28
Qtcreator indent size fix and added upload feature

kevindehecker:
Now you can directly upload from Qtc. Also, the wrong style template was selected overriding the indent size wrongly


gautierhattenberger:
The opencv submodule is updated and I guess it shouldn't be needed

Issue #1697, opened 2016-06-01
NatNet2ivy should request version number from optitrack


kirkscheper:
The current software has hardcoded version numbers. This should be made more portable by requesting the sdk version number from natnet using a PING request.

OpenUAS:
Some old links but still mabe useful about the dissussion with Optitrack devs
https://forums.naturalpoint.com/viewtopic.php?f=59&t=13272

gautierhattenberger:
This should work already with latest natnet add the python tool NatNet3

PR #2198, opened 2017-11-28
[natnet] add a NatNet3 to IVY bridge in Python

gautierhattenberger:
The new version of the Optitrack software (Motive 2.0) comes with NatNet
version 3. A Python module is provided with NatNet for
serialization/deserialization of the data streamed over the network.
This module is slightly modified here for a better handling of the
initialization and for stopping the threads.
The PPRZLINK message sent is REMOTE_GPS_LOCAL, so the transformation to
global coordinate system is left to the airborne part.

### DIFF
--- a/sw/ground_segment/python/natnet3.x/NatNetClient.py
+++ b/sw/ground_segment/python/natnet3.x/NatNetClient.py
@@ -128,7 +128,8 @@ class NatNetClient:
                     self.__trace( "\tMarker Size", i, ":", size[0] )
                     
         # Skip padding inserted by the server
-        offset += 4
+        if( self.__natNetStreamVersion[0] > 2 ):
+            offset += 4
 
         if( self.__natNetStreamVersion[0] >= 2 ):
             markerError, = FloatValue.unpack( data[offset:offset+4] )
@@ -235,7 +236,7 @@ class NatNetClient:
                 offset += 4
 
                 # Version 2.6 and later
-                if( ( self.__natNetStreamVersion[0] == 2 and self.__natNetStreamVersion[1] >= 6 ) or self.__natNetStreamVersion[0] > 2 or major == 0 ):
+                if( ( self.__natNetStreamVersion[0] == 2 and self.__natNetStreamVersion[1] >= 6 ) or self.__natNetStreamVersion[0] > 2):
                     param, = struct.unpack( 'h', data[offset:offset+2] )
                     offset += 2
                     occluded = ( param & 0x01 ) != 0
@@ -243,7 +244,7 @@ class NatNetClient:
                     modelSolved = ( param & 0x04 ) != 0
 
                 # Version 3.0 and later
-                if( ( self.__natNetStreamVersion[0] >= 3 ) or  major == 0 ):
+                if( self.__natNetStreamVersion[0] >= 3 ):
                     residual, = FloatValue.unpack( data[offset:offset+4] )
                     offset += 4
                     self.__trace( "Residual:", residual )
@@ -313,7 +314,7 @@ class NatNetClient:
             offset += 4
 
         # Hires Timestamp (Version 3.0 and later)
-        if( ( self.__natNetStreamVersion[0] >= 3 ) or  major == 0 ):
+        if( self.__natNetStreamVersion[0] >= 3 ):
             stampCameraExposure = int.from_bytes( data[offset:offset+8], byteorder='little' )
             offset += 8
             stampDataReceived = int.from_bytes( data[offset:offset+8], byteorder='little' )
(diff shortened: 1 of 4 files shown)


gautierhattenberger:
@kirkscheper, can someone from your lab test this code with your current natnet version (especially if you are using 2.x) ? I only have the latest natnet to test it, but it might be compatible with older version as well. Thanks

kirkscheper:
@gautierhattenberger no prob.

kirkscheper:
I needed to make a few changes for this to work. see changes at https://github.com/kirkscheper/paparazzi/tree/natnet3_python

One of the significant changes are to rotate the optitrack frame to ENU which was not done as well as some changes to the natnetclient that were crashing for me. 

Another issue is from #1697, if we will have multiple systems using these scripts it is really important the version number is auto identified. I tried to get the ping to work but could not succeed, maybe you can have better luck than me.

gautierhattenberger:
Actually, with Motive 2.0 you can output directly in ENU, that's why the rotation was not needed for me. I don't know if you can do that with older version. Also, it can only received rigid bodies, otherwise it is crashing. Your data size checks are probably related to that.
Maybe I can add the options for rotation and Z/Y up setting and keep the rest like this for now ?

kirkscheper:
I pushed the minimum changes required to make the script run for v2. The size checks are for safety, the check for len(data) > 0 is not smart if the next line you do is data[0:2] then data[2:4]... The next size check is that the entire message is received, this is also important as the parser does not check this anywhere else and simply tries to access the data without checking that there is enough.

I also rebased on master

Note that this will now run for Natnet version < 3 but the gps_local result sent to the vehicle will not work with motive version < 2.

sw/ground_segment/python/natnet3.x/NatNetClient.py
```diff
@@ -235,15 +236,15 @@ def __unpackMocapData( self, data ):
                 offset += 4
 
                 # Version 2.6 and later
-                if( ( self.__natNetStreamVersion[0] == 2 and self.__natNetStreamVersion[1] >= 6 ) or self.__natNetStreamVersion[0] > 2 or major == 0 ):
+                if( ( self.__natNetStreamVersion[0] == 2 and self.__natNetStreamVersion[1] >= 6 ) or self.__natNetStreamVersion[0] > 2):
```
gautierhattenberger: it actually seems that these `or major == 0` are a bad copy/paste from the C/C++ example. It should have been `or self.__natNetStreamVersion[0] == 0`. I don't know if we should keep them or not.

gautierhattenberger:
is GPS_LOCAL message not working only because there is not rotation or axis correction, or something else ?

kirkscheper:
I am not sure about Motive v2 but in v1 optitrack y axis in vertical with xz on the ground plane, therefore the drone just gets totally wrong position with this script.

The rotation to north that I applied is more so that the GCS and reality line up. I am not aware of how you apply the rotation in motive (or if it is possible in v1).

gautierhattenberger:
With v2 you can choose the vertical axis and eventually perform extra rotation/translation.
For now, we are not flying true north but using the optitrack yaw angle and north/east aligned with optitrack axis. Of course we'll need the rotation angle at some point, or init the ground frame aligned to true north.
As I said, we can add options for those to parameters, especially if you want to use this Python tool instead of the C version, otherwise there is no hurry.

kirkscheper:
I am very much in favour of using one tool, less maintenance (especially as I am the one who has to do it, hahaha)! This way we can share the misery. 

I think that once this works for you guys and it is well tested, we can also migrate to v2.

kirkscheper:
After speaking with @matejkarasek, I have a bit more info. Here is a summary.

We currently have two independent optitrack systems (one using Flex13 and the other the Prime17W). The Prime17Ws are now in our cyber zoo and for that we can upgrade to Motive version 2.0. We use the Flex13s in the windtunnel or other temporary setups but that license key can't be upgraded to Motive v2.0 (without paying more).

We have had multiple stability issues with Motive so we don't want to upgrade to v2 just yet but once the guys at ENAC show that its working well we will upgrade our system to v2 and try to use the same software as ENAC to reduce maintenance efforts. 

We would however like the C code that is setup to work with the tudelft cameras to remain in pprz so that we can continue to use our Flex13 with master (I think I will add a readme to that folder to explain that it is basically only directly useful if you are flying with airframes in Delft and applicable/tested software combinations ect.).

So yea, @gautierhattenberger if this works for you feel free to merge and test and keep up abreast of your progress. If you need us to test anything let me know.

That is all...

gautierhattenberger:
tested and working, so merging as is

PR #2200, opened 2017-11-29
[INDI] flexible priorities for WLS allocation

EwoudSmeur:


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -403,8 +403,12 @@ static void stabilization_indi_calc_cmd(struct Int32Quat *att_err, bool rate_con
     du_pref[i] = act_pref[i] - actuator_state_filt_vect[i];
   }
 
+#ifndef STABILIZATION_INDI_WLS_PRIORITIES
   //State prioritization {W Roll, W pitch, W yaw, TOTAL THRUST}
   static float Wv[INDI_OUTPUTS] = {1000, 1000, 1, 100};
+#else
+  static float Wv[INDI_OUTPUTS] = STABILIZATION_INDI_WLS_PRIORITIES;
```
kirkscheper: static const float? You may also have to change the definition of wls_alloc to define the input const, but that depends on if it should be constant or not.

And shuldn't this be in the else of the #if STABILIZATION_INDI_ALLOCATION_PSEUDO_INVERSE a bit lower down?

EwoudSmeur:
Thanks Kirk, I moved the specific calculations within the #else, and put Wv of the top, it seemed cleaner. With const the compiler gives a warning because it is passed as a pointer, at which point the const qualifier is lost apparently.
I also noticed du_pref was not used

PR #2201, opened 2017-12-03
Universal secure pprzlink

podhrmic:
This PR adds a capability of using secure telemetry with various encryption/authentication protocols.

Currently a trivial example is provided.

Works with transparent telemetry and also in NPS.

Once merged, I ll make another PR with galois embedded crypto implementation.


conf/modules/telemetry_transparent_secure_example.xml
```diff
@@ -3,20 +3,26 @@
...
+
+      However, this example should make it easier for users to device their own encryption scheme.      
     </description>
     <configure name="MODEM_PORT" value="UARTx" description="UART where the modem is connected to (UART1, UART2, etc)"/>
     <configure name="MODEM_BAUD" value="B57600" description="UART baud rate"/>
   </doc>
-  <autoload name="telemetry" type="nps"/>
+  <!-- autoload secure_nps_example module to allow encryption for simulation and testing -->
+  <!-- change accordingly when devising your own secure link -->
+  <autoload name="telemetry" type="secure_nps_example"/>
```
kirkscheper: As far as I can see the difference between this and the 'secure_nps_example' is the peripheral (so may need some different defines) but you autoload the module so why not just make one module? 
For nps|hitl you use the UDP defines, for the others you use the UART defines. It would functionally be the same as what you have now except you wouldn't add an extra file which has to be maintained.

conf/modules/telemetry_transparent_secure_example.xml
```diff
@@ -3,20 +3,26 @@
...
   <doc>
     <description>
-      Telemetry using secure PPRZ protocol over UART
-      Similar to telemetry_transparent, except it uses message scheduling and encryption
+      Example of using secure telemetr over PPRZ protocol over UART
+      
+      This file serves only as an example and should not be used in real applications, 
+      because it only passes though all data (so it is equivalent to using plaintext data).
+
+      However, this example should make it easier for users to device their own encryption scheme.
```
kirkscheper: can you pass this through a spell check?
though->through
device->devise
ect.

conf/modules/telemetry_transparent_secure_example.xml
```diff
@@ -3,20 +3,26 @@
 <module name="telemetry_transparent_secure" dir="datalink" task="datalink">
```
gautierhattenberger: I think the module (and file) names should reflect the type of encryption in case we want to implement another one

podhrmic:
Closing as this PR was moved to #2205

PR #2204, opened 2017-12-06
[tmtc] add pprzlink_proxy tool

gautierhattenberger:
PPRZLINK Proxy is a tool to allow multiple aircraft simulations
with air-to-air communication using the PPRZLINK v2 protocol
using UDP.
As aircarft can't be identified by their IP addresses (most of the
time, all simulators are running on the same computer), each agent (GCS,
aircraft) are using different ports and the mapping between ports and IDs
is done by this tool.

### DIFF
--- a/sw/ground_segment/tmtc/pprzlink_proxy.py
+++ b/sw/ground_segment/tmtc/pprzlink_proxy.py
@@ -1,0 +1,207 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2017 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+#
+# This file is part of paparazzi.
+#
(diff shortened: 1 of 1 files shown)


hooperfly:
Very nice. I will try and use this in the next week or so for some testing. Thanks @gautierhattenberger

podhrmic:
Would it be possible to mention this on the wiki (perhaps [here](https://wiki.paparazziuav.org/wiki/Pprzlink#Differences_between_version_1_and_2) would be a good place) and add some examples (although I noticed you have some examples in the comments)

noether:
I will add in the wiki an example using this proxy after this pull request :P
https://github.com/paparazzi/paparazzi/pull/2111

noether:
Tested with the distributed formation control module. It works with the nps simulator.

PR #2206, opened 2017-12-20
Circular formations. More elegant script and examples

noether:
I have rewritten (and tested) the python script for the circular formations (centralized version). It is prettier now xD. I have also added two examples for the configuration file.

This script sends two messages for setting 'gvf_ell_a' and 'gvf_ell_b'. Because in this version we deal only with circles, both messages have the same payload (the radius of the circle to be tracked). However, it is possible to cover the more general case of ellipses with a small modification, and then 'gvf_ell_a' and 'gvf_ell_b' will be different.

This modification is something that I have to think a bit how to include it :P. If the bandwitdh is an issue, then the user should use the module for the decentralized version.

edit
-----

I have also rewritten (and tested) the initialization of tables for the decentralized case. Now both cases, centralized and decentralized, can use the same config files thanks to jetson.


noether:
mmm no idea why Travis is complaining

Issue #2097, opened 2017-08-03
(UPDATED) Optical flow code sends NaN: 1) to INS 2) to Other modules


TitusBraber:
https://github.com/paparazzi/paparazzi/blame/master/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c#L376
Recently added in #2079 , this line of code fills the result structure with 0's, including the FPS parameter when not enough corners are found. This happens when the drone is close to the ground, or no features are visible.

This FPS parameter is then used in the Kalman filter:
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c#L802
Where it tries to compute 1.0f / fps

Naturally this is not intended behavior, as it results in NaN velocities that the kalman filter outputs. These are send over the ABI message into the INS, causing the INS to go: "I don't know what to do, I'll just make everything 0". 

The result is the INS flipping between 0 and the actual values received from OptiTrack for example.
The behavior can be seen in this video:
https://www.youtube.com/watch?v=8EhzH8DJy_w

When flying this gives some trouble, i.e. twitching, but the real trouble comes when landing, at which point the drone really flips out and flies away out of control.

This is confirmed for both Ardrone2 and Bebop, that I tested together with @OpenUAS over the last two days.

@manoskyr I am not entirely sure which variables you want to reset and why. Can you explain this perhaps? We might be better off just resetting certain variables or not even resetting them. Alternatively it might be possible to skip the kalman step too when not enough corners are found?

ETitus:
With https://github.com/paparazzi/paparazzi/blame/master/sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c#L376 uncommented I decided to test https://github.com/paparazzi/paparazzi/pull/2087 and I found more NaN values seeping through from the optical flow code.

I am not entirely sure why this is happening, but my (partial) analysis:
It originates from this division:
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/opticflow/size_divergence.c#L117
Somehow after the vision updates it is possible for distance_1 to be 0.

It is of course possible to make a check if distance_1 is 0, and not divide by it if that is the case, but I am not sure what the value of divs[sample] should be set to in that case, 0? This "solves" the NaN, but the computed result might be wrong. 
Optionally a new pair of i,j can be forced by returning? 

However, as this never occurred in previous versions of the optical flow code, I dived deeper:

[...]

PR #2209, opened 2018-01-02
added a divide by 0 failsafe

TitusBraber:
Referencing back to https://github.com/paparazzi/paparazzi/issues/2097#issuecomment-320302913 again, there is still a divide by 0 possible and actually happening.
If for all samples in  
for (uint16_t sample = 0; sample < n_samples; sample++)
 distance_1 is 0, then the newly introduced used_samples remains 0, leading to a division by 0 in the mean_f function in:
https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/computer_vision/opticflow/size_divergence.c#L118
This value then messes up filters where the size_divergence is used, for example OF_Landing and forthcoming OF_Hover.

The suggested fix sets the mean_divergence to 0. Though I am NOT sure if this is the correct behavior to fix it.

### DIFF
--- a/sw/airborne/math/pprz_stat.c
+++ b/sw/airborne/math/pprz_stat.c
@@ -38,14 +38,19 @@
  */
 int32_t mean_i(int32_t *array, uint32_t n_elements)
 {
-  // determine the mean for the vector:
-  float sum = 0.f;
-  uint32_t i;
-  for (i = 0; i < n_elements; i++) {
-    sum += (float)array[i];
-  }
 
-  return (int32_t)(sum / n_elements);
+  if (n_elements == 0) {
+    // Note that something else is wrong if you want the mean of 0 samples.
+    return 0;
+  } else {
+    // determine the mean for the vector:
+    float sum = 0.f;
+    uint32_t i;
+    for (i = 0; i < n_elements; i++) {
+      sum += (float)array[i];
+    }
+    return (int32_t)(sum / n_elements);
+  }
 }
 
 /** Compute the variance of an array of values (integer).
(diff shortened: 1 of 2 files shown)


kirkscheper:
You not put the failsafe on the mean function?

TitusBraber:
I thought about that @kirkscheper, I even looked if the function was being used by other files (it is not).
I wasn't sure, but judging from that it doesn't cause a segfault, instead returning a NaN, it is a float division, I thought it might be better to have this for people that might use that function and not expect a 0 entries gives 0 result. Also I wasn't sure if an if check in a math function was preferred.

If the putting the failsafe on the mean function is the preferred behavior I will change it to that instead, and include a comment above the mean_f function to describe the behavior. What do you think?

kirkscheper:
When is nan ever good on a flying vehicle? Can you update this?

kirkscheper:
If your drone is falling out of the sky due to a Nan I don't think that it is very useful for debugging. I don't really care what you do to show debugging info to the user, up to you.

sw/airborne/math/pprz_stat.c
```diff
@@ -73,18 +78,22 @@ int32_t variance_i(int32_t *array, uint32_t n_elements)
...
-  // Determine the covariance:
-  uint32_t i;
-  for (i = 0; i < n_elements; i++) {
-    sumX += (float)array1[i];
-    sumY += (float)array2[i];
-    sumXY += (float)(array1[i]) * (float)(array2[i]);
+  if (n_elements == 0) {
+    // Note that something else is wrong if you want the covariance of 0 samples.
+    return 0;
+  } else {
```
kirkscheper: Writing it like this is not good practice as the second return only happens if the else is called but the return uses things that are defined inside the else, not good.

You should have the if with the n_elements then the rest follow without the else and brackets.

sw/airborne/modules/computer_vision/opticflow/size_divergence.c
```diff
@@ -114,8 +114,17 @@ float get_size_divergence(struct flow_t *vectors, int count, int n_samples)
...
   }
 
-  // calculate the mean divergence:
-  float mean_divergence = mean_f(divs, used_samples);
+  float mean_divergence;
+  // calculate the mean divergence or use 0 when used_samples is still 0
+  if(used_samples==0)
+  {
+    // Note that this shouldn't happen
+    mean_divergence = 0.0;
```
kirkscheper: is this necessary?

sw/airborne/math/pprz_stat.c
```diff
@@ -140,16 +154,20 @@ float variance_f(float *array, uint32_t n_elements)
...
-
-  // Determine the covariance:
-  uint32_t i;
-  for (i = 0; i < n_elements; i++) {
-    sumX += arr1[i];
-    sumY += arr2[i];
-    sumXY += arr1[i] * arr2[i];
+  if (n_elements == 0) {
+    // Note that something else is wrong if you want the covariance of 0 samples.
+    return 0.0;
```
kirkscheper: this needs to return a float, 0.0 is a double, use 0.f

TitusBraber:
Thanks @kirkscheper for the feedback and explanation. I implemented all proposed fixes.

sw/airborne/modules/computer_vision/opticflow/size_divergence.c
```diff
@@ -114,8 +114,9 @@ float get_size_divergence(struct flow_t *vectors, int count, int n_samples)
     }
   }
 
-  // calculate the mean divergence:
-  float mean_divergence = mean_f(divs, used_samples);
+  float mean_divergence;
+  // calculate the mean divergence or use 0 when used_samples is still 0
+  mean_divergence = mean_f(divs, used_samples);
```
kirkscheper: why this change?

TitusBraber: It looks like that is the remainder from when I fixed it in size_div. Instead of unfixing it I changed it again. Next commit should restore it to the original state.

PR #2211, opened 2018-01-04
Fix mission fw module 

podhrmic:
- Fix in mission fw module (so it compiles)
- Add test for fixedwing mission fw module
- Better heading for simulation purposes


podhrmic:
Compiles (tested locally), the issue is with hacl, will fix in another PR

PR #2212, opened 2018-01-07
Arming status msg

kevindehecker:
After spending (again #1011 ) an hour or so figuring out why my drone would not take off, only to find that a small accidental trim in the joystick is not allowed by the arming procedure, I thought to finally create a message such that this info can be shown to the user.

The accompanying pprzlink PR: https://github.com/paparazzi/pprzlink/pull/75

Currently only the arming by throttle is fully done: the kill-switch-arming is not implemented, and  the yaw-arming is not tested.

About the yaw arming: the procedure has changed slightly: before this PR centering all sticks was only required during the initialization state. I think it should be required during all arming steps, otherwise it can easily happen that the drone will violently flip at arming. I think it has happened to me often actually. Now that the user is informed about why the drone is not arming, I think that cancels the biggest disadvantage of this extra hassle.
Let me know if this is acceptable, and I will try to test yaw arming ASAP as well.

Question: the throttle arming checks the mode and whether the AP has unarmed, yaw-arming doesn't do that... Does not seem correct...?

In the future I suggest to also not allow arming the drone if the battery is critical or outside the geofence. Any other cases?


sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -74,6 +75,32 @@ static inline void autopilot_arming_set(bool motors_on)
   }
 }
 
+#define YAW_MUST_BE_CENTERED true
```
kirkscheper: AP_ARMIING_YAW_MUST_BE_CENTERED or AP_ARMIING_REQ_YAW_CENTERED?

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
```diff
@@ -64,6 +65,30 @@ static inline void autopilot_arming_set(bool motors_on)
   }
 }
 
+/*Checks all arm requirements and returns false if OK and true otherwise.
+ * Also sets the arming status to provide information to the user
+ */
+static inline bool autopilot_arming_check_arm_requirements(void)
```
kirkscheper: This name isn't really quite logical with the description of the return. Additionally I would suggest reversing the logic, seems more intuitive. Eg. something like autopilot_arming_requirements_met(void)

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -74,6 +75,32 @@ static inline void autopilot_arming_set(bool motors_on)
...
+#define YAW_MUST_BE_PUSHED false
+/*Checks all arm requirements and returns false if OK and true otherwise.
+ * Also sets the arming status to provide information to the user
+ */
```
kirkscheper: include the description of the input here.

sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
```diff
@@ -149,10 +150,10 @@ static void send_fp_min(struct transport_tx *trans, struct link_device *dev)
...
 #endif
   pprz_msg_send_ROTORCRAFT_FP_MIN(trans, dev, AC_ID,
-                              &(stateGetPositionEnu_i()->x),
-                              &(stateGetPositionEnu_i()->y),
-                              &(stateGetPositionEnu_i()->z),
-                              &gspeed);
+                                  &(stateGetPositionEnu_i()->x),
+                                  &(stateGetPositionEnu_i()->y),
+                                  &(stateGetPositionEnu_i()->z),
+                                  &gspeed);
```
kirkscheper: Is this respacing required?

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
```diff
@@ -64,6 +65,30 @@ static inline void autopilot_arming_set(bool motors_on)
...
+{
+  if (!PITCH_STICK_CENTERED()) {
+    autopilot.arming_status = AP_ARMING_STATUS_PITCH_NOT_CENTERED;
```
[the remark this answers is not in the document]
gautierhattenberger: I agree that if we don't need more than 8 flags, that would work.

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
kevindehecker: There is only one correct result (return 0 or false) but there can be multiple errors. (return 1=Error1, 2=Error_something_else, etc). 

So, this is logical, and as far as I know the standard.

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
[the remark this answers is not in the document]
kevindehecker: The use case is sequential. More importantly, with my solution you get to directly see the arming message string in the messages window.

sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
kevindehecker: It's what the autoscript did, so I don't really care. But if I change it back we probably have this issue again in my next PR ;)

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
kevindehecker: Hmmm, you really think this makes things more clear? It's only a local define setting designed so that the arming code becomes a bit more readable (instead of using true and falses with comments each time).

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
gautierhattenberger: When a function returns a value, I think it is fine to have 0 for valid and non-zero for error codes. But since it returns a boolean, I would use true for valid and false if an error occurred.
What about `autopilot_arming_check_valid(void)` with the logic above?

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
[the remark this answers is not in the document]
gautierhattenberger: It seems you have defined a lot a failure cases, more than what can fit in 8 bits flag. So I guess we have to keep it like this for now.

kirkscheper:
I have no prob with this here, I was just saying that since it is Inthe
header that it should be a bit more specific with the define starting with
the file name.

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
kirkscheper: ok, the logic makes sense then but the return conditions then doesn't really follow from the function name (and return, uint8_t instead of bool). I dono just seems weird :p

gautierhattenberger:
Due to an annoying merge sequence in pprzlink, I need to merge this. See https://github.com/paparazzi/paparazzi/commit/2ebd36e08a5250ec1a1e751d83e9446c907754de for the changes according to the remarks.

gautierhattenberger:
closing in favor of #2243

Issue #2210, opened 2018-01-04
Recursivity in the new python paparazzi center


Fabien-B:
At the starting of the new python paparazzi center (in sw/supervision/python/), the program can be stucked in a recursivity loop by the sel call of the `init_hmi_data()` method at line 321.

The original bug seems coming from the `load_sessions_and_programs` in parser.py : when the number of control panel configuration file present is different than 1, none of them is loaded.

For exemple, the program find two files : ['conf/control_panel.xml', conf/userconf/tudelft/control_panel.xml']
What should I do ?
I think I should take the higher file in the hierarchy.
Is that ok to do it?

PR #2213, opened 2018-01-09
Fix in python supervision (#2210)

Fabien-B:
Remove the recursivity in initialisation methods and only search for "conf/control_panel.xml" as control panel configuration file.
If this file is not present, an exception is raised.


Fabien-B:
Close issue #2210

PR #2079, opened 2017-06-16
Optic flow changes

manoskyr:
This pull request tries to make the Lucas-Kanade optical flow algorithm more robust and extentable by keeping track of the same features over time instead of re-detecting every frame. Re-detection will now happen when tracked corners reach a low threshold, and only in the region(s) that have a low amount of corners.
It also fixes the already existing method which detected corners on the new frame and then projected them to the previous one to track them again.
The second commit introduces an FPS parameter to all computer vision modules such that the video listeners that they attach to the camera with the `cv_add_to_device` function can operate at a lower frequency than the camera frame rate. This can be useful in cases like the optical flow where a high frame rate could accumulate error. Note that many of those modules could not be compiled out of the box, but the feature was added so that the call to the `cv_add_to_device` would not be a source of error later on.

PR #2215, opened 2018-01-10
ARDrone vision fix

TitusBraber:
The ARDrone in paparazzi Master was flying at 2-5 FPS, which is insufficient for proper optical flow.
Upon investigating it turns out that this is the case since https://github.com/paparazzi/paparazzi/pull/2079
where the feature management was introduced, later fixed in https://github.com/paparazzi/paparazzi/pull/2087. However the feature management is turned off by default.

Furthermore the pyramid was set two 2 layers standard.

Testing on an ARDrone and Bebop alongsides showed that the Bebop had no issue with the code and was still flying arround 90 FPS(camera maximum) while the ARDrone dropped from 30-60 FPS(camera max) to 2-5 FPS. 

In this pullrequest I set the pyramid back to 0, which improved FPS to about 10-15 FPS. I also tried to turn on feature management, which had the same result. Combining these two changes returned the FPS to 30-60 FPS for an ARDrone.

I changed the setting in all ARDrone conf files that include the vision, to leave Bebop untouched.

### DIFF
--- a/conf/airframes/tudelft/ardrone2_opticflow_indi.xml
+++ b/conf/airframes/tudelft/ardrone2_opticflow_indi.xml
@@ -38,7 +38,7 @@
     <module name="cv_opticflow">
       <define name="OPTICFLOW_CAMERA" value="bottom_camera"/>
       <!-- ARDrone2 FPS improvements -->
-	  <define name="OPTICFLOW_PYRAMID_LEVEL" value="0"/>
+      <define name="OPTICFLOW_PYRAMID_LEVEL" value="0"/>
       <define name="OPTICFLOW_FEATURE_MANAGEMENT" value="1"/>
     </module>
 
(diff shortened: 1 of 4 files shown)


conf/airframes/tudelft/ardrone2_opticflow_indi.xml
```diff
@@ -37,6 +37,9 @@
     <module name="pose_history"/>
     <module name="cv_opticflow">
       <define name="OPTICFLOW_CAMERA" value="bottom_camera"/>
+      <!-- ARDrone2 FPS improvements -->
+	  <define name="OPTICFLOW_PYRAMID_LEVEL" value="0"/>
```
kirkscheper: Remove extra tab here in all updated files.

TitusBraber: Strange, this is what I see in eclipse, but on github there is an extra tab. I will run the fix code style to try and fix it!
screenshot from 2018-01-16 12-02-59

TitusBraber: Nevermind, fix code style is not for xml *facepalm*

kirkscheper:
Is this flight tested?

TitusBraber:
Yes, I wanted to know whether it was just a number increasing or actually returning performance. Luckily it added the performance back to normal levels.
Only tested on an ARDrone2, because it was only changed on an ARDrone2. Bebop1 quality didn't degrade.

PR #2218, opened 2018-01-19
Fix bebop_orangeavoid_course2017 defines

tomvand:
Fixed the MT9F002 defines in the bebop airframe in preparation for the 2018 MAVLab course. Set video streaming back to broadcast to simplify streaming. Test-flown and tagged accordingly in the conf file.


dewagter:
Thanks, Tom.

PR #2219, opened 2018-01-23
[gazebo] Add actuator dynamics and bebop model

tomvand:
New:
- Added actuator dynamics (low-pass filter) and spin-up torques to gazebo fdm
- Added a simple bebop model and two example airframe files (not added to conf since they should be drop-in replacements for the existing jsbsim airframes). The dynamics of the model should be more-or-less correct, the camera will be tweaked in a later PR.

Fixed:
- Actuator torques now get their value from NPS_ACTUATOR_TORQUES instead of _THRUSTS
- Updated documentation


conf/airframes/examples/bebop2_indi_gazebo.xml
```diff
@@ -0,0 +1,217 @@
...
+    <module name="gps" type="ublox"/>
+    <module name="stabilization" type="indi_simple"/>
+    <module name="ahrs" type="int_cmpl_quat">
+      <configure name="USE_MAGNETOMETER" value="TRUE"/>
+      <define name="AHRS_USE_GPS_HEADING" value="FALSE"/>
+    </module>
+    <module name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
```
gautierhattenberger: Can you move the modules from this node to the `firmware` node ? This section is supposed to be removed in future version.

tomvand: Sure, fixed.

conf/airframes/examples/bebop2_indi_gazebo.xml
```diff
@@ -0,0 +1,215 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="bebop2_indi">
```
kirkscheper: Is it really necessary to have a separate airframe just for gazebo? I still think that this is a dangerous trend, we will end up with duplicate files for all airframes.

conf/airframes/examples/bebop_gazebo.xml
```diff
@@ -0,0 +1,216 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="bebop">
+
+  <firmware name="rotorcraft">
```
kirkscheper: See what I mean...

conf/airframes/examples/bebop_gazebo.xml
```diff
@@ -0,0 +1,216 @@
...
+    <define  name="CLIMB_VSPEED" value="2.5" />
+    <define  name="DESCEND_VSPEED" value="-1.0" />
+  </section>
+
+  <section name="SIMULATOR" prefix="NPS_">
+    <define name="ACTUATOR_NAMES" value="nw_motor, ne_motor, se_motor, sw_motor" type="string[]"/>
+    <define name="ACTUATOR_THRUSTS" value="2.09, 2.09, 2.09, 2.09" type="float[]"/>
+    <define name="ACTUATOR_TORQUES" value="0.0250, 0.0250, 0.0250, 0.0250" type="float[]"/>
+    <define name="ACTUATOR_TIME_CONSTANTS" value="0.02, 0.02, 0.02, 0.02" type="float[]"/>
+    <define name="ACTUATOR_MAX_ANGULAR_MOMENTUM" value="0.19, 0.19, 0.19, 0.19" type="float[]"/>
```
kirkscheper: Can't these be added to the sdf model instead. I would image that they are vehicle specific and not airframe specific. Not sure if you want these to be easily configurable.

Alternatively, they could be settings in a header file stored with the simulation model which will be included in the gazebo file.

sw/airborne/filters/high_pass_filter.h
```diff
@@ -30,6 +30,60 @@
...
+static inline void init_first_order_high_pass(struct FirstOrderHighPass *filter,
+    float tau, float sample_time, float value)
+{
+  filter->last_in = value;
+  filter->last_out = value;
+  filter->time_const = 2.0f * tau / sample_time;
+}
+
+/** Update first order high pass filter state with a new value.
+ *
```
kirkscheper: @param descriptors?

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -102,10 +104,13 @@ struct gazebo_actuators_t {
   string names[NPS_COMMANDS_NB];
   double thrusts[NPS_COMMANDS_NB];
   double torques[NPS_COMMANDS_NB];
+  struct FirstOrderLowPass lowpass[NPS_COMMANDS_NB];
+  struct FirstOrderHighPass highpass[NPS_COMMANDS_NB];
+  float max_ang_momentum[NPS_COMMANDS_NB];
```
kirkscheper: float or double?

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -375,6 +396,7 @@ static void init_gazebo(void)
 
   for (uint8_t i = 0; i < NPS_COMMANDS_NB; i++) {
     gazebo_actuators.torques[i] = -fabs(gazebo_actuators.torques[i]) * yaw_coef[i] / fabs(yaw_coef[i]);
+    gazebo_actuators.max_ang_momentum[i] = -fabs(gazebo_actuators.max_ang_momentum[i]) * yaw_coef[i] / fabs(yaw_coef[i]);
```
kirkscheper: if float you need fabsf

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -536,10 +558,30 @@ static void gazebo_read(void)
...
+#endif
+    double thrust = gazebo_actuators.thrusts[i] * u;
+    double torque = gazebo_actuators.torques[i] * u;
+
+#ifdef NPS_ACTUATOR_MAX_ANGULAR_MOMENTUM
+    // Spinup torque
+    double udot = update_first_order_high_pass(&gazebo_actuators.highpass[i],
+        sp);
+    double spinup_torque = gazebo_actuators.max_ang_momentum[i] /
+    (2 * sqrtf(u > 0.05 ? u : 0.05)) * udot;
```
kirkscheper: indentation

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -536,10 +558,30 @@ static void gazebo_read(void)
...
+#else
+    double u = sp;
+#endif
+    double thrust = gazebo_actuators.thrusts[i] * u;
+    double torque = gazebo_actuators.torques[i] * u;
+
+#ifdef NPS_ACTUATOR_MAX_ANGULAR_MOMENTUM
+    // Spinup torque
+    double udot = update_first_order_high_pass(&gazebo_actuators.highpass[i],
+        sp);
```
kirkscheper: This sp can fit on the line above :p

conf/airframes/examples/bebop2_indi_gazebo.xml
tomvand: Nope. The gazebo airframes were provided only as an example, so it's fine by me to copy the necessary defines back to the original airframes, delete the gazebo ones and let the user change the fdm as you suggest.

conf/airframes/examples/bebop_gazebo.xml
tomvand: Agreed, the sdf might be a better place for this. Since the sdf is already read at the start of the simulation, I'll see if I can get these values from there instead

sw/simulator/nps/nps_fdm_gazebo.cpp
tomvand: Try telling eclipse that :P

conf/airframes/examples/bebop_gazebo.xml
tomvand: SDF does not accept elements that are not defined in the spec, instead it shows a warning and the element is not parsed.

Instead, I have now moved the gazebo defines to xml files in conf/simulator/gazebo/airframes that should be included from the main airframe file. Let me know what you think.

tomvand:
Minor issues are fixed now. Moved the gazebo-specific defines to shared .xml files in conf/simulator/gazebo/airframes. I've removed the new _gazebo aiframes but left the existing ardrone2_gazebo.xml and ardrone2_gazebo_cyberzoo.xml in place for now, let me know if these should be removed as well.

conf/airframes/examples/bebop2_indi.xml
```diff
@@ -140,7 +146,7 @@
     <define name="G1_R" value="0.0022"/>
 
     <!-- For the bebop2 we need to filter the roll rate due to the dampers -->
-    <define name="FILTER_ROLL_RATE" value="TRUE"/>
+    <!-- FILTER_ROLL_RATE moved to firmware targets -->
```
kirkscheper: Why has this been moved to nps target only? Not sure if that's a good idea. Why does this have to be turned off for Gazebo?

conf/simulator/gazebo/airframes/bebop.xml
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE airframe SYSTEM "../../../airframes/airframe.dtd">
+
+<airframe>
+  <section name="SIMULATOR" prefix="NPS_">
+    <define name="ACTUATOR_THRUSTS" value="2.09, 2.09, 2.09, 2.09" type="float[]"/>
+    <define name="ACTUATOR_TORQUES" value="0.0250, 0.0250, 0.0250, 0.0250" type="float[]"/>
+    <define name="ACTUATOR_TIME_CONSTANTS" value="0.02, 0.02, 0.02, 0.02" type="float[]"/>
+    <define name="ACTUATOR_MAX_ANGULAR_MOMENTUM" value="0.19, 0.19, 0.19, 0.19" type="float[]"/>
+    <define name="GAZEBO_AC_NAME" value="bebop" type="string"/>
+    <define name="BYPASS_AHRS" value="1"/>
```
kirkscheper: is it required to bypass the ahrs? If not I would leave this out of the default settings.

conf/simulator/gazebo/airframes/bebop.xml
```diff
@@ -0,0 +1,12 @@
...
+  </section>
+</airframe>
```
kirkscheper: you are missing some line endings in these files. Add a new line a the end of the file. You can also set this in Eclipse to be done automatically.

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -536,10 +558,29 @@ static void gazebo_read(void)
...
+    double u = sp;
+#endif
+    double thrust = gazebo_actuators.thrusts[i] * u;
+    double torque = gazebo_actuators.torques[i] * u;
+
+#ifdef NPS_ACTUATOR_MAX_ANGULAR_MOMENTUM
+    // Spinup torque
+    double udot = update_first_order_high_pass(&gazebo_actuators.highpass[i], sp);
+    double spinup_torque = gazebo_actuators.max_ang_momentum[i] /
+        (2 * sqrtf(u > 0.05 ? u : 0.05)) * udot;
```
kirkscheper: should use sqrt for doubles, not sqrtf. Also can you augment 2 * to 2. * (saves a background conversion)

sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -102,10 +104,13 @@ struct gazebo_actuators_t {
...
   double torques[NPS_COMMANDS_NB];
+  struct FirstOrderLowPass lowpass[NPS_COMMANDS_NB];
+  struct FirstOrderHighPass highpass[NPS_COMMANDS_NB];
+  double max_ang_momentum[NPS_COMMANDS_NB];
 };
 
-struct gazebo_actuators_t gazebo_actuators = {NPS_ACTUATOR_NAMES, NPS_ACTUATOR_THRUSTS, NPS_ACTUATOR_THRUSTS};
-
+struct gazebo_actuators_t gazebo_actuators = { NPS_ACTUATOR_NAMES,
+    NPS_ACTUATOR_THRUSTS, NPS_ACTUATOR_TORQUES, { }, { }, { } };
```
kirkscheper: Can you write { 0 } here to explicitly initialise the arrays? I guess it already works like this but since you only set these values inside an ifdef I think it would be better to explicitly show the default values here.

tomvand: No, this causes a warning that not all fields of the filters are initialized.

conf/simulator/gazebo/airframes/bebop.xml
tomvand: Yes, AHRS is not supported yet. I tried removing the bypass, but this results in an incorrect heading estimate.

conf/airframes/examples/bebop2_indi.xml
tomvand: Setting FILTER_ROLL_RATE to TRUE causes oscillations around the roll axis. I thought this was a gazebo problem, but I've just found out this also happens in jsbsim so it might actually be a problem with this filter instead.

Could you please try to replicate this as follows:
1. Start a flight with the examples/bebop2_indi airframe and rotorcraft_basic flight plan (from master). The flight will be simulated using jsbsim, not gazebo.
2. Follow the flight plan until the standby block.
3. Check the PFD in GCS, does the drone oscillate around the roll axis?
4. Disable the ROLL_FILTER, does the drone remain stable now?

kirkscheper: @EwoudSmeur I guess you implemented the jbsim indi stuff, can you check this?

conf/simulator/gazebo/airframes/bebop.xml
kirkscheper: Ok, just make sure to update this once it is working.

conf/airframes/examples/bebop2_indi.xml
tomvand: Discussed with @EwoudSmeur , enabling the roll filter could indeed cause these oscillations. Use of the filter on the real bebop2 is also unnecessary when the dampers are locked (which I've been told is usually the case). How about setting the roll rate filter to off by default, with a comment hinting at its use when flying with unlocked dampers (see latest commit)?

Issue #2140, opened 2017-10-25
Deprecated Warnings for Ocaml ≥ 4.04


rijesha:
Ubuntu 17.10 ships with ocaml v4.04.0

Strings.capitalize has become String.capitalize_ascii
Strings.lowercase has become String.lowercase_ascii

deprecated: Scanf.fscanf
Use Scanning.from_channel then Scanf.bscanf.

Warning 52: Code should not depend on the actual values of
this constructor's arguments. They are only for information
and may change in future versions. (See manual section 8.5)

A similar issue for ocaml ≥ 4.02
https://github.com/paparazzi/paparazzi/issues/1703

rijesha:
@gautierhattenberger  Do you have a preference on how to fix this? I can do the same thing that I did last time with 4.02. The compat module will get a lot more complex.

gautierhattenberger:
I don't have a better option for `String` for now. For `Scanf` it depends if the alternate option is supported by other 4.0x versions or not.

BTW, do you compile with ocaml from Ubuntu packages or from opam ? It seems that system packages have an issue with ocamlbuild. Do you have that as well ?

rijesha:
Very late reply, but recently I have not had any issues with system or opam versions of ocaml

PR #2221, opened 2018-01-25
Ocaml 405 fix

rijesha:
Various ocaml fixes for ocaml build warnings.

Warning 52 was fixed by running the exception for all the possible cases. Most of these were for fairly trivial errors ie:
Failure ("int_of_string")  changed to Failure _

Code was tested with ocaml 4.01, 4.02, 4.04 and 4.05.

See issue #2140

### DIFF
--- a/sw/ground_segment/cockpit/page_settings.ml
+++ b/sw/ground_segment/cockpit/page_settings.ml
@@ -55,7 +55,7 @@ object
           let v = float_of_string s in
           last_known_value <- Some v;
           set_default v
-        with Failure "float_of_string" -> ()
+        with Failure _ -> ()
       end
 end
 
--- a/sw/ground_segment/cockpit/page_settings.ml
+++ b/sw/ground_segment/cockpit/page_settings.ml
@@ -249,7 +249,7 @@ let one_setting = fun (i:int) (do_change:int -> float -> unit) ac_id packing dl_
 
   (** Insert the related buttons in the strip and prepare the papgets DnD *)
   List.iter (fun x ->
-    match Compat.bytes_lowercase (Xml.tag x) with
+    match Compat.lowercase_ascii (Xml.tag x) with
         "strip_button" ->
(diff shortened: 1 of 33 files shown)


sw/lib/ocaml/compat.ml
```diff
@@ -75,3 +69,39 @@ let bytes_set = fun s n c->
 
 let bytes_iter = fun f s->
   BYTES.iter f s
+
+IFDEF OCAML_V404 THEN
```
gautierhattenberger: Since uppercase/lowercase/capitalize functions from Bytes and String modules seems to have the same effect, I think we can make something simpler like:
```
IFDEF OCAML_V404 THEN
let lowercase_ascii = BYTES.lowercase_ascii
let uppercase_ascii = BYTES.uppercase_ascii
let capitalize_ascii = BYTES.capitalize_ascii
ELSE
let lowercase_ascii = BYTES.lowercase
let uppercase_ascii = BYTES.uppercase
let capitalize_ascii = BYTES.capitalize
END
```
This way the functions from `Compat` will match the new standards and `BYTES` will pick the correct module.

sw/ground_segment/cockpit/page_settings.ml
```diff
@@ -55,7 +55,7 @@ object
           let v = float_of_string s in
           last_known_value <- Some v;
           set_default v
-        with Failure "float_of_string" -> ()
+        with Failure _ -> ()
```
gautierhattenberger: why did you remove all these filtering ?

gautierhattenberger:
I merged the corresponding pprzlink PR, please update for the correct revision

rijesha:
I think this branch is already set for the latest pprzlink commit.

gautierhattenberger:
since I did a merge and rebase, latest commit is now https://github.com/paparazzi/pprzlink/commit/25b7024ad46964b4c99c998b75eca6db34ceab5b

PR #2226, opened 2018-02-02
Command filter for vibrating airframes

matejkarasek:
Added optional low pass filtering of the commands, essential for vibrating frames such as flapping wing MAVs.


podhrmic:
@matejkarasek I am not sure that filtering the _output_of the control system is the best way to tackle the vibration issue. Is your IMU physically dampened? Are the IMU filters set up properly? Have you tried lowering the control system frequency?

gautierhattenberger:
You should not change versions of submodules. And indentation should be fixed at some places, in particular in xml files.

matejkarasek:
@podhrmic We need this for flapping wing platforms, where the vibration is quite bad. Yes, the autopilot/IMU is on a vibration isolator (otherwise the sensors saturate). We could do some filtering on the sensor level, but we also want the unfiltered signals for our research (in-flight force estimation). Plus we want to know if the sensor saturates, or not.

We decided to filter the control system output, as this implementation can be used with any stabilization law, no matter what signals are used in the control law and it does not affect AHRS (we also want to log unfiltered attitude estimates). Plus, we only filter once, which is also important for the autopilots with low computational resources (Lisa/S).

@gautierhattenberger The style, as well as the submodules, should have been fixed now

conf/airframes/tudelft/delfly_lisas.xml
```diff
@@ -0,0 +1,257 @@
...
+
+  <modules main_freq="512">
+    <module name="send_imu_mag_current"/>
+    <!-- When sd logger is used, then radio should be DelTang (PPM) and telemetry should
+      be transparent (cable/WiFi) -->
+    <module name="logger_sd_spi_direct.xml">
+      <define name="SDLOGGER_ON_ARM" value="TRUE"/>
+      <!--define name="LOGGER_LED" value="3"/-->
+    </module>
+  </modules>
```
kirkscheper: We are moving away from using a separate modules section, can you move to contents of this to the firmware section above?

conf/radios/delfly_Rx31_DEVO10.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<!DOCTYPE radio SYSTEM "radio.dtd">
+<radio name="FlyElectric Rx31" data_min="900" data_max="2050" sync_min ="5000" sync_max ="15000" pulse_type="NEGATIVE">
+	<channel ctl="A" function="THROTTLE" min="1110" neutral="1110" max="1900" average="0"/>
+	<channel ctl="D" function="ROLL"     max="1110" neutral="1500" min="1900" average="0"/>
+	<channel ctl="C" function="PITCH"    max="1110" neutral="1500" min="1900" average="0"/>
+	<channel ctl="B" function="GEAR"     max="1110" neutral="1500" min="1900" average="0"/>
+	<channel ctl="G" function="FLAP"     min="1110" neutral="1500" max="1900" average="1"/>
+	<channel ctl="E" function="MODE"     min="1110" neutral="1500" max="1900" average="1"/>
+	<channel ctl="F" function="ELEV_DR"  min="1110" neutral="1500" max="1900" average="1"/>
```
kirkscheper: Indent

PR #2229, opened 2018-02-05
Jevois integration

gautierhattenberger:
Add a module to parse standardized messages from a JeVois smart camera (http://jevois.org).
It is also possible to change the running algorithm (mapping) and start/stop streaming.

A demo module is using the data to track an object or a face by controlling the heading of a rotorcraft. This demo could be improved in the future to replace the very basic control / heading estimation by a proper PID controller and/or target position estimator.

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -124,6 +124,33 @@
       <field name="vel_body_z_FF" type="float" unit="m/s"/>
     </message>
 
+    <message name="JEVOIS_MSG" id="21">
+      <!--description>
+        Standardized serial messages sent by the Jevois smart camera
+        http://jevois.org/doc/UserSerialStyle.html
+      </description-->
+      <field name="type" type="uint8_t">Standardized message type JEVOIS_MSG_[T1|N1|D1|T2|N2|D2|F2|T3|N3|D3|F3]</field>
+      <field name="id" type="char *">Text string describing the reported object</field>
+      <field name="nb" type="uint8_t">Number of elements in the coordinates array</field>
+      <field name="coord" type="int16_t *">List of coordinates corresponding to 1D, 2D or 3D messages</field>
+      <field name="dim" type="uint16_t *">1, 2 or 3D dimension</field>
+      <field name="quat" type="struct FloatQuat">Quaternion that relates the object's frame to the camera's frame, if appropriate</field>
+      <field name="extra" type="char *">Additional text information</field>
+    </message>
+
+    <message name="FOLLOW_TARGET" id="22">
+      <field name="id" type="uint32_t">Object ID to track</field>
+      <field name="frame" type="uint8_t">
+        bitmask to select relative coordinates (0) or global coordinates (1)
+        bit 0: bearing
+        bit 1: height
+        bit 2: distance
+      </field>
+      <field name="bearing" type="float" unit="rad">Direction of object to track</field>
+      <field name="height" type="float" unit="m">Height of object to track</field>
+      <field name="distance" type="float" unit="m">Distance of object to track</field>
+    </message>
+
   </msg_class>
 
 </protocol>
--- a/sw/airborne/modules/ctrl/object_tracking.c
+++ b/sw/airborne/modules/ctrl/object_tracking.c
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2018 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ *
+ * This file is part of paparazzi
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file "modules/ctrl/object_tracking.c"
+ * @author Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ * Control a rotorcraft heading to track an object detected by a camera
+ */
+
+#include "modules/ctrl/object_tracking.h"
+
+#include "firmwares/rotorcraft/navigation.h"
+#include "subsystems/abi.h"
+#include "generated/airframe.h"
+#include "generated/modules.h"
+
+// ABI message binding ID
+#ifndef OBJECT_TRACKING_ID
+#define OBJECT_TRACKING_ID ABI_BROADCAST
+#endif
+
+// Timout in seconds before entering search mode
+#ifndef OBJECT_TRACKING_TIMEOUT
+#define OBJECT_TRACKING_TIMEOUT 3.0f
+#endif
+
+// Turn rate in tracking mode (rad/s)
+#ifndef OBJECT_TRACKING_RATE
+#define OBJECT_TRACKING_RATE RadOfDeg(10)
+#endif
+
+// Turn rate in search mode (rad/s)
+#ifndef OBJECT_TRACKING_SEARCH_RATE
+#define OBJECT_TRACKING_SEARCH_RATE RadOfDeg(20)
+#endif
+
+// Send debug message
+#ifndef OBJECT_TRACKING_DEBUG
+#define OBJECT_TRACKING_DEBUG FALSE
+#endif
+
+#if OBJECT_TRACKING_DEBUG
+#include "subsystems/datalink/downlink.h"
+#include "pprzlink/messages.h"
+#include "mcu_periph/uart.h"
+#endif
+
+float object_tracking_rate; // in rad/s
+float object_tracking_search_rate; // in rad/s
+
+static uint8_t object_frame; // relative or global coordinates
+static float object_bearing;
+static float object_height;
+static float timeout;
+
+abi_event object_ev;
+
+static const float nav_dt = 1.f / NAV_FREQ;
+
+// callback on follow target message
+static void get_object(uint8_t sender_id __attribute__((unused)),
+    uint32_t id __attribute__((unused)),
+    uint8_t frame, float bearing, float height,
+    float distance __attribute__((unused)))
+{
+  object_frame = frame;
+  object_bearing = bearing;
+  object_height = height;
+  timeout = 0.f;
+}
+
+void object_tracking_init(void)
+{
+  object_tracking_search_rate = OBJECT_TRACKING_SEARCH_RATE;
+  object_tracking_rate = OBJECT_TRACKING_RATE;
+
+  object_frame = 0; // relative coordinates
+  object_bearing = 0.f;
+  object_height = 0.f;
+  timeout = OBJECT_TRACKING_TIMEOUT; // start in search mode
+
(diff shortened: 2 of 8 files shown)


kirkscheper:
A couple of things here:

- I am not sure the jevois stuff should go into the computer_vision section as it has to do more with sensor interfacing than any computer vision computation. Recommend moving it to the modules/sensor folder.
- As it is, the face_tracker depends on the jevois but I am not sure why. It looks like in principle it could work with any method that outputs an image coordinate. To maximize functionality, you could change this up that instead of a face tracker you have a image_reference_tracker (name obviously optional). Move the get_face function to the jevois.c and change the abi to a image_tracking_reference (with angular offsets instead of pixel offsets). That way this tracker could be used with different reference generators out of the box.
- nav_heading isn't bound in face_tracking_run()

noether:
I agree with Kirk. It seems that jevois.{c,h} are only focused on interfacing hardware, i.e., a driver. Maybe the computer_vision section is not the place for this, but modules/sensor/cameras ?

gautierhattenberger:
I agree that `computer_vision` is not nice. I would prefer `sensors/cameras` as suggested by Hector, `sensors` is too crowded.

The purpose of the `jevois` is to be agnostic about what the camera is sending. So I wouldn't mode the `get_face` there. Since the format and information in JeVois normalized message is pretty specific, it is not so easy to make it really generic. So options could be:
- provide several ABI message, one specific to JeVois format, and a second one more generic that could be reused with other cameras
- bind the tracking module to any camera specific format that could come up in the future and who wants to use this

If possible, I would like to avoid choosing for now, as none are satisfying to me. We only have one smart cam, and it is just a demo that needs improvements. If we ever face situation to add a new generic ABI message for that, it will be easier at that moment rather than trying to predict it now.

I can still make this tracker a bit more generic by:
- changing the name for `vision_ref_tracker` (any better idea?)
- passing the `nav_heading` as argument so it is not directly related to the module (we don't have a nice command API, so options are limited there)
- changing `get_face` by `get_ref`

What do you think ?

kirkscheper:
I think you misunderstood what I meant about the get_face. What I meant is simply to add a msg decoder to the jevois.c. Basically combining the driver and message decoder like with stereocam.

The parser will fetch the data from the stream and generate the message (as is done now in the jevois_parse) and the message parser will do something with that message (what is currently in the get_face). Of course you can always broadcast the jevois message over abi and have a the message parser in a different place but it would seem strange to split it. As you said the normalized message is quite specific and I don't think any other function would need this raw 'unprocessed' message. 

(alternatively you could also do both, like with the datalink parsers, they decode some messages and then forward the data buffer to other modules who might also want it. There is make more sense than here though as those messages all use the same shared pprzlink structure to read the message and messages are auto generated.)

The abi message then would not have to be specific to the jevois at all but rather just the information needed for the tracker.

gautierhattenberger:
Actually, I got you I think. But what you suggest is that the jevois driver should know the content of the message to process it, and not only the "raw" decoded frame. This is what I didn't want. Because content can be anything.
My idea is that the camera and the "end" module know what they are suppose to do, but the driver in the middle don't care about the content. The obvious drawback is that modules using this information are kind of specifically written for jevois.
An alternative I suggested above is that the driver sends both JEVOIS ABI message and eventually a more generic message. But I wouldn't do that too early, since only one module is using this for now.

kirkscheper:
Seems quite strange that the messages are specific to the jevois but you don't decode them there... I understand that the message can contain anything but the messages are still encased in the jevois structure so not useful to any other function out of the box, you will still have to write a decoder, why not have them all in one place.

I refer you again to the stereocam, there we can also have any message. We just extend the message parser when we invent a new message.

If you want it that people don't edit the file after the parser is made you can just add a second c file called jevois_msg_parser.c where you can add the parser. But decoding the message in the tracker is quite strange. I don't see the added value there just unnecessary dependencies.

gautierhattenberger:
`stereocam` have the benefit to use pprzlink messages, so with a message ID. Here, we will need to choose at compile time how to interpret the message. Any better idea ?

About providing implementation of some parsing functions like the follow_me function in stereocam modules, I would actually choose to use ABI. In this specific case, a FOLLOW_ME ABI message could be used by both modules (even if I'm only controlling heading in my case).

Maybe I can come up with something in that direction. Still, some stuff are not completely satisfying, like if we want a better target tracker (not yet speaking about control), where should it be ? What would be the input ? What should be the units in ABI messages (SI, normalized, unscaled, whatever) ?

kirkscheper:
I thought the the jevois has a type and an id to describe the received message, can't you use that (not exactly sure what is coming out of the camera of course but 'id' look like it is configurable it just won't be autogenerated)?

For the units, when sending messages in the camera frame, convert the pixel coordinates to angles from the center of the frame (so scale by the fov). Same with pixel dimensions (like height or width of a visual feature) That is the most portable to different camera resolutions ect.

I think all functions that use images should be in the computer vision folder, otherwise they can be somewhere else...

gautierhattenberger:
Would have been nice, but JeVois messages are not exactly working like this. In particular, depending of the message type, you may not have the ID field, and if you have it, it will be something dependent of the algorithm running behind. Like the "value" encoded in a ArUco tag. But usually it is not set by the user.
Also, their "basic" demo algorithms (like tag or color detection) are sending data between -/+1000. So neither pixel nor angles. Of course you can always convert to angles (that's currently done in my control module).

sw/airborne/modules/ctrl/object_tracking.c
```diff
@@ -0,0 +1,161 @@
...
+    float msg[] = {
+      rate,
+      object_lat,
+      object_vert,
+      timeout
+    };
+    DOWNLINK_SEND_PAYLOAD_FLOAT(DefaultChannel, DefaultDevice, 4, msg);
+#endif
+  }
+  nav_heading += ANGLE_BFP_OF_REAL(rate * nav_dt);
```
kirkscheper: I still think yo should bound this in this function. Add INT32_COURSE_NORMALIZE(nav_heading) at the end of the function.

Another thing is I am not sure it is a great idea to increment the nav_heading set-point using a fixed rate when getting the target from a camera pointing in the real vehicle heading. Imagine the case when the vehicle can't track the heading set-point defined by nav_heading, then nav_heading will increment faster than the vehicle and probably lead to overshooting the target. It would seem safer to use the true heading as a reference, something like: 

`nav_set_heading_deg(stateGetNedToBodyEulers_f()->psi + rate * nav_dt)`

Alternatively, you can limit the maximum angle diff between nav_heading and the current heading to make sure it doesn't run away, similar to stabilization with STABILIZATION_ATTITUDE_SP_PSI_DELTA_LIMIT. You could even use the object_tracking_deadband or (object_tracking_deadband / 2) to limit it, then the overshoot should be limited to within the deadband.

sw/airborne/modules/ctrl/object_tracking.c
```diff
@@ -0,0 +1,161 @@
...
+#else // use calculated heading
+
+  if (timeout < OBJECT_TRACKING_TIMEOUT) {
+    timeout += nav_dt;
+    // compute expected heading
+    float target_heading = stateGetNedToBodyEulers_f()->psi + object_lat;
+    float diff = target_heading - ANGLE_FLOAT_OF_BFP(nav_heading);
+    FLOAT_ANGLE_NORMALIZE(diff);
+    BoundAbs(diff, object_tracking_rate * nav_dt)
+    nav_heading += ANGLE_BFP_OF_REAL(diff);
```
kirkscheper: Note that using the difference between the target_heading and the nav_heading rather than the actual vehicle heading will limit the set-point rate not the real rate of the vehicle. I don't think this particularly a problem here, basically reference generator for the set-point. My guess is that was intentional but just wanted to mention it here for future reference.

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -362,4 +362,13 @@
 #define RANGE_FORCEFIELD_ID 1
 #endif
 
+/*
+ * ID's for computer vision systems
```
kirkscheper: Not sure if this an appropriate category given the other abi naming. There are already some other computer vision abi messages here but are named based on their message type not the sensor. I have no problem with naming the sensor rather than the use but then the other ids should be updated. You can also just make an issue of it to be updated later.

kirkscheper:
The angles for communication I meant were for within paparazzi, so the abi messages should be in angles. The camera driver should scale whatever the inputs are to angles rather than pixel locations. Now the axis definition and orientation is a different discussion (see below).

conf/abi.xml
```diff
@@ -124,6 +124,28 @@
...
+      <field name="dim" type="uint16_t *">1, 2 or 3D dimension</field>
+      <field name="quat" type="struct FloatQuat">Quaternion that relates the object's frame to the camera's frame, if appropriate</field>
+      <field name="extra" type="char *">Additional text information</field>
+    </message>
+
+    <message name="FOLLOW_ME" id="22">
+      <field name="id" type="uint32_t">Object ID to track</field>
+      <field name="frame" type="uint8_t">0: relative to camera frame, 1: global coordinates</field>
+      <field name="heading" type="float" unit="rad">Direction of object to track</field>
+      <field name="height" type="float" unit="m">Height of object to track</field>
```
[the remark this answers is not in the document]
gautierhattenberger: Since I'm not a computer vision guy at all, can you explain me the difference between heading, bearing, azimuth, etc in this field ?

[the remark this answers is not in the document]
gautierhattenberger: assuming it as the same meaning than in aeronautics, I changed to bearing, and the frame is now a bitmask telling for each value if it is relative (hence the control should track 0 to follow the target) or in absolute coordinates.

sw/airborne/modules/ctrl/object_tracking.c
```diff
@@ -0,0 +1,161 @@
...
+#include "mcu_periph/uart.h"
+#endif
+
+int16_t object_tracking_deadband;
+float object_tracking_rate;
+float object_tracking_search_rate;
+
+float object_lat;
+float object_vert;
+float timeout;
```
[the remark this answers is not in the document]
gautierhattenberger: see previous comment about frames
camera frame is never really used in the end

sw/airborne/modules/ctrl/object_tracking.c
gautierhattenberger: I removed the rate mode for now, it wasn't performing well (a real PID might help, but I don't have the time to do it now)

sw/airborne/modules/ctrl/object_tracking.c
gautierhattenberger: it was intentional, it is the user that should set this rate limit compatible with rotorcraft capabilities

sw/airborne/subsystems/abi_sender_ids.h
gautierhattenberger: changed it to `CAM_JEVOIS_ID`

gautierhattenberger:
tested and still working

conf/abi.xml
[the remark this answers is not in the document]
kirkscheper: Sorry, got distracted by some other work. 

Bearing is relative to the current vehicle heading (so relative to the x axis in the x-y plane. Azimuth is more optional, its an angle I typically use to describe the relative y angle from the horizon.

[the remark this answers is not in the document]
gautierhattenberger: I changed heading for bearing, but kept height. Is it good enough for now ? I don't even use it yet, so might be changed in the future if an angle is preferred to altitude.

gautierhattenberger:
Based on this https://i.stack.imgur.com/3Xqzo.jpg, I assumed that you could consider a relative and global (or magnetic) bearing for the target. I can change height for azimuth as well.

kirkscheper:
yea yea, good enough!

PR #2230, opened 2018-02-06
[gazebo] Cyberzoo model

tomvand:
New TU Delft Cyberzoo world and props for Gazebo:

cyberzoo_preview1
cyberzoo_preview2
cyberzoo_all_props
cyberzoo_preview_orangeavoid


kirkscheper:
oew... not sure its a good idea to include large gazebo models here... It should probably be downloaded during the make or included in a submodule which can be optionally checked out.

@gautierhattenberger advice here?

tomvand:
FYI, the gazebo folder is now 22.8MB

podhrmic:
Looking nice! What about `git lfs`? That would be a good place to store things. Or a separate repo as @kirkscheper suggested (perhaps similar to https://github.com/AggieAir/Models) and some script that downloads this on demand...

noether:
Wow, it looks very realistic to the real one xD

gautierhattenberger:
I also think it shouldn't be in the main repository. So either an external one or git lfs (I didn't know before). We could also use a variable env to specify a non standard path for Gazebo models. Otherwise we may end up with a lot of submodules.

tomvand:
Ok, so by now there are 4 different options to include these (and future) models:
1. **Add as submodule**. Advantages: all models are downloaded automatically and there is a form of version control that keeps the right versions of fdm_gazebo, airframes, flight plans and models together. Disadvantages: the submodules might get cluttered unless all models are stored in a single repo (e.g. paparazzi/gazebo_models). Alternatively, this can partially be solved by placing the gazebo submodules in a different directory (conf/simulator/gazebo, sw/ext/gazebo_models/...), although this will not prevent .gitmodules from growing. All models get downloaded, which will increase disk usage of paparazzi (optional submodules might be possible by setting update=none and ignore=all in .gitmodules, but I've never tried this...). [Shallow clones for submodules](https://stackoverflow.com/a/38895397) should be used to prevent downloading earlier revisions as well.
2. **Git LFS**. Avoids the need to make a separate repo and submodules, while still providing version control. Requires users to install git lfs if they want to use the gazebo models, but once installed they should be easy to check out. As with the submodules all models will be downloaded, but not their history. [Github provides 1 GB of storage and 1 GB/month of bandwidth for LFS](https://help.github.com/articles/about-storage-and-bandwidth-usage/), which allows only 45 checkouts per month @22MB before the bandwidth is reached.
3. **Download manually (or simple script)**. This option gives the most control over which models get downloaded. Once downloaded, the models can be placed outside paparazzi with an env variable pointing towards them, or perhaps in var or a gitignored folder under conf/simulator/gazebo. However, this solution can result in missing dependencies (e.g. the course2017_avoid_orange_cyberzoo.xml flightplan refers to the cyberzoo world that might not be installed). Version conflicts could be prevented by pointing a download script provided with paparazzi at a specific commit inside the model repo (almost like a manual submodule...)
4. **Download automatically during make**. Very similar to submodules. Like submodules and LFS, all models will be downloaded. Pointing the download script at a specific version of the model repo prevents version conflicts.

I think adding this as a submodule would be the best solution in the long run, as it works out-of-the-box, prevents missing dependencies and is included in the version control. I do not think clutter will be a big problem as I do not expect the number of large gazebo models to grow quickly. If the disk usage or total download size should be kept to a minimum, option 3 would also be possible. What are your views on this?

dewagter:
IMHO cmake / ccmake on the root would be quite good to have the user select
what is available on his machine, or what to download and which
dependencies there are, and hopefully also which packages you have to
install.

I think this is cleaner than to automatically download while a user clicks
compile.

e.g.:

OPTION  USE_OPENCV
OPTION  USE_STM32
OPTION  USE_ARM
OPTION  USE_CHIBIOS (depends on STM32)
OPTION  USE_SIMULATION (depends on  PAPARAZZI-JSB)
OPTION  USE_GAZEBO (depends on gzclient7 to 8)
OPTION  USE_GAZEBO_EXTRA_MODELS (depends on  OPTION  USE_GAZEBO)

PS: I'm not (yet) talking about compiling paparazzi with c-make. To start
it can be called as an external project, but submodules could quite easily
be removed from the paparazzi makefile already.

-Christophe

gautierhattenberger:
I'm not in favor of options 2 and 4. For option 3, there is also the `paparazzi_pkgman.py` tool that could be used, although I'm not completely sure what you can/cannot do with this.
If we go for option 4, fetching this submodule should not be automatic I think.

OpenUAS:
+1 as submodule

tomvand:
Finally had time to work on this again :). I have removed the new models from this pull request and pushed them to [a new repository](https://github.com/tudelft/gazebo_models) that is included as a submodule. The submodule can be downloaded manually if required using `git submodule update --init -- sw/ext/tudelft_gazebo_models/`

I made minor changes to the code to support multiple model and world folders. I also added a helper script to launch gzclient with the correct environment variables.

In short:
- All new models are moved to the tudelft/gazebo_models repository, which is included as a submodule in sw/ext/tudelft_gazebo_models/
- Changed vehicle and world SDF loading to use URI's, these can be used with multiple search paths
- Added gzclient_launcher.sh script that sets up environment variables before launching gzclient. This prevents errors if the user forgets to add the paths to .bashrc. Added as a tool to control_panel_example.xml

OpenUAS:
Taken all remarks into consideration and all fine, thanks!

Issue #2207, opened 2017-12-20
Many implicit-fallthrough warnings triggered by GCC7


gautierhattenberger:
See https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/ for details about this type of error.

A (temporary) solution would be to add `-Wimplicit-fallthrough=0` apparently, but since this option is not recognized by older versions it needs to be filtered and used only with GCC7.
For long term solution, the code should be updated to add the proper comments.
Any other suggestions ?

podhrmic:
The `Label(x)` still triggers warning. I tried to fix it, but now I think that this switch + goto combination is something that GCC can't handle.

Getting rid of gotos in flightplan is out of question, at least for now. Perhaps we can switch to an older GCC? 

Here is the error I am getting:
```
paparazzi/var/aircrafts/Ark_Hexa_1.8/ap/generated/flight_plan.h: In function 'auto_nav':
./subsystems/navigation/common_flight_plan.h:45:1: warning: this statement may fall through [-Wimplicit-fallthrough=]
 nav_init_stage();
 ^~~~~~~~~~~~~~~~
./subsystems/navigation/common_flight_plan.h:57:16: note: in expansion of macro 'InitStage'
 { nav_stage++; InitStage(); }
                ^~~~~~~~~
paparazzi/var/aircrafts/Ark_Hexa_1.8/ap/generated/flight_plan.h:207:9: note: in expansion of macro 'NextStage'
         NextStage();
         ^~~~~~~~~
./subsystems/navigation/common_flight_plan.h:65:1: note: here
 label_ ## x:
 ^
paparazzi/var/aircrafts/Ark_Hexa_1.8/ap/generated/flight_plan.h:208:7: note: in expansion of macro 'Label'
       Label(while_1)
       ^~~~~
```

podhrmic:
@gautierhattenberger When you compile with ChibiOS it has own fail through warnings. Indeed it was fixed in the new versions, but that means we would have to update the repo. And that means redoing the tfls module. What do you think?

Also, we could revert to an older version of gcc?

gautierhattenberger:
We have planned to update ChibiOS when we have enough time to test things properly. In the meantime, they are only warnings. If it is blocking, using GCC6 should be good enough.

podhrmic:
Not blocking, just an observation. There is still the issue with 'Label' triggering the warning, the seems to be fine.

OpenUAS:
[...]

PR #2233, opened 2018-02-08
[build] handle fallthru warning in flight plans due to NextStage

gautierhattenberger:
This should fix most remaining warnings from flight plans.
Also provide a convenience macro depending of the compiler version.

see #2207

### DIFF
--- a/sw/airborne/subsystems/navigation/common_flight_plan.h
+++ b/sw/airborne/subsystems/navigation/common_flight_plan.h
@@ -43,16 +43,16 @@ void nav_goto_block(uint8_t block_id);
 
 #define InitStage() nav_init_stage();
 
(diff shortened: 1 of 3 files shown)


podhrmic:
Thanks for the fix @gautierhattenberger !
My only question is - do we care about correct spelling?

gautierhattenberger:
you mean FALLTHRU instead of FALLTHROUGH ? I took the define from a colleague and didn't change it as it is one of the "valid" spelling was using comments to suppress the warning (see https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/).
I don't mind to change it thought.

kirkscheper:
travis says this works so merging

PR #2237, opened 2018-02-14
Fix image.c compilation warnings

tomvand:
Fixes uninitialized fields and signed/unsigned comparison compilation warnings in image.c.


kirkscheper:
Actually, the plan here is to remove the subpixel field here, it should not be in this structure, it's for optical flow not for images. It accidentally got merged.

tomvand:
Ok

PR #2238, opened 2018-02-14
Move dummy waypoint to 0, 0

tomvand:
Should prevent "waypoint 'dummy' too far from HOME" warnings when
max_dist_from_home is small (e.g. cyberzoo flightplans) and when
HOME is set appropriately close to the flight plan origin.

### DIFF
--- a/sw/tools/generators/gen_flight_plan.ml
+++ b/sw/tools/generators/gen_flight_plan.ml
@@ -738,8 +738,8 @@ let check_geo_ref = fun wgs84 xml ->
 let dummy_waypoint =
   Xml.Element ("waypoint",
                ["name", "dummy";
-                "x", "42.";
-                "y", "42." ],
+                "x", "0.";
+                "y", "0." ],
                [])
 
 
--- a/sw/tools/generators/gen_flight_plan.ml
+++ b/sw/tools/generators/gen_flight_plan.ml
@@ -735,14 +735,6 @@ let check_geo_ref = fun wgs84 xml ->
   let x = ExtXml.subst_child "waypoints" wpts xml in
   x
 
-let dummy_waypoint =
-  Xml.Element ("waypoint",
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
I think we should investigate the impact of completely removing this dummy waypoint. But concerning the warning, I would prefer to skip this point in the check rather than moving it to a "better" place. It covers more cases but not all.

EwoudSmeur:
I also think we should remove it

tomvand:
Ok. I've removed the dummy waypoint from the flight plan generator. I'm not sure what it was originally used for so I can't guarantee this won't cause errors, but make and build do not seem to miss it.

gautierhattenberger:
Although it might be interesting to removing it, it is not possible just like this because it is used by:
- gen_flight_plan.ml (https://github.com/paparazzi/paparazzi/blob/master/sw/tools/generators/gen_flight_plan.ml#L437) as a dummy point if `wp` is not specified in `go` instruction but `x` and `y` (never hear of this feature ? it is not supported anymore by the DTD file !)
- nav_survey_rectangle.c, cartography.c, nav_survey_rectangle_rotorcraft.c as dummy waypoints for internal computation (and hopefully only that)

So unless those files are updated, we can't merge like this.

podhrmic:
In the meantime, having it at (0,0) should work the same as (42,42) - which is a random value anyway - and will help @tomvand so I would be happy just moving it.

gautierhattenberger:
or move this line https://github.com/paparazzi/paparazzi/blob/master/sw/tools/generators/gen_flight_plan.ml#L1041 three lines below after the distance check so that the dummy waypoint is not concerned by that anymore

gautierhattenberger:
closed with cf9401e5dfcc7b668e9a2d40567b072e1923cc70

PR #2240, opened 2018-02-16
suggestion to make command laws more flexible

EwoudSmeur:
It seems to me the inputs to this macro cause problems and are not very useful, as you can also just put 'commands' in the command_laws. Only fbw uses 'trimmed_commands' instead of 'commands', but this can be dealt with by putting 'trimmed_commands' in the command laws instead of 'commands'.

This does require editing all airframes...

### DIFF
--- a/conf/airframes/ENAC/cyfoam.xml
+++ b/conf/airframes/ENAC/cyfoam.xml
@@ -8,6 +8,8 @@
 
   <firmware name="rotorcraft">
     <target name="ap" board="chimera_1.0">
+      <configure name="PERIODIC_FREQUENCY"  value="500"/>
+
       <module name="radio_control" type="sbus">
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
Changing all airframe files is always the most difficult. Did you try to hide `values` in a macro to solve your particular issue ? Something like
```
#define RCValue(_x) radio_control.values[_x]
```
You can probably do this in your airframe file, or add this to `radio_control.h` if you think it can be useful.

EwoudSmeur:
Thanks for this suggestion, indeed this does compile. Defining it in the airframe file works, but indeed I think it would be convenient to define it in `radio_control.h`

Though to me it seems a bit strange to fix a macro problem with more macros XD

PR #2246, opened 2018-03-19
[swarm] add formation control for rotorcraft

gautierhattenberger:
Based on Ewoud Smeur (INDI) and Hector Garcia de Marina (Formation
control) work. See
https://blog.paparazziuav.org/2017/12/02/pilot-a-super-rotorcraft
- several UAV can be control from the ground (using a joystick for
  instance) by sending proper acceleration setpoint to the INDI guidance
  controller
- configuration of the formation is done from a JSON file (2 example
  files provided)
Some extra changes:
- update accel from IMU in GPS passthrough INS
- add accel setpoint to ABI messages
- by to accel setpoint in INDI guidance
- send ground reference from natnet2ivy
- possibility to have a joystick labeled 'GCS' with input2ivy


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -145,29 +164,49 @@ void guidance_indi_run(bool in_flight, float heading_sp) {
...
-  sp_accel.z = (speed_sp_z - stateGetSpeedNed_f()->z) * guidance_indi_speed_gain;
+  // If the acceleration setpoint is set over ABI message
+  if (indi_accel_sp_set_2d) {
+    sp_accel.x = indi_accel_sp.x;
+    sp_accel.y = indi_accel_sp.y;
+    // In 2D the vertical motion is derived from the flight plan
+    sp_accel.z = (speed_sp_z - stateGetSpeedNed_f()->z) * guidance_indi_speed_gain;
+    float dt = get_sys_time_float() - time_of_accel_sp_2d;
+    if (dt > 0.5) {
+      indi_accel_sp_set_2d = false;
```
kirkscheper: This seems strange... 

I am not sure where else 'sp_accel' used but is overwritten somewhere else? Else it should be sufficient to immediately set indi_accel_sp_set_2d = false; without the dt check.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -318,5 +359,25 @@ void stabilization_attitude_set_setpoint_rp_quat_f(struct FloatEulers* indi_rp_c
...
-  QUAT_BFP_OF_REAL(stab_att_sp_quat,q_sp);
+  QUAT_BFP_OF_REAL(stab_att_sp_quat, q_sp);
 }
+
+/**
+ * ABI callback that obtains the acceleration setpoint from telemetry
+ */
+static void accel_sp_cb(uint8_t sender_id, struct FloatVect3 *accel_sp)
+{
+  if (sender_id == 1) {
```
kirkscheper: I guess these should check for ACCEL_SP_3D_ID and ACCEL_SP_2D_ID

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
gautierhattenberger: if the guidance control loop is faster than the setpoint update, you want to avoid to switch the 2D/3D control and the derivate from speed until a timeout is reached

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
gautierhattenberger: actually, I didn't pay enough attention at how the message and ID wax used, it is now done correctly with a flag instead of abusing the sender ID

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
noether: I remember that when we wrote this condition for dt was also for safety porpuses. In particular, if the commands from the ground stop, then the rotorcraft resumes the flight plan (for example, hovering in a position with the guidance from INDI). Indeed, if the setpoint update is slower than the guidance, you do not want to have a "misbehavior" by switching between two set points from different systems (formation control or guidance).

I agree with kirk that this part could be more clear... in fact, I do not even remember how we wrote it. Although the behavior was as expected (including the safety).

EwoudSmeur: Indeed, the setpoint is updated over telemetry, which will update much slower than the periodic frequency.

gautierhattenberger: merging with an extra comment

PR #2247, opened 2018-03-20
[doc] start sphynx doc for paparazzi

gautierhattenberger:
This pull request is to start a discussion around building a sphinx documentation for Paparazzi alongside the wiki and more focused on the installation and user operation.
It can also have a bit of developer information, tutorials and so on.
Main benefit is to keep in sync the most important part of the doc with the actual code and features of paparazzi.

What do you think about the general idea and the proposed organization of the files for this doc ?


saaib:
I think this is a great idea!

In my job I created a Sphinx project for HTML and XML output generation of code, which then is used to compile the User and Development guide, then the XML output is used to build a dynamic RESTful service which publishes the system API were users can pull method references.

All the documentation is focused to maintain in a single place (code) with the most detail to format and description. 

Do it ONCE, do it RIGHT!

Regards,

hooperfly:
Do you have an example/use case of how you want to use sphinx docs with respect to Paparazzi that provides for improved feature/code synchronization? Are you going to generate the documentation from the code base? I'm curious as our biggest challenge at this point is to get folks to write the necessary documentation for the various features/installation/etc.... It would be nice to have more thorough and pertinent documentation as it relates to Paparazzi usage and feature sets.

podhrmic:
@gautierhattenberger I appreciate the effort towards better documentation, but I have a few concerns.
I guess the main one is that by introducing yet another type of documentation, it will be even more confusing and difficult to find/update anything.

For example right now we have:
- [wiki](https://wiki.paparazziuav.org)
- [Doxygen docs](http://docs.paparazziuav.org/)
- Github markdown (mainly the python examples [here](https://github.com/paparazzi/pprzlink/blob/master/README.md))
- [Sphix for pprzlink](https://pprzlink.readthedocs.io/en/latest/) (Mostly empty)
- technically you can also count the autogenerated Rust docs for [pprzlink-rust](https://docs.rs/pprzlink/0.3.1/pprzlink/parser/index.html) (although minimal at the moment)

I agree with the need for autogenerated docs, reflecting the current code and relevant APIs. It can be hosted on an external site (like readthedocs.org or docs.rs) to make the maintenance easier. But I am not in favor of adding tutorials, extensive install info and such into it. That should go to wiki. Just compare the entries for pprzlink in [wiki](https://wiki.paparazziuav.org/wiki/Pprzlink) and in the [sphix doc](https://pprzlink.readthedocs.io/en/latest/). The latter has much better Python API description, but is otherwise just a placeholder.

Paparazzi already uses Doxygen for code/api docs, so how would be Sphix useful here? I would prefer putting the effort into updating instructions on Wiki, and perhaps add pointers to relevant API docs (like Doxygen).

gautierhattenberger:
The wiki have been a nice way to build documentation for years, but lately it is hard to find people updating it. Moreover it is really annoying to specify each time for which version an option is available or not.
A documentation directly linked to the source code have the benefit to avoid these issues. If a change is made that should update the doc, it can be "mandatory" to accept the corresponding pull request.

Regarding the other sources of documentation:
* the wiki would be the main source until the sphinx doc is good enough
* doxygen doc doesn't cost a lot to maintain and their are some plugins for sphinx to integrate it if we need to merge the two
* examples in github markdown should be integrated to sphinx
* pprzlink sphinx was a trial to see if it can be used for paparazzi as well, there is no reason to merge the two

For your information, this work is the result of a common work from my colleagues and myself. We are trying to set regular working days (at least one per month) to work on Paparazzi issues, PR, doc, etc. If we keep going this way, I hope that we will update this on a regular basis, mostly getting the proper information from the wiki and adding it to sphinx. For instance, we have several pprzlink tutorials almost ready to be integrated.

Note that other popular projects (ArduPilot, PX4) are using this type of documentation, I have to admit that for general information, it is nicer to read than our wiki.

I'm not saying that it should replace existing doc right away. But I think we should give a try. I'll try to handle the beginning with my colleagues. Of course, after that any help is appreciated.

podhrmic:
Thanks for the explanation. I agree that some user-friendly documentation for new users is a weak point of paparazzi, and this could be a good solution. I would be very excited if we could merge doxygen into sphinx so +1 for that one. Just for inspiration, I find the Rust documentation very easy to read and navigate through, for example: https://doc.rust-lang.org/std/vec/struct.Vec.html 

PX4/ArduPilot docs are definitely more user-friendly, but perhaps hyper-optimized for new/beginner users - I found it often difficult to find deeper/adequate description of anything beyond the basic topics. Paparazzi is perhaps skewed towards the other extreme (not enough simple and easy to read instructions), so it would be good to find a balance between those two.

kirkscheper:
We have also dedicated one day per week with (most of) the lab to tackle some paparazzi issues. A few weeks ago we tried to clean up the wiki and there were many many outdated pages. 

I am very in favour of documentation that is linked to the code. Let us know when and how we can contribute!

BTW, I htink we should start using the Projects functionality in github to track changes like this and other long term developments in pprz. Should help us keep an overview as to the direction of code devel.

gautierhattenberger:
The idea was to make a general structure, then people can start filling the pages whenever they have some time. In order to avoid duplicate work, I guess it is better to work on small part at a time and make the pull request as soon as possible.
If you all agree on that, I can merge this one and we'll see how it goes.
I can also setup a Project regarding documentation to test the functionality. If you have other projects in mind don't hesitate.

noether:
For me, the most significant inconvenience is not to update the documentation on the wiki, but to find every entry that would need to be updated. In the end, one always misses updating something in the wiki because you do not know it is merely there.

I think we do not have any tool for solving this mentioned issue, so for sure I can try sphinx (for example, in the guidance vector field code) and give some feedback.

gautierhattenberger:
I'm merging this for now, I will then continue to update from master

PR #2249, opened 2018-03-31
Guidance INDI change to YXZ eulers

EwoudSmeur:
This representation is better for guidance purposes, as with this representation, changes in heading do not affect the thrust vector.

- New control effectiveness matrix
- Added quat to euler_yxz
- Added euler_yxz to quat

Tested in simulation, not yet in real life


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -229,10 +234,9 @@ void guidance_indi_run(bool in_flight, float heading_sp)
 
   AbiSendMsgTHRUST(THRUST_INCREMENT_ID, euler_cmd.z);
 
-  guidance_euler_cmd.phi = roll_filt.o[0] + euler_cmd.x;
-  guidance_euler_cmd.theta = pitch_filt.o[0] + euler_cmd.y;
-  //zero psi command, because a roll/pitch quat will be constructed later
-  guidance_euler_cmd.psi = 0;
+  guidance_euler_cmd.theta = pitch_filt.o[0] + euler_cmd.x;
+  guidance_euler_cmd.phi = roll_filt.o[0] + euler_cmd.y;
```
tcunis: why are `euler_cmd.x` and `euler_cmd.y` changed here?

EwoudSmeur: That is because of the rotation order, first around Y (pitch) then X (roll) and then Z (yaw). Correspondingly, I changed the input vector to be [pitch, roll, thrust]. Pitch is therefore the first element of the vector (x) and roll the second (y). It used to be [roll, pitch, thrust] for the ZYX Eulers, corresponding to [roll, pitch, yaw].

EwoudSmeur:
@tcunis Thanks for your comments! I'll update the description :)

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -104,7 +104,7 @@ Butterworth2LowPass thrust_filt;
 
 struct FloatMat33 Ga;
 struct FloatMat33 Ga_inv;
-struct FloatVect3 euler_cmd;
+struct FloatVect3 control_increment; // [dtheta, dphi, dthrust]
```
tcunis: I like the renaming. It helps not to confuse people.

EwoudSmeur:
Hm, interesting option. Indeed, we could make a struct for each euler sequence, but note that there are 6 of those. Personally, I think the risk of making this mistake is small, and will be easily detected. Maybe @gautierhattenberger can shine his light on this?

tcunis:
> we could make a struct for each euler sequence, but note that there are 6 of those

well, we don't need a struct for each of the 6 Euler sequences as long as only two are used... I think we should keep `FloatEuler` for the usual ZYX-scheme and only add those schemes other than that if necessary (and used). If you want to be strict, we can define a template-like macro for that, too.

I'd be also interested in a third opinion. ;-)

EwoudSmeur:
You are forgetting the ZXY order, which is already there without a special struct ;)

I am not saying it is a bad idea, I can see the benefit. But we'd be making structs with exactly the same elements only for that purpose.  For instance compare it to FloatVect3, which is used for position, speed and acceleration vectors (and maybe more?). Should we also change that?
I think it would be good to get some additional opinions.

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -520,6 +520,33 @@ void float_quat_of_eulers_zxy(struct FloatQuat *q, struct FloatEulers *e)
...
+ * See https://en.wikipedia.org/wiki/Euler_angles
+ *
+ * @param q Quat output
+ * @param e Euler input
+ */
+void float_quat_of_eulers_yxz(struct FloatQuat *q, struct FloatEulers *e)
+{
+  const float phi2   = e->phi / 2.0;
+  const float theta2 = e->theta / 2.0;
+  const float psi2   = e->psi / 2.0;
```
kirkscheper: it would be better if this was 2.f instead of 2.0 (which is a double)

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -638,6 +665,38 @@ void float_eulers_of_quat(struct FloatEulers *e, struct FloatQuat *q)
...
+  const float qiqx = q->qi * q->qx;
+  const float qiqy = q->qi * q->qy;
+  const float qiqz = q->qi * q->qz;
+  const float qxqy = q->qx * q->qy;
+  const float qxqz = q->qx * q->qz;
+  const float qyqz = q->qy * q->qz;
+  const float r11  = 2 * (qxqz + qiqy);
+  const float r12  = qi2 - qx2 + qy2 + qz2;
+  const float r21  = -2 * (qyqz - qiqx);
+  const float r31  = 2 * (qxqy + qiqz);
```
kirkscheper: same here, 2.f

sw/airborne/math/pprz_algebra_float.c
EwoudSmeur: I didn't know this, maybe also good to keep this in mind with the defines in airframe files?

sw/airborne/math/pprz_algebra_float.c
EwoudSmeur: Without the point, wouldn't 2 be an int that is then promoted to float?

kirkscheper:
True, but if you put 2.f you skip that promotion step. It is also a nice
reinforcement to the coder that the values are floats.

sw/airborne/math/pprz_algebra_float.c
EwoudSmeur: Actually, there are a lot more of these numbers even in the same file, change those as well? I did see that for chibios and the F7 the flag -fsingle-precision-constant is defined, which should make those numbers single precision, but I don't know what happens for other configurations.

kirkscheper:
Tbh, I don't think it's that critical. It was just the only thing I saw to
comment on :p

tcunis:
Thanks for raising the issue. Good work with the INDI. :-)

Cheers,
Torbjørn
--
Torbjørn CUNIS
B.Sc. B.Sc. (JMUW), M.Sc. (RWTH)

8 bis Impasse Barthe
31200 Toulouse, FR
torbjoern.cunis@gmail.com
+49.(0)1.57.03.03.15.09
+33.(0)7.53.11.00.85

PR #2250, opened 2018-04-10
[radio] rewrite the Spektrum driver to be generic

gautierhattenberger:
This is based on #1799 and replace it.
Now, the normal UART driver is used instead of the specific code. It
makes it possible to use on only arch, including ChibiOS.
Binding is also possible, but for boards using ChibiOS and SD logger, it
is recommanded to disable it for the binding process as the LSE clock
required by the RTC is longer to start (bind pulses should be issued
before 200ms after power up).
The configuration is backward compatible and some defines in the boards'
header could be even removed. Only Apogee, Chimera and Elle0 have been
updated as it is the only board tested for now. There is no harm to keep
the old defines.


podhrmic:
Travis agains exceeds the maximum time for a single job (50min) and fails. Perhaps we need to remove yet more tests?

Issue #599, opened 2013-12-16
stabilization_attitude_rc_setpoint should not pull in autopilot and guidance


flixr:
sw/airborne/firmware/stabilization/stabilization_attitude_rc_setpoint.c pulls autopilot_mode, guidance_h_mode, transition_theta_offset.

This should rather be factored out in to separate read_rc functions and then called from guidance as appropriate.
##

PR #2253, opened 2018-04-25
[rotorcraft] fix dependancy of stabilization on guidance

kirkscheper:
Should fix the #599. Most of it was already done but this removes the dependency of the stabilization on the guidance.


EwoudSmeur:
Looks good to me

PR #2256, opened 2018-04-26
bebop2 airframe file and python param estimation

EwoudSmeur:
- Upgrade tudelft bebop2 airframe file to use indi with WLS control allocation.
- Also added a python script to estimate the control effectiveness for all actuators


EwoudSmeur:
Thanks for your comments @kirkscheper ! I didn't know about the future module, but even without using it I can run the code with python2 without errors (not sure why?). However, the estimation gives different results then for some reason. Unless you feel that we really need support for python2 I am fine with not supporting it.

EwoudSmeur:
Ah I see now that print() works for both python 2 and 3, whereas print ... only works for python 2

PR #2260, opened 2018-05-07
add blocking option to i2c, remove unnecesary GNU_SOURCE and protect  pthread_setname from mac

kirkscheper:
The cameras for the bebop need blocking to write the configuration. Additionally, actuators have a multi-write protocol so also need blocking write.

This also moves the GNU_SOURCE def to the makefile and places ifdef protection for pthread_setname on mac.


gautierhattenberger:
I'm just making the timeout configurable and merging since without this patch, several things are broken, including bebop2 flights.

PR #2261, opened 2018-05-08
Pullrequest bluegiga stdma dongle

coppolam:
This tool allows Bluegiga dongle to communicate using STDMA (Self-organized Time-Division Multiple Access). Using this, drones with a Bluegiga USB Bluetooth dongle can communicate with eachother and exchange their RSSI (i.e., signal strength, which in a lot of cases can be used as a measure of distance). This has been tested with up to 3 ARDrones.


coppolam:
This tool allows Bluegiga dongle to communicate using STDMA (Self-organized Time-Division Multiple Access). Using this, drones with a Bluegiga USB Bluetooth dongle can communicate with eachother and exchange their RSSI (i.e., signal strenth, which is a measure of distance). This has been tested with up to 3 ARDrones.

kirkscheper:
Some of this code is auto-generated, can you add some info on where the origin? Also, can you add some info on the cdc-acm.ko?

coppolam:
The two auto-generated files are cmd_def.c and cmd_def.h, I added header info to the top of the files to explain where they originate from (note that the similar files can be found in the bluegiga_usb_dongle tool, so I took the headers from there). The cdc-acm.ko is a kernel object file. Once again, this can also be found in the bluegiga_usb_dongle tool (I assume it is relevant when uploading to the module).

kirkscheper:
On a second look, why did you make a new folder copying all the code from the bluegiga_usb_dongle and not just expand the code base there to include this? You can make two targets, one for the dongle and another for the STDMA dongle.

coppolam:
That's a fair point. The reason for duplicating is that the files are the same, but not one to one. There are differences in the code when you compare them. I'm personally not sure on the impact of the differences and how they may affect how well the code functions. Do you maybe know something about it?

kirkscheper:
I am inclined to merge this for now that it is in the project but I think that this should eventually be merged with the bluegiga_usb_dongle code. This will require some development and testing which is too much for this pr.

@gautierhattenberger if you agree can you merge?

sw/tools/STDMA_dongle/Makefile
```diff
@@ -0,0 +1,76 @@
...
+$(TARGET): $(OBJS)
+	@echo 'Building target: $@'
+	$(CC) -o"$(TARGET).elf" $(OBJS) $(LDFLAGS)
+	@echo 'Finished building target: $@'
+
+clean:
+	-$(RM) $(OBJS) $(DEPS) $(TARGET).elf
+
+#include $(PAPARAZZI_SRC)/conf/Makefile.linux
+DRONE = $(PAPARAZZI_HOME)/sw/tools/parrot/ardrone2.py
```
gautierhattenberger: Is this only available for ardrone2 ? not available for bebop 1/2 or other ?

gautierhattenberger:
I don't mind to merge this, but if it is compatible with other airframe than ardrone2, it is worth mentioning it and at least make this configurable with `DRONE ?= ...`

coppolam:
I think it should be possible to also use this with Bebops, but it has only ever been directly tested on ARDrones. I can test with Bebop2 before changing to DRONE ?= ... and finalizing the pull request.

gautierhattenberger:
someone to finish this ?

kirkscheper:
I will try but was taking a look at the other pull requests from Christophe first.

sw/tools/STDMA_dongle/README.md
```diff
@@ -0,0 +1,37 @@
...
+make clean && make --> makes it for the computer
+EMBED=1 --> this builds it for the drone, rather than for your computer
+Once connected to the drone --> make upload_program EMBED=1 HOST=192..........
+
+### Setting up the communication between laptops/drones
+Turn on the drones and go to:
+
+	cd $PAPARAZZI_HOME/sw/tools/STDMA_dongle
+
+The command
```
dewagter:
```suggestion

Select your **DRONE** type in the Makefile (tested with ARDrone)

The command
```

gautierhattenberger: why not also adding `DRONE ?= ...` so that we can change the type of drone without necessarily changing the file itself ?

gautierhattenberger: hum, I wanted to make a modification proposal but apparently it directly applies it. @dewagter what is the trick to make a commit suggestion ?

dewagter: The modification proposal (insert a suggestion) is available when you make a comment about a line of code and hit Ctrl-G or press the +- icon and change the code within.

PR #2263, opened 2018-05-09
Fix pthread_setname_np conflicts for Mac OS X with respect to simulation.

hooperfly:
Bracket pthread_setname_np calls used by the simulation with __APPLE__ to avoid conflicts. Tested on Mac OS X system.

### DIFF
--- a/sw/airborne/arch/linux/mcu_periph/sys_time_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/sys_time_arch.c
@@ -103,9 +103,7 @@ void sys_time_arch_init(void)
     perror("Could not setup sys_time_thread");
     return;
   }
-#ifndef MACOSX
   pthread_setname_np(tid, "pprz_sys_time_thread");
-#endif
 }
 
 static void sys_tick_handler(void)
--- a/sw/airborne/arch/linux/mcu_periph/uart_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/uart_arch.c
@@ -60,7 +60,7 @@ void uart_arch_init(void)
     fprintf(stderr, "uart_arch_init: Could not create UART reading thread.\n");
     return;
   }
-#ifndef MACOSX
+#ifndef __APPLE__
   pthread_setname_np(tid, "pprz_uart_thread");
 #endif
 }
--- a/sw/airborne/arch/linux/mcu_periph/udp_arch.c
+++ b/sw/airborne/arch/linux/mcu_periph/udp_arch.c
@@ -59,7 +59,9 @@ void udp_arch_init(void)
     fprintf(stderr, "udp_arch_init: Could not create UDP reading thread.\n");
     return;
   }
+#ifndef __APPLE__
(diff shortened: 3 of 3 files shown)


sw/airborne/arch/linux/mcu_periph/sys_time_arch.c
```diff
@@ -103,9 +103,7 @@ void sys_time_arch_init(void)
     perror("Could not setup sys_time_thread");
     return;
   }
-#ifndef MACOSX
```
gautierhattenberger: why no `__APPLE__` here ?

hooperfly: ... by the way, there are many other instances of calls to `pthread_setname_np` in the code base.

```
./airborne/arch/linux/mcu_periph/udp_arch.c:  pthread_setname_np(tid, "pprz_udp_thread");
./airborne/arch/linux/mcu_periph/uart_arch.c:  pthread_setname_np(tid, "pprz_uart_thread");
./airborne/arch/linux/mcu_periph/sys_time_arch.c:  pthread_setname_np(tid, "pprz_sys_time_thread");
./airborne/arch/linux/mcu_periph/i2c_arch.c:  pthread_setname_np(tid, "pprz_i2c_thread");
./airborne/arch/sim/mcu_periph/udp_arch.c:  pthread_setname_np(tid, "pprz_udp_thread");
./airborne/arch/sim/mcu_periph/uart_arch.c:  pthread_setname_np(tid, "pprz_uart_thread");
./airborne/boards/swing/board.c:  pthread_setname_np(bat_thread, "pprz_bat_thread");
./airborne/boards/swing/board.c:  pthread_setname_np(button_thread, "pprz_button_thread");
./airborne/boards/swing/baro_board.c:  pthread_setname_np(baro_thread, "pprz_baro_thread");
./airborne/boards/ardrone/navdata.c:  pthread_setname_np(navdata_thread, "pprz_navdata_thread");
./airborne/modules/computer_vision/video_thread.c:    pthread_setname_np(tid, "pprz_camera_thread");
./airborne/modules/sonar/sonar_bebop.c:  pthread_setname_np(tid, "pprz_sonar_thread");
```

hooperfly: ... I probably don't run into those issues because my `arch` is `stm32`

hooperfly: @gautierhattenberger I'm happy to wrap the other instances of these calls if you like. There are other options as well. We could define an OS abstraction layer for these types of issues and resolve the differences there. Also, there are BSD thread libraries available for Mac OS X that define `pthread_setname_np`, I'm not sure how they differ from their POSIX thread support in implementation.

gautierhattenberger: If it is only a SIM/NPS issue on Mac, their is probably no need to protect things that would not be part of a simulation. If you think some of these calls should be protected, please do it, otherwise it is file like this.

Issue #2187, opened 2017-11-16
Segfault in cv_async_function() when video changes size


tomvand:
I'm getting a segfault when my module attempts to change the resolution of the images returned from its video callback:

Asynchronous video listeners keep their own copy of the current video frame. This image is created in [`cv_async_function()` (cv.c line 114)](https://github.com/paparazzi/paparazzi/blob/d2079a780f316ef79c55ae30cc4e2e07a8c74144/sw/airborne/modules/computer_vision/cv.c#L114) with the size of the current frame `img`.
When new frames arrive they are copied to this image in [line 119](https://github.com/paparazzi/paparazzi/blob/d2079a780f316ef79c55ae30cc4e2e07a8c74144/sw/airborne/modules/computer_vision/cv.c#L119).

This fails when the size of the video frames changes between subsequent calls. [`image_copy()`](https://github.com/paparazzi/paparazzi/blob/d2079a780f316ef79c55ae30cc4e2e07a8c74144/sw/airborne/modules/computer_vision/lib/vision/image.c#L89) does not check that the size of the source image is the same as the target one, so when the new image is bigger, the memcpy on [line 89](https://github.com/paparazzi/paparazzi/blob/d2079a780f316ef79c55ae30cc4e2e07a8c74144/sw/airborne/modules/computer_vision/lib/vision/image.c#L89) writes beyond the allocated buffer and causes a segfault.

The problem can be fixed by re-creating the asynchronous copy of the image using the following code (inserted at line 116 of cv.c):
```
  if (async->img_copy.buf_size != img->buf_size) {
    printf("[cv_async_function] BUFFER SIZE CHANGED!\n");
    printf("Old:\t%d x %d\n", async->img_copy.w, async->img_copy.h);
    printf("New:\t%d x %d\n", img->w, img->h);
    image_free(&async->img_copy);
    image_create(&async->img_copy, img->w, img->h, img->type);
  }
```

Should this be fixed so that the size of the video can be changed, or should I not have attempted this in the first place? I could not find any documentation that this is not allowed. (Unless I missed something there is very little documentation on the video callback system of cv.c and video_thread.c).

Secondly, should image_copy check that the images' buffer sizes are equal before copying, so that segfaults like these are prevented or at least easier to find? Or should image_copy allocate a new buffer when the images' types or sizes are not equal?

kirkscheper:
[...]

PR #2269, opened 2018-05-30
unwrap_pano module for panoramic cameras

tomvand:
This module unwraps images taken through a panoramic lens:
image

Also includes a fix for #2187. This fix must be enabled explicitly as not all existing modules might work with changing video sizes.


sw/airborne/modules/computer_vision/cv.c
```diff
@@ -113,6 +113,13 @@ int8_t cv_async_function(struct cv_async *async, struct image_t *img)
   if (async->img_copy.buf_size == 0) {
     image_create(&async->img_copy, img->w, img->h, img->type);
   }
+#if CV_ALLOW_VIDEO_TO_CHANGE_SIZE
+  // Note: must be enabled explicitly as not all modules may support this. (See issue #2187)
+  if (img->buf_size > async->img_copy.buf_size) {
```
dewagter: only grow?

conf/modules/pano_unwrap.xml
```diff
@@ -0,0 +1,82 @@
...
+    </dl_settings>
+  </settings>
+  <depends>video_thread</depends>
+  <header>
+    <file name="pano_unwrap.h"/>
+  </header>
+  <init fun="pano_unwrap_init()"/>
+  <makefile>
+    <file name="pano_unwrap.c"/>
+    <define name="CV_ALLOW_VIDEO_TO_CHANGE_SIZE" value="1"/>
```
dewagter: When you define this, then all modules that do not support changing size will still crash if I am not mistaken...

sw/airborne/modules/computer_vision/cv.c
tomvand: Yes, as suggested by @kirkscheper  [here](https://github.com/paparazzi/paparazzi/issues/2187#issuecomment-345455394). It prevents some unnecessary free'ing and alloc'ing at the cost of a larger memory consumption (that was already required to hold the larger frame anyway).

PR #2273, opened 2018-06-19
State return fix

gautierhattenberger:
- return before setting flag if conversion is not done
  if bit is set for a given representation while conversion was not  possible, this may impact future conversions in other representation  as well, depending of the priority order (note that proper errno is still not set at the moment)
-  add more conversions for LLA format, especially handle NED and UTM initialization


podhrmic:
Looks good to me.

PR #2274, opened 2018-06-20
[gcs] add shapes to the background layer to keep WP and A/C on top

### LINKED ISSUE #2272: Drawn shapes on GCS blocks Waypoint edit
When drawing shapes on the GCS using the IVY SHAPE message, the shapes appear on the foreground of the GCS. It can hide the aircraft symbol as depicted in the picture. When waypoints are located underneath these shapes, I'm not able to edit those points from the GCS. 

In my opinion, the shapes should be plotted on the background of the GCS in order to overcome those problems.

issue_layers

gautierhattenberger:
fix #2272


OpenUAS:
Yes, it works, big ithanks for your fast fix

PR #2276, opened 2018-06-23
Rover firmware integration

gautierhattenberger:
Basic support for rover firmware: https://www.youtube.com/watch?v=CJ1q4373pB8

Several improvements should be added later for better support of some telemetry, ins, ahrs, actuators, ...
Navigation is based on a new idea with registered patterns (function pointers) to allow switching between different algorithms (like carrot-based, GVF, ...).
Only available with ChibiOS for now.
Autopilot is based on AP generation.


fvantienen:
Is this working and final, so does it need a review?

gautierhattenberger:
It is working for sure. There are some works to do on the messages but this is a bit out of the scope of the PR.
I was more or less looking for comments on the implementation of the navigation layer, as this could be used in the other firmwares.

PR #2277, opened 2018-06-26
[XSens] Move parser variables to struct, and call the decoder after p…

dewagter:
…arsing.


sw/airborne/modules/ins/ins_xsens.c
```diff
@@ -77,11 +77,11 @@ void ins_xsens_init(void)
 
 void ins_xsens_event(void)
 {
-  xsens_event();
-  if (xsens.msg_received) {
```
dewagter: this variable was never set to true as the parser used a global variable

PR #2279, opened 2018-06-27
[CV] Improve timing stability

kirkscheper:
Here I tried to remove some mallocs and frees and only do them when the image sizes have changed. 

Moving the timing from the start to the end of the cv computation loop helps the timing of the loop when the computational time is varying.

I also made the requested frequency of the bottom camera settable with MT9V117_TARGET_FPS. It is more efficient to limit the speed with this rather than with OPTICFLOW_FPS if no other processes are using the bottom camera.

Running the main loop at double the ap frequency seems more appropriate than the arbitrary 1000Hz that was there before. This is now also settable using the POLLING_PERIOD define.

Lastly, there is an interesting program on the bebop (2) called ephemerisd. When the parrot dragon software is running it uses about 2% of the cpu but once paparazzi is launched it jumps to 50%. I am not entirely sure what this program does or why it uses to much processing. If you Google ephemeris you get something with GPS satellite tables so I think it is safe to say that if you are flying inside you can safely close this program. For now I put it inside of an i#f PRIMARY_GPS == GPS_DATALINK. Let me know if you have a better protection check.


gautierhattenberger:
looks fine.

would be nice to test killing `ephemerisd` program when using the gps, because it shouldn't be need in our case at least

PR #2280, opened 2018-06-27
[rotorcraft] update yaw to arm

kirkscheper:
Somehow the yaw to arm was broken, there was quite a bit of strange behavior.

With this, you give a side stick to arm the motors, this will start the motors at minimum speed. Your stick controls will not be fed through to the controller till you return the stick to the center. When disarming, you reduce the throttle to minimum and give full side stick till the motors stop. You first have to return the stick to center before you can re-arm.

I also add an explicit autopilot_set_in_flight(false) command when starting the motors. This is not strictly necessary since the in_flight check checks if the throttle is retarded but I think this is nice just in case the in_flight code is changed in the future.


gautierhattenberger:
I don't have much time to test, but if you can confirm that it works, we can merge

PR #2281, opened 2018-06-27
[mcu_periph] add full duplex named pipe

kirkscheper:
This allows for high speed, lossless fifo communication between two programs running on a linux platform using a two half duplex named pipes.

### DIFF
--- a/sw/airborne/arch/sim/mcu_periph/pipe_arch.c
+++ b/sw/airborne/arch/sim/mcu_periph/pipe_arch.c
@@ -1,0 +1,1 @@
+../../linux/mcu_periph/pipe_arch.c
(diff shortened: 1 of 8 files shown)


sw/airborne/arch/sim/mcu_periph/pipe_arch.c
```diff
@@ -0,0 +1 @@
+/home/kirk/mavlab/paparazzi/sw/airborne/arch/linux/mcu_periph/pipe_arch.c
```
gautierhattenberger: you can't use absolute path here

sw/airborne/arch/sim/mcu_periph/pipe_arch.h
```diff
@@ -0,0 +1 @@
+/home/kirk/mavlab/paparazzi/sw/airborne/arch/linux/mcu_periph/pipe_arch.h
```
gautierhattenberger: same thing, no absolute path

PR #2282, opened 2018-06-27
[logging] added ability to give linux file logger a datetime stamp name for convenience

kirkscheper:
Will also attempt to make the requested directory if it doesn't already exist.


gautierhattenberger:
looks good

kirkscheper:
tested... working

Issue #2284, opened 2018-06-28
XBEE_API not working in Fixedwing - LPC21


dewagter:

dewagter:
- hardware-debugged the wait function: it works exactly as expected

dewagter:
- hardware debugged: when sending "+++" config string, the put_bytes is called 3 times but only the first "+" is actually sent over the serial. The XBee obviously never replies.

dewagter:
- while the XBee is being configured, there is also NO PWM so it looks like some master interrupt function is still off.

gautierhattenberger:
This should be checked, but maybe the `xbee_init` function is called (from the `modules_init`) before the `mcu_int_enable` function. It does nothing on stm32, but for lpc it means the uart interrupts won't be called, so nothing will happen after the first character.

dewagter:
See https://github.com/paparazzi/paparazzi/pull/2285

PR #2285, opened 2018-06-28
[xbee_api] fixedwing LPC21 xbee_api broken

dewagter:
- interrupts not enabled during xbee setup (which is called from modules_init)
 - thereby only the first byte is sent (in LPC: the tx-complete interrupt loads the next byte)
 - and configuration fails

 - moving modules_init() after the interrupts solves this
 - it is unclear how much this will break elsewhere.

### DIFF
--- a/conf/airframes/tudelft/yapa_xsens.xml
+++ b/conf/airframes/tudelft/yapa_xsens.xml
@@ -18,7 +18,7 @@ YAPA + XSens + XBee
     <define name="AGR_CLIMB"/>
     <module name="radio_control" type="ppm"/>
     <!-- Communication -->
-    <module name="telemetry" type="transparent">
+    <module name="telemetry" type="xbee_api">
       <configure name="MODEM_BAUD" value="B57600"/>
     </module>
     <!-- Actuators -->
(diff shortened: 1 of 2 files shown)


dewagter:
https://github.com/paparazzi/paparazzi/issues/2284

gautierhattenberger:
according to the comment below (https://github.com/paparazzi/paparazzi/pull/2285/files#diff-f8e90c989a729602c5d1e55cc81e6722L183) it will break things for sure.
I think it is `mcu_int_enable` that should be moved instead. On stm32 and linux boards, things are already running during init and it seems to work, so it is worth trying it on lpc21 boards (so enabling IRQ at the and of `mcu_init`.

dewagter:
That makes sense. So I moved the enable interrupts. This does not change anything for STM and Linux since the enable_isr is empty.

According to my tests, I see no problem at all enabling the interrupts earlier. This also solves the xbee_api problem. All timers and pulses seem OK.

gautierhattenberger:
If you can confirm with hardware that it is still working, I'm fine with this

dewagter:
I checked on a yapa2.0 board (electrically identical to the tiny1.3) and it works.

I also checked the init code that is now after the interrupt enable: trig, ahrs, state, baro, autopilot, settings are all safe to do with the interrupts enabled. Some modules only work after, which is now fixed. 

The only possible issue was registering the timers, which could have race conditions:

 - but the timers are first setup completely, and only afterwards enabled using an atomic operation, so it should be fine
 - if this was a problem, it would also be a problem for STM boards that enable their interrupts at the start too

martinmm:
@dewagter thanks a lot

Issue #2286, opened 2018-06-28
In Fixedwing, when RC is lost in MANUAL mode and throttle is red, the motor keeps turning with the last commanded throttle


dewagter:

dewagter:
- creates very high risk of fly-away

dewagter:
fly-away condition: (e.g. tudelft_conf SkyWalker)

 - fly in manual with default/versatile flight plan (still in holding point)
 - loose RC -> GCS shows RC =  none, throttle is red, but stays at the last commanded manual throttle value. Mode stays Manual.
 -  stabilization is on (failsafe value?)

=> stabilized with power = fly-away

dewagter:
See pull request #2287

PR #2287, opened 2018-06-28
Fix: In Fixedwing, when RC is lost in MANUAL mode and throttle is red, the motor keeps turning with the last commanded throttle and does not go to RC_LOST_MODE

dewagter:
Solves #2286

### DIFF
--- a/sw/airborne/firmwares/fixedwing/autopilot_static.c
+++ b/sw/airborne/firmwares/fixedwing/autopilot_static.c
@@ -114,16 +114,20 @@ void autopilot_static_on_rc_frame(void)
   uint8_t mode_changed = false;
   copy_from_to_fbw();
 
-  /* really_lost is true if we lost RC in MANUAL or AUTO1 */
-  uint8_t really_lost = bit_is_set(imcu_get_status(), STATUS_RADIO_REALLY_LOST) &&
+  /* rc_lost_while_in_use is true if we lost RC in MANUAL or AUTO1 */
+  uint8_t rc_lost_while_in_use = bit_is_set(imcu_get_status(), STATUS_RADIO_REALLY_LOST) &&
                         (autopilot_get_mode() == AP_MODE_AUTO1 || autopilot_get_mode() == AP_MODE_MANUAL);
 
+  /* RC_LOST_MODE defaults to AP_MODE_HOME, but it can also be set to NAV_MODE_NAV or MANUAL when the RC receiver is well configured to send failsafe commands */
+  if (rc_lost_while_in_use) { // Always: no exceptions!
+    mode_changed = autopilot_set_mode(RC_LOST_MODE);
+  }
+
+  /* If in-flight, with good GPS but too far, then activate HOME mode
+   * In MANUAL with good RC, FBW will allow to override. */
   if (autopilot_get_mode() != AP_MODE_HOME && autopilot_get_mode() != AP_MODE_GPS_OUT_OF_ORDER && autopilot.launch) {
     if (too_far_from_home || datalink_lost() || higher_than_max_altitude()) {
       mode_changed = autopilot_set_mode(AP_MODE_HOME);
-    }
-    if (really_lost) {
-      mode_changed = autopilot_set_mode(RC_LOST_MODE);
     }
   }
   if (bit_is_set(imcu_get_status(), AVERAGED_CHANNELS_SENT)) {
--- a/sw/airborne/firmwares/fixedwing/nav.c
+++ b/sw/airborne/firmwares/fixedwing/nav.c
@@ -424,7 +424,12 @@ void nav_home(void)
   NavCircleWaypoint(WP_HOME, FAILSAFE_HOME_RADIUS);
   /** Nominal speed */
   nav_pitch = 0.;
-  v_ctl_mode = V_CTL_MODE_AUTO_ALT;
+  if (autopilot_in_flight() && autopilot.launch) {
+    v_ctl_mode = V_CTL_MODE_AUTO_ALT;
+  } else {
+    v_ctl_mode = V_CTL_MODE_AUTO_THROTTLE;
+    v_ctl_throttle_setpoint = 0;
+  }
   nav_altitude = ground_alt + HOME_MODE_HEIGHT;
   compute_dist2_to_home();
   dist2_to_wp = dist2_to_home;


dewagter:
- if RC lost while in manual, then always set the RC_LOST_MODE
 - if RC_LOST_MODE is AP_MODE_HOME, then when not launched, this it will be killed HOME mode which is no problem
 - if you were in NAV mode, nothing changes
 - if you want to go to NAV mode on RC loss, this can be set

 - so I see no reason why not to always check the safety, and I do see big risks

dewagter:
Any objection to merge this?

dewagter:
I personally think it is  worse that on RC loss in MANUAL you do NOT enter the configured RC_LOST_MODE because it creates danger for third parties in normal operating conditions while the throttle on the ground HOME MODE creates danger for yourself in bad practice conditions. Now without GPS or in case of GPS problem, or in case you get RC loss with high headwind you do not enter the RC_LOST_MODE.

First, if you are in manual mode, it is good practice to switch off you plane before your transmitter.
But to reduce chances of personal damage due to HOME modes being triggered after landing and switching off the RC while in MANUAL, I would prefer another solution like:

**on RC-LOSS with zero throttle from MANUAL or AUTO1 and entering HOME_MODE then kill.**

dewagter:
Here is a new proposition: (see code)

 - No more conditions to not have RC_LOST_MODE activate from MANUAL or AUTO1. Even without GPS, even without launch, also during static testing on the ground while doing safety checks before take-off, etc... ALWAYS.
 - When on the ground in manual, there was a risk of accidentally entering HOME mode (with throttle) when the RC was switched off before the plane after having been launched. To solve this problem, when entering RC_LOST_MODE, if the throttle was OFF, then the throttle is killed, so even in HOME or NAV mode, the plane will not throttle up. If you switch off your RC from manual/auto1 after landing, it will never throttle up.

gautierhattenberger:
If I got it right, If you are running the motor on ground in manual and loose the RC, it will enter HOME with throttle, which is not completely solving the initial issue

dewagter:
Yes, you are right:

 - if you are on the ground in MANUAL
 - after a flight or having launch=1 (which is not the case when you power up your fixedwing)
 - and you switch off your RC WHILE giving throttle

Then, HOME mode will become active without kill mode, and throttle, (which you already had) will increase to HOME-MODE throttle values.

The current situation is when you lose RC in manual with throttle, than throttle also stays on.

dewagter:
@gautierhattenberger : if you have another suggestion, you are very welcome but: 
 - Right now we are not allowed to fly with master in fixed-wing as it fails basic safety checks***.
 - Even worse: the problem you mention (sudden throttle on ground due to HOME mode on RC loss) can ALSO happen with the current master code: namely if you pressed launch in the GCS or have flown with GPS before the RC loss.
 - Finally, the exceptions based on certain conditions make it hard to explain to users.

This pull request solves all this and makes nothing worse as far as I can see. The behavior becomes:
 - 1) simple rule: RC lost in MANUAL/AUTO1 -> RC_LOST_MODE. Always. Like it was in 2005.
 - 2) simple rule: RC lost without throttle, then throttle stays off. Always.

*** every model airfield wants to see: turn on the transmitter, turn on plane, give a bit of throttle, turn off the transmitter, the throttle must turn off. right now paparazzi fixed-wing does not do that.

gautierhattenberger:
For now, I don't have a magic solution. What is really annoying me is that an aircraft not kill in MAN with a bit of throttle, in case of RC lost will:
- keep the current throttle (current situation, not satisfactory)
- switch to RC_LOST, so most likely HOME which will increase the throttle to try to climb to safe alt (your proposed modification)

I can see big issues in both case.
Something we could do at least (close to your second rule): when RC is lost in MAN before takeoff, don't switch to RC_LOST/HOME (current behavior) but at least stop throttle if kill mode is still enabled.

dewagter:
With this pull request: if you lose RC with a bit of throttle BEFORE launch, then it switches to HOME mode WITHOUT throttle.

Without this pull request, with a bit of throttle, on RC_LOST, then throttle stays until RC is restored. Even the kill button on the ground station does not work.

The only problem that remains is that after launch (and if you flightplan did not have kill on land), if you had throttle in RC while losing RC, and you have RC_LOST_MODE as HOME or NAV and you have good GPS, then will it throttle up, while before this pull request it would stay at the last received RC throttle value. However, IMHO, this is not a normal condition:

Normal use conditions:
 - switching off your plane before your transmitter is the best practice in any mode (no problem)
 - switching off your transmitter before your plane without throttle (now solved: throttle will always stay off: before it would go to RC_LOST mode since launch is active)
 - switching off your transmitter with throttle on, before launch (solved: with PR: it enters RC_LOST_MODE + launch(0) = no throttle. Current situation is last manual throttle command is FROZEN, GCS can not kill)
 - real RC-loss on a long final in MAN/AUTO1 mode = RC_LOST_MODE (same as now in most cases, but now there are exceptions, and this PR removes them)

Problem cases:
 - real RC loss in flight at a moment where zero manual throttle is given (which can happen in fixedwing), but the pilot expects HOME mode (now the plane will be in kill mode; you need to unkill through telemetry)
 - real RC loss while taxing back from the landing location with a bit of throttle: the plane might take-off again while entering HOME. Do you have experience with this? Since the transmitter is close, we never had this before.
 - very bad practice situation where the pilot is in MANUAL, gives throttle after a flight which did not do AUTO2 landing with a kill throttle at the end and switches off his transmitter expecting the throttle will switch off. To me, this case can be neglected.

Possible solutions:

 - add a landing detector like in rotorcraft to give the safest response both in flight and on the ground (risk: get false detections inflight)

gautierhattenberger:
I agree that the proper solution would be an efficient and robust "in_flight" detector, which is not so easy on a fixedwing platform. I guess we could start listing the possible criteria for this.

Some comments:
- the `autopilot_set_motors_on` function is empty on fixedwing firmware at the moment, the only way to stop the motor is to overwrite the throttle command I think (we could change this of course).
- the threshold on 1% is really low and barely never happen if you move the throttle stick by hand

What about:
```
if (rc_lost_while_in_use) {
 if (launch) {
  mode_changed = autopilot_set_mode(RC_LOST_MODE);
 } else {
   autopilot_set_motors_on(false); /* assuming the function is implemented... */
 }
}
```
So if RC is lost in MAN/AUTO1, go to RC_LOST_MODE unless you have not start flying (then stop motors). Problems that may arise after flight are less likely to happen as you said, and a "in_flight" detection might improve this later.

dewagter:
- as far as I can see, `autopilot_set_motors_on` from autopilot.c calls the `autopilot_static_set_motors_on` (which is empty in fixedwing) but also sets `autopilot.kill_throttle` which should set the `v_ctl_throttle_setpoint` to zero. Will test again.

dewagter:
Any idea to reintegrate this pull request? We can not fly fixedwing with master until this is fixed. Our minimum viable requirement is:

 - before takeoff, in manual, when throttle is applied and the RC is lost, then throttle is removed.

(This should also solve the fly-away condition that master has now when flying without GPS: then inflight is not set and throttle stays with stabilization)

dewagter:
Extra detail: we set RC_LOST_MODE to failsafe.

gautierhattenberger:
what about the last code I proposed ? if rc lost and not launched, turn off motor (either directly or by implementing the proper function in autopilot.c if setting kill_throttle is not enough)

gautierhattenberger:
side note: with autopilot generation, you can define the failsafe behavior you like ;)

dewagter:
@gautierhattenberger Your proposed solution would be an improvement compared to the current master.

Nevertheless, I still think it has several problems:

 - the GCS shows an undefined mode
 - when RC_LOST_MODE is failsafe, the throttle is cut but the failsafe actuators are not set
 - when no GPS is plugged, then launch is not set

The only logical solution to me (and the ones I asked) is

 - RC lost while in use => RC_LOST_MODE  (ALWAYS)
 - If RC_LOST_MODE == mode that can set throttle => extra safety to avoid setting throttle on the ground, e.g. with a launch == 0 -> kill and maybe additionally kill if rc_throttle == 0  for after manual landings.

gautierhattenberger:
So something like
```
if (rc_lost_while_in_use) {
  mode_changed = autopilot_set_mode(RC_LOST_MODE);
}
```
and replace https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/fixedwing/autopilot_static.c#L235
with
```
if (autopilot.launch) {
  nav_home();
} else {
  stop motor
}
```
?

dewagter:
- RC lost when in use? -> Always activate ``RC_LOST_MODE``
 - ``RC_LOST_MODE`` == ``MANUAL``: works fine if you configure your RC to send zero throttle like in model aircraft
 - ``RC_LOST_MODE`` == ``NAV``? If you selected this, then go to NAV. If NAV gives throttle, that is because you asked it to...
 - ``RC_LOST_MODE`` == ``HOME_MODE``: Home mode should not give throttle when on the ground. So a fix for Home mode was proposed.

PR #2288, opened 2018-06-29
[datalink] Set Xbee channel

Fabien-B:
Configure the xbee channel at modem initialization.
Use option -ch to change xbee channel.
Default is 0x0C.

### DIFF
--- a/sw/ground_segment/tmtc/link.ml
+++ b/sw/ground_segment/tmtc/link.ml
@@ -45,7 +45,7 @@ let transport_of_string = function
 
 
 type ground_device = {
-  fd : Unix.file_descr; transport : transport ; baud_rate : int ; channel : int
+  fd : Unix.file_descr; transport : transport ; baud_rate : int ; channel : int option
 }
 
 (* We assume here a single modem is used *)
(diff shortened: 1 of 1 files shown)


dewagter:
Will it change the channel of the XBee if no option is given?

dewagter:
Great, thanks!

PS: what is the corresponding define to add to your airframe to change the airborne side?

Fabien-B:
In one of my airframes I have this line:
`<define name="XBEE_INIT" value="\"ATID3332\rATCH17\rATP1\rATBD6\r\""/>`
But I think this only necessary stuff is :
`<define name="XBEE_INIT" value="\"ATCH17\r\""/>`
I may add a tag to the telemetry configuration to make ot easier.

dewagter:
Thanks. Looks great. Info in config xml is also a good idea.

gautierhattenberger:
will merge after testing on hardware

Fabien-B:
Tested on hardware, it works !
The tests I made :
Aircraft on channel 0x10, modem on channel 0x0C:
    launched without arguments : not working (as expected)
    launched with arg _-ch 0x10_ : it works
Modem also on channel 0x10 , launched without argument : it works (so it don't change the channel if the argument is not given)

PR #2290, opened 2018-07-02
[airframe] Adds airborne xbee channel configuration.

Fabien-B:
Allows airborne xbee channel configuration easier than with XBEE_INIT.
To change airborne xbee channel to e.g. 0x17, use :
```
<module name="telemetry" type="xbee_api">  
  <configure name="MODEM_CHANNEL"  value="0X17"/>
</module>
```

Is this the good way to do it ?
Should I add `MODEM_CHANNEL ?= 0` to every board makefile ?

Needs paparazzi/pprzlink#89.


conf/modules/telemetry_xbee_api.xml
```diff
@@ -24,6 +25,7 @@
     <define name="DOWNLINK_DEVICE" value="$(MODEM_PORT_LOWER)"/>
     <define name="XBEE_UART" value="$(MODEM_PORT_LOWER)"/>
     <define name="XBEE_BAUD" value="$(MODEM_BAUD)"/>
+    <define name="XBEE_CHANNEL" value="$(MODEM_CHANNEL)"/>
```
fvantienen: Why do you want to change it to a `configure`? This is not general for modems and currently only applicable to xbee's. My suggestion is to just remove this line, together with all the board file changes.

conf/boards/apogee_1.0.makefile
```diff
@@ -36,6 +36,7 @@ SYS_TIME_LED       ?= 1
 
 MODEM_PORT ?= UART1
 MODEM_BAUD ?= B57600
+MODEM_CHANNEL ?= 0
```
fvantienen: This is not board specific, but drone specific. A bit strange to have this configure here.

sw/airborne/modules/datalink/xbee_dl.c
```diff
@@ -44,10 +44,10 @@ void xbee_dl_init(void)
 #if USE_HARD_FAULT_RECOVERY
   if (recovering_from_hard_fault)
     // in case of hardfault recovery, we want to skip xbee init which as an active wait
-    xbee_transport_init(&xbee_tp, &((XBEE_UART).device), AC_ID, XBEE_TYPE, XBEE_BAUD, NULL, XBEE_INIT);
+    xbee_transport_init(&xbee_tp, &((XBEE_UART).device), AC_ID, XBEE_TYPE, XBEE_BAUD, XBEE_CHANNEL, NULL, XBEE_INIT);
```
fvantienen: If you remove the configure then add an `#if !defined XBEE_CHANNEL` that does `#define XBEE_CHANNEL 0` it would solve all your problems.

gautierhattenberger:
A less intrusive solution would be to generate or extend the `XBEE_INIT` string to pass the channel settings. This would require no modifications of pprzlink and leave the board files and airframe files unchanged on paparazzi side.

Fabien-B:
I have tested it, it works, and I just added the define documentation.

PR #2296, opened 2018-07-17
[chibios] upgrade to chibios 18.2.2

gautierhattenberger:
port code and boards to latest stable chibios
also fix bug when time_t is 64 bits wide with meteostick
it should also remove most of the warnings with recent GCC 7


podhrmic:
Will do.

podhrmic:
Works on Lisa MX (didn't test HITL, but flashing and telemetry works OK). 
VMS_ECU: I don't have the hardware anymore, so cannot test.
PX4: The ChibiOS port wasn't complete (I think PWM didn't work for me properly), so I suggest merging as is and we can revisit ChibiOS on PX4 later.
Didn't test on Lia/Lisa M but I don't expect any surprises there.
Thanks for making the PR!

sw/airborne/boards/chimera/chibios/v1.0/board.cfg
```diff
@@ -0,0 +1,146 @@
...
+# 'FSMC' => ['ALTERNATE', 'PUSHPULL', 'SPEED_HIGH', 'FLOATING', 'LEVEL_HIGH'],
+# 'SDIO' => ['ALTERNATE', 'PUSHPULL', 'SPEED_HIGH', 'PULLUP', 'LEVEL_HIGH'],
+# 'SDIOCK' => ['ALTERNATE', 'PUSHPULL', 'SPEED_HIGH', 'FLOATING', 'LEVEL_HIGH'],
+# 'CAN' => ['ALTERNATE', 'PUSHPULL', 'SPEED_HIGH', 'FLOATING', 'LEVEL_HIGH'],
+# 'DCMI' => ['ALTERNATE', 'PUSHPULL', 'SPEED_HIGH', 'FLOATING', 'LEVEL_HIGH'],
+# 'LED' => ['OUTPUT', 'PUSHPULL', 'SPEED_VERYLOW', 'FLOATING', 'LEVEL_LOW'],
+# 'PASSIVE' => ['INPUT', 'PUSHPULL', 'SPEED_VERYLOW', 'FLOATING', 'LEVEL_LOW']);
+#
+# SYSTEM
+A13     SWDIO     SYS  AF:SYS_JTMS-SWDIO
```
podhrmic: good to see that ChibiOS came up with a better format for the board config files

gautierhattenberger: Actually, this is not from ChibiOS, it is from my colleague when he get bored building board.h files by hand...

PR #2297, opened 2018-07-17
Solving problems flying with optical flow

guidoAI:
In the current master version, vertical velocity is also determined by the optical flow module, with the help of divergence. However, the current code is not robust - leading to faulty large altitude estimates and very large velocities. This leads to large control inputs, which result in worse vision, which result in even worse estimates... you get the drift.

So, in this pull request, I just changed some things in the bebop2_opticflow.xml (it was set not to update sonar on AGL, etc.) and the opticflow_module.c files to make the hover mode work again out of the box (this was broken due to the addition of the vertical velocity). 

In the future, this pull request should be followed up by a more fundamental solution that takes the reliability of the divergence into account for estimating the vertical velocity well. For now, though, this pull request will allow people to fly with optical flow again.

### DIFF
--- a/sw/airborne/modules/computer_vision/opticflow_module.c
+++ b/sw/airborne/modules/computer_vision/opticflow_module.c
@@ -126,7 +126,7 @@ void opticflow_module_run(void)
                                   0.0f, //opticflow_result.vel_body.z,
                                   opticflow_result.noise_measurement,
                                   opticflow_result.noise_measurement,
-                                  1000.0f //opticflow_result.noise_measurement
+                                  -1.0f //opticflow_result.noise_measurement // negative value disables filter updates with OF-based vertical velocity.
                                  );
     }
     opticflow_got_result = false;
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -123,10 +123,10 @@ void opticflow_module_run(void)
       AbiSendMsgVELOCITY_ESTIMATE(VEL_OPTICFLOW_ID, now_ts,
                                   opticflow_result.vel_body.x,
                                   opticflow_result.vel_body.y,
-                                  opticflow_result.vel_body.z,
+                                  0.0f, //opticflow_result.vel_body.z,
                                   opticflow_result.noise_measurement,
                                   opticflow_result.noise_measurement,
-                                  opticflow_result.noise_measurement
+                                  1000.0f //opticflow_result.noise_measurement
```
kirkscheper: The vertical velocity update can be disabled using a negative value for the noise.

kirkscheper: @guidoAI can you update this?

PR #2299, opened 2018-07-19
[module] add option to digital cam to send second DC_SHOT message

gautierhattenberger:
this is useful to trigger or just send information to an embedded companion CPU in addition to the GCS


gautierhattenberger:
tested and working

PR #2305, opened 2018-07-31
Adding the option to track back flow, adding an error to flow vectors…

guidoAI:
…, adding the show flow option to the xml and user interface.

### DIFF
--- a/sw/airborne/modules/computer_vision/lib/vision/image.c
+++ b/sw/airborne/modules/computer_vision/lib/vision/image.c
@@ -27,6 +27,7 @@
 #include "image.h"
 #include <stdlib.h>
 #include <string.h>
+#include "lucas_kanade.h"
 
 /**
  * Create a new image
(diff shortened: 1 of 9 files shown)


sw/airborne/modules/computer_vision/lib/vision/image.c
```diff
@@ -27,6 +27,7 @@
 #include "image.h"
 #include <stdlib.h>
 #include <string.h>
+#include "lucas_kanade.h"
```
kirkscheper: This dependency is a little strange, in principle the image.c/h should not be dependent on a specific optical flow file. That said, I also don't think the image_show_flow should be in this file either as it is specific to optical flow and not generic images. In principle, the draw vector should be here and the image_show_flow should be wherever the flow functions are implemented.

I won't ask you to change this here as you didn't introduce the core problem and I know you don't have time to fix it but consider this in the future.

kirkscheper:
If you clean up the indentation and run fix_code_style I will merge this.

guidoAI:
Done! 

Indeed a good idea to place the show flow somewhere else. Will look at it in the future (time permitting).

PR #2314, opened 2018-08-20
[gcs] Add auto center aircraft

fvantienen:
This does work but it sometimes draws the waypoints and geofences on the top left corner of the map. 

@gautierhattenberger do you have any idea why this happens?
I guess it is some multi-threading issue, but my ocaml and GCS knowledge is not that great to solve it. Similar behaviour happens when continuously pressing 'ctrl+c' without this feature.


fvantienen:
If you load tiles the problem is really annoying

fvantienen:
Found the issue and fixed ;) So only need to check it now

OpenUAS:
YaY! :+1:

PR #2315, opened 2018-08-23
Derotate flow front camera

guidoAI:
Derotation is also needed when using flow from the frontal camera. However, a simple correction factor applied to the median flow will not work for broad field-of-view lenses such as the Bebop 2 lens. The best way then, is to undistort and derotate individual vectors. This is what I implemented in this pull request. The parameters for the undistortion are stored in the camera-specific file and can hence be set in the airframe file.

### DIFF
--- a/sw/airborne/boards/bebop/mt9f002.c
+++ b/sw/airborne/boards/bebop/mt9f002.c
@@ -67,12 +67,12 @@ struct video_config_t front_camera = {
   .cv_listener = NULL,
   .fps = MT9F002_TARGET_FPS,
   .camera_intrinsics = {
-        .focal_x = MT9F002_FOCAL_X;
-        .focal_y = MT9F002_FOCAL_Y;
-        .center_x = MT9F002_CENTER_X;
-        .center_y = MT9F002_CENTER_Y;
-        .Dhane_k = MT9F002_DHANE_K;
-    }
+    .focal_x = MT9F002_FOCAL_X;
+    .focal_y = MT9F002_FOCAL_Y;
+    .center_x = MT9F002_CENTER_X;
+    .center_y = MT9F002_CENTER_Y;
+    .Dhane_k = MT9F002_DHANE_K;
+  }
 };
 
 /**
--- a/sw/airborne/boards/bebop/mt9f002.c
+++ b/sw/airborne/boards/bebop/mt9f002.c
@@ -67,11 +67,11 @@ struct video_config_t front_camera = {
   .cv_listener = NULL,
   .fps = MT9F002_TARGET_FPS,
   .camera_intrinsics = {
-    .focal_x = MT9F002_FOCAL_X;
-    .focal_y = MT9F002_FOCAL_Y;
-    .center_x = MT9F002_CENTER_X;
-    .center_y = MT9F002_CENTER_Y;
-    .Dhane_k = MT9F002_DHANE_K;
+    .focal_x = MT9F002_FOCAL_X,
+    .focal_y = MT9F002_FOCAL_Y,
+    .center_x = MT9F002_CENTER_X,
+    .center_y = MT9F002_CENTER_Y,
+    .Dhane_k = MT9F002_DHANE_K
   }
 };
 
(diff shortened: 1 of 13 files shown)


sw/airborne/boards/bebop/mt9f002.c
```diff
@@ -65,7 +65,14 @@ struct video_config_t front_camera = {
...
   .cv_listener = NULL,
-  .fps = MT9F002_TARGET_FPS
+  .fps = MT9F002_TARGET_FPS,
+  .camera_intrinsics = {
+    .focal_x = MT9F002_FOCAL_X;
+    .focal_y = MT9F002_FOCAL_Y;
+    .center_x = MT9F002_CENTER_X;
+    .center_y = MT9F002_CENTER_Y;
+    .Dhane_k = MT9F002_DHANE_K;
+  }
```
kirkscheper: You should define this for the bottom cam as well.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
```diff
@@ -528,29 +534,75 @@ bool calc_fast9_lukas_kanade(struct opticflow_t *opticflow, struct image_t *img,
...
+
+    // determine the roll, pitch, yaw differencces between the images.
+    float phi_diff = opticflow->img_gray.eulers.phi - opticflow->prev_img_gray.eulers.phi;
+    float theta_diff = opticflow->img_gray.eulers.theta - opticflow->prev_img_gray.eulers.theta;
+    float psi_diff = opticflow->img_gray.eulers.psi - opticflow->prev_img_gray.eulers.psi;
+
+    if (strcmp(OPTICFLOW_CAMERA.dev_name, "/dev/video0") == 0) {
+      // bottom cam: just subtract a scaled version of the roll and pitch difference from the global flow vector:
+      diff_flow_x = phi_diff * OPTICFLOW_FX;
+      diff_flow_y = theta_diff * OPTICFLOW_FY;
```
kirkscheper: These parameters now look redundant, maybe you can reuse the parameters from the new struct you added. Otherwise people might end up setting the wrong parameters in their airframe. For this, you will have to update the variable names in the airframes that use the old parameter with the new one and update the documentation of the optical flow module.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
```diff
@@ -528,29 +534,75 @@ bool calc_fast9_lukas_kanade(struct opticflow_t *opticflow, struct image_t *img,
...
+      for (int i = 0; i < result->tracked_cnt; i++) {
+        // subtract the flow:
+        vectors[i].flow_x -= predicted_flow_vectors[i].flow_x;
+        vectors[i].flow_y -= predicted_flow_vectors[i].flow_y;
+      }
+    }
   }
 
   float rotation_threshold = M_PI / 180.0f;
+
```
kirkscheper: The code below and above look like they should be rearranged slightly. I would propose, first check that there are enough corners and that the derotation can be applied then apply it with an if else for the camera type. You will have to check and confirm if this will have the desired behaviour.

sw/airborne/boards/bebop/mt9f002.c
guidoAI: At the moment the bottom cam is derotated in the old way. It can be added later in my eyes.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
guidoAI: FX and FY are indeed redundant, but not the correction factor. Also, all airframe files that use OPTICFLOW_FX and OPTICFLOW_FY would have to be changed. The current idea in the pull request is to keep handling the bottom cam as was (since this worked), and just to add a method to derotate flow for the frontal cam.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
guidoAI: Yes, will do.

sw/airborne/modules/computer_vision/opticflow/opticflow_calculator.c
guidoAI: PS: will in any case have a look at it

guidoAI:
Wait one second - still have to fix one thing.

guidoAI:
Thanks for the comments, implemented them. I also removed a double definition of bool success, which led to a warning.

PR #2316, opened 2018-08-24
[mission] add support for custom mission patterns or actions

gautierhattenberger:
Specific navigation functions or even payload actions can be registered in the mission module in order to be called from the mission controller like the built-in navigation patterns.

Each custom element (callback function) is registered/called with an unique string identifier of 5 char max. A maximum of 12 parameters can be passed to the callback function.

An example is provided with the nav_flower navigation function.

### DIFF
--- a/sw/airborne/modules/mission/mission_common.c
+++ b/sw/airborne/modules/mission/mission_common.c
@@ -39,6 +39,14 @@ void mission_init(void)
   mission.insert_idx = 0;
   mission.current_idx = 0;
   mission.element_time = 0.;
+
+  // FIXME
+  // we have no guarantee that nav modules init are called after mission_init
+  // this would erase the already registered elements
+  //for (int i = 0; i < MISSION_REGISTER_NB; i++) {
+  //  mission.registered[i].cb = NULL;
+  //  memset(mission.registered[i].type, '\0', MISSION_TYPE_SIZE);
+  //}
 }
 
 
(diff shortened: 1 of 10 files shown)


podhrmic:
@gautierhattenberger is the idea here that with the custom mission we can create/control the flight plan more or less dynamically and add/remove elements as needed in flight?

sw/airborne/modules/mission/mission_common.c
```diff
@@ -39,6 +39,14 @@ void mission_init(void)
   mission.insert_idx = 0;
   mission.current_idx = 0;
   mission.element_time = 0.;
+
+  // FIXME
```
fvantienen: Can't you just check at the start of the `mission_register` if the initialisation is done, and else perform it?

gautierhattenberger: As long as the status of the memory at startup is not predictable, I don't see how you can tell if a non-zero value is legitimate or not.
I'm still working on the build system to integrate a dependency mechanism in the end. Then it will be possible to guarantee that mission is initialized for other modules using it. In the meantime, I don't have a clean solution.

gautierhattenberger:
@podhrmic the mission mode itself is made to control a sequence of action (instead of flight plan blocks). The custom patterns just extend the set of basic patterns (point, circle, line) to virtually any kind of patterns and even actions. It is really dedicated to automatic task and mission planing systems.

sw/airborne/modules/mission/mission_common.c
fvantienen: I though about adding a static value `initialised` from start which you set to `false`. Then you can check this during the adding.

fvantienen: And to ensure that it is false fro . the start you can do something like this:
```
struct _mission mission = {
  .insert_idx = 0,
  .current_idx = 0,
  .element_time = 0,
  .initialised = 0
};
```

gautierhattenberger: That's an idea, thank. Actually it seems I can set the whole structure to 0 statically with `{0}`

PR #2319, opened 2018-09-03
Added Disco camera support and more fixes

OpenUAS:


gautierhattenberger:
it would be much better if you don't mix several pull requests into one

sw/airborne/modules/digital_cam/dc_ctrl_parrot_mykonos.c
```diff
@@ -0,0 +1,311 @@
...
+
+/* Execute the Shoot, Record and Stream commands */
+void dc_ctrl_parrot_mykonos_command(enum dc_ctrl_parrot_mykonos_status cmd)
+{
+  int ret __attribute__((unused));
+
+  dc_ctrl_parrot_mykonos.status = cmd;
+  switch (cmd) {
+    case DC_CTRL_PARROT_MYKONOS_RECORD_START:
+#ifndef SITL
```
gautierhattenberger: you should make a function or a macro that does the system calls or nothing when doing sim

OpenUAS: Yes, already stage 3 and 4 in the making with more changes.  A separate pull  was the original plan but already so much other branches changes etc. "rotting" in the repo since it kind of busy. And then another week goes by, and another. Still not even came around a PR for all the Parrot minidrones and that was last year October.. Well will find a way to have more frequent to the point PR's somehow

OpenUAS: Basically was rewriting the whole DC.h interface and have universal API and support for multiple cam with same DC routines for e.g. Distance and feedback this to the GCS GUI. Halfway the change realized the other changes will not be in master anytime soon if this would be added. A better flying ETECS Disco is needed for tests for the OBC2018 dynamic avoidance. So ripped out the new work on DC so Disco fixes would land up in master soon.

gautierhattenberger:
@OpenUAS again, please don't merge your PR yourself, especially with changes requests pending, unless you receive explicit agreement from other developers.
This is not the first time this summer, and last time it broke the default behavior of the airspeed sensor used on chimera boards.

PR #2320, opened 2018-09-05
Mekf Wind INS

gautierhattenberger:
This new module is a full INS including wind estimation based on IMU/GPS/baro + airspeed/AoA/Sideslip measurements with MEKF filter.
The integration is using the C++ Matrix library Eigen included as a submodule.

Some minor changes:
- airspeed and incidence sensors using ABI
- improve some drivers and system configuration
- example airframe, tested in flight

Reference:
Martin Brossard, Jean-Philippe Condomines, Silvère Bonnabel. Tightly coupled navigation and wind estimation for mini UAVs. AIAA GNC 2018, AIAA Guidance, Navigation, and Control Conference, Jan 2018, Kissimee, United States.
https://hal-enac.archives-ouvertes.fr/hal-01814261

### DIFF
--- a/sw/airborne/pprz_syscalls.c
+++ b/sw/airborne/pprz_syscalls.c
@@ -1,0 +1,174 @@
+/*
+ * Copyright (C) 2017 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Fake system calls
+ *
+ * based on:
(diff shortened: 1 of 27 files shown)


sw/airborne/pprz_syscalls.c
```diff
@@ -0,0 +1,174 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <signal.h>
+
+/***************************************************************************/
+
+#pragma GCC diagnostic ignored "-Wmissing-prototypes"
+
+#ifndef USE_CHIBIOS_RTOS // already defined by chibios syscalls
```
podhrmic: just to clarify, is this file used as a compatibility layer for some modules that use ChibiOS? If so - wouldn't it make more sense to throw a compilation error, since without ChibiOS such code might compile but won't run properly?

gautierhattenberger: I did that some time ago, so I might be wrong, but I actually think the problem comes from Eigen. This `ifndef` makes it possible to use the code on top of ChibiOS or bare-metal. ChibiOS already defines some syscalls so they will be used in this case. If you use libopencm3 you need to define all of them. Still some of them were missing from ChibiOS version (getpid, kill, exit).
The best would actually be to disable them completely from Eigen but apparently some print/assert can't be turned off.

noether:
I am thrilled about seeing this PR! 

By the way, what is the status of using Eigen while debugging? Can you compile with -o0, make some complex operations such as inverses and still print the values of matrices? I remember you were facing the same problems as me several years ago with Eigen in the microcontroller (and I never figured out the why of the problem).

noether:
Fair enough indeed. Yes, the problems were/are in the ARM, not in the simulation (and to be honest, it is where I check/validate the numerical implementation of an algorithm).

Thanks again for this PR. In fact, it would make much easier the implementation of the GVF in 3D for rotorcraft.

gautierhattenberger:
@podhrmic do you validate the PR ?

podhrmic:
Yes I do:-)

PR #2325, opened 2018-09-11
Ransac

guidoAI:
Added an easy wrapper around the pprz_svd_solve_float method, called fit_linear_model, to which you can pass a 1D target array and a 2D array of n_samples x Dimension for making a linear fit.

Around this linear fit function, I also made a random sample consensus method (RANSAC), which is a robust way of doing the linear fit. The fit will then still give good results for a modest number of outliers.

### DIFF
--- a/sw/airborne/math/RANSAC.c
+++ b/sw/airborne/math/RANSAC.c
@@ -25,6 +25,11 @@
  * The concept is to select (minimal) subsets of the samples for learning
  * a model, then determining the error on the entire data set, and selecting the best fit. In determining the error, the individual sample errors
  * are capped, so that outliers can be identified and have only a modest influence on the results.
+ *
+ * Read: Fischler, M. A., & Bolles, R. C. (1981). Random sample consensus: a paradigm for model fitting with applications to image analysis and automated cartography.
+ * Communications of the ACM, 24(6), 381-395.
+ *
+ * This file depends on the function fit_linear_model in math/pprz_matrix_decomp_float.h/c
  */
 
 
@@ -119,6 +124,7 @@ void RANSAC_linear_model(int n_samples, int n_iterations, float error_threshold,
  * @param[in] sample The sample vector of size D
  * @param[in] weights The weight vector of size D+1
  * @param[in] D The dimension of the sample.
+ * @return The predicted value
  */
 float predict_value(float *sample, float *weights, int D, bool use_bias)
 {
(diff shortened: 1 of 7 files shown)


sw/airborne/math/RANSAC.c
```diff
@@ -0,0 +1,169 @@
...
+  return sum;
+}
+
+/** Get indices without replacement.
+ *
+ * @param[in] indices_subset This will be filled with the sampled indices
+ * @param[in] n_samples The number of samples / indices.
+ * @param[in] count The function will sample n_sample numbers from the range 1, 2, 3,..., count
+ */
+
```
fvantienen: This can optionally be removed for style ;)

fvantienen:
I see now that the `math` folder is compiled during the tests with `gnu99`. In order to fix the continuous integration it might be best to change this to the new c++11? @gautierhattenberger

sw/airborne/math/RANSAC.c
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  // printf("Prediction = %f\n", sum);
+
+  return sum;
+}
+
+/** Get indices without replacement.
+ *
+ * @param[in] indices_subset This will be filled with the sampled indices
```
gautierhattenberger: this param should be labeled `[out]`

sw/airborne/math/RANSAC.c
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+}
+
+/** Predict the value of a sample with linear weights.
+ *
+ * @param[in] sample The sample vector of size D
+ * @param[in] weights The weight vector of size D+1
+ * @param[in] D The dimension of the sample.
+ */
```
gautierhattenberger: add `@return` doc

fvantienen:
`gnu99` and `c99` are not the same, so since we compile all our other code with `gnu99` my advise is to go with that.

gautierhattenberger:
@kirkscheper yes, this fix should work

sw/airborne/math/RANSAC.c
guidoAI: I don't know what you mean with this... part of the doxygen?

sw/airborne/math/RANSAC.c
guidoAI: Done

sw/airborne/math/RANSAC.c
guidoAI: Done

guidoAI:
Added the requested lines to the Makefile, and modified the comments according to the suggestions.

Issue #2270, opened 2018-05-31
Spektrum RX with Spektrum serial out not working


OpenUAS:
![spektrumrxtest](https://user-images.githubusercontent.com/483944/40798387-38db43b2-650b-11e8-8549-9b71f9fddaa6.jpg)

v5.12 all OK, switching to master no RC anymore... Exact same hardware, just reflashed master instead of v5.12.

Using Official Spektrum RX, but have about 30+ different Receiver to try if needed. Also +5 different TX is needed.

More info can be added to this issue when needed, plz ask. 

Ongoing investigation. Will look at all (board)code, do logic analyser dump, etc. etc.

As Autopilot boards to test on we have all variants of Lisa M (v1.0, V2, v2,1, MX etc. also other AP boards when needed ) Well borrowed out the original Lisa L, but that is fine i guess ;() ...

gautierhattenberger:
closed with #2326

PR #2326, opened 2018-09-13
[radio] use DSM2 as default binding protocol

gautierhattenberger:
This was the previous behavior before rewritting the spektrum driver.
DSMX mode should only be used for compatible receivers and transmitters,
otherwise the system type will be be detected properly and frames will
be discarded.
This should solve #2270


gautierhattenberger:
@fvantienen or @OpenUAS can you check that this is solving the Spektrum issue mentioned in #2270 ?

gautierhattenberger:
can I merge this one ?

gautierhattenberger:
someone to check this ? otherwise I'll merge this since it is not working at the moment anyway.

gautierhattenberger:
tested again and really needed, DSM2 is default like it was before

Issue #2328, opened 2018-09-14
git branch cleaning


gautierhattenberger:
We have a bunch of branches in the paparazzi repository that are not pending pull request. I suggest that the authors of these branches decide if they are worth keeping or not:

  remotes/paparazzi/bitshift_cleanup
  remotes/paparazzi/blos
  remotes/paparazzi/campaign2011
  remotes/paparazzi/campaign2013
  remotes/paparazzi/coverity_scan
  remotes/paparazzi/cpp_support
  remotes/paparazzi/crypto_link
  remotes/paparazzi/cyberzoo_gazebo
  remotes/paparazzi/dshot_f4
  remotes/paparazzi/feature4_fixedwing_in_rotor
  remotes/paparazzi/gec_module_makefile
  remotes/paparazzi/lpe
  remotes/paparazzi/message_scheduling
  remotes/paparazzi/messages
  remotes/paparazzi/mission_simulation
  remotes/paparazzi/ocamlbuild
  remotes/paparazzi/rc_datalink
  remotes/paparazzi/revert-2010-fix_configs
  remotes/paparazzi/secure_link
  remotes/paparazzi/secure_link_tmp
  remotes/paparazzi/sonarqube_test
  remotes/paparazzi/state_utm

Some of them seems interesting, so they could be opened as PR

dewagter:
cleaned up mine

podhrmic:
cleaned up mine, see the related PR #2332 
also, keeping the `coverity_scan` branch for now - not sure if anyone (@kirkscheper ?) is using it.

gautierhattenberger:
closing this as most of the branches have been either removed or kept on purpose

PR #2332, opened 2018-09-19
Local Position Estimator

podhrmic:
Based on #2328 I am putting this PR here for discussion.
This is an old attempt to copy the Local Position Estimator (INS) from Pixhawk into Paparazzi. Back in time I didn't have C++ support, now it would be much nicer with Eigen. LPE uses lidar and sonar data on top of (optional) GPS to be used for sub-meter precision.

If I remember it right,  the performance was comparable with the other INS's, but I am not sure if it was actually great.

Anybody interested in polishing this or using it?

### DIFF
--- a/sw/airborne/modules/ins/ins_lpe.c
+++ b/sw/airborne/modules/ins/ins_lpe.c
@@ -1,0 +1,263 @@
+/*
+ * Copyright (C) 2016 Michal Podhradsky <michal.pohradsky@aggiemail.usu.edu>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+/**
+ * @file modules/ins/ins_lpe.c
+ *
+ * Local Position Estimator
+ */
+#include "modules/ins/ins_lpe.h"
+#include "subsystems/abi.h"
+#include "mcu_periph/sys_time.h"
+#include "message_pragmas.h"
+
+#include "state.h"
+
+#ifndef USE_INS_NAV_INIT
+#define USE_INS_NAV_INIT TRUE
+PRINT_CONFIG_MSG("USE_INS_NAV_INIT defaulting to TRUE")
+#endif
+
+/*
+ * ABI bindings
+ */
+/** baro */
+#ifndef INS_LPE_BARO_ID
+#if USE_BARO_BOARD
+#define INS_LPE_BARO_ID BARO_BOARD_SENDER_ID
+#else
+#define INS_LPE_BARO_ID ABI_BROADCAST
+#endif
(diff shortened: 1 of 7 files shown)


sw/airborne/modules/ins/ins_lpe.c
```diff
@@ -0,0 +1,533 @@
...
+#include "message_pragmas.h"
+#include "subsystems/imu.h"
+
+#ifndef USE_INS_NAV_INIT
+#define USE_INS_NAV_INIT TRUE
+PRINT_CONFIG_MSG("USE_INS_NAV_INIT defaulting to TRUE")
+#endif
+
+#ifndef P_MAX // max allowed value in state covariance
+#define P_MAX 1.0e6
```
gautierhattenberger: for single precision numerical constant, it is usually better to specify it with `1.0e6f`, same for the rest of the code

sw/airborne/modules/ins/ins_lpe.c
```diff
@@ -0,0 +1,533 @@
...
+
+
+/**
+ * Reset states (matrices and vectors) back to
+ * their intial value
+ */
+void ins_lpe_zero_matrices(void)
+{
+    // initialize process matrix A 9x9
+    ins_lpe.A.initialized = FALSE; // not initialized yet
```
gautierhattenberger: it is advise to use lowercase `true` and `false` now

sw/airborne/modules/ins/ins_lpe.c
```diff
@@ -0,0 +1,533 @@
...
+    ins_lpe.u[U_az] = 0;
+  }
+
+  // update A with new body-to-NED roration matrix
+  ins_lpe_update_ssstates();
+
+
+  // continuous time kalman filter prediction
+  // integrate runge kutta 4th order
+  // TODO move rk4 algorithm to matrixlib
```
gautierhattenberger: there is already a RK lib in paparazzi: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/math/pprz_rk_float.h
it is used in the ins_invariant filter: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/ins/ins_float_invariant.c#L347

sw/airborne/modules/ins/ins_lpe.c
```diff
@@ -0,0 +1,533 @@
...
+/**
+ * @file modules/ins/ins_lpe.c
+ *
+ * Local Position Estimator
+ */
+#include "modules/ins/ins_lpe.h"
+#include "subsystems/abi.h"
+#include "mcu_periph/sys_time.h"
+#include "message_pragmas.h"
+#include "subsystems/imu.h"
```
gautierhattenberger: it shouldn't be needed when using ABI messages

sw/supervision/paparazzicenter.glade
```diff
@@ -1171,6 +1171,7 @@
                                 <property name="can_focus">True</property>
                                 <property name="receives_default">False</property>
                                 <property name="use_action_appearance">False</property>
+                                <accelerator key="c" signal="activate" modifiers="GDK_MOD1_MASK"/>
```
gautierhattenberger: it would be better if it was in a separated pull request since not related to the ins

sw/airborne/modules/ins/ins_lpe.h
```diff
@@ -0,0 +1,151 @@
...
+};
+
+struct InsOpticflow {
+  bool data_available;
+  uint32_t timestamp;
+  struct FloatVect3 vel_body;
+};
+
+
+struct InsLpe {
```
gautierhattenberger: is it really necessary to have all these structure public ? It would be a bit safer to have them in the .c file and only have here the required extern definition if any

podhrmic:
@gautierhattenberger I am leaning towards closing this, and perhaps make a new/better position estimator in Rust later.

gautierhattenberger:
fine, closing this then

PR #2334, opened 2018-09-25
Improvements to snake gate detection and subsequent pose estimation

guidoAI:
Improvements to the gate detection and pose estimation, including:
- Allowing snake gate to produce multiple simultaneous gate detections
- Better detection of slanted gates by remembering the snake end locations
- Bug fixes in the gate pose estimation


sw/airborne/modules/computer_vision/snake_gate_detection.c
```diff
@@ -46,14 +46,14 @@
 #include "paparazzi.h"
 
 // to debug the algorithm, uncomment the define:
-// #define DEBUG_SNAKE_GATE
+#define DEBUG_SNAKE_GATE
```
gautierhattenberger: do you really want to keep this enabled by default ? all the other ones are disabled

guidoAI: Good catch! Changed it.

PR #2335, opened 2018-10-04
Gazebo model of the 2018 dronerace track

dewagter:


gautierhattenberger:
it seems that the 'push' build is fine, so it is probably an error on travis/ubuntu side, or you think it is something else ?

kirkscheper:
Ok! Merge when ready

PR #2337, opened 2018-10-04
Spike filter for sonar: e.g. to pass through drone-race gates

dewagter:


sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -205,3 +226,37 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
+  }
+  previous_distance = current_distance;
+  if (distance > 5) {
+    distance = 5;
+  }
+  return distance;
+}
+
+void sonar_bebop_clock()
+{
```
kirkscheper: Why not sure the functions in sys_time?

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -205,3 +226,37 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
+    through_gate_green_light = 0;
+  }
+
+  if (through_gate_green_light == 1) {
+    distance = z0;
+  } else {
+    distance = distance_sonar;
+  }
+  previous_distance = current_distance;
+  if (distance > 5) {
```
kirkscheper: This is very strange hard coded value. There are already defines in the ins to only consider sonar values in a certain range.

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -205,3 +226,37 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
+  float distance;
+  current_distance = distance_sonar;
+  diff_pre_cur = current_distance - previous_distance;
+  if (diff_pre_cur < -0.4) {
+    z0 = previous_distance;
+    through_gate_green_light = 1;
+    counter_temp3 = 0;
+    time_temp3 = 0;
+  }
+  if (diff_pre_cur > 0.4 || time_temp3 > 1.0) {
```
kirkscheper: Both these parameters should be user adjustable

dewagter:
- opt-in added with a description in module.xml: default = not active
 - parameters are now configurable
 - periodic function changed to sys_time
 - bound of 5m is replaced by the default max-sonar behavior
 - removed some variables

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -208,3 +216,43 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
+
+static float sonar_filter_narrow_obstacles(float distance_sonar)
+{
+  static float previous_distance = 0;
+  static float z0 = 0;
+  bool through_gate_green_light;
+  float diff_pre_cur = distance_sonar - previous_distance;
+  if (diff_pre_cur < -SONAR_BEBOP_FILTER_NARROW_OBSTACLES_JUMP) {
+    z0 = previous_distance;
+    through_gate_green_light = 1;
```
kirkscheper: Variable names with gate here are out of place, maybe use obstacle, also what is green light supposed to mean here?

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -208,3 +216,43 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
+    SysTimeTimerStart(sonar_bebop_spike_timer);
+  }
+  previous_distance = distance_sonar;
+  float time_since_reset = SysTimeTimer(sonar_bebop_spike_timer);
+  if ((diff_pre_cur > SONAR_BEBOP_FILTER_NARROW_OBSTACLES_JUMP) || (time_since_reset > USEC_OF_SEC(SONAR_BEBOP_FILTER_NARROW_OBSTACLES_TIME)) ) {
```
kirkscheper: Do you think the usec(fixed_define) gets properly optimised as a constant by the compiler? Of not consider precomputing the constant or using get_sys_time_float instead of the local timer.

dewagter: Looking at the code, I think the precompiler does convert it. ```get_sys_time_float``` on the other hand needs runtime conversions.

dewagter:
- rebased
 - no more gate variable names

dewagter:
About the INS:

 I totally agree that the INS has the most information and thereby can best interpret the distance measurement. This includes information about attitude and the distance not being perpendicular.

I also agree that this code might be used with other distance sensors like laser range.

Add on top of that, that the sonar should actually find several echo's. When it is passing a gate, it also has an echo for the ground.

On the down-side:
 - there are many INS and it should be added to all
 - the INS is already very complex
 - the sonar should actually give a list of echo's instead of just 1

Therefore I still opt to make a simple spike filter which is adapted to the given sensor, with a documented define in the module header to give the best chance of correct interpretation and use, given the complexity of the project.

Should anyone have the possibility to address the upgrades discussed above, please feel free to do so.

PR #2338, opened 2018-10-04
Camera intrinsics are not initialized for Gazebo

dewagter:


sw/simulator/nps/nps_fdm_gazebo.cpp
```diff
@@ -693,6 +693,11 @@ static void init_gazebo_video(void)
     cameras[i]->sensor_size.h = cam->ImageHeight();
     cameras[i]->crop.w = cam->ImageWidth();
     cameras[i]->crop.h = cam->ImageHeight();
+    cameras[i]->camera_intrinsics.focal_x = cameras[i]->output_size.w/2.0f;
+    cameras[i]->camera_intrinsics.center_x = cameras[i]->output_size.w/2.0f;
+    cameras[i]->camera_intrinsics.focal_y = cameras[i]->output_size.h/2.0f;
+    cameras[i]->camera_intrinsics.center_y = cameras[i]->output_size.h/2.0f;
+    cameras[i]->fps = cam->UpdateRate();
```
kirkscheper: Why did you move this line? Like this it is not clear that the fps is not overwritten if the MT9F002 cam is used.

dewagter: I don't know why Guido did this but I agree FPS should always be written.

PR #2340, opened 2018-10-04
Linear-fit update

dewagter:
- safeguards for divide by zero
 - silent compile warnings
 - linear-fit with prior information


dewagter:
@guidoAI can you please update this

dewagter:
- fixed linear_fit_with_prior
 - DEBUG_PRINT to allow compiling on MCU
 - all comments that will not be needed for using the code or updating the code (e.g. to priors with D > 2) are removed.

PR #2345, opened 2018-10-05
[fix] some extra warning, error and failing conf fixed

gautierhattenberger:
some extra errors and warnings from CI builds


gautierhattenberger:
thanks for adding missing airframe

PR #2347, opened 2018-10-08
[ublox] Hw-Sw version info and Config Info on startup

dewagter:
Since the Ublox 7 series, several constellations are supported.
Since the Ublox 8 series, constellations can be enabled at the same time.
It now becomes important to know which version is in your plane and which constellation is activated.
A dedicated message will be sent on startup.


dewagter:
Depends on https://github.com/paparazzi/pprzlink/pull/93

PR #2349, opened 2018-10-18
Minor stylistic fixes in vectornav and nps

podhrmic:


sw/simulator/nps/nps_autopilot_fixedwing.c
```diff
@@ -157,8 +157,8 @@ void nps_autopilot_run_step(double time)
     float dist = (float) sensors.sonar.value;
     AbiSendMsgAGL(AGL_SONAR_NPS_ID, dist);
 
-    uint16_t foo = 0;
 #ifdef SENSOR_SYNC_SEND_SONAR
+    uint16_t foo = 0;
```
gautierhattenberger: this fix is probably not rebased since this is already fixed in master

podhrmic:
hmm,  so maybe I already pushed it. Weird...

PR #2350, opened 2018-11-05
Supervision tools

Fabien-B:
The purpose of this PR is to move the tools configuration (partially) out of the control_panel.xml.
The default tool configuration will be in individual files (one for each tool) in conf/tools/.
You can then override the default configuration in your control_panel.xml (eg. to change the default arguments or add custom tools). You can also add the `blacklisted="true"` attribute to the tools in your control_panel.xml to hide it in the paparazzi center.

This way, you will be able to see the new tools without changing your control_panel.xml, and you can make a custom configuration without affecting the default one.

### DIFF
--- a/sw/supervision/pc_control_panel.ml
+++ b/sw/supervision/pc_control_panel.ml
@@ -51,14 +51,19 @@ let tools_xml = List.map (fun f -> ExtXml.parse_file f) tool_files
 let programs =
   let h = Hashtbl.create 7 in
   let s = ExtXml.child ~select:(fun x -> Xml.attrib x "name" = "programs") control_panel_xml "section" in
+  let b = List.filter (fun p -> String.equal (ExtXml.attrib_or_default p "blacklisted" "false") "true") (Xml.children s) in (*List blacklisted programs*)
   (*Adds tools to h*)
   List.iter
     (fun p -> Hashtbl.add h (ExtXml.attrib p "name") p)
     tools_xml;
-    (*Overwrite tools in h by the custom configuration from control_panel.xml*)
+  (*Overwrite tools in h by the custom configuration from control_panel.xml*)
   List.iter
     (fun p -> Hashtbl.replace h (ExtXml.attrib p "name") p)
     (Xml.children s);
+  (*Remove blacklisted programs*)
+  List.iter
+    (fun p -> Hashtbl.remove h (ExtXml.attrib p "name"))
+    b;
   h
 
 let program_command = fun x ->
--- a/sw/supervision/pc_control_panel.ml
+++ b/sw/supervision/pc_control_panel.ml
@@ -28,14 +28,32 @@ module Utils = Pc_common
 
 let (//) = Filename.concat
 
+(*Search recursively files in a directory*)
+let dir_contents dir =
+  let rec loop result = function
+    | f::fs when Sys.is_directory f ->
+          Sys.readdir f
+          |> Array.to_list
+          |> List.map (Filename.concat f)
+          |> List.append fs
+          |> loop result
+    | f::fs -> loop (f::result) fs
+    | []    -> result
+  in
+loop [] [dir]
+
 let control_panel_xml_file = Utils.conf_dir // "control_panel.xml"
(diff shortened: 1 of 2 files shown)


sw/supervision/pc_control_panel.ml
```diff
@@ -28,14 +28,42 @@ module Utils = Pc_common
...
 let control_panel_xml_file = Utils.conf_dir // "control_panel.xml"
 let control_panel_xml = ExtXml.parse_file control_panel_xml_file
+let xml_regex = Str.regexp ".*xml"
+let tool_files = List.filter (fun s -> Str.string_match xml_regex s 0) (dir_contents (Utils.conf_dir // "tools")) (*List all xml files in conf/tools/*)
+let tools_xml = List.map (fun f -> ExtXml.parse_file f) tool_files 
+
 let programs =
   let h = Hashtbl.create 7 in
   let s = ExtXml.child ~select:(fun x -> Xml.attrib x "name" = "programs") control_panel_xml "section" in
+  let b = List.filter (fun p -> String.equal (ExtXml.attrib_or_default p "blacklisted" "false") "true") (Xml.children s) in (*List blacklisted programs*)
```
gautierhattenberger: replace `String.equal` with `String.compare (...) "true" == 0`

sw/supervision/pc_control_panel.ml
```diff
@@ -28,14 +28,42 @@ module Utils = Pc_common
...
+          |> loop result
+    | f::fs -> loop (f::result) fs
+    | []    -> result
+  in
+loop [] [dir]
+
 let control_panel_xml_file = Utils.conf_dir // "control_panel.xml"
 let control_panel_xml = ExtXml.parse_file control_panel_xml_file
+let xml_regex = Str.regexp ".*xml"
+let tool_files = List.filter (fun s -> Str.string_match xml_regex s 0) (dir_contents (Utils.conf_dir // "tools")) (*List all xml files in conf/tools/*)
```
gautierhattenberger: also possible with `List.filter (Filename.check_suffix ".xml") (...)`

sw/supervision/pc_control_panel.ml
```diff
@@ -28,14 +28,42 @@ module Utils = Pc_common
 
 let (//) = Filename.concat
 
+(*Search recursively files in a directory*)
+let dir_contents dir =
+  let rec loop result = function
+    | f::fs when Sys.is_directory f ->
+          Sys.readdir f
```
gautierhattenberger: just in case this syntax doesn't work with older Ocaml versions: https://rosettacode.org/wiki/Walk_a_directory/Recursively#OCaml

sw/supervision/pc_control_panel.ml
Fabien-B: Done

sw/supervision/pc_control_panel.ml
Fabien-B: Done

sw/supervision/pc_control_panel.ml
Fabien-B: No longer needed thanks to the new method for walking in directories.

gautierhattenberger:
Maybe the conversion script should be applied to the control_panel example file (and remove the corresponding section in the file)

gautierhattenberger:
execution fails if directory `conf/tools` doesn't exist...

Fabien-B:
I propose I run the script myself with the latest control_panel_example.xml, commit all the tool files, then remove the tools from the control_panel_exemple.xml (I will leave a commented custom configuration example).
Current configuration of every body will not be changed, and new users will use the tools files.

gautierhattenberger:
@dewagter @podhrmic What do you think about this evolution ? The improvements are:
- it is easier to add new tools that will be immediately visible for the users
- you can hide tools that you don't use
- you can set different default using the control_panel file
- it is backward compatible so you you don't change your current control_panel you keep every like before

gautierhattenberger:
anyone for a comment of this modification ?

kirkscheper:
Looks good!

podhrmic:
Looks good to me:-)

PR #2353, opened 2018-11-24
Indicate when HITL is not getting AP data

### LINKED ISSUE #2265: Indicate when HITL is not getting AP data
ideally through some kind of HITL message showing that the autopilot data are missing. That helps with debugging.

podhrmic:
Fixes #2265

### DIFF
--- a/sw/simulator/nps/nps_main_hitl.c
+++ b/sw/simulator/nps/nps_main_hitl.c
@@ -214,9 +214,24 @@ void *nps_ap_data_loop(void *data __attribute__((unused)))
   struct pprz_transport pprz_tp_logger;
 
   pprz_transport_init(&pprz_tp_logger);
+  uint32_t rx_msgs = 0;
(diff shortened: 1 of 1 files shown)


sw/simulator/nps/nps_main_hitl.c
```diff
@@ -214,9 +214,24 @@ void *nps_ap_data_loop(void *data __attribute__((unused)))
   struct pprz_transport pprz_tp_logger;
 
   pprz_transport_init(&pprz_tp_logger);
+  uint32_t rx_msgs = 0;
+  uint32_t rx_commands = 0;
+  uint32_t rx_motor_mixing = 0;
+  uint8_t show_stats = 1;
```
kirkscheper: Seems useful if show_stats was configurable from the airframe file.

podhrmic: I know, this is just the bare minimum fix to give the user at least some feedback.

PR #2355, opened 2018-11-25
Corroded Paparazzi

podhrmic:
This is a tracking PR for my efforts of bringing Rust to Paparazzi. Why is this useful? Besides being an interesting demonstration of technology, the main benefit is that Rust brings high level language abstraction (types, traits, generics) into embedded world without the need of garbage collection. Since Rust is memory safe, we don't have to worry (at least not too much) about using heap for data allocation/deallocation. Rust is also fast (zero-cost abstraction, no GC), so performance is comparable with native C code. And the Rust compiler checks lots of things for you, so it is harder to make a mistake. So I am hoping Rust can make writing autopilot code more productive, and also safer.

The current state is: 
- Rust's allocator is backed by ChibiOS's API for heap allocation/deallocation
- Simple rust module is being called 

TODO:
- better way of building the rust libraries (currently I need to call `xargo` separately, not sure yet how to wrap it into the build system)
- create a paparazzi-compatibility Rust crate that has the relevant C calls (for state/control etc)

The goal is to create a new autopilot (similar to the "static autopilot") with perhaps some new cool control system that is easier to create in Rust.

I am looking for early feedback. @gautierhattenberger


gautierhattenberger:
I'm fine with the general idea but will need more time to look at this.
Some general comments:
 - is it really necessary to add some many airframe files ? if yes, please follow recommendation to merge modules to firmware node (and resolve conflicts anyway)
 - what about key generator ? is it incompatible or handled in a more generic way ?
 - I don't know much about rust build system, can you explain the difference between Cargo and Xargo ?

podhrmic:
> is it really necessary to add some many airframe files ? if yes, please follow recommendation to merge modules to firmware node (and resolve conflicts anyway)

My bad, I will cut it down to a reasonable sample.

> what about key generator ? is it incompatible or handled in a more generic way ?

I am planning to redo it anyway, I don't like the way it is structured now. I left some comments in the Makefile, hopefully explaining some ideas.

> I don't know much about rust build system, can you explain the difference between Cargo and Xargo ?

In short, `cargo` is the main build tool, while `xargo` is a "cargo that works for embedded environments". However, `xargo`'s functionality is being merged into `cargo`, so this is sort of an intermediate step. Eventually all `xargo` calls will be replaced with `cargo` calls. Sorry it is confusing, but Rust is still figuring out the best way to handle embedded targets.

podhrmic:
Update: I removed the excess airframe configs, and cleaned up the build process for rust components (I can now use Cargo only). Note about the keygen - since it can be useful in many situations where encryptions is needed, I keep it as a separate one-off build. I can't think of other Rust modules that I would need to run at compilation but not need to link them to the final binary.

The next step is to include Chibios-Contrib repo so I can have random number generator on STM32F4.

podhrmic:
@gautierhattenberger This is ready for a review. The main contribution is adding cargo builds into paparazzi build process, so you can add rust modules. The provided demo module doesn't do anything useful, it is for demonstration only. I will add more useful modules later.

There is a few paparazzi compatibility modules, that make writing Rust for Paparazzi simpler. Small changes and added API in those are expected (like add more state interface functions, or mutexes, etc).

I will also make a wiki page with proper documentation.

I would be stoked to have this in 5.14

sw/airborne/arch/chibios/halconf.h
```diff
@@ -258,6 +257,48 @@
...
+/*===========================================================================*/
+/* DAC driver related settings.                                              */
+/*===========================================================================*/
+
+/**
+ * @brief   Enables synchronous APIs.
+ * @note    Disabling this option saves both code and data space.
+ */
+#if !defined(DAC_USE_WAIT) || defined(__DOXYGEN__)
+#define DAC_USE_WAIT                        TRUE
```
gautierhattenberger: how are this options related to the current PR ? Is it generating extra code if the DAC driver is not activated ?

podhrmic: Fixed now.

PR #2357, opened 2018-11-26
Update gazebo from v7 to v8/v9 compatibility

kirkscheper:
Gazebo 7 is not supported on Ubuntu 18.04 (Bionic), this PR updates the function calls in the nps fdm to be compatible with Gazebo 8 and Gazebo 9. With this, gazebo is no longer compatible with installations on systems earlier than Ubuntu 16.04 (Xenial).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: c
 compiler:
   - gcc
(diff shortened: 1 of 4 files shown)


kirkscheper:
The docker for semaphoreci needs updating, @gautierhattenberger any advice on how I can do that?

dewagter:
also change gazebo7 into 9 in https://github.com/paparazzi/paparazzi/blob/master/.travis.yml

gautierhattenberger:
I think you need to make a PR to update this file: https://github.com/paparazzi/paparazzi/blob/master/docker/ci/Dockerfile
Once in master, it should trigger a new build of the docker image

kirkscheper:
Ok, well in that case this should be ready to merge, semaphore should update once merged.

Issue #2313, opened 2018-08-09
Vsupply resolution increase to 0.01V?


matejkarasek:
Currently the resolution of battery voltage (electrical.vsupply) is 0.1V, which is rather low if you fly on 1S battery. 
https://github.com/paparazzi/paparazzi/blob/018df2fb301c5960f8b857f03498e64d4cff6050/sw/airborne/subsystems/electrical.c#L128

How about increasing it to 0.01V? uint16 would still allow to store max 655V, so more than enough... Are there some reasons against this?

podhrmic:
You would have to change the units for each pprzlink message that uses `vsupply`: https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml#L803 (that is easy).

The bigger task is to change failsafe code that compares vsupply to some threshold to make it aware that we have a different resolution now (or convert from centivolts to decivolts as needed).

Indeed this wouldn't be backwards compatible (so we should make users aware of this change).

Otherwise I don't have any objections.

matejkarasek:
The failsafe code seems to be in the same file, with the hardcoded multiplication by 10
https://github.com/paparazzi/paparazzi/blob/018df2fb301c5960f8b857f03498e64d4cff6050/sw/airborne/subsystems/electrical.c#L184
So an easy fix there, just a different multiplier...

grep on 'electrical.vsupply' doesn't give that many hits, so even the rest of the code can be fixed easily

Of course, backward compatibility would be an issue (that's why I asked here...)

podhrmic:
I think it is a fine change for v5.13 release, just have to be transparent about it. @gautierhattenberger @fvantienen what do you think?

gautierhattenberger:
All the ground code should also be checked for that of course. I'm not really against this but I was hoping to improve the BAT/ENERGY messages in general. Their is no particular reason to keep both, the uint16 for energy in ENERGY msg can be a problem. I would be in favor of merging them and using float everywhere (and reflect that in the airborne code of course).
We could also find less intrusive solution like changing the airborne part for higher resolution (or float) and use the ENERGY message for display instead of BAT. This way, messages are unchanged.

fvantienen:
For me it is also fine to change it. But we could indeed better chamge it to float directly

OpenUAS:
[...]

PR #2360, opened 2018-11-27
upgrade voltage and current to float

kirkscheper:
This should comprehensively address #2313. This was much more work than I had first thought it would be but I think this updates the voltage and current in every part of pprz. Additionally, I clarified the difference between charge and energy in the electrical module and messages.

Note, some of the components of the BAT message have been removed as they were already contained in the (ROTORCRAFT_)NAV_STATUS messages.

I tested this in sim, nps and on a bebop but this will need a thorough review before merging. I look forward to your comments!


conf/gcs/papgets.xml
```diff
@@ -14,25 +14,26 @@
...
+        <property name="field" value="ENERGY:voltage"/>
+        <property name="scale" value="1."/>
         <property name="min" value="0."/>
         <property name="max" value="15."/>
         <property name="size" value="50."/>
         <property name="text" value="Bat(V)"/>
       </papget>
-      <papget type="message_field" display="gauge" x="120" y="414">
-        <property name="field" value="BAT:throttle"/>
-        <property name="scale" value="0.0104167"/>
```
OpenUAS: correct?

conf/gcs/solar.xml
```diff
@@ -2,8 +2,8 @@
   <rows>
     <widget NAME="map2d" SIZE="500">
       <papget type="message_field" display="text" x="5" y="125">
-        <property name="field" value="BAT:energy"/>
-        <property name="scale" value="1"/>
+        <property name="field" value="Energy:charge"/>
```
OpenUAS: Uppercase

kirkscheper: tnx

conf/gcs/papgets.xml
kirkscheper: I think so, why would it not be?

kirkscheper: I think so, why would it not be?

conf/gcs/solar.xml
kirkscheper: tnx

sw/ground_segment/tmtc/server.ml
```diff
@@ -438,7 +435,7 @@ let send_aircraft_msg = fun ac ->
                   "temp", f a.temp;
                   "bat", f a.bat;
                   "amp", f a.amp;
-                  "energy", PprzLink.Int a.energy] in
+                  "charge", f a.charge] in
```
gautierhattenberger: why charge here instead of keeping energy ? also unit in message definition is not matching

kirkscheper: Energy is power x time in Wh and charge is current x time in Ah. In the code, energy was sometimes wrongly used as charge. I corrected this, renaming everything that was measuring or displaying charge to be called charge. I also added a true measure of energy.

Here, charge was being sent so I updated it. So either both the name and the unit were wrong or the server was sending the wrong value. Do you know if this should have been energy or charge? Doesn't look like this value was being used anywhere.

This also reminds me I forgot to make a pull request for the pprzlink side of this.

kirkscheper:
Any more requests/comments here?

PR #2364, opened 2018-12-03
Restructure bebop camera settings and ISP

kirkscheper:
This is my effort at completing #2180. I implemented the changes that were made there (that I believed to be correct) as well as the recommended preset resolutions (VGA, 720p, 1080p and full resolution).

One of the changes was in how the camera settings were used, I built upon these suggestions and updated the settings. Now the user only sets the output resolution, image zoom and centre offsets and the rest is computed to generate a good output.

As the isp can only be used with the bebop, I moved the isp files to the bebop board folder.

In the process of doing this I realised that the documentation for the bebop cameras was quite poor and decided to make a module for the bebop cameras so that the settings can be auto documented. This results in an additional module to include in the airframe. To make this a little easier I changed all video_thread includes in bebop airframes to bebop_cam (which autoloads video_thread, plz let me know if you disagree or have an alternative).

There are also some miscellaneous changes that I found while implementing the upgrades (e.g. a setting for camera with the bebop_ae_awb even though it explicitly only works with the front camera).

Apologies for another large PR but I think that this simplifies and clarifies the camera usage for the bebop.


conf/airframes/tudelft/guido_ardrone2_optitrack.xml
```diff
@@ -30,7 +30,7 @@ ARDrone2 with optical_flow landing.
     <module name="ins" type="extended"/>
 
     <module name="bat_voltage_ardrone2"/>
-    <module name="video_thread"/>
```
gautierhattenberger: is it really a bebop cam for an ardrone2 ?

kirkscheper: Whoops, thanks. I will still wait for @dewagter to take a look at it before I merge as it is quite a lot.

sw/airborne/boards/bebop/mt9f002.c
```diff
@@ -27,20 +27,151 @@
...
+#define MT9F002_GAIN_RED 1.4
+#endif
+
+#ifndef MT9F002_GAIN_BLUE
+#define MT9F002_GAIN_BLUE 2.7
+#endif
+
+// parameters for undistortion
+#ifndef MT9F002_FOCAL_X
+#define MT9F002_FOCAL_X 311.59304538f
```
dewagter: default to WIDTH/2

kirkscheper:
Thanks, I will complete this in the new year after the stable release.

PR #2367, opened 2018-12-11
[math] fix quat_of_axis functions

### LINKED ISSUE #2331: quat_of_axis_angle does not yield a normalized quaternion
Both the int32_quat_of_axis_angle and float_quat_of_axis_angle don't return a normalized quaternion. This is due to the fact that

1. The axis is not normalized inside the function. The documentation does state that it uses a unit axis.
2. (only for the int32 variant) The PPRZ_ITRIG_COS and PPRZ_ITRIG_SIN do not give the angles that correspond with the quaternion fractions (i.e. PPRZ_ITRIG_COS(0) != QUAT1_BFP_OF_REAL(1))

Is this behavior as designed?
If it is, I think the function documentation should mention this behavior more clearly.
If not, setting the cosine and sine of the angle to the proper frac may do the job, although it's not clear to me how to work with a unit axis that is not either the x,y or z axis.

Anyway I could try to solve it/update the documentation if deemed required. Right guidance_hybrid_set_cmd_i in guidance_hybrid.c outputs a non-normalized quaternion. This is not a problem since the reference generator in stabilization_quat_int.c normalizes it. Normalizing q_yaw_sp in guidance_hybrid_set_cmd_i also works if the behavior  of int32_quat_of_axis_angle shouldn't change.

gautierhattenberger:
Improve doc and choose a more reasonable output resolution of the integer version.

The only possible issue might be in the hybrid guidance:
- quat setpoint should now have the proper value, but since the "wrong" value was currently compensated later in the code it shouldn't change anything
- euler setpoint should be correct now, but it is unclear if it was even used with hybrid guidance (it shouldn't)
- hybrid guidance from master is currently not much used apparently and might need some extra cleaning and better integration anyway, better have this part correct now at least

I suggest to merge this now for next release even if hybrid guidance can't be fully checked before that.
fix #2331

### DIFF
--- a/sw/airborne/math/pprz_algebra_int.c
+++ b/sw/airborne/math/pprz_algebra_int.c
@@ -412,10 +412,10 @@ void int32_quat_of_axis_angle(struct Int32Quat *q, struct Int32Vect3 *uv, int32_
   PPRZ_ITRIG_SIN(san2, (angle / 2));
   int32_t can2;
   PPRZ_ITRIG_COS(can2, (angle / 2));
-  q->qi = can2;
-  q->qx = san2 * uv->x;
-  q->qy = san2 * uv->y;
-  q->qz = san2 * uv->z;
+  q->qi = (can2 << (INT32_QUAT_FRAC-INT32_TRIG_FRAC));
+  q->qx = (san2 << (INT32_QUAT_FRAC-INT32_TRIG_FRAC)) * uv->x;
+  q->qy = (san2 << (INT32_QUAT_FRAC-INT32_TRIG_FRAC)) * uv->y;
+  q->qz = (san2 << (INT32_QUAT_FRAC-INT32_TRIG_FRAC)) * uv->z;
 }
 
 void int32_quat_of_rmat(struct Int32Quat *q, struct Int32RMat *r)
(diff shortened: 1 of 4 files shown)


sw/airborne/math/pprz_algebra_int.c
```diff
@@ -412,10 +412,10 @@ void int32_quat_of_axis_angle(struct Int32Quat *q, struct Int32Vect3 *uv, int32_
   PPRZ_ITRIG_SIN(san2, (angle / 2));
   int32_t can2;
   PPRZ_ITRIG_COS(can2, (angle / 2));
-  q->qi = can2;
-  q->qx = san2 * uv->x;
-  q->qy = san2 * uv->y;
-  q->qz = san2 * uv->z;
+  q->qi = (can2 << (INT32_QUAT_FRAC-INT32_TRIG_FRAC));
```
kirkscheper: Else where in this file, INT_MULT_RSHIFT is used to complete multiplications which multiplies then scales, any specific reason that isn't done here?

gautierhattenberger: it should be INT_MULT_LSHIFT eventually, but this one doesn't exist...

gautierhattenberger:
it just have no side effect on the rest of the code and it's barely not used anyway

PR #2368, opened 2018-12-11
[conf] remove conf failing the CI tests

gautierhattenberger:
they can be added back once the issues are properly handled
- some errors with opencv compilation not producing the correct lib
- some error in the xml

At least try to have a CI build that doesn't fail. The actual problems are not that clear, and it may come from the CI configuration as well. Until this is properly identified, there is no point keeping them in the nightly build.


OpenUAS:
@kirkscheper local build of EFlite-UMX_Sbach_342 (all 4! targets) build perfectly. A CI error log really appriciated. Could setup my own travis-ci server... but a bit extreme to debug this right ;)?

PR #2370, opened 2018-12-14
[rc] Fix SBUS initialization synchronising with timeout

fvantienen:
Fix bug when SBUS packets contain the `SBUS_START_BYTE` and can not synchronise at boot.

### DIFF
--- a/sw/airborne/subsystems/radio_control/sbus_common.h
+++ b/sw/airborne/subsystems/radio_control/sbus_common.h
@@ -85,6 +85,7 @@ struct Sbus {
   uint8_t buffer[SBUS_BUF_LENGTH];  ///< Input buffer
   uint8_t idx;                      ///< Input index
   uint8_t status;                   ///< Decoder state-machine status
+  uint32_t start_time;              ///< Decoder start time
 };
 
 /**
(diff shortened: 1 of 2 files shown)


sw/airborne/subsystems/radio_control/sbus_common.h
```diff
@@ -85,6 +87,7 @@ struct Sbus {
   uint8_t buffer[SBUS_BUF_LENGTH];  ///< Input buffer
   uint8_t idx;                      ///< Input index
   uint8_t status;                   ///< Decoder state-machine status
+  uint32_t start_time;              ///< Decoder start time
```
kirkscheper: Can you also initialize this in the sbus_common_init?

sw/airborne/subsystems/radio_control/sbus_common.h
```diff
@@ -58,9 +58,11 @@
  * depending of the RC transmitter.
  * The radio XML file is used to assign the
  * input values to RC channels.
+ * 4MS timeout
  */
 #define SBUS_BUF_LENGTH 24
 #define SBUS_NB_CHANNEL 16
+#define SBUS_TIMEOUT_MS 4
```
kirkscheper: Why is this one in the header? If not needed outside of sbus_common.c, can you move it to the source file?

It might also be nice to have this configurable, maybe the delay can change depending on the application. If you add the setting, also add it to the module description.

fvantienen: Not really needed to be configurable, since the baudrate of SBUS and the frame length are defined this timeout does never change

OpenUAS:
Did not work well on Disco

OpenUAS:
Will retest on all new Disco

PR #2373, opened 2018-12-18
[python] added explicit pyhton2 binding where ambiguous

kirkscheper:
Many python scripts in paparazzi are not designed to run on python3 yet have shebang lines with simply python. If a user computer has python3 as default this will cause many scripts (including parrot_utils.py) to not work. This PR makes all scripts default to python2.

I also added a simple script to try to identify which scripts might not be python3 compatible for when we will try to upgrade the scripts.

I also tried to update some print functions to python3 using the suggestions from 2to3.

Please advise if any of these files are known to actually be fully functional under both versions and I will revert those files before merging.

### DIFF
--- a/start.py
+++ b/start.py
@@ -1,7 +1,6 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 from __future__ import print_function
-
 
 import pygtk
 import gtk
--- a/sw/airborne/modules/loggers/logger_uart_parse.py
+++ b/sw/airborne/modules/loggers/logger_uart_parse.py
@@ -1,3 +1,5 @@
+#! /usr/bin/env python
+
 import struct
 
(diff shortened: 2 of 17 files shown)


kirkscheper:
@gautierhattenberger, you have experience with python, what do you think?

gautierhattenberger:
I'm really not sure. It probably doesn't harm to apply this, but also most of this scripts probably just work fine with Python 3 and it is too bad to stick to python 2.
It looks a bit hard to test all of them one by one. Maybe we could only change the one that are not even passing your test script.
Would be nice to have @flixr advices here :)

kirkscheper:
OK, I have only added the shebang where I am pretty sure it doesn't work with python3 (namely all modules that use pygtk). I have added future includes for print and updated all python files that I could easily find with the test tool. Should be good to merge once it passes the tests.

PR #2377, opened 2019-01-10
Autonomous Drone Race 2018 Code

dewagter:


kirkscheper:
There is quite a lot wrong with this, is this supposed to be reviewed for merging?

dewagter:
Yes, it would be nice to have drone racing code in master at some point. Can you try to summarize in a few points what the main problems are?

sw/airborne/modules/ctrl/dronerace/filter.h
```diff
@@ -0,0 +1,42 @@
...
+
+
+extern void filter_reset(void);
+
+extern void filter_predict(float phi, float theta, float psi, float dt);
+extern void filter_correct(void);
+extern float mx;
+extern float my;
+extern float filteredX;
+extern float filteredY;
```
kirkscheper: Where are this used outside of filter.c? Try to only include variables that are needed by other source of header files here. It makes understanding how variables are used/set/reset much more clear (and safe).

podhrmic:
@kirkscheper @dewagter are you planning to add some info about running the drone race / how the code works on wiki? For example I would be excited to replicate an autonomous race with paparazzi!

OpenUAS:
@podhrmic Yes, planning to add info about running the drone race on wiki. In the ideal docs case indeed replicate an autonomous race with paparazzi should be there in wiki and links to relevant papers. Busy helping out people involved but will take at least 3 weeks

PR #2380, opened 2019-01-16
Adding DelFly Nimble

matejkarasek:
Adding a DelFly Nimble with a Lisa/MXS autopilot, the necessary changes include:

- a DelFly Nimble specific board file
- an extra feedforward term on rate added to stabilization_euler_int.c (needed due to high damping of the flapping wing platform)


conf/boards/lisa_mxs_1.0_nimble.makefile
```diff
@@ -0,0 +1,50 @@
...
+BARO_LED           ?= none
+AHRS_ALIGNER_LED   ?= none
+GPS_LED            ?= none
+SYS_TIME_LED       ?= 1
+LOGGER_LED         ?= 2
+
+FLASH_MODE ?= SWD_NOPWR
+
+# disabling PWM to prevent interference when PA3 remapped to ADC
+$(TARGET).CFLAGS+=-DUSE_PWM5=0
```
podhrmic: not sure if it is worth having a new board file just for Delfly - maybe a better way would be to define the changes (no LEDs, USE_PWM5) in the airframe config file and use regular lisa_mxs makefile

conf/flight_plans/tudelft/flight_plan.dtd
```diff
@@ -0,0 +1,341 @@
+<!-- Paparazzi flight plan DTD -->
```
podhrmic: Better to use the `flight_plan.dtd` that is in `conf/flight_plans` than to copy the file over.

conf/flight_plans/tudelft/delfly_nimble_cyberzoo.xml
```diff
@@ -0,0 +1,112 @@
+<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
```
podhrmic:
```suggestion
<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
```

conf/modules/guidance_rotorcraft.xml
```diff
@@ -26,10 +26,10 @@
   <settings target="ap|nps">
     <dl_settings>
       <dl_settings NAME="Vert Loop">
-        <dl_setting var="guidance_v_kp" min="0" step="1" max="600" shortname="kp" param="GUIDANCE_V_HOVER_KP" persistent="true" module="guidance/guidance_v"/>
-        <dl_setting var="guidance_v_kd" min="0" step="1" max="600" shortname="kd" param="GUIDANCE_V_HOVER_KD" persistent="true"/>
-        <dl_setting var="guidance_v_ki" min="0" step="1" max="300" shortname="ki" handler="SetKi" param="GUIDANCE_V_HOVER_KI" persistent="true" module="guidance/guidance_v"/>
-        <dl_setting var="guidance_v_nominal_throttle" min="0.2" step="0.01" max="0.8" shortname="nominal_throttle" param="GUIDANCE_V_NOMINAL_HOVER_THROTTLE" persistent="true"/>
+        <dl_setting var="guidance_v_kp" min="0" step="1" max="2000" shortname="kp" param="GUIDANCE_V_HOVER_KP" persistent="true" module="guidance/guidance_v"/>
```
podhrmic: do you need these higher limits for Delfly? If so, it might make sense to have a separate guidance module, because this change would make it harder to adjust the controller for regular airframes (the range would be too large).

sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
```diff
@@ -96,6 +96,10 @@ PRINT_CONFIG_VAR(GUIDANCE_V_ADAPT_THROTTLE_ENABLED)
 #define GUIDANCE_V_MAX_SUM_ERR 2000000
 #endif
 
+#ifndef GUIDANCE_V_MAX_CMD
+#define GUIDANCE_V_MAX_CMD 8640
```
podhrmic:
```suggestion
#define GUIDANCE_V_MAX_CMD 0.9*MAX_PPRZ
```

conf/flight_plans/tudelft/delfly_nimble_cyberzoo.xml
matejkarasek: fixed

conf/modules/guidance_rotorcraft.xml
matejkarasek: Yes, we need much higher gains, but for now I return it back to the standard values

conf/boards/lisa_mxs_1.0_nimble.makefile
matejkarasek: Yeah, I wasn't either... Discussed this with @kirkscheper and we agreed this will be the best solution since the autopilot connections on the Nimble are fixed now (and are quite different from the standard mapping). 

We could also do this in the airframe files, but this would mean a lot of copy-pasting for every new DelFly (they are all slightly different, so typically require different airframes...)

podhrmic: OK, lets leave this as is.

matejkarasek:
If all is fine, can this be pulled into master? :)

conf/modules/guidance_rotorcraft.xml
```diff
@@ -64,4 +64,4 @@
     <file name="guidance_v_adapt.c" dir="$(SRC_FIRMWARE)/guidance"/>
     <file name="guidance_flip.c" dir="$(SRC_FIRMWARE)/guidance"/>
   </makefile>
-</module>
+</module>
```
kirkscheper: Can you undo this change?

matejkarasek: fixed!

PR #2387, opened 2019-02-06
[autopilot] flight times and kill_throttle were not available in GCS …

gautierhattenberger:
…anymore since #2360

works with https://github.com/paparazzi/pprzlink/pull/97


gautierhattenberger:
is it okay if I merge this ? as it has an impact on the perception of the AP status by the operator, it is quite critical I think.

PR #2389, opened 2019-02-12
use hard-float compiler for parrot disco

bartslinger:
Using `arm-2016.02-linaro` compiler from https://github.com/Parrot-Developers/toolchains


bartslinger:
Never mind, this produces a segfault

PR #2390, opened 2019-02-12
[linux] Fix uart/serial waiting

fvantienen:
Whenever the UART does give an event you need to read all the bytes and not only a single byte. Next to that opening the uart without having a buffer removes delays.


fvantienen:
I tested it on a disco with an SBUS receiver and works. This could solve a lot of UART problems like GPS, external usb UARTs etc.

PR #2392, opened 2019-02-19
add new sensors

gautierhattenberger:
add two new sensors:
- lis3mdl magnetometer
- sdp3x differential pressure sensor

### DIFF
--- a/conf/modules/mag_lis3mdl.xml
+++ b/conf/modules/mag_lis3mdl.xml
@@ -1,0 +1,42 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="mag_lis3mdl" dir="sensors">
+  <doc>
+    <description>
+      ST LIS3MDL magnetometer.
+    </description>
+    <configure name="MAG_LIS3MDL_I2C_DEV" value="i2cX" description="I2C device to use"/>
+    <define name="MODULE_LIS3MDL_SYNC_SEND" value="TRUE|FALSE" description="Send IMU_RAW message with each new measurement (default: FALSE)"/>
+    <define name="MODULE_LIS3MDL_UPDATE_AHRS" value="TRUE|FALSE" description="Copy measurements to imu and send as ABI message (default: FALSE)"/>
+    <define name="LIS3MDL_CHAN_X_SIGN" value="+|-" description="Reverse polarity of x axis (default: +)"/>
+    <define name="LIS3MDL_CHAN_Y_SIGN" value="+|-" description="Reverse polarity of y axis (default: +)"/>
+    <define name="LIS3MDL_CHAN_Z_SIGN" value="+|-" description="Reverse polarity of z axis (default: +)"/>
+    <define name="LIS3MDL_CHAN_X" value="0|1|2" description="Channel id of x axis (default: 0)"/>
+    <define name="LIS3MDL_CHAN_Y" value="0|1|2" description="Channel id of y axis (default: 1)"/>
+    <define name="LIS3MDL_CHAN_Z" value="0|1|2" description="Channel id of z axis (default: 2)"/>
+    <section name="MAG_LIS3MDL" prefix="LIS3MDL_">
+      <define name="MAG_TO_IMU_PHI" value="0.0" description="Rotation between sensor frame and IMU frame (phi angle)"/>
+      <define name="MAG_TO_IMU_THETA" value="0.0" description="Rotation between sensor frame and IMU frame (theta angle)"/>
+      <define name="MAG_TO_IMU_PSI" value="0.0" description="Rotation between sensor frame and IMU frame (psi angle)"/>
+    </section>
+  </doc>
+  <header>
+    <file name="mag_lis3mdl.h"/>
+  </header>
+  <init fun="mag_lis3mdl_module_init()"/>
+  <periodic fun="mag_lis3mdl_module_periodic()" freq="60"/>
+  <periodic fun="mag_lis3mdl_report()" freq="10" autorun="FALSE"/>
+  <event fun="mag_lis3mdl_module_event()"/>
+  <makefile target="ap">
+    <file name="mag_lis3mdl.c"/>
+    <file name="lis3mdl.c" dir="peripherals"/>
+    <raw>
+      ifeq ($(MAG_LIS3MDL_I2C_DEV),)
+        $(error mag_lis3mdl module error: please configure MAG_LIS3MDL_I2C_DEV)
+      endif
+    </raw>
+    <configure name="MAG_LIS3MDL_I2C_DEV" case="upper|lower"/>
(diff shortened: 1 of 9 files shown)


sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      int16_t p_raw = ((int16_t)(buf[0]) << 8) | (int16_t)(buf[1]);
+      sdp3x.pressure = ((float)p_raw / sdp3x.pressure_scale) - sdp3x.pressure_offset;
+
+      int16_t t_raw = ((int16_t)(buf[3]) << 8) | (int16_t)(buf[4]);
+      sdp3x.temperature = (float)t_raw / SDP3X_SCALE_TEMPERATURE;
+
+      // Send (differential) pressure via ABI
+      AbiSendMsgBARO_DIFF(SDP3X_SENDER_ID, sdp3x.pressure);
```
mrtbrnz: Adding the AbiSendMsgBARO_DIFF and AbiSendMsgTEMPERATURE inside the next "#if USE_AIRSPEED_SPD3X" condition lets us use the sensor as an external sensor when required. For example, we may use the MS45XX airspeed sensor for flight which updates the state and use the SDP3x for other purpose ?

gautierhattenberger: If you have several sensor sending the same type of data but you only want to have one of them used, you should filter from the receiver side, so not bind to ABI_BROADCAST but to the specific sensor

mrtbrnz: Ok!, I get it now...

fvantienen:
Looks good, better than my implementation of the LIS3MDL driver which I wanted to merge this week. I will quickly test this driver as well

conf/modules/mag_lis3mdl.xml
```diff
@@ -0,0 +1,42 @@
...
+  <event fun="mag_lis3mdl_module_event()"/>
+  <makefile target="ap">
+    <file name="mag_lis3mdl.c"/>
+    <file name="lis3mdl.c" dir="peripherals"/>
+    <raw>
+      ifeq ($(MAG_LIS3MDL_I2C_DEV),)
+        $(error mag_lis3mdl module error: please configure MAG_LIS3MDL_I2C_DEV)
+      endif
+    </raw>
+    <configure name="MAG_LIS3MDL_I2C_DEV" default="TRUE" case="upper|lower"/>
```
fvantienen: default TRUE?

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -0,0 +1,223 @@
...
+      // Send temperature as float in deg Celcius via ABI
+      AbiSendMsgTEMPERATURE(SDP3X_SENDER_ID, sdp3x.temperature);
+      // Compute airspeed
+      sdp3x.airspeed = sqrtf(Max(sdp3x.pressure * sdp3x.airspeed_scale, 0));
+
+#if USE_AIRSPEED_SDP3X
+      stateSetAirspeed_f(sdp3x.airspeed);
```
fvantienen: Isn't it nicer to remove this and instead use the `air_data` module?

sw/airborne/modules/sensors/mag_lis3mdl.c
```diff
@@ -0,0 +1,137 @@
...
+#define LIS3MDL_CHAN_X_SIGN +
+#endif
+#ifndef LIS3MDL_CHAN_Y_SIGN
+#define LIS3MDL_CHAN_Y_SIGN +
+#endif
+#ifndef LIS3MDL_CHAN_Z_SIGN
+#define LIS3MDL_CHAN_Z_SIGN +
+#endif
+
+#if MODULE_LIS3MDL_UPDATE_AHRS
```
fvantienen: Why do we need this `UPDATE_AHRS`? If we can just listen to the correct ABI id instead. Maybe we could add an `ABI_ID_NONE` if you don't wanna listen to any magnetometer.
This could make it more future proof if we wanna combine multiple magnetometers.

sw/airborne/modules/sensors/mag_lis3mdl.c
```diff
@@ -0,0 +1,137 @@
...
+#else
+    // unscaled vector
+    VECT3_COPY(imu.mag_unscaled, mag);
+#endif
+    // scale vector
+    imu_scale_mag(&imu);
+
+    AbiSendMsgIMU_MAG_INT32(MAG_LIS3MDL_SENDER_ID, now_ts, &imu.mag);
+#endif
+#if MODULE_LIS3MDL_SYNC_SEND
```
fvantienen: Is this really needed? I know that you prolly c+p this from the other magnetometer like me.
But can we just remove this and if we want this feature add an ABI listener which does this in for example the `imu.c`?

sw/airborne/peripherals/lis3mdl.c
```diff
@@ -0,0 +1,160 @@
...
+ * @file peripherals/lis3mdl.c
+ *
+ * ST LIS3MDL 3-axis magnetometer driver interface (I2C).
+ */
+
+#include "peripherals/lis3mdl.h"
+
+#define LIS3MDL_ENABLE_AUTO_INC   (0x1<<7)
+
+#define LIS3MDL_STATUS_ZYXOR      0b10000000
```
fvantienen: Shouldn't we move these registers to a separate file?

sw/airborne/modules/sensors/airspeed_sdp3x.c
gautierhattenberger: I agree. I guess other sensors should be updated the same way, but I'm afraid that it will break some airframe files since air_data is not always loaded in "historical" config file.

sw/airborne/modules/sensors/mag_lis3mdl.c
gautierhattenberger: At the moment we are still accessing the global `imu` structure from mag modules, so only one should be explicitly set to be used in AHRS/INS. Once we have a proper way to handle several sensors of the same type, we can get rid of this. Note that pprzlink v2 introduced a component id for this purpose (regarding the message part) but this is not used at all at the moment.

And `ABI_NONE` is already the binding id for receiving nothing (id 0) :)

sw/airborne/modules/sensors/mag_lis3mdl.c
gautierhattenberger: only for debugging, merging this possibility for all sensors requires a proper use of the component IDs

sw/airborne/peripherals/lis3mdl.c
gautierhattenberger: that will make sense if we ever have two sensors with the same registers but for I don't think it is necessary

PR #2394, opened 2019-02-22
[chibios] Add PX4FMU V4 (pixracer)

fvantienen:
This adds support for the Pixracer (PX4FMU) using chibios.


conf/boards/px4fmu_4.0_chibios.makefile
```diff
@@ -0,0 +1,89 @@
...
+SBUS_PORT ?= UART6
+RADIO_CONTROL_SPEKTRUM_PRIMARY_PORT   ?= UART6
+
+#The modem serial on px4 is called serial 1, but connected to uart2 on the f4
+MODEM_PORT ?= UART2
+MODEM_BAUD ?= B57600
+
+#The GPS serial on px4 is called serial 3, but connected to uart4 on the f4
+GPS_PORT ?= UART4
+GPS_BAUD ?= B38400
```
gautierhattenberger: maybe we can set a default speed of 57600 for recent board

sw/airborne/boards/px4fmu/chibios/v4.0/board.c
```diff
@@ -0,0 +1,263 @@
...
+
+  stm32_gpio_init();
+  stm32_clock_init();
+}
+
+#if HAL_USE_SDC || defined(__DOXYGEN__)
+/**
+ * @brief   SDC card detection.
+ */
+bool sdc_lld_is_card_inserted(SDCDriver *sdcp) {
```
gautierhattenberger: if you have a SD card on this board, you probably need to implement this function like: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/boards/chimera/chibios/v1.0/board.c#L220

fvantienen: Yes I thought so as well, but these lines aren't connected on the pixracer so this is not possible. I looked at ardupilot and they have it like this and it works as well.

conf/boards/px4fmu_4.0_chibios.makefile
fvantienen: I will change this

PR #2398, opened 2019-02-27
Pullrequest bebop2 optitrack

coppolam:
Added an airframe + flight plan that flies a stable flight using Optitrack in a small space. Gains have been tuned to keep the maximum velocity capped (0.5m/s), no overshooting of waypoints (GUIDANCE_H_REF_ZETA=1), and only switch waypoints when 0.2m away from it.

### DIFF
--- a/conf/airframes/tudelft/bebop2_optitrack.xml
+++ b/conf/airframes/tudelft/bebop2_optitrack.xml
@@ -1,0 +1,228 @@
+<!DOCTYPE airframe SYSTEM "../../airframe.dtd">
+
+<airframe name="bebop2_optitrack">
+  <description>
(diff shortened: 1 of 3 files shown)


conf/airframes/tudelft/bebop2_optitrack.xml
```diff
@@ -0,0 +1,228 @@
...
+      <configure name="USE_MAGNETOMETER" value="FALSE"/>
+      <define name="AHRS_USE_GPS_HEADING" value="TRUE"/>
+    </module>
+    <module name="ins" type="extended"/>
+  </firmware>
+
+  <modules main_freq="512">
+    <module name="geo_mag"/>
+    <module name="air_data"/>
+    <module name="send_imu_mag_current"/>
```
kirkscheper: We are trying to move away from the modules section, can you move these to the firmware section?

PR #2399, opened 2019-02-27
[gazebo] add ROS support and sonar sensor

tomvand:
This pull request adds:
- ROS support for Gazebo. Use the tools/gzserver_ros_launcher script to launch gazebo with the ROS libraries.
- Support for sonar sensors. If no sonar sensor is found, the gazebo fdm falls back to the default behavior of using the Z coordinate as sonar measurement.
- Bebop 2 and SLAMDunk models for gazebo. The SLAMDunk publishes the camera images as ROS topics.


tomvand:
Sure, I'll add it to my to-do list.

PR #2402, opened 2019-03-05
[ins] Add EKF2 from the PX4 ECL library

fvantienen:
This adds the EKF2 from the PX4 ECL library at https://github.com/px4/ecl.

It is tested with a bebop2 and the splash3 rotorcraft als NPS simulation works.
- No support yet for fixedwing, but could be added easily (don't forget to set set_is_fixedwing to the estimator).
- No support yet for optical flow and other visual aids
- ~~Warnings for c++ unused attributes are disabled, see https://github.com/PX4/ecl/issues/581~~
- Origin comes from the EKF and can not be reset or set by flightplan/user. Should add support to the estimator interface to enable this feature.
- The `hacc` and `vacc` from GPS are only supported for ublox gps systems for now (also no simulated values)

### DIFF
--- a/sw/airborne/c++.cpp
+++ b/sw/airborne/c++.cpp
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2019 Freek van Tienen <freek.v.tienen@gmail.com>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+#include <stdlib.h>
+
+/**
+ * Replace new and delete operators of C++
+ */
+
+void * operator new(size_t size)
+{
+  if (size < 1) {
+    size = 1;
+  }
+  return(calloc(size, 1));
+}
+
+void operator delete(void *p)
+{
+  if (p) free(p);
+}
+
+void * operator new[](size_t size)
+{
+  if (size < 1) {
+    size = 1;
+  }
+  return(calloc(size, 1));
+}
+
+void operator delete[](void * ptr)
+{
+  if (ptr) free(ptr);
+}
--- a/sw/airborne/modules/ctrl/vertical_ctrl_module_demo.c
+++ b/sw/airborne/modules/ctrl/vertical_ctrl_module_demo.c
@@ -48,7 +48,7 @@ PRINT_CONFIG_VAR(VERTICAL_CTRL_MODULE_AGL_ID)
 static abi_event agl_ev; ///< The altitude ABI event
 
 /// Callback function of the ground altitude
-static void vertical_ctrl_agl_cb(uint8_t sender_id __attribute__((unused)), float distance);
+static void vertical_ctrl_agl_cb(uint8_t sender_id, uint32_t stamp, float distance);
 
 struct VerticalCtrlDemo v_ctrl;
 
@@ -85,7 +85,7 @@ void vertical_ctrl_module_run(bool in_flight)
   }
 }
 
-static void vertical_ctrl_agl_cb(uint8_t sender_id, float distance)
+static void vertical_ctrl_agl_cb(__attribute__((unused)) uint8_t sender_id, __attribute__((unused)) uint32_t stamp, float distance)
 {
   v_ctrl.agl = distance;
 }
--- a/sw/airborne/modules/lidar/lidar_lite.c
+++ b/sw/airborne/modules/lidar/lidar_lite.c
@@ -137,8 +137,9 @@ void lidar_lite_periodic(void)
         }
       }
       break;
-    case LIDAR_LITE_PARSE:
+    case LIDAR_LITE_PARSE: {
       // filter data
+      uint32_t now_ts = get_sys_time_usec();
       lidar_lite.distance_raw = update_median_filter_i(
                                   &lidar_lite_filter,
                                   (uint32_t)((lidar_lite.trans.buf[0] << 8) | lidar_lite.trans.buf[1]));
@@ -154,12 +155,13 @@ void lidar_lite_periodic(void)
 
       // send message (if requested)
       if (lidar_lite.update_agl) {
-        AbiSendMsgAGL(AGL_LIDAR_LITE_ID, lidar_lite.distance);
+        AbiSendMsgAGL(AGL_LIDAR_LITE_ID, now_ts, lidar_lite.distance);
       }
 
       // increment status
       lidar_lite.status = LIDAR_LITE_INIT_RANGING;
       break;
+    }
     default:
       break;
   }
--- a/sw/airborne/modules/lidar/lidar_sf11.c
+++ b/sw/airborne/modules/lidar/lidar_sf11.c
@@ -97,9 +97,10 @@ void lidar_sf11_periodic(void)
       lidar_sf11.trans.buf[0] = 0; // set tx to zero
       i2c_transceive(&LIDAR_SF11_I2C_DEV, &lidar_sf11.trans, lidar_sf11.addr, 1, 2);
       break;
-    case LIDAR_SF11_READ_OK:
+    case LIDAR_SF11_READ_OK: {
       // process results
       // filter data
+      uint32_t now_ts = get_sys_time_usec();
       lidar_sf11.distance_raw = update_median_filter_i(
                                   &lidar_sf11_filter,
                                   (uint32_t)((lidar_sf11.trans.buf[0] << 8) | lidar_sf11.trans.buf[1]));
@@ -115,12 +116,13 @@ void lidar_sf11_periodic(void)
 
       // send message (if requested)
       if (lidar_sf11.update_agl) {
-        AbiSendMsgAGL(AGL_LIDAR_SF11_ID, lidar_sf11.distance);
+        AbiSendMsgAGL(AGL_LIDAR_SF11_ID, now_ts, lidar_sf11.distance);
       }
 
       // reset status
       lidar_sf11.status = LIDAR_SF11_REQ_READ;
       break;
+    }
     default:
       break;
   }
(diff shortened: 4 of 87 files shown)


sw/airborne/c++.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+#include <stdlib.h>
+
+/**
+ * Replace new and delete operators of C++
+ */
+
+void * operator new(size_t size)
```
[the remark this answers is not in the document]
fvantienen: In the `ins_ekf2.cpp` and inside the `ecl` library itself multiple new and delete oprations are done

sw/airborne/modules/lidar/lidar_sf11.c
```diff
@@ -97,9 +97,10 @@ void lidar_sf11_periodic(void)
       lidar_sf11.trans.buf[0] = 0; // set tx to zero
       i2c_transceive(&LIDAR_SF11_I2C_DEV, &lidar_sf11.trans, lidar_sf11.addr, 1, 2);
       break;
-    case LIDAR_SF11_READ_OK:
+    case LIDAR_SF11_READ_OK: {
```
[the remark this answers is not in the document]
fvantienen: This is needed if you start with a variable definition.

sw/airborne/modules/lidar/lidar_lite.c
```diff
@@ -137,8 +137,9 @@ void lidar_lite_periodic(void)
         }
       }
       break;
-    case LIDAR_LITE_PARSE:
+    case LIDAR_LITE_PARSE: {
```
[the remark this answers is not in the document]
fvantienen: Same as the other needed if you start with a variable definition.

fvantienen:
Fixed new lines

podhrmic:
@fvantienen this is exciting and I am hoping to read through the code more thoroughly soon.
One immediate question though - can I use this on Lisa MX? Does it work with ChibiOS?

fvantienen:
It works with chibios and linux (tested). Should also work with libopencm3 stm32 (untested).

So you can use this on a lisa MX. But make sure the sensor ranges, scalings and update rate are correct(I did not test well what happens with all these different settings).

podhrmic:
LGTM! Although I haven't tried running it yet.

sw/airborne/modules/ctrl/vertical_ctrl_module_demo.c
```diff
@@ -48,7 +48,7 @@ PRINT_CONFIG_VAR(VERTICAL_CTRL_MODULE_AGL_ID)
 static abi_event agl_ev; ///< The altitude ABI event
 
 /// Callback function of the ground altitude
-static void vertical_ctrl_agl_cb(uint8_t sender_id __attribute__((unused)), float distance);
+static void vertical_ctrl_agl_cb(uint8_t sender_id, uint32_t stamp, float distance);
```
gautierhattenberger: missing `unused` attribute

sw/airborne/subsystems/ins/ins_ekf2.cpp
```diff
@@ -0,0 +1,509 @@
...
+  bool got_imu_data;
+};
+static struct ekf2_t ekf2;  ///< Local EKF INS description
+
+#if PERIODIC_TELEMETRY
+#include "subsystems/datalink/telemetry.h"
+
+static void send_ins(struct transport_tx *trans, struct link_device *dev)
+{
+  /*pprz_msg_send_INS(trans, dev, AC_ID,
```
gautierhattenberger: remove if not used ?

sw/airborne/subsystems/ins/ins_ekf2.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+static void send_ins_z(struct transport_tx *trans, struct link_device *dev)
+{
+  /*pprz_msg_send_INS_Z(trans, dev, AC_ID,
+                      &ins_int.baro_z, &ins_int.ltp_pos.z, &ins_int.ltp_speed.z, &ins_int.ltp_accel.z);*/
+}
+
+static void send_ins_ref(struct transport_tx *trans, struct link_device *dev)
+{
+    float qfe = 101325.0; //TODO: this is qnh not qfe?
```
gautierhattenberger: this should be initialized from the baro measurements during the init phase

sw/airborne/subsystems/ins/ins_ekf2.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+/* Update the INS state */
+void ins_ekf2_update(void)
+{
+  /* Set EKF settings */
+  ekf.set_in_air_status(autopilot_in_flight());
+
+  /* Update the EKF */
+  if(ekf2.got_imu_data && ekf.update()) {
+    filter_control_status_u control_status;
```
gautierhattenberger: check indentation and formatting

sw/airborne/subsystems/ins/ins_ekf2.cpp
```diff
@@ -0,0 +1,509 @@
...
+    stateSetNedToBodyQuat_f(&ltp_to_body_quat);
+
+    /* Check the quaternion reset state */
+    float delta_q_reset[4];
+    uint8_t quat_reset_counter;
+    ekf.get_quat_reset(delta_q_reset, &quat_reset_counter);
+
+    // If reset update the setpoint heading
+    if(ekf2.quat_reset_counter < quat_reset_counter) {
+      float psi = matrix::Eulerf(matrix::Quatf(delta_q_reset)).psi();
```
gautierhattenberger: I still don't get why the filter should have access to the control part. It breaks the separation between estimation and control and make it unusable for a future fixed-wing usage or when using a custom control loop.

sw/airborne/subsystems/ins/ins_ekf2.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+/* Update INS based on Baro information */
+static void baro_cb(uint8_t __attribute__((unused)) sender_id, uint32_t stamp, float pressure)
+{
+  // Calculate the air density
+  float rho = pprz_isa_density_of_pressure(pressure, 20.0f); // TODO: add temperature compensation now set to 20 degree celcius
+  ekf.set_air_density(rho);
+  
+  // Calculate the height above mean sea level based on pressure
+  float height_amsl_m = pprz_isa_height_of_pressure_full(pressure, 101325.0); //101325.0 defined as PPRZ_ISA_SEA_LEVEL_PRESSURE in pprz_isa.h
```
gautierhattenberger: I think amsl should be calculated from QNH, not mean sea level pressure (this one gives flight level).
Actually, maybe you should get this information from the air_data module (temperature, qnh, ...)

sw/airborne/modules/ctrl/vertical_ctrl_module_demo.c
fvantienen: This is only needed in the real implementation not the definition.

sw/airborne/subsystems/ins/ins_ekf2.cpp
fvantienen: Fixed and added a AHRS status message

sw/airborne/subsystems/ins/ins_ekf2.cpp
fvantienen: This will be replaced by an ABI message later on, but for now is really needed for safe operation.

The problem is as following (without this if statement):
- Initial magnetometer measurements are incorrect (for example due to metal in the ground or other interference)
- Therefor the initial heading is incorrect (for example east) and the setpoint is set to this incorrect heading
- At an height of 1.5m the EKF2 filter resets the heading (because it knows initial heading on the ground can be bad)
- Now a correct heading is found (for example north)
- In all control modes, even in attitude you get a sudden yaw movement of the difference between the initial and correct heading. (From north to east) Caused by the incorrect setpoint at boot.

So to solve this correctly there needs to be made a consideration in every control mode if you want to correct the setpoint or not whenever the EKF2 has an quaternion reset. Similar problems occur in hover mode where you want the position setpoint to be updated on a position reset of the EKF2.

sw/airborne/subsystems/ins/ins_ekf2.cpp
fvantienen: I will calculate the QNH for now myself, but maybe the air_data module should be updated to output ABI messages. Because I don't like adding inputs which are not coming from ABI.

sw/airborne/subsystems/ins/ins_ekf2.cpp
fvantienen: Fixed

sw/airborne/subsystems/ins/ins_ekf2.cpp
fvantienen: I checked the PX4 code they use the same pressure for their calculations. So I think this is fine for now and can be updated when air_data outputs abi messages. https://github.com/PX4/Firmware/blob/c415477caf00b9acdb2f3215267bccbb5b859045/src/modules/sensors/sensor_params.c#L119

fvantienen:
I fixed most things and I think it can be merged. The `splash3` airframe file contains an NPS simulation with the EKF2 that works, for the people interested.
It is tested on quadrotors and it is advised to calibrate both you magnetometer and accelerometer. The INS_EKF2 message gives feedback about what is wrong based on the bitfields described in `common.h` and is extremely useful if it does not work properly.

sw/ext/Makefile
```diff
@@ -77,6 +77,10 @@ TRICAL: TRICAL.update
 
 fatfs: fatfs.update
 
+ecl: ecl.update
+
+matric: matrix.update
```
podhrmic: You have a typo here, should be `matrix`

podhrmic:
You have two warnings showing up when building `Splash3`. Looks like just a missing cast. I think it would be good to fix it before merging. Otherwise it looks good and the simulation is running well.
```
modules/lidar/tfmini.c: In function 'tfmini_send_lidar':
modules/lidar/tfmini.c:55:23: warning: passing argument 6 of 'pprz_msg_send_LIDAR' from incompatible pointer type [-Wincompatible-pointer-types]
                       &tfmini.parse_status);
                       ^
In file included from /paparazzi/var/include/pprzlink/messages.h:236:0,
                 from modules/lidar/tfmini.c:35:
/paparazzi/var/include/pprzlink/telemetry/LIDAR.h:75:20: note: expected 'uint8_t * {aka unsigned char *}' but argument is of type 'enum TFMiniParseStatus *'
 static inline void pprz_msg_send_LIDAR(struct transport_tx *trans, struct link_device *dev, uint8_t ac_id, float *_distance, uint8_t *_status, uint8_t *_trans_status) {
                    ^
modules/lidar/tfmini.c: In function 'tfmini_parse':
modules/lidar/tfmini.c:95:3: warning: switch missing default case [-Wswitch-default]
   switch (tfmini.parse_status) {
   ^
```

fvantienen:
@podhrmic I will make a separate PR for that since I added that somewhere else.

PR #2406, opened 2019-03-14
[driver] add BMP3 (Bosh-Sensortech) baro driver

gautierhattenberger:
- peripheral and test module
- only I2C interface for now
- configured with recommended settings for drone applications
- tested with BMP388

### DIFF
--- a/conf/modules/baro_bmp3.xml
+++ b/conf/modules/baro_bmp3.xml
@@ -1,0 +1,27 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="baro_bmp3" dir="sensors">
+  <doc>
+    <description>
+      Bosch-Sensortech BMP3xx pressure sensor
+    </description>
+    <configure name="BMP3_I2C_DEV" value="i2cX" description="select which i2c peripheral to use (default i2c1)"/>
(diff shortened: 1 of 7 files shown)


conf/modules/baro_bmp3.xml
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="baro_bmp3" dir="sensors">
+  <doc>
+    <description>
+      Bosch-Sensortech BMP3xx pressure sensor
+    </description>
+    <configure name="BMP3_I2C_DEV" value="i2cX" description="select which i2c peripheral to use (default i2c1)"/>
+    <define name="BMP3_SLAVE_ADDR" value="BMP3_I2C_ADDR|BMP3_I2C_ADDR_ALT" description="i2c slave address (default BMP3_I2C_ADDR)"/>
```
fvantienen: Does the value describe all different values? because with the compensation this is not done.

gautierhattenberger: There is no real guidelines for that, but since i2c addresses have only two possible values, I thought it was okay. I usually do the same for true/false, but not more that 3 possible elements.

PR #2411, opened 2019-03-27
Fixed ACT corner bug

guidoAI:
ACT corner used to increment the number of agents looking for corners in case of little texture. It had a maximum of 1000 agents, while the fast 9 corner structure array was only 512 long. So, after long periods of little texture, the ACT corner file would write outside the memory, causing troubles with video capturing. Easy to find bug, which has now been removed.


OpenUAS:
Thx.

PR #2413, opened 2019-04-01
[filter] IMU prefiltering using the One Euro filter

gautierhattenberger:
One Euro filter is a very simple adaptive low pass filter that can efficiently remove the noise from the gyros and accelerometers without adding to much lag.
As it is a separated module connected to the ABI bus, it can be used with any IMU and AHRS/INS combination by selecting the correct data source for the estimation filters.

The plot below shows the recording of the roll gyro, with the blue line the original noisy signal while the red line is the filtered value. The noise level is much lower but their is almost no lag during fast rotations.
gyro_time_one_euro

The FFT of the above signal shows the noise from the motors between 150 and 200 Hz that is almost completely removed.
gyro_fft_one_euro

This filter is very cheap in terms of computation and also very easy to tune with only two parameters.
See http://cristal.univ-lille.fr/~casiez/1euro/ for more details.

### DIFF
--- a/conf/modules/filter_1euro_imu.xml
+++ b/conf/modules/filter_1euro_imu.xml
@@ -1,0 +1,51 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="filter_1euro_imu" dir="imu">
+  <doc>
+    <description>
+      Prefiltering for IMU data using 1euro filter.
+
+      The 1 Euro filter is an first order low pass filter with an adaptive cutoff frequency.
+      It is able to efficiently remove high frequency noise when the average signal is stable
+      with very small lag by increasing the bandwith to capture faster movements.
+
+      Only two parameters are required to tune the filter:
+      - mincutoff that defines the minimum bandwith: reduce it if noise at low speed is a problem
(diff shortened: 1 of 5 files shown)


EwoudSmeur:
Cool, adaptive cutoff! However, the data that you show I wouldn't classify as particularly noisy :wink:

@mrtbrnz mentioned that you want to use it in combination with INDI, but I am not sure if this will work well; any lag introduced by filtering the gyroscopes is compensated by applying the same filter on the incremented inputs. If the same filter is not really the same, because the cutoff is adaptive and may be different, it may fail to synchronize the incremented inputs with the measurements.

gautierhattenberger:
I choose a part of the signal where you can actually see something :)
You can use this with any IMU/AHRS combination, so not only on quads using INDI. I know that even small lags can make a huge difference there. I tried to run the calibration script of INDI using the raw and the filtered data, and the efficiency gains are pretty similar. But since I'm not use to INDI, I can't tell for now if it is problem or not. I will try to make some flights for real if I find the time...
For now it seems to work with regular control at least.

PR #2418, opened 2019-05-02
[rotorcraft] don't reset vertical ref in guided mode

### LINKED ISSUE #2410: Vertical reference is wrong in guided mode
In my module, I keep call autopilot_guided_goto_ned(0.0,0.0,-1.5,0) with fixed frequency. However, the reference is not kept at -1.5. You can find the log here.
Screenshot from 2019-03-26 10-45-34

The problem might be that in **guidance_v_set_guided_z(float z)**, there are two lines

guidance_v_z_sum_err = 0; 
GuidanceVSetRef(stateGetPositionNed_i()->z, stateGetSpeedNed_i()->z, 0);

which update the reference based on current altitude.

gautierhattenberger:
When calling the set_z or set_vz functions in guided mode, the reference
should not be called as it would make them ineffective when the control
rate is high

close #2410

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c
@@ -560,9 +560,6 @@ bool guidance_v_set_guided_z(float z)
     /* reset speed setting */
     guidance_v_zd_sp = 0;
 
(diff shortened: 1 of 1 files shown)


fvantienen:
@ls90911 Can you test this?

PR #2422, opened 2019-05-13
[ahrs] add filter on accel heuristic of DCM AHRS

gautierhattenberger:
the same filter is already applied in the int_quat ahrs and can prevent
incorrect measurement rejection when strong structural vibrations
increase the IMU noise


gautierhattenberger:
I will test this next Thursday and let you know. In the lab, it makes a huge difference with high level of vibrations.

OpenUAS:
Appreciated :1st_place_medal:

gautierhattenberger:
I did a flight with this, but the plane I used started a strong flutter during the "reference flight", that was suppose to demonstrate that this was not changing the normal behavior of a plane. I don't think the flutter is related to the modification (it was a flying wing not stiff enough). The result is that the AHRS finally diverge after some time, but from the logs, I'm pretty sure it would have happen even earlier without this (I recorded almost 3G of noise at the end of the flight).
Since this is not completely conclusive, I'll make an other flight once I have a proper plane for this test.

gautierhattenberger:
This is now tested and working

PR #2430, opened 2019-05-16
Removal of unused flightplans

danielwillemsen:
There are many flight plans which are not used in any user configs, nor are included in any other flight plans. For some of these flight plans, it is highly unlikely that they could ever have any usage, as they are old, rarely updated and uncommented. Therefore I would propose to remove these flight plans, all of which are not used in any user configs:

**1. legacy flight plans
2. very old competition flight plans
3. some seperate flight plans namely:**
        - sectors.xml:  contains only some sectors in France, last commit 2007.
        - failsafe_modes.xml: last commit 2006.
        - huit .xml: after 2006 only some "routine changes" with which I mean changes that affected many flight plans at once.
        - kv_svalbard.xml: after 2008 only some "routine changes".
        - poles.xml: after 2009 only some "routine changes"


gautierhattenberger:
I don't mind removing the legacy flight plans and the old stuff in main dir, but eventually I'd like to keep the competitions.

danielwillemsen:
> I don't mind removing the legacy flight plans and the old stuff in main dir, but eventually I'd like to keep the competitions.

I can understand that. I have undone the removal of the competitions.

PR #2433, opened 2019-05-29
Improved automated python tools for maintenance

matteobarbera:
Extended start.py and paparazzi_health.py by adding routines to give paparazzi users a better overview of old and unused files. New features:

start.py:
- New "Maintenance Tools" button that opens a separate window with new maintenance features Maintenance Tools window 

Maintenance Tools Window:
- Generate an html table that gives an overview of all unique modules used by the airframes of a specific conf xml
- Create a list of airframes, flight plans and boards files that are not tested by any conf, together with their last commit date
- Create a list of all modules in paparazzi with information on how many times they are used by airframes, their last commit date, and if their corresponding source file is missing or not


sw/lib/python/paparazzi_health.py
```diff
@@ -123,114 +230,362 @@ def airframe_details(self, xmlname):
...
+        mods = dict()
+        fps_usage = dict()
+        conf_files = paparazzi.get_list_of_conf_files()
+
+        if show_boards:
+            bs = self.find_board_files()
+
+        # Generation of list of all files in the sw directory, for checking with the modules.
+        if show_modules:
+            module_sw_dir = paparazzi.PAPARAZZI_HOME + "/sw/"
```
gautierhattenberger: it should be `PAPARAZZI_SRC` here

matteobarbera:
I implemented most suggested changes from Codacy, except for security related issues. I could replace lxml with defused equivalents if necessary but I would not know what to do with the subprocesses

PR #2435, opened 2019-05-31
[module] WS2812 led module

gautierhattenberger:
WS2812 are RGB LEDs that can be controlled in series with a single wire. A demo module makes a group of LEDs continuously changing color.

Using the ChibiOS DMA driver.


gautierhattenberger:
I didn't do the driver myself, but it is good to know. If we keep using this, we might add the SPI alternative as well then. Thanks

PR #2436, opened 2019-06-13
Filtering option added to SDP3X airspeed sensor

mrtbrnz:
SDP3X airspeed sensors seem to be very fast to respond and this makes it difficult to use for maintaining constant airspeed for example... So a bit of optional filtering helps.

### DIFF
--- a/conf/modules/airspeed_sdp3x.xml
+++ b/conf/modules/airspeed_sdp3x.xml
@@ -8,8 +8,8 @@
     <define name="SDP3X_I2C_DEV" value="i2cX" description="set i2c peripheral (default: i2c2)"/>
     <define name="SDP3X_I2C_ADDR" value="0x50" description="i2c slave address (default: 0x50)"/>
     <define name="SDP3X_SYNC_SEND" value="TRUE|FALSE" description="flag to enable sending every new measurement via telemetry (default:FALSE)"/>
-    <define name="SDP3X_PRESSURE_SCALE" value="1.05" description="pressure scaling factor to convert raw output to Pa (default: set according to pressure range and output type according to datasheet)"/>
-    <define name="SDP3X_PRESSURE_OFFSET" value="8618.4" description="pressure offset in Pa (default: set according to pressure range and output type according to datasheet)"/>
+    <define name="SDP3X_PRESSURE_SCALE" value="1.0" description="pressure scaling factor to convert raw output to Pa (default: set according to pressure range and output type according to datasheet)"/>
+    <define name="SDP3X_PRESSURE_OFFSET" value="0.0" description="pressure offset in Pa (default: set according to pressure range and output type according to datasheet)"/>
     <define name="SDP3X_AIRSPEED_SCALE" value="1.6327" description="quadratic scale factor to convert differential pressure to airspeed"/>
     <define name="SDP3X_LOWPASS_TAU" value="0.15" description="Time constant for second order Butterworth low pass filter, cut-off freq of 1/(2*pi*tau) ~= 1Hz"/>
     <define name="USE_AIRSPEED_SDP3X" value="TRUE|FALSE" description="Set airspeed in state interface for this sensor"/>
--- a/conf/modules/airspeed_sdp3x.xml
+++ b/conf/modules/airspeed_sdp3x.xml
@@ -8,15 +8,18 @@
     <define name="SDP3X_I2C_DEV" value="i2cX" description="set i2c peripheral (default: i2c2)"/>
     <define name="SDP3X_I2C_ADDR" value="0x50" description="i2c slave address (default: 0x50)"/>
     <define name="SDP3X_SYNC_SEND" value="TRUE|FALSE" description="flag to enable sending every new measurement via telemetry (default:FALSE)"/>
-    <define name="SDP3X_PRESSURE_SCALE" value="1.0" description="pressure scaling factor to convert raw output to Pa (default: set according to pressure range according to datasheet)"/>
-    <define name="SDP3X_PRESSURE_OFFSET" value="0." description="pressure offset in Pa (default: set according to pressure range  according to datasheet)"/>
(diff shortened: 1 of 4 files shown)


conf/modules/airspeed_sdp3x.xml
```diff
@@ -8,15 +8,18 @@
     <define name="SDP3X_I2C_DEV" value="i2cX" description="set i2c peripheral (default: i2c2)"/>
     <define name="SDP3X_I2C_ADDR" value="0x50" description="i2c slave address (default: 0x50)"/>
     <define name="SDP3X_SYNC_SEND" value="TRUE|FALSE" description="flag to enable sending every new measurement via telemetry (default:FALSE)"/>
-    <define name="SDP3X_PRESSURE_SCALE" value="1.0" description="pressure scaling factor to convert raw output to Pa (default: set according to pressure range according to datasheet)"/>
-    <define name="SDP3X_PRESSURE_OFFSET" value="0." description="pressure offset in Pa (default: set according to pressure range  according to datasheet)"/>
+    <define name="SDP3X_PRESSURE_SCALE" value="1.05" description="pressure scaling factor to convert raw output to Pa (default: set according to pressure range and output type according to datasheet)"/>
+    <define name="SDP3X_PRESSURE_OFFSET" value="8618.4" description="pressure offset in Pa (default: set according to pressure range and output type according to datasheet)"/>
```
gautierhattenberger: this example value is way to far from realistic values, better keep it to zero, which is the current default

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -35,6 +36,23 @@
...
 
+#ifndef USE_AIRSPEED_SDP3X
+#if USE_AIRSPEED
+#define USE_AIRSPEED_SDP3X TRUE
+PRINT_CONFIG_MSG("USE_AIRSPEED_SDP3X set to TRUE since this is set USE_AIRSPEED")
+#endif
+#endif
+
+#if USE_AIRSPEED_SDP3X
+#include "state.h"
```
gautierhattenberger: At the moment, most airspeed sensor drivers are setting directly the state interface, which is not really nice. I would prefer that sensor just send an ABI message (either AIRSPEED or BARO_DIFF depending on the sensor) and the air_data module set the state interface in the end.

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -91,9 +109,22 @@ PRINT_CONFIG_VAR(SDP3X_PRESSURE_OFFSET)
...
 #endif
 
+/** Time constant for second order Butterworth low pass filter
+ * Default of 0.15 should give cut-off freq of 1/(2*pi*tau) ~= 1Hz
+ */
+#ifndef SDP3X_LOWPASS_TAU
+#define SDP3X_LOWPASS_TAU 0.15
+#endif
+
+#define SDP3X_SENDER_ID 42 //FIXME
```
gautierhattenberger: this should be a `AIRSPEED_SDP3X_ID` in the `subsystems/abi_sender_ids.h` file (and used to send ABI message)

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -163,13 +198,17 @@ void sdp3x_periodic(void)
   }
 }
 
+#define AUTOSET_NB_MAX 20
```
gautierhattenberger: is it really needed to have the auto offset for this sensor ?

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -195,6 +255,9 @@ void sdp3x_event(void)
       // Compute airspeed
       sdp3x.airspeed = sqrtf(Max(sdp3x.pressure * sdp3x.airspeed_scale, 0));
 
+#if USE_AIRSPEED_SDP3X
+      stateSetAirspeed_f(sdp3x.airspeed);
```
gautierhattenberger: this should be replaced by something like https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/sensors/airspeed_adc.c#L89

conf/modules/airspeed_sdp3x.xml
mrtbrnz: Corrected them to one and zero.

sw/airborne/modules/sensors/airspeed_sdp3x.c
mrtbrnz: Added the ABI msg id in the airspeed msg

sw/airborne/modules/sensors/airspeed_sdp3x.c
mrtbrnz: Most of the time the zero is really zero, however it may be good to have it incase we need it on the filed... I will keep an eye on the utilization during flights, and if we really never use it in real applications, then I will PR again to erase it.

sw/airborne/modules/sensors/airspeed_sdp3x.c
mrtbrnz: Done, and tested on an Apogee v1.0 board

sw/airborne/modules/sensors/airspeed_sdp3x.c
```diff
@@ -128,7 +157,7 @@ static void sdp3x_downlink(struct transport_tx *trans, struct link_device *dev)
 void sdp3x_init(void)
 {
   sdp3x.pressure = 0.;
-  sdp3x.temperature = 0;
+  sdp3x.temperature = 0.;
   sdp3x.airspeed = 0.;
   sdp3x.pressure_scale = SDP3X_PRESSURE_SCALE;
   sdp3x.pressure_offset = SDP3X_PRESSURE_OFFSET;
```
gautierhattenberger: init autoset_offset to false is missing

PR #2438, opened 2019-06-14
Improvement to start.py maintenance tools

matteobarbera:
The maintenance tools in start.py now allows the HTML table for the module overview to be generated based on an airframe board file, to quickly compare the used modules in all airframes of a certain board across all conf files.


fvantienen:
Could you maybe squash this commit? Because this commit history seems unusable.

tomvand:
Hi Matteo, why did you close this pull request? FYI, we can squash all your commits when we merge your PR, so you do not need to create a new, clean branch if that is the problem ;). Just let us know when your work is ready to be reviewed, as you still seem to regularly update this branch.

matteobarbera:
I had originally closed it while I tried to figure out how to squash the last two commits, which were done to resolve some conflicts we had with the master branch. If it's possible to do it on your end that would be great, I can't seem to rebase the branch without creating a new one

Fabien-B:
I never really use this script (what a shame) so I will give my observations "as is" so they might not be all relevant :

- At startup, it would be best if the pre-selected items in the combo boxes match the current configuration
- When hiting the "Set Active" button, it works well, but the items in the combo boxes shift by one above (lower index)
- The airframes used by a configuration file are shown, which is good. But when hiting "More info", the file is not regenerated, and all airframes are shown, which is confusing.

In the maintenance tools specifically :
- Same thing for the Conf combo box: the default should be the active one
- When selecting an item in a combo box, the other combobox automatically reset to "--None Selected--". Is this a normal behavior ? I don't really understand how it is meant to be used.
- The Module usage is great. I might have so suggestions for improvement, but later.
- I haven't tested the "Show Untested Files" because of a bug :
`Traceback (most recent call last):
  File "./start.py", line 143, in show_untested
    show_modules=data["Modules"].get_active())
  File "/home/fabien/paparazzi/sw/lib/python/paparazzi_health.py", line 626, in run
    self.not_tested_html(f, show_airframes, show_flightplans, show_boards, show_modules)
  File "/home/fabien/paparazzi/sw/lib/python/paparazzi_health.py", line 523, in not_tested_html
    afs_sorted = self.generate_sorted_list(afs)
  File "/home/fabien/paparazzi/sw/lib/python/paparazzi_health.py", line 351, in generate_sorted_list
    reverse=True)
  File "/home/fabien/paparazzi/sw/lib/python/paparazzi_health.py", line 350, in <lambda>
    file_date_lst = sorted(zip(lst, commit_dates), key=lambda x: datetime.datetime.strptime(x[1], '%d-%m-%Y'),
  File "/usr/lib/python2.7/_strptime.py", line 332, in _strptime
    (data_string, format))
ValueError: time data '00-00-0000' does not match format '%d-%m-%Y'`

matteobarbera:
@Fabien-B Thanks for the comments and suggestions. 
I have made a few changes:

- The pre-selected option in the combo boxes now matches the active selection
- Using "Set Active" no longer causes the displayed option in the combo box to switch
- The "Untested file" button should now be able to deal with untracked files and will display them at the top of the generated list
- The combo boxes automatically switching to None Selected is intended, as the tool currently either displays information about airframes that belong to a specific conf or all airframes across all confs with a specfic board. I added a description, the usage is hopefully clearer. In the future I would like to add the option to filter by board for the airframes in a specific conf

I have not made changes to the "More Info" button, it has always displayed all airframes of all confs. I am not sure if only displaying the airframes of the selected conf is desirable, I can see advantages to both seeing all airframes and just seeing additional info on the airframes of a single conf.

Fabien-B:
I think that's good to be merged, @gautierhattenberger ?

PR #2442, opened 2019-06-19
fix Tawaki RTC

gautierhattenberger:
- Fix config for RTC on Tawaki board
- Fix small bug in chibios sd log module
- Update ChibiOS for latest bug fixes


gautierhattenberger:
@podhrmic do you mind if I update ChibiOS here ? It is only to catch latest bug fix in current stable branch.

podhrmic:
@gautierhattenberger not at all, thanks for asking:-)

PR #2443, opened 2019-06-21
[ocaml] add pointeur dereferencing keyword to avoid conflict with XML invalid characters. Follow #2434.

Fabien-B:
You can write `foo @DEREF bar` in place of `foo->bar`.
Maybe too verbose ? Would `@DR` be better ?


gautierhattenberger:
not sure all users know what "deref" means, but I'm pretty sure they will be lost with "DR"... any other opinion ?

Fabien-B:
So, should this PR be merged ?
Cleaning a bit my stuff ;-)

Issue #2093, opened 2017-07-25
Improve Python Paparazzi Center


gautierhattenberger:
A new paparazzi center is available since https://github.com/paparazzi/paparazzi/pull/1811. It is written in Python and can be used as an alternate to the legacy Ocaml version.
Some improvements in the coding and the interface could be done. The following list is not exhaustive and users are welcome to add their comments and requirements.

- [x] at the moment, can only be run from its own directory: paths should be improved
- [x] some graphical elements like the right panels should have the possibility to be hidden
- [x] the list of tools could be more user friendly (a good example I think is the tools selection in latest Matlab, with large icons grouped by themes and the possibility to select favorites that then appears at the top)
- [x] improve some progress bars
- [x] saved XML files are not matching the one from Ocaml xml-light library, it is then annoying when switching from one center to an other. Not sure if we can do something about that

Don't hesitate to add more.

gautierhattenberger:
More changes to come, but closing this for now

PR #2445, opened 2019-06-24
python paparazzi center tools impovement

Fabien-B:
Contribute to #2093 : improve tools display.
This branch sleeps on my computer for too long, I updated it to match how tools are now managed (not done in master).
Can it be merged, even if work still has to be done on the python paparazzi center?


fvantienen:
I'm also ok with merging this but smaphoreci is blocking this.. @gautierhattenberger can you fix this?

PR #2447, opened 2019-07-01
Enac student projects integration

gautierhattenberger:
Several results coming from student projects at Enac:
- update and fix for the JeVois smart cam driver
- improvement of the UWB-based localization, either with local or global positioning
- trajectory tracking based on shift around a reference
- some small corrections here and there

### DIFF
--- a/conf/flight_plans/rover_optitrack_PIR.xml
+++ b/conf/flight_plans/rover_optitrack_PIR.xml
@@ -1,0 +1,72 @@
+<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
+
+<flight_plan alt="148" ground_alt="146" lat0="43.5640917" lon0="1.4829201" max_dist_from_home="20" name="Rover Optitrack (ENAC)" security_height="0.3">
+  <header>
+
+</header>
+  <waypoints>
+    <waypoint name="HOME" x="0.0" y="0.0"/>
+    <waypoint name="GOAL" x="2.0" y="2.0"/>
+    <waypoint name="STDBY" x="-0.7" y="-0.8"/>
+    <waypoint name="TD" x="0.8" y="-1.7"/>
+    <waypoint name="P1" x="-1" y="-4"/>
+    <waypoint name="P2" x="-1" y="4"/>
+    <waypoint name="_S1" x="3" y="4"/>
+    <waypoint name="_S2" x="3" y="-4"/>
+    <waypoint name="_S3" x="-3" y="-4"/>
+    <waypoint name="_S4" x="-3" y="4"/>
+    <waypoint name="_N1" x="4.5" y="5.2"/>
+    <waypoint name="_N2" x="4.5" y="-5.2"/>
+    <waypoint name="_N3" x="-4.5" y="-5.2"/>
+    <waypoint name="_N4" x="-4.5" y="5.2"/>
+  </waypoints>
+  <sectors>
+    <sector name="Net" color="red">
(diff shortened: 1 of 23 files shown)


noether:
Happy to review them, but it will take me some time. A lot of lines of code :P.

noether:
@gautierhattenberger . Is it possible in the .xml (or somewhere else in the wiki) a link to the student reports linked with these modules? Even in French for me is okei :P. The idea is to follow better what are the different functionalities available in the module, and to understand faster the code eventually by others.

gautierhattenberger:
the student reports are not that useful... but I have https://hal-enac.archives-ouvertes.fr/hal-01936955 which explains a part of the job done here. Not sure the wiki is the best place and I even removed some parts from the final version of my paper.

conf/flight_plans/rover_optitrack_PIR.xml
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE flight_plan SYSTEM "flight_plan.dtd">
```
noether: What does it mean "PIR" ?

I suppose that in the near future we should have ./plans/flightplans ./plans/"groundplans"/ etc?

sw/airborne/filters/pid.h
```diff
@@ -0,0 +1,317 @@
...
+ * @param pid pointer to PID structure
+ */
+static inline void reset_pid_f(struct PID_f *pid)
+{
+  pid->u = 0.f;
+  pid->e[0] = 0.f;
+  pid->e[1] = 0.f;
+  pid->sum = 0.f;
+}
+
```
noether: I like this tool :P

What about a function for resetting only the integral part (sum)? Or better if you can set the value of "sum" to whatever value you wish, i.e., its initial condition. (I believe Simulink has it too).

conf/flight_plans/rover_optitrack_PIR.xml
gautierhattenberger: who knows...
PIR stands for Projet Initiation Recherche. It is not really necessary to keep this one, but it has some nice moves for holonomic rover.

sw/airborne/filters/pid.h
gautierhattenberger: If you only need to reset/set the integral, maybe you can do it directly by changing the `sum` field ? Do you really think it is worth a dedicated function ? (probably better I guess...)

gautierhattenberger:
I'll had the ref to one of module description

conf/flight_plans/rover_optitrack_PIR.xml
```diff
@@ -0,0 +1,72 @@
...
+        <go from="P2" hmode="route" wp="P1" approaching_time="0"/>
+      </while>
+    </block>
+    <block name="circle">
+      <circle radius="nav.radius" wp="STDBY"/>
+    </block>
+    <block name="line_P1_P2_turn">
+      <go wp="P1" approaching_time="0"/>
+      <while cond="TRUE">
+        <go from="P1" hmode="route" wp="P2" approaching_time="0" pre_call="nav_set_heading_deg(turn_rate*block_time)"/>
```
noether: Out of curiosity. I am not familiar with pre_call inside the block. Is this called several times while in the "go"? What is the behavior of the rover here?

sw/airborne/filters/pid.h
noether: Not really. I think setting the field "sum" directly should be fine although not elegant :P

PR #2448, opened 2019-07-04
Created a default Bebop aircraft using optitrack and full INDI

danielwillemsen:
1. Created a default indoor optitrack bebop which uses full indi and can directly be used to fly in the cyberzoo. Control effectiveness values have been determined by using the adaptive INDI during test flights on 2 different bebops.
2. Added a session in the TU Delft userconf which is suitable for flight with joystick and optitrack.


danielwillemsen:
> Is it for Bebop1 or Bebop2?
It is for Bebop1, should I make this more clear in the airframe file?

noether:
> > Is it for Bebop1 or Bebop2?
> 
> It is for Bebop1, should I make this more clear in the airframe file?

Yes please, some people might understand by 'Bebop' the latest one.
Thanks!

PR #2449, opened 2019-07-04
Add e-identification module.

Fabien-B:
A law will be publish soon to comply with the EU regulation that will require every drones to broadcast its position over wifi beacon frames.
This module format a compliant frame and send it through UART so an external chip (e.g. esp8266) can broadcast it over wifi.


podhrmic:
Interesting! So instead of [ADS-B ](https://www.faa.gov/nextgen/programs/adsb/) the users in France (or whole EU?) will be using wifi? @Fabien-B do you have more info about the technology? I am just imagining that the wifi beacons won't have very long reach?

Fabien-B:
It is not really an UTM thing (or U-space as its called in EU) to track drones and give services, its much more a law enforcement tool, where policemen will be able to report the ID of a given drone, and easily locate the pilot (then some kind of registrar will exist to link the ID back to the owner).
In the required data to send, there is of course the drone position and its height, but also its take-off position.
The reach is not really specified but it is written that the device must comply with the power limits on the frequency, and emit at the maximum allowed power, so I assume its a way to get a good enough range.
The good thing with wifi is that much of the current drones already have it so its not a strong requirement for the manufacturers, and any phone can be turned into a receptor, so its cheap to implement, and good enough for the "every day law enforcement".

Also, ADS-B is not seriously (et least for many people, not all) considered as a good option for drone position reporting because of the frequency over use, and its poor precision (as it was made for planes. DLR made an experiment in which the drone position was "jumping" because of not enough digits in the ADS-B)

sw/airborne/modules/e_identification_fr/e_identification_fr.h
```diff
@@ -0,0 +1,52 @@
...
+  E_ID_LON_TO = 9,
+  E_ID_H_SPEED = 10,
+  E_ID_ROUTE = 11
+  // 12 to 200 are reserved for future use
+};
+
+extern void e_identification_fr_init(void);
+extern void e_identification_fr_periodic(void);
+
+#define E_Identification_Fr_Start() (e_identification_fr_e_identification_fr_periodic_status = MODULES_START)
```
fvantienen: Are these defines really needed?

conf/modules/e_identification_fr.xml
```diff
@@ -0,0 +1,30 @@
...
+  </doc>
+  <header>
+    <file name="e_identification_fr.h"/>
+  </header>
+  <init fun="e_identification_fr_init()"/>
+  <periodic fun="e_identification_fr_periodic()" freq="4" autorun="FALSE"/>
+  <makefile target="ap">
+    <configure name="E_ID_PORT"  case="upper|lower"/>
+    <configure name="E_ID_BAUD" default="B115200"/>
+    <define name="USE_$(E_ID_PORT_UPPER)"/>
```
fvantienen: You can also disable the RX of this UART because you don't seem to use it.

sw/airborne/modules/e_identification_fr/e_identification_fr.h
```diff
@@ -0,0 +1,52 @@
...
+#ifndef E_IDENTIFICATION_FR_H
+#define E_IDENTIFICATION_FR_H
+
+enum e_id_type {
+  // 0 reserved for future use
+  E_ID_PROTOCOL_VERSION = 1,
+  E_ID_ID_FR = 2,
+  E_ID_ID_ANSI_UAS =3,
+  E_ID_LAT = 4,
+  E_ID_LON =5,
```
fvantienen: Please run formatting on all files

sw/airborne/modules/e_identification_fr/e_identification_fr.c
```diff
@@ -0,0 +1,181 @@
...
+    buf[1] = offset + 2;  //include from STX to checksum_b
+    
+    uint8_t ck_a = 0;
+    uint8_t ck_b = 0;
+    for(int i=1; i<offset; i++) {
+      ck_a += buf[i];
+      ck_b += ck_a;
+    }
+    buf[offset++] = ck_a;
+    buf[offset++] = ck_b;
```
fvantienen: If you are using PPRZ messages why don't you add some intermcu messages and use the pprzlink library?

sw/airborne/modules/e_identification_fr/README.md
```diff
@@ -0,0 +1,6 @@
+E-Identification-FR
+===================
+
+French law required that every drone broacast its position over wifi beacon frames.
+Use this module with an ESP8266 with and code : https://github.com/enacuavlab/FrDroneID/tree/master/airborne_e_id
```
fvantienen: Maybe add a copy at the paparazzi organisation?

sw/airborne/modules/e_identification_fr/e_identification_fr.c
```diff
@@ -0,0 +1,181 @@
...
+#include "subsystems/navigation/waypoints.h"
+#endif
+
+#define MAX_BUF_LEN 74
+
+#define LEN_ID_FR 30
+#define LEN_ID_SERIAL 24
+
+struct uart_periph *dev;
+char id_fr[LEN_ID_FR];
```
fvantienen: can be static

sw/airborne/modules/e_identification_fr/e_identification_fr.c
```diff
@@ -0,0 +1,181 @@
...
+#if defined(ROTORCRAFT_FIRMWARE)
+#include "subsystems/navigation/waypoints.h"
+#endif
+
+#define MAX_BUF_LEN 74
+
+#define LEN_ID_FR 30
+#define LEN_ID_SERIAL 24
+
+struct uart_periph *dev;
```
fvantienen: can be static, and should be static with such a generic name. Could also be initialised directly here.

sw/airborne/modules/e_identification_fr/e_identification_fr.h
Fabien-B: The module should be started at take-off and stopped at landing.
Its much more easier to start and stop the module from a flight plan with these defines.

conf/modules/e_identification_fr.xml
Fabien-B: Ok, did the same as in actuators_sbus.xml

sw/airborne/modules/e_identification_fr/e_identification_fr.h
Fabien-B: Should be ok now. Is there a tool to do it automagically ?

sw/airborne/modules/e_identification_fr/e_identification_fr.c
Fabien-B: ok

sw/airborne/modules/e_identification_fr/e_identification_fr.c
Fabien-B: ok

sw/airborne/modules/e_identification_fr/README.md
Fabien-B: Ok, the best may be to fork it ?

sw/airborne/modules/e_identification_fr/e_identification_fr.c
Fabien-B: Yes, I did not think about it, I will do it soon when I will have some time !

gautierhattenberger:
I suggest to use this https://github.com/enacuavlab/paparazzi/commit/40abf31fda8ff86480165b36f37b6560f7e85ca2 instead of firmware defines in this specific module. A more generic solution will be useful in many cases.

sw/airborne/modules/e_identification_fr/e_identification_fr.h
gautierhattenberger: `./fix_code_style.sh sw/airborne/modules/e_identification_fr/e_identification_fr.*`

PR #2450, opened 2019-07-04
[fix] fix some DMA issues and motor pins for DShot on Tawaki

gautierhattenberger:
Now, the servo B connector of the tawaki is the default for DShot.
Motor count starts at 1, instead of 0, I guess it is more common.

Also:
- activate the mag ahrs update by default on this board.
- board is used not rotated by default (when nothing is defined)


gautierhattenberger:
tested and working

PR #2452, opened 2019-07-09
TU Delft MAVLab course 2019

tomvand:
This pull request contains all changes for this year's MAVLab course, which was given in February.

In short:
- Two example airframes:
  - orangeavoid: uses waypoints for navigation
  - orangeavoid_guided: uses body-frame guidance commands. Could potentially be used without OptiTrack (not tested). Uses a front-camera green detector to stay within the flight area.
  - Camera defines have been updated after #2364. The color detector has been re-tuned afterwards.
  - All aircraft are test-flown and tagged.
- New joystick confs:
  - sm600.xml: for the new usb transmitters in the lab. Still a bit buggy sometimes, but flyable.
  - turnigy, xbox
- The orange avoider is updated and separated into two modules: a colored object detector and an avoidance module. The modules communicate over ABI. The avoidance logic has been moved from the airframe to the avoidance module to make to code easier to understand and adapt. (All credit goes to @kirkscheper for this one).
- opencv_bebop is updated to 3.4.5 and now also includes the contrib modules. OpenCV is now also compiled for NPS, so that the same configuration is used in the sim as on the real drone.
- video_capture has been updated to capture image sequences in addition to still frames.
- file_logger has been partially rewritten for readability.
- A python script is added to measure the flown distance of an aircraft. This script can be run on any pc as long as it's connected to OptiTrack (thanks @dewagter and Shushuai Li).

The course documentation can be found here: https://tudelft.github.io/coursePaparazzi/


conf/joystick/xbox_gamepad.xml
```diff
@@ -84,7 +84,7 @@ so e.g. HatDown(dpad)
 
     <!-- go to "land here" block if in AUTO2 and pressing down on dpad -->
     <message class="ground" name="JUMP_TO_BLOCK" on_event="(mode > 1) && HatDown(dpad)">
-      <field name="block_id" value="IndexOfBlock('land here')"/>
+      <field name="block_id" value="IndexOfBlock('Land here')"/>
```
gautierhattenberger: This will make this unusable for all other flight plans except yours. I don't mind changing the name of the land block, but then it has to be done everywhere.

conf/telemetry/default_rotorcraft.xml
```diff
@@ -13,7 +13,7 @@
       <message name="ALIVE"                    period="2.1"/>
       <message name="INS_REF"                  period="5.1"/>
       <message name="ROTORCRAFT_NAV_STATUS"    period="1.6"/>
-      <message name="WP_MOVED"                 period="1.3"/>
```
gautierhattenberger: Why is this needed ? WP_MOVED is normally sent as an acknowledgment when a WP is moved. The message in telemetry file should be a low freq update in case a message is missed or when the ref point is changed.
This might significantly increase the telemetry rate, so I don't think it is good to have it in the default file. If it is really needed, then make a dedicated file, or make a module that directly send the WP_MOVED acknowledgment message at the correct rate.

sw/airborne/modules/loggers/file_logger.c
```diff
@@ -106,56 +140,10 @@ void file_logger_stop(void)
...
-  static uint32_t counter;
-  struct Int32Quat *quat = stateGetNedToBodyQuat_i();
-
-#ifdef COMMAND_THRUST //For example rotorcraft
-  fprintf(file_logger, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n",
-          counter,
-          imu.gyro_unscaled.p,
-          imu.gyro_unscaled.q,
-          imu.gyro_unscaled.r,
-          imu.accel_unscaled.x,
```
gautierhattenberger: accel and mag are not part of this log anymore ?

sw/tools/opti_dist/NatNetClient.py
```diff
@@ -0,0 +1,438 @@
...
+def trace( *args ):
+    pass # print( "".join(map(str,args)) )
+
+# Create structs for reading various object types to speed up parsing.
+Vector3 = struct.Struct( '<fff' )
+Quaternion = struct.Struct( '<ffff' )
+FloatValue = struct.Struct( '<f' )
+DoubleValue = struct.Struct( '<d' )
+
+class NatNetClient:
```
gautierhattenberger: it seems to be a duplicate of https://github.com/paparazzi/paparazzi/blob/master/sw/ground_segment/python/natnet3.x/NatNetClient.py
Please merge if needed and integrate to current NatNet3 support.

sw/tools/opti_dist/dist.py
```diff
@@ -0,0 +1,66 @@
...
+# Configure the streaming client to call our rigid body handler on the emulator to send data out.
+streamingClient.newFrameListener = receiveNewFrame
+streamingClient.rigidBodyListener = receiveRigidBodyFrame
+# Start up the streaming client now that the callbacks are set up.
+# This will run perpetually, and operate on a separate thread.
+streamingClient.run()
+
+time.sleep(2)
+print('Start tracking')
+# file = open('opti.csv', 'w')
```
gautierhattenberger: I suggest to use runtime argument to specify the log file name and write the data when needed rather than editing this file.

conf/joystick/sm600.xml
```diff
@@ -0,0 +1,47 @@
...
+    <message class="datalink" name="RC_4CH" send_always="true">
+      <field name="throttle" value="Fit(LeftStickVertical,-90,127,0,127)"/>
+      <field name="roll" value="RightStickHorizontal"/>
+      <field name="pitch" value="-RightStickVertical"/>
+      <field name="yaw" value="LeftStickHorizontal"/>
+      <field name="mode" value="2-2*VRA"/>
+    </message>
+  </messages>
+
+</joystick>
```
gautierhattenberger: missing newline at end of file

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -429,11 +429,21 @@
...
-
 #ifndef DETECT_GATE_ABI_ID
 #define DETECT_GATE_ABI_ID 33
 #endif
 
+/*
+ * VISUAL_DETECTION communication (message 27)
+*/
+#ifndef COLOR_OBJECT_DETECTION1_ID
+#define COLOR_OBJECT_DETECTION1_ID 1
```
gautierhattenberger: Wouldn't it be more extensible to add an ID field (camera, object or whatever is needed) to the ABI message rather than multiple sender ID ?

sw/airborne/modules/loggers/file_logger.c
tomvand: No, these were not relevant for the course. Instead I added the drone's velocity, attitude and rates. I assumed that people would re-add these measurements if they need them, as they will probably modify this file to their own needs anyway. It's not a problem to re-add them though, do you think I should do that?

conf/joystick/xbox_gamepad.xml
tomvand: I changed this back to the lowercase 'land here' (also in the flight plan) to keep things compatible with other flight plans that may not have been merged into master yet.

conf/telemetry/default_rotorcraft.xml
tomvand: The avoidance module used the waypoints_set_ functions which do not send this acknowledgement. I was not aware of this so I increased the update rate.

I have implemented the missing waypoint_move_ functions which *do* send this message, and restored the WP_MOVED update rate to its original value.

conf/joystick/sm600.xml
tomvand: fixed

sw/tools/opti_dist/dist.py
tomvand: I have cleaned up the script and added the output file as an argument. Will re-test asap.

sw/tools/opti_dist/NatNetClient.py
tomvand: I do not know what changed in this version of the NatNetClient, so I won't merge it. I have rewritten the dist.py script to load the client in the natnet3.x folder instead.

sw/airborne/subsystems/abi_sender_ids.h
tomvand: No. The number of filters is also limited by the inability of the airframe and module settings to handle an arbitrary number of defines/variables. (For instance, right now the filter number is hardcoded into the define name, as in COLOR_OBJECT_DETECTOR_LUM_MIN**1**). This is already causing some code duplication inside the detect_color_object module.

I did some experiments where a single define could set up an arbitrary number of filters. This is technically possible but I ran into some problems (#2454). It would also not provide a solution for the settings in the GCS.

I can move the ID inside the ABI message (in fact, this is already done by the detector) and test for this value inside a single callback, but this would not make the system any more extensible.

sw/airborne/modules/loggers/file_logger.c
gautierhattenberger: at some point, this file logger should be a generic API to write anything and the data to be written defined elsewhere, but that would be for another time

sw/airborne/subsystems/abi_sender_ids.h
gautierhattenberger: does my comments for #2454 helps you in any way ?

tomvand: It made it clear why my attempts to fix this did not work, but even if #2454 could be fixed it would still not be possible to add a variable number of settings to the module.
I would rather leave the code as-is for now as this has already been testflown and handling the IDs differently would not really improve much.

PR #2461, opened 2019-10-04
FIX usb_transparent STM32F4 GPIOA clock not enabled

tomvand:
Fixes USB serial port not working on the Eachine Trashcan CrazybeeF4FR board. The GPIOA peripheral clock enable from the [STM32F4 example](https://github.com/libopencm3/libopencm3-examples/blob/f9713ff86508a522a2c02bde947b57a871bc34bc/examples/stm32/f4/stm32f4-discovery/usb_cdcacm/cdcacm.c#L231) was missing.


podhrmic:
LGTM, thanks for the PR!

PR #2464, opened 2019-10-08
[module] add target_localization module

gautierhattenberger:
Provide the position of a target on a flat ground based on detection
from a camera (VISUAL_DETECTION ABI message) and the current position
and orientation of the vehicle.
Developed and used during IMAV2019 competition.

### DIFF
--- a/sw/airborne/modules/computer_vision/cv_target_localization.h
+++ b/sw/airborne/modules/computer_vision/cv_target_localization.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2019 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ *
+ * This file is part of paparazzi
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/computer_vision/cv_target_localization.h
```diff
@@ -0,0 +1,46 @@
...
+
+extern void target_localization_init(void);
+extern void target_localization_report(void);
+
+// settings and handlers
+extern uint8_t target_localization_mark;
+extern void cv_target_localization_report_mark(uint8_t mark);
+extern bool target_localization_update_wp;
+
+// TODO move functionality to the camera driver
```
podhrmic: Should the TODO be addressed before merging?

gautierhattenberger: I may not have time to do it soon, so I want to keep it as a reminder and test it properly before making a second PR.

PR #2465, opened 2019-10-10
[fix] bungee_takeoff was not turning motor on

gautierhattenberger:
Two issues:
- launch variable was not set to true which prevents the control to set throttle command
- the autopilot_set_kill_throttle had not effects on fixedwing (in fact always setting to true). We probably missed this because this function is mostly called to kill but not to resurrect motor...

Second point is a major bug I think...

### DIFF
--- a/sw/airborne/firmwares/fixedwing/autopilot_generated.c
+++ b/sw/airborne/firmwares/fixedwing/autopilot_generated.c
@@ -82,9 +82,10 @@ void autopilot_generated_set_mode(uint8_t new_autopilot_mode)
 }
 
 
-void autopilot_generated_set_motors_on(bool motors_on __attribute__((unused)))
+void autopilot_generated_set_motors_on(bool motors_on)
 {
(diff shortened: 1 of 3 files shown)


OpenUAS:
A autopilot.launch=1 could also have been set in the flightplan... (  <var="autopilot.launch" set value="1"  />  ) . Now we lose a bit of control since the module sets it, which is a bit more risky in case the launch fails. Btw  autopilot_set_kill_throttle only set it to idle on gas engines ;) Anyhow none cares likely except me no one uses it. If it bothers in the future I'll make an improved PR...

gautierhattenberger:
We actually had launch failures because `launch` was not set.
Also I don't agree that setting from the flight plan is the same. If you set before calling the run function, it means you are waiting with the launch variable already set to true, and if you set it after, the motor won't start when requested and the procedure is likely to fail. With this solution, `launch` is set to true right when you need it, not before, and you will not forget to do it in the FP or by hand.

PR #2467, opened 2019-10-17
[pprz] change the old paparazzi launch script from ocaml to python

gautierhattenberger:
This new script is doing the same than before but don't rely on the
ocaml install which was a problem when mixing opam and system install.

Note that when updating, git will complain about the already existing
file 'paparazzi'. So it is necessary to delete the old file (or run
'make clean') before updating the code.


noether:
Oh,

I did not check that you requested @fvantienen to check the commit before merging the pull request. Sorry guys, hopefully, it is okei for you.

noether:
Apparently, it does not work in Ubuntu 18.04

Traceback (most recent call last):
  File "./paparazzi", line 20, in <module>
    os.execve(path, args, env)
ValueError: execve: argv must not be empty

If you execute it with Python2 is okei. Gautier tested it in ubuntu 16.04 with Python3 and it works fine too.

PR #2469, opened 2019-10-29
Collective tracking control

noether:
This code is quite old, and I polished it a bit. 

It is the implementation of a tracking algorithm coming from this paper https://arxiv.org/abs/1810.00182 .

I have written some instructions in the wiki https://wiki.paparazziuav.org/wiki/Module/collective_tracking_control

For sure I am missing documentation. To be honest, I forgot exactly all the necessary doc items. So, if you do not mind, just tell me what is missing :P.

The "attached" Python simulation to tune the gains quickly is hardcoded for three aircraft + target. One day I should make it more general... but I believe it is easy to modify it by somebody if (s)he is interested in the algorithm.


conf/modules/collective_tracking_control.xml
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="collective_tracking_control" dir="ctc">
+  <doc>
+  <description>Distributed algorithm for tracking a velocity/position reference signal. This module needs of custom messages in messages.xml. Please, check the wiki https://wiki.paparazziuav.org/wiki/Module/collective_tracking_control for more information about the messages and how to tune the gains.
```
gautierhattenberger: When the doc is generated, the first line is used as "short description" and the rest as "detailed description". Can you separate the first line from the rest ?

conf/modules/collective_tracking_control.xml
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="collective_tracking_control" dir="ctc">
```
gautierhattenberger: In order to limit the number of directory in the modules folder, would it be possible to move this to either `multi` or `ctrl` dir ? Same for the target part (and DCF, but that is for another PR ;) )

conf/modules/collective_tracking_control_target.xml
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="collective_tracking_control_target" dir="ctc">
+  <doc>
+	  <description>Distributed algorithm for tracking a velocity/position reference signal. Module only for the target to broadcast its position and velocity to the pursuers.
```
gautierhattenberger: Same comment that the other xml file

sw/airborne/modules/ctc/ctc.h
```diff
@@ -0,0 +1,69 @@
...
+#define CTC_H
+
+#include "std.h"
+
+typedef struct {
+  float k1;
+  float k2;
+  float alpha;
+  uint16_t timeout;
+  float p_centroid_x;
```
gautierhattenberger: don't forget that you can use the `struct FloatVect2` and corresponding math operations from `math/pprz_algebra.h` and `math/pprz_algebra_float.h` to have a more compact code.

noether:
Sure, it makes sense. Especially, to move to another directory the "multi" algorithms (yep, for the next PR).

I will write two functionalities more too:

- I will add the option to set a fixed target from settings or from a waypoint (right now it is always another vehicle). 
- I will add a Python script to start the algorithm (send navigation block) simultaneously from the ground.

sw/airborne/modules/ctc/ctc.c
```diff
@@ -0,0 +1,285 @@
...
+    ctc_control.target_py = moving_target_py;
+    ctc_control.target_vx = moving_target_vx;
+    ctc_control.target_vy = moving_target_vy;
+
+    collective_tracking_control();
+
+    return true;
+}
+
+bool collective_tracking_waypoint(uint8_t wp)
```
gautierhattenberger: `wp` is not used here ? same for `x,y` below ?

sw/airborne/modules/ctc/ctc.c
```diff
@@ -0,0 +1,285 @@
...
+      tableNei[i][2] = (int16_t)(DL_CTC_INFO_TO_NEI_vy(dl_buffer) * 100);
+      tableNei[i][3] = (int16_t)(DL_CTC_INFO_TO_NEI_px(dl_buffer) * 100);
+      tableNei[i][4] = (int16_t)(DL_CTC_INFO_TO_NEI_py(dl_buffer) * 100);
+      break;
+    }
+}
+
+void parse_ctc_TargetInfo(void)
+{
+  float moving_target_px = DL_CTC_INFO_FROM_TARGET_px(dl_buffer);
```
gautierhattenberger: these are local variables, you are not changing the global one

noether:
Do not look at these commits yet. I had to push them from my laptop because I had to stop coding (baby alarm :P). I work from two or three computers, so every push in my local repo shows up here, that is why.

Once I finish the changes, I will let you know to review them.

Thanks!

sw/airborne/modules/ctc/ctc.c
noether: fixed

sw/airborne/modules/ctc/ctc.c
noether: fixed

sw/airborne/modules/ctc/ctc.h
noether: I totally forgot about it. I would prefer to do not touch this part of the code since it is working (validated).

I have another algorithm to implement later. I will use these libraries for sure.

conf/modules/collective_tracking_control_target.xml
noether: Sure, I agree. I will do that for the next PR.

conf/modules/collective_tracking_control.xml
noether: Sure, I agree.

conf/modules/collective_tracking_control.xml
noether: Oh, I did not do that in my last push. I will make the change and push it.

gautierhattenberger:
It seems you need to rebase or merge with master and remove conflicts before I can merge it

PR #2471, opened 2019-11-15
[jevois] move some functionnality to jevois cam driver

gautierhattenberger:
The Jevois driver has the option to send some state information such as altitude, position and orientation.

The extraction of ID number is skipping the potential letters at the beginning of the ID string, as well as trailing characters if any.
This is allows to use the default Aruco demo for instance since the ID reported begins with 'U'.


gautierhattenberger:
This code is compatible with most of the Jevois standard detector. We are usually using the Aruco marker with this one.
There are some custom modules here: https://github.com/paparazzi/pprz-jevois

PR #2472, opened 2019-11-15
[build] fix build dep for tools

### LINKED ISSUE #2228: Build dependencies (seems) not correctly setup
When running a make with -j8 on an eight core machine I get the following error:

[error.txt](https://github.com/paparazzi/paparazzi/files/1694479/error.txt)

It seems like that the dependencies of the pprzlink build is not correctly setup. I was looking into this specifically to try to speed up the travis build so that we can use as much of the build time for airframes. I took a quick look at it but not that experienced with makefiles so not too efficient at debugging. Can someone take a look at this and suggest a solution?

gautierhattenberger:
fix #2228

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ static: cockpit tmtc generators sim_static joystick static_h
 
 libpprzlink:
 	$(MAKE) -C $(EXT) pprzlink.update
-	$(Q)Q=$(Q) DESTDIR=$(PPRZLINK_INSTALL) PPRZLINK_LIB_VERSION=${PPRZLINK_LIB_VERSION} $(MAKE) -C $(PPRZLINK_DIR) libpprzlink-install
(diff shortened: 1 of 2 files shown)


OpenUAS:
Did:

 $ git fetch upstream
 $ git checkout -b fix-build-dep upstream/fix-build-dep
 $ git merge master

Tested with :
 $ make clean -9 && make -j9  -> OK
 
and with:
 $ make clean -j9 && make libpprzlink -j9  -> OK

one small part is then still forced -j1... Well OK did not even notice the difference in buildspeed ;)

So yeah all very fine AFAICS

make_before
clean_after
make_after

OpenUAS:
Future improvment maybe inhibit the parallelism in makefile with the .NOTPARALLEL  .. and maybe per default add -j

PR #2475, opened 2019-12-06
Testfly parrot disco phase2

OpenUAS:
Have a dozen other fixes in various branches never landing up in master so wil propose all those in reasonable order as PR. we would lose else fixes and lots of work done by TU students... anyhow first then this. Tonight then RTCM3 fixes and more EKF2 fixes as PR


sw/airborne/boards/bebop/mt9v117.c
```diff
@@ -37,6 +37,14 @@
 #include <linux/videodev2.h>
 #include <linux/v4l2-mediabus.h>
 
+#include "generated/airframe.h"
+#ifdef BOARD_DISCO
+#include "boards/disco.h"
```
gautierhattenberger: same comments than mt9f002.c

sw/airborne/boards/disco.h
```diff
@@ -53,10 +53,12 @@
 #endif
 
 /* Default actuators driver */
+//#ifdef BOARD_DISCO
```
gautierhattenberger: please remove unneeded commented code from this file

sw/airborne/modules/sonar/sonar_bebop.c
```diff
@@ -206,20 +225,21 @@ void *sonar_bebop_read(void *data __attribute__((unused)))
...
       // Send ABI message
       uint32_t now_ts = get_sys_time_usec();
       AbiSendMsgAGL(AGL_SONAR_ADC_ID, now_ts, sonar_bebop.distance);
-#ifdef SENSOR_SYNC_SEND_SONAR
-      // Send Telemetry report
-      DOWNLINK_SEND_SONAR(DefaultChannel, DefaultDevice, &sonar_bebop.meas, &sonar_bebop.distance);
-#endif
+
 #ifndef SITL
+    #ifdef SENSOR_SYNC_SEND_SONAR
```
gautierhattenberger: why not making this message available even in SITL ?

sw/airborne/peripherals/ak8963.c
```diff
@@ -37,10 +37,11 @@ void ak8963_init(struct Ak8963 *ak, struct i2c_periph *i2c_p, uint8_t addr)
   ak->i2c_p = i2c_p;
   /* set i2c address */
   ak->i2c_trans.slave_addr = addr;
-  ak->i2c_trans.status = I2CTransDone;
+  ak->i2c_trans.status = I2CTransFailed;
```
gautierhattenberger: why setting this to Failed then Done at the end of the init ?

sw/tools/parrot/disco.py
```diff
@@ -38,7 +38,7 @@ def uav_status(self):
     def init_extra_parser(self):
 
         # Parse the extra arguments
-        self.parser.add_argument('--min_version', metavar='MIN', default='1.4.1',
+        self.parser.add_argument('--min_version', metavar='MIN', default='1.7.1',
```
gautierhattenberger: why rising the min version ? is it really not compatible anymore ?

sw/tools/parrot/disco.py
```diff
@@ -51,5 +51,4 @@ def parse_extra_args(self, args):
 if __name__ == "__main__":
     disco = Disco()
     disco.parse_args()
-    exit(0)
-
+    exit(0)
```
gautierhattenberger: a bunch of files have lost their newline at end of file, any reason for that ?

sw/airborne/boards/bebop.h
```diff
@@ -50,10 +50,12 @@
 #endif
 
 /* Default actuators driver */
+#ifdef BOARD_BEBOP
```
gautierhattenberger: this file can be used with something else than a bebop ?

sw/airborne/modules/sonar/sonar_bebop.c
OpenUAS: Yes, good idea

sw/tools/parrot/disco.py
OpenUAS: It is, as far as our last test compatible for upload but we do never fly anymore with a lower v than 1.7.1 The reason is if we want to use the Hardware encoding is could cause .ko/.so entrypoint issues again, And since Parrot stopped making the Disco v1.7.1 will be the last firmware ever. All Discos we test on (8) have v1.7.1. If there would be a benefit I'm fine with 1.4.1 but have no plan to test it since there is no benefit ATM and will heighten the risk someting goes wrong. Where it boiled down to was we want to limit the amount of firmware we need to testfly on just to wards one perfect flying Disco AC

sw/airborne/boards/disco.h
OpenUAS: OK

sw/airborne/peripherals/ak8963.c
OpenUAS: Its been a wilde ago, but there was someting, dont recall, so will have to do a new longflight > 45 min extended batt to validate.

gautierhattenberger: this is the init function, only called once after boot, so I don't see how it is related to long flight ?

sw/tools/parrot/disco.py
gautierhattenberger: Increasing min version don't allow old airframe that might not have been updated to fly. I would not change that unless necessary because it is just not necessary. I don't think we need to test the older firmware either. If someone report an issue, we might increase the value later instead of supporting something not produced anymore.

sw/airborne/modules/sonar/sonar_bebop.c
OpenUAS: ....Will then also enable SONAR_COMPENSATE_ROTATION and SONAR_BEBOP_FILTER_NARROW_OBSTACLES to be used in SITL

sw/airborne/boards/bebop/mt9f002.c
```diff
@@ -38,6 +38,13 @@
 #include <linux/i2c-dev.h>
 #include <linux/videodev2.h>
 
+#include "generated/airframe.h"
+#ifdef BOARD_DISCO
+#include "boards/disco.h"
```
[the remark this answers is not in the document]
OpenUAS: Yes correct. Howver recall there was some issue if BOARD_CONFIG was used after all the cam refactor Will update in multiple PR to get better Parrot support.

sw/airborne/peripherals/ak8963.c
OpenUAS: But there was something like  ak->init_status = AK_CONF_UNINIT; the evet got already called and there it wa not in AK_STATUS_IDLE or AK_STATUS_READ thus to default but there only   I2CTransSuccess or I2CTransFailed make it go to idle...so if not set to I2CTransFailed in the begin it is never reached. I will change and retest, but this was one of the once in so many times cases I remember.

sw/tools/parrot/disco.py
OpenUAS: I understand, so fine byme I'll leave to v1.4.1 as min, not however we will only testfly v1.7.1 thus we cannot guarantee v1.4.1 still works 100% with PPRZ Master then. But it will not crash for us either then because of that ;)

sw/tools/parrot/disco.py
OpenUAS: Except for a setting mistake swapping between PC's and someones different IDE or editors no.. so will re-add

sw/airborne/peripherals/ak8963.c
gautierhattenberger: how is it possible that `event` functions are been called before and of `init` ? (https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/fixedwing/main.c#L45)

PR #2478, opened 2019-12-18
Fixes PR #2475

OpenUAS:
Made requested changes


sw/airborne/peripherals/ak8963.c
```diff
@@ -114,11 +114,11 @@ void ak8963_event(struct Ak8963 *ak)
         ak->data.vect.x = Int16FromBuf(ak->i2c_trans.buf, 0);
         ak->data.vect.y = Int16FromBuf(ak->i2c_trans.buf, 2);
         ak->data.vect.z = Int16FromBuf(ak->i2c_trans.buf, 4);
-        ak->data_available = true;
 
         // Read second status register to be ready for reading again
-        ak->i2c_trans.buf[0] = AK8963_REG_ST2;
+        ak->i2c_trans.buf[0] = AK8963_REG_ST2; // Data overflow bit 3 and data read error status bit 2
         i2c_transceive(ak->i2c_p, &(ak->i2c_trans), ak->i2c_trans.slave_addr, 1, 1);
+        ak->data_available = true;
```
gautierhattenberger: New data flag should only be set to true one time. If you keep it here and then request the status to either reset it to true or false, you might either read the data twice or read false values. Before, the error flag was just ignored, we can change that but if should be consistent.

OpenUAS: Ok, thanks for looking into it. So I'll remove line 121 ( ak->data_available = true;  ) is that correct what you request?

OpenUAS: Whatever solution is fine

> This contains a lot of changes

Yeah, sorry, this comes since I set sub projects like e.g. "Make disco fly 100% using all useful sensor" then one comes across many issue, only flight testable often when multiple things are fixed. I'll see if in the future I'll can spit PR's. There are limited resources for PPRZ fixing and testflying things. This was Stage 2...

PR #2482, opened 2020-01-24
Refresh WMM data for 2020 to 2025

OpenUAS:
With help of the COF file from https://www.ngdc.noaa.gov/geomag/WMM/soft.shtml  updated data.

Great care was taken that all data is moved to the array correctly.  It would be really good someone else **revalidates** the tables.

Since the tables are build exactly like in 2010,2015 but ordering of field was **never validated** compared to the use of g1, g2 in the **source code** my me.

It would be a **must** IMHO to make sure the field coefficients ordering in the array are correct in the first place since the WMM.COF file does not have much of a description of fields...

- [ ] Re-validated table correctness by 2nd person
- [ ] Validated array field ordering correctness by 2nd person

Based on this file [WMM2020COF.zip](https://github.com/paparazzi/paparazzi/files/4107238/WMM2020COF.zip)

_it is a dirty job but someone has to do it ;)  then we are off the hook for 5 years again..._


gautierhattenberger:
Please don't remove unrelated files (the `*_ui.py` files from the new create module tool)

OpenUAS:
I thought they were dynamically re-created at build time, if not, my bad.

PR #2484, opened 2020-01-26
Wls update

EwoudSmeur:
I wrote a test for the WLS control allocation and found a mistake with the initialization of one of the arrays. The fix gives the correct result in the test, but is not flight tested.

Something else, but related:
I also realized that the use of Variable Length Arrays (VLA) leads to a lot more assembly code than fixed size arrays, and in this case it is not really needed if you call the algorithm with arguments of the same size. (Usage example: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/stabilization/wls/wls_alloc.c#L109)
I think that only if you would have two constrained least squares problems of different size to solve it could make sense to do it like this. Should we restrict it to 'single use' in order to make it faster?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/wls/wls_alloc.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/wls/wls_alloc.c
@@ -170,7 +170,7 @@ int wls_alloc(float* u, float* v, float* umin, float* umax, float** B,
   for (int i = n_v; i < n_c; i++) {
     memset(A[i], 0, n_u * sizeof(float));
     A[i][i - n_v] = Wu ? Wu[i - n_v] : 1.0;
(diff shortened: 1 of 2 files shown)


noether:
@EwoudSmeur it looks good to me. May you make happy Travis and his semaphore? :P

gautierhattenberger:
Do we use variable length arrays in airborne code ?

EwoudSmeur:
Thanks!

@gautierhattenberger Yes so maybe I wasn't completely clear, I was only talking about the WLS code. Some years ago I replaced the use of real dynamic memory in the WLS code with these VLA's, but now I think it is better not to use them, especially because it is not strictly needed. Judging from your comment you would agree? :)

gautierhattenberger:
indeed, if you can use fixed size arrays, it is better

PR #2487, opened 2020-02-07
Dynamic grid in rectangle surveys & New flight_plan feature

jburns11:
This edit allows the user survey a rectangle and change the sweep width (grid argument) without restarting the survey. Care was taken to preserve backwards compatibility with calling the periodic rectangle survey function with two waypoint arguments. This means that all the old flight_plans can still be compiled without edit.

This new feature can also be examined by using the new flight plan directive as such...

<survey_rectangle_dynamic wp1="p1" wp2="p2" grid="var_grid" orientation="NS"/>

This would use the variable var_grid as the sweep argument. If we adjust the var_grid in the middle of our survey, the survey will not be interrupted. Instead the updated argument will be used with the subsequent periodic calls to the survey rectangle function.

This appears to work with rotor crafts, fixed wings, and with either orientation.

### DIFF
--- a/sw/airborne/modules/nav/nav_survey_rectangle_rotorcraft.c
+++ b/sw/airborne/modules/nav/nav_survey_rectangle_rotorcraft.c
@@ -163,7 +163,7 @@ void nav_survey_rectangle_rotorcraft_setup(uint8_t wp1, uint8_t wp2, float grid,
 
 bool nav_survey_rectangle_rotorcraft_run(uint8_t wp1, uint8_t wp2)
 {
-  return nav_survey_rectangle_rotorcraft_run_dynamic(uint8_t wp1, uint8_t wp2, -1);
+  return nav_survey_rectangle_rotorcraft_run_dynamic(wp1, wp2, -1);
 }
 
 bool nav_survey_rectangle_rotorcraft_run_dynamic(uint8_t wp1, uint8_t wp2, float grid)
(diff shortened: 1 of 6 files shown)


jburns11:
http://lists.paparazziuav.org/Allowing-variable-grid-argument-in-rectangle-survey-td18501.html

sw/airborne/modules/nav/nav_survey_rectangle_rotorcraft.c
```diff
@@ -161,9 +161,18 @@ void nav_survey_rectangle_rotorcraft_setup(uint8_t wp1, uint8_t wp2, float grid,
...
 
-
 bool nav_survey_rectangle_rotorcraft_run(uint8_t wp1, uint8_t wp2)
 {
+  return nav_survey_rectangle_rotorcraft_run_dynamic(wp1, wp2, -1);
+}
+
+bool nav_survey_rectangle_rotorcraft_run_dynamic(uint8_t wp1, uint8_t wp2, float grid)
+{
+  if (grid > 0) {
```
noether: May you use the ternary operator here? I think it is easier to read the code than to have two ifs and one else.

sw/airborne/subsystems/navigation/nav_survey_rectangle.c
```diff
@@ -86,9 +86,18 @@ void nav_survey_rectangle_init(uint8_t wp1, uint8_t wp2, float grid, survey_orie
...
 
-
 void nav_survey_rectangle(uint8_t wp1, uint8_t wp2)
 {
+  nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, -1)
+}
+
+void nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, float grid)
+{
+  if (grid > 0) {
```
noether: Same here about using the ternary operator if you do not mind.

jburns11:
Thanks for the feedback. I will add these changes to the run functions.

PR #2488, opened 2020-02-07
Dynamic grid in rectangle surveys & New flight_plan feature

jburns11:
This edit allows the user survey a rectangle and change the sweep width (grid argument) without restarting the survey. Care was taken to preserve backwards compatibility with calling the periodic rectangle survey function with two waypoint arguments. This means that all the old flight_plans can still be compiled without edit.

This new feature can also be examined by using the new flight plan directive as such...

<survey_rectangle_dynamic wp1="p1" wp2="p2" grid="var_grid" orientation="NS"/>

This would use the variable var_grid as the sweep argument. If we adjust the var_grid in the middle of our survey, the survey will not be interrupted. Instead the updated argument will be used with the subsequent periodic calls to the survey rectangle function.

This appears to work with rotor crafts, fixed wings, and with either orientation.

P.S. Sorry I created and closed a previous PR. I realized I had a couple silly errors so I fixed them and created a new request.

### DIFF
--- a/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
+++ b/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
@@ -34,6 +34,7 @@ static struct point survey_from;
 static struct point survey_to;
 static bool survey_uturn __attribute__((unused)) = false;
 static survey_orientation_t survey_orientation = NS;
+float sweep_var;
 
 #define SurveyGoingNorth() ((survey_orientation == NS) && (survey_to.y > survey_from.y))
 #define SurveyGoingSouth() ((survey_orientation == NS) && (survey_to.y < survey_from.y))
@@ -82,20 +83,17 @@ void nav_survey_rectangle_init(uint8_t wp1, uint8_t wp2, float grid, survey_orie
     }
   }
   nav_survey_shift = grid;
+  sweep_var = grid;
   survey_uturn = false;
   LINE_START_FUNCTION;
 }
 
 void nav_survey_rectangle(uint8_t wp1, uint8_t wp2)
 {
-  nav_survey_rectangle_dynamic(wp1, wp2, -1);
-}
-
-void nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, float grid)
-{
-  if (grid > 0) {
-    if (nav_survey_shift > 0) nav_survey_shift = grid;
-    else nav_survey_shift = -grid;
+  /* Update sweep_width if "extern float sweep" has been updated */
+  if (sweep != nav_survey_shift && sweep != -nav_survey_shift)
+  {
+    nav_survey_shift = (nav_survey_shift > 0 ? sweep_var : -sweep_var);
   }
 
   static float survey_radius;
--- a/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
+++ b/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
@@ -88,7 +88,7 @@ void nav_survey_rectangle_init(uint8_t wp1, uint8_t wp2, float grid, survey_orie
 
 void nav_survey_rectangle(uint8_t wp1, uint8_t wp2)
 {
-  nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, -1);
+  nav_survey_rectangle_dynamic(wp1, wp2, -1);
 }
 
 void nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, float grid)
--- a/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
+++ b/sw/airborne/subsystems/navigation/nav_survey_rectangle.c
@@ -88,7 +88,7 @@ void nav_survey_rectangle_init(uint8_t wp1, uint8_t wp2, float grid, survey_orie
 
 void nav_survey_rectangle(uint8_t wp1, uint8_t wp2)
 {
-  nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, -1)
+  nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, -1);
 }
 
 void nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, float grid)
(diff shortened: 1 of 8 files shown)


jburns11:
This makes more sense to me! I will adjust the gen_flight_plan.ml and flight_plan.dtd files.

noether:
> First, please note that you don't need to close/reopen a PR to make changes. You can just push (or even force push) new commit to the same branch and this will update the PR.
> 
> I'm mostly wondering if we should create a new xml element (`survey_rectangle_dynamic` in this case), let the users call the procedure by hand (like in https://github.com/paparazzi/paparazzi/blob/master/conf/flight_plans/rotorcraft_survey.xml), or add an attribute to select the type of survey (e.g. `type="dynamic"`).
> The later one has the benefit to limit the number of duplicated lines in flight plan generator (testing which "run" function should be called) and can be extended in the future for other types with minimal changes.
> @jburns11 @noether what do you think of this solution ?
> 
> You can also look at @noether comments in previous PR.

I would avoid the first solution. Otherwise, we would fall into having an endless list. Especially, if the new element is just an extension of another one (to do not mention extensions of extensions, mixed extensions, etc)

If it is just an "extension/modification", in my opinion, it should be "called_once" in the flight plan as something "experimental" (as we have with the GVF for tracking a circle for example). The ideal solution (if possible) would be to integrate such a modification in the original function/module as an attribute/type. If the user needs to set/adjust a variable for the new functionality, we always have the dl_settings for that, so it can be change dynamically during the flight as well.

Does this make sense to you guys?

sw/airborne/subsystems/navigation/nav_survey_rectangle.c
```diff
@@ -82,20 +83,17 @@ void nav_survey_rectangle_init(uint8_t wp1, uint8_t wp2, float grid, survey_orie
...
-  nav_survey_rectangle_dynamic(wp1, wp2, -1);
-}
-
-void nav_survey_rectangle_dynamic(uint8_t wp1, uint8_t wp2, float grid)
-{
-  if (grid > 0) {
-    if (nav_survey_shift > 0) nav_survey_shift = grid;
-    else nav_survey_shift = -grid;
+  /* Update sweep_width if "extern float sweep" has been updated */
+  if (sweep != nav_survey_shift && sweep != -nav_survey_shift)
```
noether: Correct me if I am wrong but nav_survery_shift is a float right? Then this condition will be "always" satisfied.

noether: For me, the rest of the PR looks okeish :P.

jburns11: Yes nav_survey_shift is a float, and  so is sweep. I do not think it will always be satisfied, unless I am missing something about float comparison.  I could change it to something like...
Solution 1:
    if (sweep != ABS(nav_survey_shift)) { ....
But I don't think this would change the logic!

Another solution might be to do (i.e. using a flag like you recommended)....
Solution 2:
    #ifdef SURVEY_RECTANGLE_DYNAMIC
        nav_survey_shift = (nav_survey_shift > 0 ? sweep_var : -sweep_var);
    #endif

This would just update nav_survey_shift every iteration, when SURVEY_RECTANGLE_DYNAMIC is defined in the flight_plan header section.

noether: Solution 2 is readable and very clean. There are enough resources to allow such an update in every iteration :P.

noether:
Cool,

Thanks a lot @jburns11 . If @gautierhattenberger agrees, this updated nav survey looks okei for me.

jburns11:
Great! Happy to contribute. This is my first paparazzi PR/Commit, but it's heavily involved in my  research, so hopefully you'll be seeing more of me on here!

...One small note. I change the name of the float argument to the prototype for nav_survey_rectangle_rotorcraft. This doesn't change any logic, and isn't required. But now it matches the argument in the function's definition header.

noether:
> Great! Happy to contribute. This is my first paparazzi PR/Commit, but it's heavily involved in my research, so hopefully you'll be seeing more of me on here!

Happy to hear that Paparazzi is helping you in your work :P. Do you have any links to your research/group? I am always curious about how Paparazzi is employed out there :P.

gautierhattenberger:
It's all good. Just one more little thing: can you add the setting for `sweep_var` to the FW module, similar to https://github.com/paparazzi/paparazzi/blob/master/conf/modules/nav_survey_rectangle_rotorcraft.xml#L36
You can also add your new flag to the 2 modules' documentation:
```
<define name="NAV_SURVEY_RECTANGLE_DYNAMIC" value="FALSE|TRUE" description="..."/>
```

jburns11:
> It's all good. Just one more little thing: can you add the setting for `sweep_var` to the FW module, similar to https://github.com/paparazzi/paparazzi/blob/master/conf/modules/nav_survey_rectangle_rotorcraft.xml#L36
> You can also add your new flag to the 2 modules' documentation:
> 
> ```
> <define name="NAV_SURVEY_RECTANGLE_DYNAMIC" value="FALSE|TRUE" description="..."/>
> ```

Great! If that is it, then I believe this can now be closed. Thank you guys for your time!

jburns11:
> > Great! Happy to contribute. This is my first paparazzi PR/Commit, but it's heavily involved in my research, so hopefully you'll be seeing more of me on here!
> 
> Happy to hear that Paparazzi is helping you in your work :P. Do you have any links to your research/group? I am always curious about how Paparazzi is employed out there :P.

So I am a first year PhD student, at Binghamton University in New York. We are a Programming Languages and Systems group, but the group's research into UAVs is very new (I am its first PhD student). The flight plan "language" is a really good sandbox to evaluate PL features in. For example, see this older paper, that mentions paparazzi, on Energy Types... http://www.cs.binghamton.edu/~davidl/papers/SEsCPS.pdf

My focus is on developing PL abstractions for adaptability, which will hopefully make programming complex UAV behaviors much more simple.

noether:
> > > Great! Happy to contribute. This is my first paparazzi PR/Commit, but it's heavily involved in my research, so hopefully you'll be seeing more of me on here!
> > 
> > 
> > Happy to hear that Paparazzi is helping you in your work :P. Do you have any links to your research/group? I am always curious about how Paparazzi is employed out there :P.
> 
> So I am a first year PhD student, at Binghamton University in New York. We are a Programming Languages and Systems group, but the group's research into UAVs is very new (I am its first PhD student). The flight plan "language" is a really good sandbox to evaluate PL features in. For example, see this older paper, that mentions paparazzi, on Energy Types... http://www.cs.binghamton.edu/~davidl/papers/SEsCPS.pdf
> 
> My focus is on developing PL abstractions for adaptability, which will hopefully make programming complex UAV behaviors much more simple.

Thanks for the info. Good luck with your PhD! And if you need help with Pprz, you know where we are :P.

PR #2489, opened 2020-02-12
Extending dynamic sweep width to poly_rotorcraft and poly_osam modules

jburns11:
I extended our variable sweep solution to poly_rotorcraft and poly_osam modules. I am still working on implementing it in survey_polygon, as this has a more complicated setup.


jburns11:
For some reason the travis build seems to be failing because of a redundant definition of sweep_var.

It is getting a redundant definition from nav.h and nav_survey_rectangle.h

However, when I pull my branch, the definition of sweep_var is not in nav_survey_rectangle.h so I am unsure why this is happening. It seems to have ignored my change to remove 'extern float sweep_var' from nav_survey_rectangle.h

noether:
Thanks for the PR. I will review the code. It should be okei after your last PR :P.

It seems that the survey functions should have a more deep revision. All of them could be superseded by poly_survey. As always, it is a matter of having time for that, and of course, the functions are working so they are not priority xD.

jburns11:
> Thanks for the PR. I will review the code. It should be okei after your last PR :P.
> 
> It seems that the survey functions should have a more deep revision. All of them could be superseded by poly_survey. As always, it is a matter of having time for that, and of course, the functions are working so they are not priority xD.

Hey again. I fixed the build error. It was a conflict with the master branch that I had to resolve. Now it appears to be building well with travis.

I'm open to the idea of spearheading the dev side of revising the survey functions. It is true that all you would really need is survey_poly.

In order to provide backwards compatibility with old flight plans, we could still use survey_poly with 2 
way points could use the current behavior of survey_rectangle.

noether:
It looks okei. Thanks a lot for the update.

Would you mind to update paparazzi/sw/airborne/modules/guidance/gvf/nav/nav_survey_polygon_gvf.c  nav_survey_polygon_gvf.h as well? They are only for fixed wings.

jburns11:
> It looks okei. Thanks a lot for the update.
> 
> Would you mind to update paparazzi/sw/airborne/modules/guidance/gvf/nav/nav_survey_polygon_gvf.c nav_survey_polygon_gvf.h as well? They are only for fixed wings.

Done! The only poly_survey that does not support dynamic sweep width is nav_survey_poly (used by fixed wings). This has a slightly more complicated setup than the other surveys. Some of this info must be recalculated upon updating sweep_var.

noether:
@jburns11 

Oh! Absolutely, if you do not mind, a demo is always welcome. Mostly because many of Pprz features are hidden for newcomers or first-time checkers.

This last commit/update on the gvf is only for fixed-wing aircraft. If nav_survey_poly is not supported yet, then the gvf version of it will not work either right?

jburns11:
I do not think that the gvf version is dependent on nav_poly_survey. The two are separate implementations, so it is possible to have dynamic sweep in gvf and not nav_poly_survey. Actually I found the nav_poly_survey to perform poorly in simulation, while the gvf version performed much better.

noether:
That is interesting. When I wrote the gvf version, I started from the nav_poly_survey and replaced the "guidance with the carrot" functions by the "guidance with the vector fields" functions. In principle, the gvf has better performance for tracking trajectories.

Your sweep variable should have the same impact on both poly_surveys (carrot and gvf). I am wondering why the carrot version of poly_survey needs a different implementation (than the gvf version) as you suggested in an earlier comment. Please, correct me if I am missing something here.

Thanks for the validation in simulation by the way!

This PR looks okei for me, I will approve it after your comments on poly_survey.

Thanks!

jburns11:
So I tried our solution on nav_poly_survey(). It seems that the u-turn diameter doesn't match with the distance between legs. Here is what I did...
```
bool nav_survey_polygon_run(void)
{
  #ifdef NAV_SURVEY_POLYGON_DYNAMIC
  survey.psa_sweep_width = (survey.psa_sweep_width > 0) ? sweep_var : -sweep_var;
  VECT2_SMUL(survey.sweep_vec, sweep, survey.psa_sweep_width);
  #endif
```


Its seems like nav_poly_survey_setup() has additional steps that the gvf version doesn't implement. 

I also tried adjusting "sweep_vec" as such...
```
bool nav_survey_polygon_run(void)
{
  #ifdef NAV_SURVEY_POLYGON_DYNAMIC
  survey.psa_sweep_width = (survey.psa_sweep_width > 0) ? sweep_var : -sweep_var;
  VECT2_SMUL(survey.sweep_vec, sweep, survey.psa_sweep_width);
  #endif
```
but it didn't seem to help much, but I don't know what else would need to be recomputed.

gautierhattenberger:
I guess you also need to recompute all parameters from `setup` functions that are directly or indirectly impacted by sweep width, like `seg_start` and `seg_end`. I suspect this is almost worth a dedicated function for that.

noether:
> So I tried our solution on nav_poly_survey(). It seems that the u-turn diameter doesn't match with the distance between legs. Here is what I did...
> 
> ```
> bool nav_survey_polygon_run(void)
> {
>   #ifdef NAV_SURVEY_POLYGON_DYNAMIC
>   survey.psa_sweep_width = (survey.psa_sweep_width > 0) ? sweep_var : -sweep_var;
>   VECT2_SMUL(survey.sweep_vec, sweep, survey.psa_sweep_width);
>   #endif
> ```
> 
> Its seems like nav_poly_survey_setup() has additional steps that the gvf version doesn't implement.
> 
> I also tried adjusting "sweep_vec" as such...
> 
> ```
> bool nav_survey_polygon_run(void)
> {
>   #ifdef NAV_SURVEY_POLYGON_DYNAMIC
>   survey.psa_sweep_width = (survey.psa_sweep_width > 0) ? sweep_var : -sweep_var;
>   VECT2_SMUL(survey.sweep_vec, sweep, survey.psa_sweep_width);
>   #endif
> ```
> 
> but it didn't seem to help much, but I don't know what else would need to be recomputed.

I see. Then, your extension for the gvf version works "out of the box" while the nav_poly_survey needs of further calculations. I did not remember that while I wrote the gvf version of the poly survey. Thanks for the clarification.

If you think that these extra calculations are no "trivial" (extra effort/time), and the gvf version with sweep_var for fixed-wing works already, then we can accept this PR, and leave the "carrot version" for another time.

jburns11:
I think that's a good idea. I need to try out a couple things before the poly_survey is fully functional. Once I get it working, I can create a PR for it.

If we are waiting on poly_survey, then this PR should be ready to be merged. Thanks again for the help, and I'm sure you'll here from me soon.

Best,
Jack

PR #2492, opened 2020-03-10
[fix] datalink_time is currently set to zero even for non-telemetry link

gautierhattenberger:
When using the `extra_dl` port for communication with an on-board computer board, the `datalink_time` variable which is used for datalink lost detection is set to zero for all new messages. This is including the onboard messages that are never lost.
This fix disable by default the `datalink_time` update for `extra_dl` and make it configurable for all over possible datalink transports.

### DIFF
--- a/sw/airborne/arch/sim/sim_ap.c
+++ b/sw/airborne/arch/sim/sim_ap.c
@@ -41,6 +41,9 @@ float roll_rate_pgain;
 uint16_t datalink_time = 0;
 uint16_t datalink_nb_msgs = 0;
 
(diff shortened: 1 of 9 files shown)


OpenUAS:
Clever, I've never thought of (mis) using extra_dl port for communication to on-board  computer, interesting.

Not simple for me to test two datalinks ATM no time, maybe next week with a Disco Wifi and 868modem both used where, I'll stop Wifi or the other serial modem.

But if you tested it.. lets honor the PR

gautierhattenberger:
This have been tested and validated by colleagues from Meteo-France

PR #2494, opened 2020-03-14
[datalink] parse the correct datalink buffer

gautierhattenberger:
When using a different datalink buffer than dl_buffer, the callback functions from the modules must use the buffer address passed as a parameter, otherwise there is no guarentee that the correct buffer will
be parsed.

This have been tested and validated.


gautierhattenberger:
the error reported by codacy is a known issue that will be solved in a future pull request

fvantienen:
Seems to be fine! Nice find

gautierhattenberger:
good idea, I'll consider that for a next PR

PR #2497, opened 2020-03-16
[link] Pprzlink router (for herelink)

fvantienen:
This is a simple pprzlink router for the Herelink, without any crc checking or parsing. There is no real manual yet, but only some adb commands which needs to be executed on the Air and Ground module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ paparazzi.sublime-workspace
 *.files
 *.includes
 
+# VSCode IDE
+.vscode/ipch/
+
 # Vagrant VM files
 /.vagrant
 
@@ -144,6 +147,10 @@ Cargo.lock
 /sw/logalizer/ahrs2fg
 /sw/logalizer/disp3d
 /sw/logalizer/sdlogger_download
+
+# /sw/misc
+/sw/misc/link/pprzlink_router
+/sw/misc/link/pprzlink_router_cc
 
 # /sw/simulator/
 /sw/simulator/gaia
--- a/sw/misc/link/Makefile
+++ b/sw/misc/link/Makefile
@@ -1,0 +1,13 @@
+# Quiet compilation
+Q=@
+
+all: pprzlink_router pprzlink_router_cc
+
+pprzlink_router: pprzlink_router.c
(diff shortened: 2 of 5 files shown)


gautierhattenberger:
The code itself looks fine, eventually you could format it with paparazzi rules.
The only thing is maybe the folder name `link` which is very generic. Could you find something a bit more specific, eventually `herelink` and add a small readme at least to tell what it is suppose to be.

fvantienen:
I will add some more information later. But I made it for the herelink, but can be used more general. It is just a simple forwarder between UART and UDP endpoints.

gautierhattenberger:
This is what I understood, but the name can be misleading with the actual link agent. Maybe a name with "bridge-something" ?

gautierhattenberger (dismissed):
Except renaming it to something more explicit than pprz_router, I think it is fine

OpenUAS:
@fvantienen plz, rename to something so it can be merged?... and I can test on a separate herelink set I have at hand.

gautierhattenberger:
@fvantienen can you address the naming issue before next release ? Thanks

fvantienen:
What would you prefer as name @gautierhattenberger , as I cannot think of any other correct name?

Fabien-B:
`bridge_udp_uart` ?

gautierhattenberger:
I'm also fine with bridge_udp-uart or udp-uart_bridge, as you like it

gautierhattenberger:
@fvantienen can you make the required changes ? I want to make the 5.18 release next week. Thanks

gautierhattenberger:
@fvantienen can you update or close this if not relevant anymore ?

dewagter:
Hi @gautierhattenberger and @Fabien-B,

The current naming is in line with the other names like ```/etc/mavlink-router``` on the herelink and can be also be used from ```UDP``` to ```UDP``` or any other combination. In this sense it is more a router than an uart-udp_brigde.

On top of that, this code is already widely distributed and documented on several wiki. Any possibility to keep this name?

gautierhattenberger (approved):
lets keep it like this then

PR #2499, opened 2020-03-16
[misc] Add GPSD to IVY functionalities for follow me

fvantienen:


gautierhattenberger:
For some reason that I have forgotten, the gpsd2ivy (and dependency to gpsd lib) was not mandatory on purpose. I think it was a request from @martinmm but I can't remember why. It can be because gpsd daemon can be in conflict with other things in some cases, so not everyone wants to install it.

fvantienen:
Ok I can remove the automatic build of it, or maybe add a check if it is installed and then only build it.

gautierhattenberger:
checking before compiling seems a good idea

fvantienen:
I now added a check before compilation

gautierhattenberger:
seems travis build is fine, but report to github failed: https://travis-ci.org/github/paparazzi/paparazzi/builds/663048143

PR #2500, opened 2020-03-16
[ekf2][sdlog] Add more ekf2 debugging information and magneto setting and add sdlogger filename

fvantienen:
This needs https://github.com/paparazzi/pprzlink/pull/105

### DIFF
--- a/conf/modules/ins_ekf2.xml
+++ b/conf/modules/ins_ekf2.xml
@@ -6,6 +6,14 @@
       simple INS and AHRS using EKF2 from PX4
     </description>
   </doc>
+  <settings>
+	<dl_settings NAME="INS">
+      <!-- EKF2 Configuration parameters -->
+      <dl_settings name="ekf2">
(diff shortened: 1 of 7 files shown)


gautierhattenberger:
pprzlink updated

fvantienen:
Then this should be able to be merged if approved. After that I will rebase https://github.com/paparazzi/paparazzi/pull/2502 which will then successfully compile.

fvantienen:
Wait I need to update the reference to pprzlink as well I see now. But this would require both the #2502 and this one in the same branch.

conf/modules/ins_ekf2.xml
```diff
@@ -6,6 +6,14 @@
       simple INS and AHRS using EKF2 from PX4
     </description>
   </doc>
+  <settings>
+	<dl_settings NAME="INS">
+      <!-- Optical flow calculations parameters -->
```
gautierhattenberger: wrong comment

fvantienen:
I have rebased this one and update pprzlink. But had to add also #2502 to prevent a compile error with the sdlogger filename added to the message.
Sorry for this mess...

conf/modules/ins_ekf2.xml
fvantienen: Resolved

PR #2502, opened 2020-03-16
[sdlog] Add sdlog filename and fix Flight Recorder maximum message rate

fvantienen:
Should be applied after https://github.com/paparazzi/paparazzi/pull/2500 and also after https://github.com/paparazzi/pprzlink/pull/105


fvantienen:
This is now included in #2500 because of pprzlink merging.

PR #2503, opened 2020-03-19
[gps] add support for UBX PVT message

gautierhattenberger:
PVT is a useful compact and complete message available since Ublox 8. It can eventually replace all other messages for fixedwing, and need extra lla to ecef for rotorcraft.

I don't have GPS module to test it right now, so any help here is welcome :)

### DIFF
--- a/conf/modules/gps.xml
+++ b/conf/modules/gps.xml
@@ -29,7 +29,7 @@
   </header>
   <init fun="gps_init()"/>
 
-  <makefile>
+  <makefile target="fbw|ap|sim|nps|hitl">
     <configure name="GPS_LED" default="none"/>
     <file name="gps.c" dir="subsystems"/>
     <define name="USE_GPS"/>
--- a/conf/modules/gps_ublox.xml
+++ b/conf/modules/gps_ublox.xml
@@ -43,8 +43,5 @@
       endif
     </raw>
   </makefile>
-  <makefile target="fbw">
-    <define name="USE_GPS"/>
-  </makefile>
 </module>
 
--- a/conf/ubx.xml
+++ b/conf/ubx.xml
@@ -48,6 +48,46 @@
       <field name="res2" format="U4"/>
     </message>
 
+    <message name="PVT" ID="0x07" length="92">
+      <field name="ITOW" format="U4" unit="ms"/>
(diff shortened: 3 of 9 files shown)


fvantienen:
Nice I can test later! I have also added this in another branch after some rewrite of stuff and will compare.

fvantienen:
On a similar note the log file conversion `sd2log` also creates incorrect dates somehow. Could this also be caused by the GPS rollover bug?

gautierhattenberger:
Not sure, but it could be. Did you check that you have a valid GPS message in the log and what is the value of the `week` field ?

fvantienen:
`41.6560 14 GPS 0 16602173 27 0 -17000 0 0 0 378282750 31 0`

Found the problem, I already implemented the `PVT` message but did not calculate the week, this is why it is `0`. Your implementation should fix the problem. Since `NAV-SOL` was removed there was no message receiving the `week`.

fvantienen:
This is what I had laying around: https://github.com/paparazzi/paparazzi/commit/cb196ca331b13b4369cc7a969c67ee36a72ebbad
It works and flew with it, but I think we should combine these.

gautierhattenberger:
I agree, but not sure if it is better to do it in one step or not

fvantienen:
On another note: FBW builds with GPS fail because I think `gps.c` isn't compiled with it which is now needed for the week calculation.

gautierhattenberger:
yes I just saw that
how are you suppose to use the `gps_ubx` part without the `gps` part ?

fvantienen:
The interMCU does: `AbiBindMsgGPS(IMCU_GPS_ID, &gps_ev, gps_cb);` So it will just forward whatever GPS as secondary GPS.

gautierhattenberger:
is it really a problem to compile gps.c in FBW part ? If it does, I guess we'll need to make gps_common lib for all the small helper functions

fvantienen:
I don't think it will be a problem, so you can test if it works.

gautierhattenberger:
I'm bringing some of your modifications in this pull request.
I didn't do the part about RTCM yet. @fvantienen  Do you think this is something you would need in this PR ?

fvantienen:
I don't need that for now :) Thanks a lot, this seems great!

PR #2504, opened 2020-03-19
[fix] split the sweep variable and let each module use its own

gautierhattenberger:
This is to prevent multiple declarations. It should also keep the variable sweep capabilities but with a different setting for each nav module.

this was introduced by #2488 and related PR.


fvantienen:
I just wanted to make an issue for it. But nice that you already fixed it :)

noether:
Cool, thanks a lot for the fix.

I did not have time to do it myself :/

PR #2506, opened 2020-03-23
[hardware] Add Pixhawk 4 autopilot

### LINKED ISSUE #2251: Support for Pixhawk 4
The release of Pixhawk 4 is planned for May: https://www.dronecode.org/pixhawk-4-coming-soon/

The question is, do we want Paparazzi to run on it? The hardware is similar to Chimera board (STM32F7 with an MPU unit), it has a bunch of IO, FBW co-processor and a SD card, so pretty much everything an autopilot needs.

The issues are:
- [ ] a couple of new sensors that need drivers
- [x] likely only for ChibiOS, to utilize memory protection unit and SD card logging (so we would need to update the ChibiOS version first)
- [x] porting ChibiOS to support a new board (pins, IO, clock setup etc)
- [x] figure out the AP/FBW uploader (does it work for ChibiOS?)

Since I don't have any immediate need to use Pixhawk 4, I wanted to see if other are interested in using it?

fvantienen:
This will close #2251 and needs #2505 

Not sure if everything is included in this pullrequest. Also no testing has been done with the PX4 flash modules, since I could not get it to work in vmware. Currently using the debug port with SWD to flash the autopilot and am not using the IO processor.


sw/airborne/arch/chibios/STM32F76xxI.ld
```diff
@@ -0,0 +1,132 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/*
+ * STM32F76xxI generic setup.
```
gautierhattenberger: Why is this file needed here ? It should be in `sw/ext/chibios/os/common/startup/ARMCMx/compilers/GCC/ld/` already, we are using it for Tawaki board.

fvantienen: I think it is because I shifted the flash 0x08008000 to keep the bootloader

gautierhattenberger: right, I was wondering where this is selected, but it is in chibios makefile in conf, so it's fine

PR #2507, opened 2020-03-24
[guidance] Update hybrid guidance for readability

fvantienen:
Updates to the hybrid guidance. There is a known bug with `int32_vect2_norm` which overflows in this guidance module, when distances get to big. 

This is Work In Progress and I want to fix both this bug and maybe also clean this module more with some struct.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
@@ -41,18 +41,39 @@
 #include "firmwares/rotorcraft/guidance/guidance_v.h"
 
 // max airspeed for quadshot guidance
+#ifndef MAX_AIRSPEED
 #define MAX_AIRSPEED 15
+#endif
+
 // high res frac for integration of angles
 #define INT32_ANGLE_HIGH_RES_FRAC 18
 
 // Variables used for settings
 int32_t guidance_hybrid_norm_ref_airspeed;
-float alt_pitch_gain = 0.3;
+float guidance_hybrid_norm_ref_airspeed_f;
 int32_t max_airspeed = MAX_AIRSPEED;
 int32_t wind_gain;
 int32_t horizontal_speed_gain;
 float max_turn_bank;
 float turn_bank_gain;
+
+#define AIRSPEED_HOVER          4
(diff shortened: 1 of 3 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
```diff
@@ -41,19 +41,40 @@
...
 
+#define AIRSPEED_HOVER          4
+#define AIRSPEED_FORWARD        12
+#define CRUISE_THROTTLE         4000
+#define FWD_SPEED_P_GAIN        30
+#define FWD_ALT_THRUST_GAIN     0.35
+#define FWD_PID_DIV             2
+#define FWD_NOMINAL_PITCH       78.0
+#define FWD_PITCH_GAIN          2.1
+#define HOVER_P_GAIN            12
```
gautierhattenberger: These defines should be configurable (and documented in xml file)

sw/airborne/firmwares/rotorcraft/guidance/guidance_hybrid.c
```diff
@@ -105,11 +126,12 @@ void guidance_hybrid_init(void)
...
-  horizontal_speed_gain = 8;
+  horizontal_speed_gain = 6;
   guidance_hybrid_norm_ref_airspeed = 0;
-  max_turn_bank = 40.0;
-  turn_bank_gain = 0.8;
-  wind_gain = 64;
+  guidance_hybrid_norm_ref_airspeed_f = 0;
+  max_turn_bank = 23.0;
+  turn_bank_gain = 0.5;
+  wind_gain = 35;
```
gautierhattenberger: That should not be hardcoded I think

gautierhattenberger:
Is this still relevant after #2533 ?

gautierhattenberger:
ping @fvantienen @EwoudSmeur  ?

EwoudSmeur:
I think it is still relevant. With the other controller better performance can be achieved, but it also needs more knowledge to set up and it is more prone to oscillations if the control effectiveness is wrong, if the airframe is flexible, etc.
I'll try to update this pull request in the near future, but this period is quite busy for me so no guarantees, unfortunately...

gautierhattenberger:
closed with #2964

PR #2511, opened 2020-03-24
[actuators] Add UAVCAN actuators for chibios

fvantienen:
This currently works, but I think we should split the UAVCAN from the actuators. That way we can also implement sensors and other UAVCAN equipment.


gautierhattenberger:
I totally agree that we should split this in two, and even hide the ChibiOS implementation behind a proper can peripheral interface. The current one is not generic at all at the moment.
So basically, I don't know what to do with this PR... Although it is a nice addition, It should be completely rewritten, but this may take a long time.

sw/airborne/arch/chibios/subsystems/actuators/actuators_uavcan.c
```diff
@@ -0,0 +1,334 @@
+/*
+ * Copyright (C) 2013 AggieAir, A Remote Sensing Unmanned Aerial System for Scientific Applications
+ * Utah State University, http://aggieair.usu.edu/
+ *
+ * Michal Podhradsky (michal.podhradsky@aggiemail.usu.edu)
```
podhrmic: I suspect the copyright can be updated as we were not involved into making this file:-)

fvantienen: Will change that ^^

fvantienen:
Depending on the time I have now I will decide if I want to merge this somewhat like this or in a better way.

gautierhattenberger:
@fvantienen @OpenUAS it seems there are plans to make a proper CAN + UAVCAN driver, should we keep this ad-hoc implementation ?

gautierhattenberger:
if you can resolve the conflict on TUdelft conf, I'm okay to merge this until there is a proper support of CAN and UAVCAN

gautierhattenberger (approved):
thank you for splitting uavcan driver from actuator module

gautierhattenberger:
CI is failing for unrelated reasons, so it should be fine since it doesn't really impact other things

PR #2513, opened 2020-03-28
#2384, Convert some scripts to Gtk3.0.

### LINKED ISSUE #2384: [python] Upgrade python GUIs from gtk+2 to gtk+3 to support python3
The scripts start.py, airframe_editor.py, gui_dialogs.py and iridium_link.py all use pygtk with gtk+2 which is not supported by python3. These should be updated to use python3-gobject instead with the following include:

import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk

Fabien-B:
Finally trying to fix #2384 before Ubuntu 20.04.
Need to check if it still works on previous ubuntu versions.
start.py have weird behavior when resized, but otherwise seems to work well.

### DIFF
--- a/start.py
+++ b/start.py
@@ -2,9 +2,9 @@
 
 from __future__ import print_function
 
-import pygtk
-import gtk
-pygtk.require('2.0')
+import gi
+gi.require_version('Gtk', '3.0')
+from gi.repository import Gtk, GdkPixbuf
 
 
 import os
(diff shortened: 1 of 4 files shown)


gautierhattenberger:
Using gtk3 requires to install the package `python3-gi`. I suggest to let the user install it manually for 18.04, and I will add the proper dependency for 20.04. Do you agree ?

dewagter:
Perfect. Totally agree.

PR #2514, opened 2020-03-31
Add software-I2C

tomvand:
This PR adds software-driven I2C in addition to the existing hardware I2C drivers. This allows I2C to be used on arbitrary GPIO pins at a limited bitrate (approx. 15kbps). The software I2C code is completely non-blocking and platform-independent. Software-I2C (softi2c0, softi2c1) can transparently replace the existing I2C peripherals.

I2C writing has been tested with a logic analyzer. Reading has not been tested yet because I do not have the necessary tools at home.

Since this pull requests also makes minor changes to the existing I2C code (use of function pointers to allow mixing of multiple I2C architectures), careful review and testing of existing airframes would be highly appreciated.

This PR is part of upcoming support for the Eachine Trashcan ([WIP](https://github.com/tomvand/paparazzi/tree/trashcan_spi_cc2500_pr)) and the bitcraze flowdeck.


tomvand:
Found a potential problem with the i2c_blocking_ functions, these will hang and timeout because the software i2c is not updated. Will push a fix asap.

sw/airborne/mcu_periph/i2c.h
```diff
@@ -33,6 +33,7 @@
 #include "std.h"
 
 #include "mcu_periph/i2c_arch.h"
+#include "mcu_periph/softi2c.h"
```
gautierhattenberger, line 36: is it really necessary to include this ?

sw/airborne/mcu_periph/softi2c.h
```diff
@@ -0,0 +1,52 @@
...
+ * Platform-independent software I2C implementation.
+ * Can be used transparently in place of the hardware I2C in i2c.h.
+ */
+
+#ifndef MCU_PERIPH_SOFTI2C_H
+#define MCU_PERIPH_SOFTI2C_H
+
+#include "mcu_periph/i2c.h"
+
+struct i2c_periph;
```
gautierhattenberger, line 34: isn't this already done by including `mcu_periph/i2c.h`

sw/airborne/mcu_periph/softi2c.c
```diff
@@ -0,0 +1,672 @@
...
+      } else {
+        softi2c_gpio_drive_low(d->sda_port, d->sda_pin);
+      }
+      d->bit_state++;
+      return false;
+    case 2:
+      // After SDA rise(/fall) and data set-up time
+      softi2c_gpio_highz(d->scl_port, d->scl_pin);
+      d->bit_state++;
+      __attribute__ ((fallthrough));
```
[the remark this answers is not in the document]
tomvand: I had tried a few comments, but none seemed to work. Turns out gcc is *really* specific about the comment. Fixed.

sw/airborne/mcu_periph/i2c.h
tomvand: It was there so that all modules currently including i2c.h can also use the softi2c peripherals. I have now added the required extern declarations and removed the include, which works just as well.

sw/airborne/mcu_periph/softi2c.h
tomvand: It was there because the header files included each other, which messed up the order of declarations. Fixed.

sw/airborne/arch/sim/mcu_periph/i2c_arch.c
```diff
@@ -27,16 +27,22 @@
 #include "mcu_periph/i2c.h"
 
 
-bool i2c_idle(struct i2c_periph *p __attribute__((unused))) { return true; }
-bool i2c_submit(struct i2c_periph *p __attribute__((unused)), struct i2c_transaction *t __attribute__((unused))) { return true;}
+static bool i2c_sim_idle(struct i2c_periph *p __attribute__((unused))) __attribute__((unused));
+static bool i2c_sim_submit(struct i2c_periph *p __attribute__((unused)), struct i2c_transaction *t __attribute__((unused))) __attribute__((unused));
+static void i2c_sim_setbitrate(struct i2c_periph *p __attribute__((unused)), int bitrate __attribute__((unused))) __attribute__((unused));
```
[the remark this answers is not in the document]
tomvand: gcc did not accept the unused attribute in the function definition, only in the declaration. Turns out it needs to be placed in front of the function name instead of behind it. Fixed.

gautierhattenberger:
Did you have the chance to make hardware test ? On which platforms ?

tomvand:
I have only done a writing test so far, on the Eachine Trashcan (CrazybeeF4FR, STM32F4) connected to a logic analyzer.
Right now I am working on the VL53L1X range sensor using the software-I2C, so I hope to get some hardware test results within a few days. I need to improvise a bit as I am stuck at home and do not have any other autopilots or i2c devices here.

gautierhattenberger:
Ok, I tested it on a Tawaki with F7 and the usual i2c is still working

PR #2520, opened 2020-04-08
[jsbsim] support of more recent jsbsim versions

gautierhattenberger:
The development of jsbsim is now switch to git with an new active team.
The stable release v1.0.0rc1 has several fix and improvements, but
requires some modification from us to properly handle the type of file
path passed as arguments to load models.

Also, the default path when pkg-config is not found was /opt, which was
used in very early versions of NPS. The default is now /usr so that it
is compatible with a standard install of the new lib version.

In addition the new version will be packaged for ubuntu 20.04 focal

### DIFF
--- a/conf/modules/fdm_jsbsim.xml
+++ b/conf/modules/fdm_jsbsim.xml
@@ -5,28 +5,38 @@
     <description>
       JSBSim FDM backend for NPS simulator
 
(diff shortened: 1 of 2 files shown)


alex31:
tested on kubuntu 20.04, it compiles without warning, and run flawlessly beebop2 target.

PR #2522, opened 2020-04-12
Fix distance counter to work with different number of messages

danielwillemsen:


sw/ground_segment/python/distance_counter/distance_counter.py
```diff
@@ -65,11 +65,14 @@ def message_recv(self, ac_id, msg):
             # graphical update
             wx.CallAfter(self.update)
         if msg.name == "ROTORCRAFT_STATUS":
+            cpu_time_idx = msg.fieldnames.index("cpu_time")
             self.msg_count_time = self.msg_count_time + 1
-            time_new = float(msg.get_field(12))
+            time_new = float(msg.get_field(cpu_time_idx))
```
gautierhattenberger, line 70: may I suggest a simple `float(msg['cpu_time'])`

danielwillemsen: yes, that's a better idea

PR #2523, opened 2020-04-12
WLS VLA

EwoudSmeur:
I took a moment to remove those variable length arrays (VLA) from the WLS control allocation code. This should speed up the computation time considerably.

I changed the variable length of the arrays to a length set by defines. The problem I ran into is that I am not sure exactly where to define these defines, as it should be possible to define them from the test, or from an airframe, directly or indirectly. If I include the airframe the test does not compile any more, while for the test I now put them in the makefile to make it work... I guess there must be a nice way of doing this right?

### DIFF
--- a/conf/airframes/tudelft/bebop_indi_actuators.xml
+++ b/conf/airframes/tudelft/bebop_indi_actuators.xml
@@ -5,12 +5,15 @@
   </description>
 
   <firmware name="rotorcraft">
-    <target name="ap" board="bebop"/>
+    <target name="ap" board="bebop">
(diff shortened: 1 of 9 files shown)


gautierhattenberger:
what about defining them in the makefile part of the modules that are using the WLS code ? (or makefile in the case of test code)

EwoudSmeur:
Thanks Gautier, I used the makefile section in the stabilization_indi module conf to have the defines. Also, I changed one of the inertia values in the bebop jsbsim file by a tiny amount, otherwise it would start upside down. I am not sure how or why...

PR #2528, opened 2020-05-05
adapt calibration script to python updates

EwoudSmeur:
The calibration script was interpreting the noise_threshold input as a string when running the script with python3. I also got problems with the deprecated pyplot.hold, which could be omitted without problems.

I'm not sure what version of python we are aiming for? As far as I know python2 is no longer maintained so I suppose python3?

### DIFF
--- a/sw/tools/calibration/calibrate.py
+++ b/sw/tools/calibration/calibrate.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
 #  Copyright (C) 2010 Antoine Drouin
 #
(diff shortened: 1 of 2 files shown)


EwoudSmeur:
Yes that makes sense.
But before we were requiring those same libs for python2, so why not now for python3?

gautierhattenberger:
It is just a matter of dependencies of the paparazzi-dev package. With Ubuntu 20.04, I have switch all python libs to Python3, but it is not done for previous versions.
It is not a "big" issue, it can be installed by hand if needed, and many people doing dev with python already have that probably. It is just annoying for newcomers using the one-liner install.
But I'm fine with switching to Pyhton3 all scripts that have been successfully tested.

EwoudSmeur:
Ok I understand, indeed maybe it will be a bit annoying for some. On the other hand it needs to happen some time, so maybe now is as good as ever.

PR #2529, opened 2020-05-08
PMW3901MB optic flow sensor driver

tomvand:
This PR adds a driver for the PMW3901MB optic flow sensor as found on the bitcraze flowdeck.

I only have access to a bench setup right now, if anyone could flight-test this it would be highly appreciated (@gautierhattenberger?). The sensor readings appear very noisy so I'm curious how well it works in closed-loop.


sw/airborne/modules/sensors/opticflow_pmw3901.c
```diff
@@ -0,0 +1,194 @@
...
+#define OPTICFLOW_PMW3901_AGL_ID ABI_BROADCAST
+#endif
+PRINT_CONFIG_VAR(OPTICFLOW_PMW3901_AGL_ID)
+
+#ifndef OPTICFLOW_PMW3901_AGL_TIMEOUT_US
+#define OPTICFLOW_PMW3901_AGL_TIMEOUT_US 500000
+#endif
+
+
+#define RAD(_x) ((_x) / 180.0 * M_PI)
```
gautierhattenberger, line 59: you should use `RadOfDeg` macro from `std.h`

tomvand:
Hi Gautier, I tested this on the crazybee f4fr pro (=stm32f4) of the eachine trashcan, with the spi bus connected to the pads of the desoldered max osd chip.
Can you confirm that both the VCC and VCOM pins of the flowdeck receive power? (https://wiki.bitcraze.io/_media/projects:crazyflie2:expansionboards:flow-deck-v2_rev-a_schematic.pdf). That's what I was really stuck on and produced the same symptoms, although i would assume that the crazyflie should already handle this. And to be sure, you also checked the nCS signal I assume? And does the logic analyzer show the correct values being sent during chip initialization? (To rule out polarity issues?)
And what revision of the flow deck are you using? I tested this on the v2, although the optic flow sensor and circuit should be the same on both versions...

gautierhattenberger:
I'm using the flow deck v2 and with crazyflie v2.1 (stm32f4) and the range sensor is working, so I guess the power on the board is correct. The chip select line seems to be correctly used and the transmitted values are the correct one.
Actually, it starts sending the register values after the 100 trials at reading the register 0x00 without success as nothing is ever received (always 0xFF).

flow_deck_spi_analyzer

tomvand:
The range sensor uses only one of the two power pins, so it is not necessarily true that the optic flow sensor should also work then. Apart from that I have no idea what could be going wrong right now, I can take a closer look after the weekend. I would really like to be sure that the two voltages are present and correct. If that doesn't work, maybe adding a pull-down on miso changes anything?

gautierhattenberger:
I checked the power levels and they seem to be all good for both of them. Some thing for the internal 1.8V regulator. This might also be a hardware issue on by sensor. I will try to test it with the original firmware.

tomvand:
Fixed all comments. Will check if sensor still responds on my test setup with latest commit.

gautierhattenberger:
I reflashed the original firmware, but none of my decks are properly detected. Do you have any idea how to check that with crazyflie firmware ? From the doc, it tries to read the one wire memory chip, but in my case there is no reply at all. Even the multi-ranger, which is working for sure.
I might have some hardware issues here, so if you confirm that this is working for you, I'll merge as his.

gautierhattenberger:
Here is a capture of the sensor (finally) working with original firmware
flow_deck_spi_analyzer2
As you can see, the SPI idle level and phase are not the one you have selected for Paparazzi. Apparently it is idle low and detect on rising edge (edge 1).

gautierhattenberger:
@tomvand your crazybee board, is it using ChibiOS or libopencm3 ? I think the problem comes from the spi driver of ChibiOS that is not implementing (yet) the select/unselect option.

gautierhattenberger:
@tomvand check the `pmw3901mb-pr` in the paparazzi repo, I made some changes to the spi driver of chibios and some settings of your driver. It is now working on crazyflie

gautierhattenberger:
@tomvand did you get a chance to test with my modifications (https://github.com/paparazzi/paparazzi/tree/pmw3901mb-pr) ? I would like to merge this before the next release (soon)
I did a short flight with a crazyflie and it was working nicely.

tomvand:
Will check and review changes this morning

tomvand:
(*bump* because I'm not sure if edited comments show up as new)

gautierhattenberger:
merged with https://github.com/paparazzi/paparazzi/commit/022e6abed977578bbdae528e44c78695f0d32d2b
thanks

Issue #2531, opened 2020-05-24
Adding fabs and explicit type conversion


joezie:
I found two bugs in navigation-related source codes. Details are listed below:

* nav_smooth.c (the earliest version with the bug: tree e2ccf1b6a317d1fb1a9b3c6d1f5376b5f94ab912 in release v5.17_devel)
https://github.com/paparazzi/paparazzi/blob/e2ccf1b6a317d1fb1a9b3c6d1f5376b5f94ab912/sw/airborne/modules/nav/nav_smooth.c#L157-L170
_delta_ is impacted by the values of _wind_east_, _wind_north_, and _airspeed_. In some cases, the computed _delta_ can be a negative value, and then we perform _sqrt()_ on it, which result to NaN error. To reproduce this error,  you can manually set _stateGetHorizontalWindspeed_f()->y_ and _stateGetHorizontalWindspeed_f()->x_ to -10.0 and 10.0 respectively at the following function call and run flight simulation.
https://github.com/paparazzi/paparazzi/blob/e2ccf1b6a317d1fb1a9b3c6d1f5376b5f94ab912/sw/airborne/modules/nav/nav_smooth.c#L189
To fix this bug, a possible solution is to perform _fabs()_ on _delta_ before performing _sqrt()_. 
`ground_speeds[i] = scal + sqrt(fabs(delta))/2.; `
But please note that the computed result might not be the ideal value.

* navigation.c (the earliest version with the bug: tree dd0c6cc1faa93b4132627334650b3243102b7073 in release v5.17_devel)
https://github.com/paparazzi/paparazzi/blob/dd0c6cc1faa93b4132627334650b3243102b7073/sw/airborne/firmwares/rotorcraft/navigation.c#L274
Here, _wp_diff_ and _pos_diff_ are two-dimensional variables representing the horizontal distance
between two waypoints _wp_start_ and _wp_end_ and between the UAV and _wp_start_ respectively.
The types of their members x and y are both signed integers. However, since _nav_leg_length_
is an unsigned integer, the result computed within the parentheses must be implicitly converted
to unsigned integer before divided by _nav_leg_length_. When the UAV is approaching the waypoint
_wp_start_, if the result computed within the parentheses happens to be a negative value whose most significant bit is set to 1, it would be interpreted as a large value after conversion to unsigned
integer. After being divided by _nav_leg_length_, the most significant bit becomes 0, and therefore,
[...]

PR #2532, opened 2020-05-24
Navigation: fix calculations of ground speeds and navigation progress

joezie:
- add fabs and explicit type conversion respectively to fix calculation errors

Fixes issue #2531

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/navigation.c
+++ b/sw/airborne/firmwares/rotorcraft/navigation.c
@@ -592,7 +592,7 @@ void nav_route(struct EnuCoor_i *wp_start, struct EnuCoor_i *wp_end)
   INT32_VECT2_RSHIFT(pos_diff, pos_diff, INT32_POS_FRAC);
   uint32_t leg_length2 = Max((wp_diff.x * wp_diff.x + wp_diff.y * wp_diff.y), 1);
   nav_leg_length = int32_sqrt(leg_length2);
-  nav_leg_progress = (pos_diff.x * wp_diff.x + pos_diff.y * wp_diff.y) / nav_leg_length;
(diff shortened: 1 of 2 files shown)


sw/airborne/modules/nav/nav_smooth.c
```diff
@@ -170,7 +170,7 @@ static void compute_ground_speed(float airspeed,
     /* g^2 -2 scal g + c = 0 */
     float scal = wind_east * cos(alpha) + wind_north * sin(alpha);
     float delta = 4 * (scal * scal - c);
-    ground_speeds[i] = scal + sqrt(delta) / 2.;
+    ground_speeds[i] = scal + sqrt(fabs(delta)) / 2.;
```
gautierhattenberger, line 173: This is not the proper way I think. The navigation routine should just fail if delta is negative.

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -592,7 +592,7 @@ void nav_route(struct EnuCoor_i *wp_start, struct EnuCoor_i *wp_end)
   INT32_VECT2_RSHIFT(pos_diff, pos_diff, INT32_POS_FRAC);
   uint32_t leg_length2 = Max((wp_diff.x * wp_diff.x + wp_diff.y * wp_diff.y), 1);
   nav_leg_length = int32_sqrt(leg_length2);
-  nav_leg_progress = (pos_diff.x * wp_diff.x + pos_diff.y * wp_diff.y) / nav_leg_length;
+  nav_leg_progress = (pos_diff.x * wp_diff.x + pos_diff.y * wp_diff.y) / (int32_t)nav_leg_length;
```
gautierhattenberger, line 595: good catch

sw/airborne/modules/nav/nav_smooth.c
joezie: In the latest commit, I changed to return false on calculation failure.

PR #2533, opened 2020-05-26
Cyclone guidance code

EwoudSmeur:
This is the code that we were using for the Cyclone, but that I still wanted to clean up before opening the merge request. Now, to get the ball rolling, I'm opening the merge request, even though still quite a few things need to be changed. I will put effort in making it more separate, such that it does not affect or influence currently working code, and making it more flexible.


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.h
```diff
@@ -0,0 +1,62 @@
...
+
+extern float guidance_indi_pos_gain;
+extern float guidance_indi_speed_gain;
+extern float guidance_indi_pos_gainz;
+extern float guidance_indi_speed_gainz;
+
+extern struct FloatVect2 desired_airspeed;
+extern int16_t update_hp_freq_and_reset;
+extern struct FourthOrderHighPass flap_accel_hp;
+extern float guidance_indi_max_airspeed;
```
gautierhattenberger, line 60: I think it is usually cleaner to use a structure for all the variables

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -76,6 +76,27 @@
...
+  float norm = FLOAT_VECT2_NORM(*vect3);
+  if(norm>1.0) {
+    float scale = max/norm;
+    vect3->x *= scale;
+    vect3->y *= scale;
+  }
+}
+
+struct FloatVect2 line_vect, to_end_vect;
+
```
gautierhattenberger, line 99: move to math lib ?

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -118,6 +139,14 @@ int32_t nav_circle_radius, nav_circle_qdr, nav_circle_radians;
 /* nav_route variables */
 struct EnuCoor_i nav_segment_start, nav_segment_end;
 
+#ifdef GUIDANCE_INDI_MAX_AIRSPEED
+#define NAV_MAX_SPEED (GUIDANCE_INDI_MAX_AIRSPEED + 10.0)
+float nav_max_speed = NAV_MAX_SPEED;
+#endif
```
gautierhattenberger, line 145: I don't really like to have ref to GUIDANCE in the nav part

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -298,6 +330,162 @@ bool nav_approaching_from(struct EnuCoor_i *wp, struct EnuCoor_i *from, int16_t
   return false;
 }
 
+/**
+ * @brief function that returns a speed setpoint based on flight plan.
```
gautierhattenberger, line 334: I don't like the way the navigation code is structured in general. So I'm not sure where to put to keep the code understandable, as it seems to be an alternate nav routine, right ?

sw/airborne/modules/air_data/air_data.c
```diff
@@ -173,6 +173,9 @@ static void send_baro_raw(struct transport_tx *trans, struct link_device *dev)
 
 static void send_air_data(struct transport_tx *trans, struct link_device *dev)
 {
+#ifdef SITL
+  air_data.airspeed = stateGetAirspeed_f();
```
gautierhattenberger, line 198: you should use `USE_NPS_AIRSPEED` in your airframe file instead

sw/airborne/modules/ctrl/ctrl_effectiveness_scheduling.c
```diff
@@ -73,7 +100,57 @@ void ctrl_eff_scheduling_periodic(void)
...
 }
+
+void ctrl_eff_scheduling_periodic_b(void)
+{
+#ifdef SITL
+  struct NedCoor_f *speed_ned = stateGetSpeedNed_f();
+  float airspeed = sqrt(speed_ned->x*speed_ned->x+speed_ned->y*speed_ned->y);
+#else
+  float airspeed = stateGetAirspeed_f();
+#endif
```
gautierhattenberger, line 115: if you enable airspeed in NPS, maybe this is not needed anymore

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.h
```diff
@@ -33,7 +33,7 @@
 #include "math/pprz_algebra_float.h"
 
 extern void guidance_indi_enter(void);
-extern void guidance_indi_run(float heading_sp);
+extern void guidance_indi_run(float *heading_sp);
```
gautierhattenberger, line 36: what is the point of making it a pointer since the value is not changed in the function ?

EwoudSmeur:
Thanks @gautierhattenberger for your comments! I will update accordingly, and indeed I agree those functions should not be in the navigation, but perhaps have a separate module or even reuse the gvf.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.h
EwoudSmeur: indeed in guidance_indi.c the value is not changed, but when using guidance_indi_hybrid.c it is. I am now thinking that the coupling between guidance and control should even be tighter, feeding the required rates for some manoeuvre (think of a coordinated turn) to the controller (maybe it should be considered one big controller?)

gautierhattenberger:
If you can move it to a module and use "custom" segment and circle functions instead of the built-in nav, it is fine of course.
For the rover navigation, the nav layer is an empty shell that point to the navigation function of your choice. I guess we should have that for all firmware.

noether:
> Thanks @gautierhattenberger for your comments! I will update accordingly, and indeed I agree those functions should not be in the navigation, but perhaps have a separate module or even reuse the gvf.

Hi Ewoud,

I am writing now code for the 3D GVF (fixed-wing). We have a deadline by the end of June, so hopefully, it will be ready before that.

The code will come together with an extension for rotorcraft (2D and 3D), that requires the INDI since it passes the desired accelerations to it (as we were doing together last time).

Let me know whether you need any "outputs", or to consider any other model for the vehicle for the GVF. Now it is the moment for that xD.

gautierhattenberger:
@EwoudSmeur @noether In the future, I would like to have each control layer as independent as possible. Meaning that they have their own inputs and outputs variables, not connected to another layer, and certainly not setting directly the underlying control loops inputs as it is currently done in guidance setting the attitude setpoint directly.
I'd rather let the AP mode state machine do the proper connections (aka "in the darkness bind them").
It would be nice to keep that in mind for future design.

noether:
> @EwoudSmeur @noether In the future, I would like to have each control layer as independent as possible. Meaning that they have their own inputs and outputs variables, not connected to another layer, and certainly not setting directly the underlying control loops inputs as it is currently done in guidance setting the attitude setpoint directly.
> I'd rather let the AP mode state machine do the proper connections (aka "in the darkness bind them").
> It would be nice to keep that in mind for future design.

Sorry for kidnapping the thread :P. We can continue the discussion of the following later in the chat or mail:

Sure, that should not be a problem for the GVF. Now the 2D GVF can "spit" the "desired heading rate". I have a student now that once the quarantine is over, he will setup a rover with Pprz. I will contact you by then to see how to integrate the GVF following such architecture, so I learn what you have in mind and eventually the idea would be to port it to fixed-wing (right?). 

The 3D GVF can spit both "heading rate + vertical speed" (fixed-wing) or "3D accelerations" (rotorcrafts). In fact, the latter sends an ABI msg to INDI (smth close to what you are suggesting).

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.h
gautierhattenberger: The coupling (or not) is highly dependent of the control stack. So having each layer properly separated allows a better and more flexible assembly in the end.
In your case, if you think that you need a dedicated control layer integrated to the guidance, I'm perfectly fine. If you need to only add a few options to the current stab or the ref generator, that could be done as well. I just want to avoid functions like the "set_guidance_from_nav" which is making tight coupling between nav, guidance and control from the implementation point of view, while they the design is decoupled from the control point of view.

EwoudSmeur: Thanks a lot for your insight. If I understand you correctly, you're saying that we can make modular layers for control, guidance and navigation, and big controllers that may do everything in one go. But for the modular code, it should not be interdependent. Is that a good summary? I would fully agree with that.

sw/airborne/modules/air_data/air_data.c
EwoudSmeur: This define is only used in fixedwing, but by adding the AbiSendMsgAIRSPEED line to the rotorcraft code in case of USE_AIRSPEED, all these ifs are no longer needed

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.h
gautierhattenberger: Yes, that's the idea.
It will make it much easier to use the "generated" autopilot stack (see http://wiki.paparazziuav.org/wiki/Autopilot_generation).

conf/Makefile.chibios
```diff
@@ -119,7 +119,7 @@ endif
 
 # C++ specific options here (added to USE_OPT).
 ifeq ($(USE_CPPOPT),)
-  USE_CPPOPT = -fno-rtti
+  USE_CPPOPT = -fno-rtti -std=c++11
```
gautierhattenberger, line 122: see #2539

gautierhattenberger:
@EwoudSmeur can you rebase this latest master ? (mostly for the chibios makefile). Then I think it can be merged

PR #2536, opened 2020-05-27
fixed ref reset on mode switch heading issue

EwoudSmeur:
Upon some mode switches, the reference attitude is reset to the current attitude. A quaternion is made from the current attitude in euler angles, but for this a 'pitch-corrected' heading is used (a heading that points in the same direction if you pass 90 deg pitch. This is a problem, because if interpreted as a normal ZYX euler angle set the reference attitude is incorrectly set. The result is a 180 yaw movement when switching between FWD and ATT with pitch < -90 deg. The real unedited ZYX euler angles should be used instead.

Actually, looking at it again, should the ref psi be reset at all? Ref phi and theta are also not reset...

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_float.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_float.c
@@ -141,6 +141,22 @@ static void send_att_ref(struct transport_tx *trans, struct link_device *dev)
                                         &att_ref_quat_f.accel.q,
                                         &att_ref_quat_f.accel.r);
 }
+
+static void send_ahrs_ref_quat(struct transport_tx *trans, struct link_device *dev)
+{
+  struct Int32Quat *quat = stateGetNedToBodyQuat_i();
+  struct Int32Quat refquat;
+  QUAT_BFP_OF_REAL(refquat, att_ref_quat_f.quat);
+  pprz_msg_send_AHRS_REF_QUAT(trans, dev, AC_ID,
+                              &refquat.qi,
+                              &refquat.qx,
+                              &refquat.qy,
+                              &refquat.qz,
+                              &(quat->qi),
+                              &(quat->qx),
+                              &(quat->qy),
+                              &(quat->qz));
+}
 #endif
 
 void stabilization_attitude_init(void)
@@ -173,6 +189,7 @@ void stabilization_attitude_init(void)
 #if PERIODIC_TELEMETRY
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_STAB_ATTITUDE_FLOAT, send_att);
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_STAB_ATTITUDE_REF_FLOAT, send_att_ref);
+  register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_AHRS_REF_QUAT, send_ahrs_ref_quat);
 #endif
 }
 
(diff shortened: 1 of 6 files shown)


EwoudSmeur:
I guess we should fix this for the other controllers as well by the way.

gautierhattenberger:
I guess you might want to properly set the ref when entering a mode that is using it. But I this might generate discontinuities in some cases.
Actually, I'm wondering would would result in just applying the current quaternion attitude as the current ref, instead of doing an half euler trick ? I don't see why not starting the ref from your current full attitude. And you shouldn't have the psi error in this case. What do you think ?

EwoudSmeur:
That would probably be the best thing to do.
I can imagine if in one mode the controller has some error with the reference, once the ref is reset to the current attitude the control effort would reduce. But this issue would be a very tiny one compared to (with the current code) switching from ATT (with some ref phi and theta) to RATE and then back to ATT: The ref will be at the old phi and theta with some new psi, if I read the code correctly... I would prefer just resetting to the current attitude. Do you think this is correct?

dewagter:
- I would highly agree to avoid any Euler conversion here: quaternion/rmat only
 - When changing mode, IMHO it's OK to ``reduce the control action''.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_ref_quat_float.h
```diff
@@ -59,7 +59,7 @@ struct AttRefQuatFloat {
 
 
 extern void attitude_ref_quat_float_init(struct AttRefQuatFloat *ref);
-extern void attitude_ref_quat_float_enter(struct AttRefQuatFloat *ref, float psi);
+extern void attitude_ref_quat_float_enter(struct AttRefQuatFloat *ref, struct FloatEulers *state_euler);
```
gautierhattenberger, line 62: Why passing an euler format here, rather than a quat ?
With the current implementation, it means that you'll still have to make the conversion at some point, but if we replace it by a ref really based on quaternions, this would make a transformation that is not needed.

EwoudSmeur: Ah I figured that since it is there in the struct, it would be best to compute both, but it is also already computed with every update. Does it still make sense to keep the euler angles in the reference struct though?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_float.c
```diff
@@ -192,7 +192,9 @@ void stabilization_attitude_enter(void)
   /* reset psi setpoint to current psi angle */
   stab_att_sp_euler.psi = stabilization_attitude_get_heading_f();
 
-  attitude_ref_quat_float_enter(&att_ref_quat_f, stab_att_sp_euler.psi);
+  struct FloatQuat *state_quat = stateGetNedToBodyQuat_f();
+
+  attitude_ref_quat_float_enter(&att_ref_quat_f, state_euler);
```
gautierhattenberger, line 197: it should be `state_quat` as second argument

EwoudSmeur: Ah of course, no as you can see not tested yet :sweat_smile: 
I can test everything in simulation, will do that right away.

EwoudSmeur:
I think this is ready to be merged, or are there any other points?

gautierhattenberger:
I wanted to test it, but couldn't find the time :(

PR #2539, opened 2020-06-19
Compile chibiOS with cpp 11

EwoudSmeur:
I get errors if I try to compile cpp code (EKF2) with chibios. Using cpp 11 solves it.

### DIFF
--- a/conf/Makefile.chibios
+++ b/conf/Makefile.chibios
@@ -79,6 +79,7 @@ endif
 #
 $(TARGET).srcs += c++.cpp pprz_syscalls.c
 $(TARGET).CXXFLAGS += -fno-sized-deallocation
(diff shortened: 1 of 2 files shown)


EwoudSmeur:
Thanks for the suggestion! I tried to change it to:
`CXXSTANDARD` ?= -std=c++11
But this does not seem to have any effect...

gautierhattenberger:
Then you have to include this flag somewhere, I don't think it is done already

EwoudSmeur:
I added the CXXSTANDARD in the EKF2 module, since that is what needs it. Somehow adding a default does not seem to be necessary, but I can still do this if needed.

gautierhattenberger:
let's start with this, we'll see if it is useful to make it a default later

PR #2540, opened 2020-06-30
Eachine Trashcan support

tomvand:
This PR adds support for the Eachine Trashcan with the following additions:
- **FrskyX SPI drivers**. Ported from betaflight. I copied the code with the least amount of editing possible to simplify later updates. RC channels work as normal, and both up- and downlink are supported in the form of a generic paparazzi serial device. I have included a module and an opentx lua script to tunnel pprzlink (up+down) over frsky to a pc. FrskyX telemetry is also available in limited form.

  I have left the betaflight headers/license (GPLv3) in place to give proper credit, and added the paparazzi license where substantial code was added. Let me know if this is the right way to handle this.

- **DFU command module**. Reset the STM32F4 to DFU using a command over the USB serial line. The command is compatible with betaflight, so drones with the betaflight firmware can be flashed directly out-of-the-box. (Note that the linux serial driver may need a while to connect, and that the trashcan sometimes needs batteries plugged in due to the FPV camera current draw). The command is automatically sent when pressing upload if the module is added to the airframe.

  The STM DFU bootloader is very particular about the processor state at boot, and support from OpenCM is pretty limited, so the code may look slightly hackish. Feedback welcome.

- **Trashcan board files and airframe**. Original files provided by @OpenUAS, thanks! Airframe tuning needs to follow in a later PR but the drone is flyable right now.

This PR has been lying on the shelf for a bit too long because corona made it difficult to test-fly, so I'd rather get it merged now than wait while it gets further out of sync with master. The only things I have not finished yet is the airframe, this can be tuned in a later PR. The drone flies ok right now, and I have already used this code during the development of the flowdeck drivers.


tomvand:
Pprzlink pull request: https://github.com/paparazzi/pprzlink/pull/118

sw/airborne/subsystems/radio_control/cc2500_frsky/cc2500_paparazzi.c
```diff
@@ -0,0 +1,76 @@
...
+
+static uint16_t frsky_raw[RADIO_CTL_NB];
+
+
+#if PERIODIC_TELEMETRY
+#include "subsystems/datalink/telemetry.h"
+
+static void send_cc2500(struct transport_tx *trans, struct link_device *dev)
+{
+  pprz_msg_send_CC2500(trans, dev, AC_ID,
```
gautierhattenberger, line 42: what about reusing a message already existing ? https://github.com/paparazzi/pprzlink/pull/118#issuecomment-651780344

OpenUAS:
Fantastic progress and PR :1st_place_medal: will test together with Nilay if TU situation permits.

sw/airborne/modules/dfu_command/dfu_command.c
```diff
@@ -0,0 +1,54 @@
...
+ */
+/**
+ * @file "modules/dfu_command/dfu_command.c"
+ * @author Tom van Dijk
+ * Read USB serial for dfu command
+ */
+
+#include "modules/dfu_command/dfu_command.h"
+
+#include "arch/stm32/mcu_arch.h"
```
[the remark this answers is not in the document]
tomvand: Agreed, but with the note that booting to DFU is only implemented for the STM32 right now.

sw/airborne/subsystems/radio_control/cc2500_frsky/cc2500_paparazzi.c
tomvand: Agreed, will swap it for the PPM message (provided that this does not cause conflicts with other modules/subsystems).

tomvand:
> In general, their is a lot of commented (and maybe unrelated) code coming with this lib. I guess it can't be helped, but it would be good to clean this at some point.

I considered it but think it will be better to leave the commented code in place, as it makes it very clear what parts of the betaflight code are actually being used and need to be supported with compatibility code. I think this will make potential upgrades a lot easier.

tomvand:
All comments fixed

conf/tools/blacklisted
```diff
@@ -0,0 +1,9 @@
+#Blacklisted tools
+#Add the name of the tool as defined in its xml file to hide it in the "Tools" menu in the paparazi center.
+#Lines starting with # are comments and will be ignored.
+#
+#Examples: (Uncomment to hide them in the paparazzi center)
+#BluegigaUsbDongleScanner
+#BluegigaUsbDongle
+#IridiumDialer
+
```
gautierhattenberger, line 9: This file should not be under revision control, only `blacklisted_example`. Can you remove it, and even better add it to `.gitignore` (should have been done before).

tomvand: I would be very happy to add it to .gitignore, I commit this file way too often...

OpenUAS:
:purple_heart:

PR #2545, opened 2020-06-30
Aircraft generator

gautierhattenberger:
This is a major rewrite of the build system and code generation of Paparazzi. Why ?

Because the current one looks like this: http://wiki.paparazziuav.org/wiki/DevGuide/AircraftBuildProcess. Basically a program is called from `make`, calling more independent generator using `make` from syscalls, getting the result from stdout, or temp files, or both. And the XML configuration files are parsed several times by different tools in this process.

The idea here, is that everything related to code generation is handled by a single program. All necessary config files are passed as input, parsed all together (with module dependency resolution someday...), then code is generated if needed. The parsing library can also be reuse by other tools such as the server or the paparazzicenter.

So what's next ?
I don't expect this code to be actually reviewed because:
- it is in ocaml
- it is in ocaml
- it is way too big

So if you want to go with this, the only way is to actually use it, and fix it if needed. It is passing the standard tests, but there are maybe some corner cases.
It should be functionally the same and should work in most cases. However, here are some changes:
- no more `submodule` allowed in airframe files
- `modules` node is still supported, but I would like to get ride of it
- some minor changes (old way still supported): `module` -> `header` in settings XML file, `ctl` not mandatory in radio files, `reverse` allowed in radio instead of swapping min/max values, ...

A nice knew addition (to be tested a bit more): you can define a PAPARAZZI_MODULES_PATH environment variable to set a list of external places where to find paparazzi modules (XML and source code).

All comments are welcome.


gautierhattenberger:
I was planing to switch to version 6 at the point were I can no longer maintain backward compatibility. Which is getting closer...
There are a few major changes, that will eventually to a new version for Paparazzi, in particular (in increasing severity):
- moving and sorting all subsystems code to modules
- a proper module dependency mechanism
- a new set of default messages

If we can get close to that, then I totally agree to dump major version and drop a few things. What do you think @dewagter  ?

dewagter:
Tested several airframes and it seems to work nicely. I like the module overview in the beginning of compiling. I added fix for an existing typo that the new compiler found in a radio file.

I was wondering if this pull-request would make our long-desired update possible: splitting airframe files: e.g. have a single BEBOP airframe with 5 different PAYLOAD files that load different modules: e.g. optic-flow and GPS bebops are 95% identical. Now there are over 10 (outdated) copies with only a minor change. 


If this is still backward compatible and does not break pending pull requests, it could be merged under version 5. I was just saying that it sounds like we are pretty close to a v6 since compiling is totally rewritten. Add the calibration files in the conf, the fixedwings+hybrid flying with rotorcraft code and drop code for hardware that can not be bought anymore (after a mailing list poll) together with the 3 points you mention and you have probably lost your backward compatibility while cleaning up the project.

gautierhattenberger:
It is not yet possible to have splited airframe, but it would make it much easier. Something like the modules in the paparazzi center where airframe is a list of files instead of a single one.
I can think about it for the next round of modifications. I want to be sure that this part is completely operational first.

dewagter:
This part worked for all airframes I tested. It also helped to find an existing bug by having better checks on the xml/modules. For me, it can be reintegrated.

PR #2547, opened 2020-07-06
[conf] cleaning some airframe files for CI builds

gautierhattenberger:
mostly by moving modules from 'modules' to 'firmware' section, which
should be done in all airframe files now.


gautierhattenberger:
Also, don't hesitate to add your own cleaning to this PR or a new one.

PR #2548, opened 2020-07-07
Added try...except to be more safe after pprzlink api changes. 

MJafarMashhadi:
xref paparazzi/pprzlink#117, [this comment](https://github.com/paparazzi/pprzlink/pull/117#issuecomment-653585123)

### DIFF
--- a/sw/ground_segment/python/guided_mode_example.py
+++ b/sw/ground_segment/python/guided_mode_example.py
@@ -166,10 +166,7 @@ class IvyRequester(object):
             self.ac_list = [int(a) for a in msg['ac_list'].split(',') if a]
             print("aircrafts: {}".format(self.ac_list))
 
-        self._interface.subscribe(aircrafts_cb, "(.*AIRCRAFTS .*)")
-        sender = 'get_aircrafts'
-        request_id = '42_1' # fake request id, should be PID_index
-        self._interface.send("{} {} AIRCRAFTS_REQ".format(sender, request_id))
+        self._interface.send_request('ground', "AIRCRAFTS", aircrafts_cb)
         # hack: sleep briefly to wait for answer
         sleep(0.1)
         return self.ac_list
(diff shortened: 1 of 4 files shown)


gautierhattenberger:
I think that if we do that, pprzlink should be updated as well, otherwise no exception will be raised

MJafarMashhadi:
Thanks!
@dewagter The messages inside the exceptions have rather good explanations (See https://github.com/paparazzi/pprzlink/blob/2e0ca565a8df9f2dc037148a837049cf6123dc60/lib/v1.0/python/pprzlink/ivy.py#L175 for example) so I didn't bother with making the error message here wordier, printing the exception would do imho. But I can add a `traceback.print_exc()` to print the stack trace for more debug info, what do you think?

@gautierhattenberger You mean updating the git submodule to point to the latest pprzlink commit on master? If that's the case let's wait for paparazzi/pprzlink#122 to finalize and then I'll update this PR.

dewagter:
@MJafarMashhadi Thanks, looks good to me.

sw/ground_segment/python/guided_mode_example.py
```diff
@@ -161,14 +161,26 @@ def shutdown(self):
...
 
         def aircrafts_cb(ac_id, msg):
+            global new_answer
             self.ac_list = [int(a) for a in msg['ac_list'].split(',') if a]
             print("aircrafts: {}".format(self.ac_list))
+            new_answer = True
 
         self._interface.send_request('ground', "AIRCRAFTS", aircrafts_cb)
         # hack: sleep briefly to wait for answer
-        sleep(0.1)
```
MJafarMashhadi, line 174: I can remove this commit if you think it's unnecessary or unrelated. I was updating the `send_request` part with the new method and noticed this can be changed a bit.

gautierhattenberger: This example should probably use the `pprz_connect` tool, but I guess this is also nice to have an example using the request API directly. So I think we can keep it

MJafarMashhadi:
Thanks!

Issue #2546, opened 2020-07-02
DShot jitters


EwoudSmeur:
@fvantienen helped me to set up dshot on the Chimera autopilot (STM F7) with ChibiOS, but I'm having some issues. I can start and stop the motors, they spin and react to throttle changes. However, they jitter when changing the command. Sometimes I can find a stick position in which they stay at a constant RPM.

I have recorded the messages and they seem correct, at least most of the time. However, I just saw that sometimes there can be an extra bit, so 17 instead of 16. I can imagine that this could cause jitter. Is this a known issue @gautierhattenberger ? I recorded some data with a logic analyzer which I can share if needed. Thanks for any help!

EwoudSmeur:
![image](https://user-images.githubusercontent.com/2464596/86414340-e8498b80-bcc3-11ea-8bf6-b334ea28945c.png)

The first 11 bits of the sent messages

gautierhattenberger:
I will forawrd the issue to my colleague who wrote the low level driver for this. If you have the recording of the logic analyzer that would help a lot.

EwoudSmeur:
Thanks Gautier!
This is the logic analyzer data and a matlab script for making the figure above:
https://filesender.surf.nl/?s=download&token=87cdfc90-049e-4494-863c-87e36948a2a0

This is how often 17 bits (34/2) are sent instead of 16:
![image](https://user-images.githubusercontent.com/2464596/86445777-46518f80-bd13-11ea-97aa-982945eb128c.png)

gautierhattenberger:
Which timer are you using for the DSHOT ? If possible, link to the board file where this is configured. Thanks

EwoudSmeur:
Thanks for your response, we were using timer 3:
https://github.com/EwoudSmeur/paparazzi/commit/d3782b2a4f78a6fd0ba73bd716e8ab8491e277fb

alex31:
I am a bit away from Logic Analyser and i cannot verify that the candidate that i have found is actually the problem you have met, but i have a candidate : the dma buffer is not declared in a static memory area that is non cached, so the driver will work on F4 but will meet cache coherency problem on F7.

could you try this : in file
sw/airborne/arch/chibios/modules/actuators/actuators_dshot_arch.c.
replace all 
DSHOTDriver DSHOTDx;
by
IN_DMA_SECTION_NOINIT(DSHOTDriver DSHOTDx);

[...]

PR #2552, opened 2020-07-08
Fix dshot and dma

alex31:
fix for issue : 
DShot jitters #2546

fix dma buffer coherency by placing it in non cachable memory section, and by declaring whole array volatile.


gautierhattenberger:
@EwoudSmeur could you test this fix ?

EwoudSmeur:
Thanks for yet another update! I tested it, but unfortunately still no luck... I do get a warning in the function dma_lld_start_transfert, because in the struct definition mem0p is not declared volaitle. @fvantienen was wondering why IN_DMA_SECTION_NOINIT is used instead of IN_DMA_SECTION? In any case I tried it but no difference.

alex31:
>  I do get a warning in the function dma_lld_start_transfert, because in the struct definition mem0p is not declared volaitle. 

Oh, of course volatile is needed at the struct definition in the header. In fact, I debug out of paparazzi, and then manually revert the fixes in paparazzi after, and i missed that warning sorry, I have commited this fix, but i just tried in my testbench to remove volatile in the definition, and i still don't have errors, so even if the fix has to be applied, it probably won't fix your issue.

> @fvantienen was wondering why IN_DMA_SECTION_NOINIT is used instead of IN_DMA_SECTION? In any case I tried it but no difference.

the struct is  initialized by the code, so there is no need to init this buffer with initial value before one enters in main.

To debug this issue, I have configured a timer in input capture to decode the dshot protocol. I send dshot commands at 1000 hz, and each time verify that the command I receive is the one I generate. Within my testbed, I had the same rate of error than you, and now with the fixes, I have no more error, but my test do nothing else than generating and receiving dshot command. If the problem is due to a bandwidth problem on the AHB bus, it will be dependent whether you use dma for other stuff : I²C, SD log, SPI, etc etc, so could you tell me what peripheral you use ?

do you compile with standard pprz level of optimisation (-O2 I think) ?

One think you can try, to release the stress on the buses, is to reduce the dshot speed to 300 or even 150 if you esc is ok with that : it's hardcoded in sw/airborne/arch/chibios/modules/actuators/esc_dshot.c:48:#define DSHOT_SPEED 600

thanks for helping !

EwoudSmeur:
Thanks a lot for your support @alex31! How strange that the fixes solve the issue for your setup but not for mine...

For the compiler optimization, I didn't specify anything, so I assume it is the default. Regarding the peripherals, I am surely using a couple of them: GPS on UART, Magnetometer on I2C1, airspeed on I2C2, IMU on SPI(?), SD logger, telemetry on UART... I think they all use DMA but not 100% sure.

I will try out the lower speed!

alex31:
here a trace for a loop that send shot value from 50 to 2000 incrementing each time, if you want to analyze it with your matlab script, the more the eyes ...
[capture.zip](https://github.com/paparazzi/paparazzi/files/4898019/capture.zip)

EwoudSmeur:
That looks fine! So you think somehow the other processes may be interfering with the DShot driver?

gautierhattenberger:
Now that you mention the speed, I remember that I had issues with some ESC with the DSHOT600 so I used DSHOT300 and it was fine. I though it was an ESC problem, but maybe not only...

alex31:
> So you think somehow the other processes may be interfering with the DShot driver?

when a peripheral is driven by dma, the peripheral ask for new value when it need it. In case of peripheral like uart, i2c, spi, etc, the peripheral gently wait for the data to arrive, if there is some lag, that will slow down a bit the transmission, but that will not break anything because, the serialization is done inside the peripheral. 

In our case, it's the dshot driver which does the serialization using a timer in pwm mode. In this mode, when the timer ask for a new value to the dma, is the value is late, the timer will repeat the last value : a bit is added, and that breaks the protocol. (_Ideally, this function should be done by a fpga, unfortunately, there is not yet fpga cell included in the MCU, and ST seems not interested to offer this option anytime soon_)

the value can be late because of contention on some resources, in our case : dma controler, and memory.

what can be done to improve the situation : 

1/ on dma controller side : better use some integrated feature like fifo, burst mode, priority : i will read again AH4031, AN4667 application note. On F7, there is 2 dma controllers, but each peripheral can use only one, so we cannot better balance their use. In mode advanced MCU like H7, there is 4 dma controllers and nearly all peripherals can use any controller.

2/ on memory side, if there is a task in you application which intensively use the memory, one could help by letting this task use a different memory section, there is 4 memory sections on the F7 : we can modify the linker script to better isolate memory section used by the dma.

3/ on bus side, we can a little bit change the priority of the access to memory between cpu and dma with cortex register AHBSCR

EwoudSmeur:
> Now that you mention the speed, I remember that I had issues with some ESC with the DSHOT600 so I used DSHOT300 and it was fine. I though it was an ESC problem, but maybe not only...

I tried DShot 150 and 300, but it did not seem to have any effect :(

>  In this mode, when the timer ask for a new value to the dma, is the value is late, the timer will repeat the last value : a bit is added, and that breaks the protocol.

Ok I understand now. But how long can the DMA controller take before it is too late? And is there some way to stretch this? And why is it only the second (or first?) bit that seems to be late?

> 2/ on memory side, if there is a task in you application which intensively use the memory, one could help by letting this task use a different memory section, there is 4 memory sections on the F7 : we can modify the linker script to better isolate memory section used by the dma.

The thing is, the actuator values are sent after the autopilot has finished its job, so I think all the intensive stuff is over. After that, only the modules have to be run. That does include all these I2C transactions with the sensors. Maybe one of those is competing for the DMA like you said?

Regarding the solution directions, I am happy to test anything you propose :smile:

alex31:
There is probably something special on your configuration that i am not able to reproduce...

I have add two tasks on my testbed : a memory intensive task, and a  dma intensive task on the same dma channel the dshot driver use : it has an effect, but is is negligible : 4 malformed dshot frames for 100.000. so we can eliminate memory bandwidth problems and probably dma bandwidth problem. I will now generate interruptions storm to see if it has an effect.

@gautierhattenberger : could you plug a logic analyser on an autopilot, capture and save dshot out and send me the capture, to see if this problem is present in any pprz configuration, or if it is specific to Ewoud's one.

@EwoudSmeur : could send me a link on an up2date branch of your project so that a can see is i find something specific that could trig the bug ?

alex31:
able to reproduce bug : i have activated ADC with adc callback triggering interruption. Now i know that having a lot of interrupts triggers the bad behavior, need to understand why since DMA transfert should be ISR aware.

alex31:
I have lowered adc irq and dma priority, lower by a factor 100 malformed frame in my testbench. @EwoudSmeur , could you try the latest commit ?

EwoudSmeur:
@alex31 Great, I'll test asap! I'm working on this branch: https://github.com/EwoudSmeur/paparazzi/tree/cyclone_feedback
Airframe: https://github.com/EwoudSmeur/paparazzi/blob/cyclone_feedback/conf/airframes/tudelft/cyfoam.xml
Chimera board configuration: https://github.com/EwoudSmeur/paparazzi/blob/cyclone_feedback/sw/airborne/boards/chimera/chibios/v1.0/chimera.h

EwoudSmeur:
I tried some more, also changing the priority of the PWM DMA used for DSHOT to 0 and removing other sensors, but no luck. I now also looked at both DShot channels, and they have the same mistakes at different times:

image

alex31:
beware that DMA priority is not like IRQ priority : 

IRQ : higher priority for 0, lower for 15
DMA : lower priority for 0, higher for 3

so don't set Dshot DMA priority to 0 !

EwoudSmeur:
Thanks @alex31 for your advice, and sorry I didn't report back in a while. I have tried the things you mentioned, and I tried removing other modules from my airframe file in an attempt to find which module is conflicting. I didn't keep track of the percentage of wrong messages in an organized way, but looking at the logs that I have there are indeed differences. Usually, the % of wrong messages is between 2 and 7%, but I managed to get one with 30% (I think it's the one where I adjusted the DMA priority the wrong way).
I also noticed that I had four dshot servos defined while only using 2. I tried to define only one dshot servo, and got an interesting result: wrong messages only every ~250 ms.
image

I am not sure if that provides some insight into where the problem may be? I can still try other things that you can think of, but at the moment I don't know what else to try.

alex31:
I have dug a little bit since, and there is good and bad news.
bad news is there is not a single error that can be fixed to resolve the problem, since it's a latency problem on the AHB bus, we can only mitigate the problem, with the hope to reach less than 0.1% of wrong frames, 0 wrong frame is probably impossible without additional hardware or autopilot redesign.

good news is that it can be mitigated applying numerous small fixes that i am still evaluating. 

waiting for theses fixes, you can try on your side theses easy configuration changes (it will not solve the problem, but should lower the number of wrong frame)

sw/airborne/arch/chibios/modules/actuators/esc_dshot.c :
_#define DSHOT_SPEED              150_
...
  _driver->dma_conf = (DMAConfig) {
    .stream = config->dma_stream,
    .channel = config->dma_channel,
    .dma_priority = 3,
    .irq_priority = 6,
    .direction = DMA_DIR_M2P,
    .psize = timerWidthInBytes,
    .msize = timerWidthInBytes,
    .inc_peripheral_addr = false,
    .inc_memory_addr = true,
    .circular = false,
    .error_cb = NULL,
    .end_cb = NULL,
    .pburst = 8,
    .mburst = 8,
    .fifo = 0
  };_


sw/airborne/arch/chibios/modules/actuators/esc_dshot.h
_#define DSHOT_CHANNELS       2_

sw/airborne/boards/chimera/chibios/v1.0/mcuconf.h : 
_#define STM32_ADC_ADCPRE                    ADC_CCR_ADCPRE_DIV8_
_#define STM32_ADC_ADC1_DMA_PRIORITY         0_
_#define STM32_SDC_SDMMC1_DMA_PRIORITY       0_


there is other fixes that will be part of the future patch, but the changes above should lower the error frame count.

alex31:
@EwoudSmeur 

which version of compiler do you use, the one packaged with ubuntu, or recommended official arm one ([https://developer.arm.com/-/media/Files/downloads/gnu-rm/9-2020q2/gcc-arm-none-eabi-9-2020-q2-update-x86_64-linux.tar.bz2?revision=05382cca-1721-44e1-ae19-1e7c3dc96118&la=en&hash=D7C9D18FCA2DD9F894FD9F3C3DC9228498FA281A](url) ?

alex31:
Based on @EwoudSmeur observation that the bits in error (due to dma latency) was always first or second transmitted bits, i have change the frame from "16 bits of dshot + 4 bits of silence" to "2 bits of silence + 16 bits of dshot + 2 bit of silence".
In my setup, with ADC+SDC+UART in dma, % of error frame drop to 0. 
@EwoudSmeur, can you test this ?

EwoudSmeur:
Hi @alex31, your fix totally solved the problem! Thanks for sticking with me all this time, you're the best! :smiley:

alex31:
Thank you for your patience, I have been really slow to find the workaround. The good side is that i now have a better understanding of the DMA subsystem, I am ready to explore the new DMA controller(s) of STM32H7, will soon be time to use these beasts :-)

EwoudSmeur:
No worries, you've been great! But now I'm stumbling on my next quest, which is to get rpm feedback from the ESC. I'm trying to define DSHOT_TELEMETRY_DEV (https://github.com/alex31/paparazzi/blob/fix_dshot_and_dma/sw/airborne/boards/apogee/chibios/v1.0/board.h#L413), but I'm not sure what to define it to. I suppose it should be a SerialDriver, but I can't find any example code with that. Would you know what I can put here? Thanks in advance!

gautierhattenberger:
I have never tried that part with Paparazzi, that's why their is no example. I guess it should be something like `&SDX` with `X` being the correct uart number.

So, can I merge this PR now ?

EwoudSmeur:
Thanks for your suggestion @gautierhattenberger I had tried it without the & but also with I get:
`./boards/chimera/chibios/v1.0/chimera.h:54:30: error: 'SD2' undeclared here (not in a function)`

In any case, I'd be happy if you merge this PR :smile:

gautierhattenberger:
You probably need to enable this uart on ChibiOS side (so from board conf), but without using the Paparazzi flag (like USE_UARTX). Or maybe you can and it doesn't matter.
Actually, configuring this probably seemed annoying, so I decided to do it later when integrating the driver last year...

alex31:
At the ChibiOS level, STM32_SERIAL_USE_USART2 should be defined to TRUE in
the mcuconf.h file, and the pin multiplexer should route USART2 TX en RX
function to the "good" pin in the board.h file. But there is probably a
PPRZ layer to do this safely. In any case, the dshot driver then configure
the serial driver, so no work on your side.
The dshot driver has been written 3 years ago, so i hope that the telemetry
protocol has not been changed since, dshot seems a moving target, i know
that there is a more modern version of dshot where telemetry is not done by
serial, but directly on the pwm pin using the same "pulse width protocol",
it means that the timer has to be reconfigured in Input Capture tu read the
telemetry.

EwoudSmeur:
Thanks for all your help! defining STM32_SERIAL_USE_USART2 to TRUE does the job, and the ESC sends telemetry back! The pins in the board.h file look alright, but reading the telemetry does not work. It seems to be stuck at this line: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/arch/chibios/modules/actuators/esc_dshot.c#L472
I tried UART1 and 2 (without flow control), but no luck. Is it something still with the configuration of the UART that could be wrong? Thanks for any tips.

alex31:
In this kind of telemetry, all the TX pins of the ESC are connected
together, which is barely a good idea with standard USART. It works because
all the ESC put their TX pins in high impedance, all but the one which is
selected to send back telemetry. The driver has to query the ESC one by one
and wait for the answer. So in this mode, is it mandatory that the macro
DSHOT_CHANNELS reflect the exact number of ESCs. If DSHOT_CHANNEL is 4 and
you only have 2 ESCs,  the driver will wait for an answer that will never
arrive.
This is perhaps not the only problem, as Gautier told you, telemetry has
been tested OK, but never within PPRZ, it will eventually work , but
perhaps after some tries ... For now i am in holidays, i will do by best to
help you, but i have no esc, no autopilot where i am to test.

Le mar. 4 août 2020 à 14:17, Ewoud Smeur <notifications@github.com> a
écrit :

EwoudSmeur:
Oh I'm sorry to disturb you during your holiday! Please don't feel obliged to spend any time on this. But your tips were very insightful, and I got it running now, totally awesome! I think the issue was that onGoingQry was not initialized false, such that he would always wait for a non-existing query.
Now that things are running and I understand the inner workings a bit better I have confidence that I can get this thing in the air. Cheers to you for your great help! :+1: :smile:

alex31:
Thanks for finding !  I have added the missing variable init in dshotStart
on my local file, will post a PR when I'll return from holidays !
I am happy that this code is finally used by someone 😃

Le mar. 4 août 2020 à 19:08, Ewoud Smeur <notifications@github.com> a
écrit :

Issue #2537, opened 2020-05-28
heading drift in simulation


EwoudSmeur:
In the nps (JSBSIM) simulator, I observe a slow heading drift of ~-20 degrees per 5 minutes for my cyfoam airframe. The scaled mag sensor values don't change as I just let the vehicle sit there on virtual ground.

The mag is rotated in the airframe, so I have in the mag configuration:
    <module name="mag" type="hmc58xx">
      <configure name="MAG_HMC58XX_I2C_DEV" value="i2c2"/>
      <define name="MODULE_HMC58XX_UPDATE_AHRS"/>

      <define name="HMC58XX_CHAN_X" value="1"/>
      <define name="HMC58XX_CHAN_Y" value="0"/>
      <define name="HMC58XX_CHAN_Z" value="2"/>

      <define name="HMC58XX_CHAN_X_SIGN" value="-"/>
      <define name="HMC58XX_CHAN_Y_SIGN" value="+"/>
      <define name="HMC58XX_CHAN_Z_SIGN" value="+"/>

      
      <define name="MPU9250_MAG_PRESCALER" value="32768"/>

      <define name="HMC58XX_MAG_TO_IMU_PHI" value="RadOfDeg\(6.5\)"/>
      <define name="HMC58XX_MAG_TO_IMU_THETA" value="RadOfDeg\(28.5\)"/>
      <define name="HMC58XX_MAG_TO_IMU_PSI" value="RadOfDeg\(181.0\)"/>
    </module>

And 

    <define name="BODY_TO_IMU_PHI"   value="0." unit="deg"/>
    <define name="BODY_TO_IMU_THETA" value="94." unit="deg"/>
    <define name="BODY_TO_IMU_PSI"   value="0." unit="deg"/>

I imagined that mag rotation to be the cause, but commenting that out as well as the sign stuff still does not work. Maybe someone has an idea? Otherwise I can investigate further...

EwoudSmeur:
Example: 
![image](https://user-images.githubusercontent.com/2464596/83189232-33a2d400-a131-11ea-9548-f0626dfee79d.png)

gautierhattenberger:
that is annoying for sure
do you have the same for a standard frame like a bebop ?
what do you have for the raw imu data ?

EwoudSmeur:
[...]

PR #2553, opened 2020-07-13
Fix modules and telemetry freq

gautierhattenberger:
With the introduction of #2545, the modules and periodic telemetry prescalers was not set properly. The problem is that the correct value is only available after generating the makefile and calling "make" again (old process), but it is not possible anymore have the information from the start, and only the default freq was used (as reported in #2537).
So, instead of a static prescaling computed by the generators, the prescaler values are now computed at compile / run time.
In terms of frequency, it should be similar to the old code as long as there is no periodic functions trying to go faster than the base frequency. It was previously reported by the generator, it won't be the case anymore. A future improvement would be to have checks at compile time eventually.
The only change I'm expecting so far, is the balancing of the functions (to avoid to have all periodic functions called in the same loop). The old way required the base frequency, so it is replaced by something slightly different that should have the same effect. However, there is no guarantee that the sequence of modules' functions will be exactly the same.

I have only tested this code in simulation at the moment, so hardware tests would be welcome.

### DIFF
--- a/conf/modules/extra_dl.xml
+++ b/conf/modules/extra_dl.xml
@@ -23,8 +23,10 @@
 
   <makefile target="ap">
     <configure name="EXTRA_DL_PORT" default="uart1" case="upper|lower"/>
+    <configure name="TELEMETRY_FREQUENCY" default="$(PERIODIC_FREQUENCY)"/>
     <define name="EXTRA_DOWNLINK_DEVICE" value="$(EXTRA_DL_PORT_LOWER)"/>
     <define name="USE_$(EXTRA_DL_PORT_UPPER)"/>
+    <define name="TELEMETRY_FREQUENCY" value="$(TELEMETRY_FREQUENCY)"/>
     <raw>
       # Check for UDP port
       ifneq (,$(findstring udp,$(EXTRA_DL_PORT_LOWER)))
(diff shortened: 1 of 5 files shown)


gautierhattenberger:
@EwoudSmeur did you have a chance to test this ?

EwoudSmeur:
I've built two aircraft, and seen how the generated modules.h file is different. Now it all seems correct, so I'm ok with merging! Great fix, thanks!

martinmm:
Could it be that this patch changes the behavior of the delay="" in the <periodic> functions of module .xml files so that either a 16 bit fraction of the freq could be set or a 0 .. 0.95 in float (which would require a change from int to float for "delay" in module.ml)?
Seems the (mostly sensor) modules using delay="" do no longer work as they asume the "delay" defines a fraction of the main_freq.

gautierhattenberger:
@martinmm yes, you are right, I forgot to update the documentation. Delay is now between 0 and 1 (float), but if it founds a value higher than 1, it divides it by 2^16 (previously the highest value allowed). So it also works with 1..65536 (int)... This is not probably not really nice for users. Do you have a preferred option ? The easiest for me is the [0..1[ option, with warning or errors if not valid. The old way (fraction of the period) was way too confusing for users except a few devs, and I don't think it is compatible anymore anyway.

martinmm:
@gautierhattenberger This mechanism is more straight forward than before where the max. value depended on the frequency. The 0..65536 works fine but it seems that the module.ml needs a change as it only accepts integers for the delay value. A float like delay="0.5" does not work (I did not manage to fix it myself, though).

PR #2560, opened 2020-07-28
[Generator] Throw error or warning for radio conf.

Fabien-B:
Fix divide by zero when neutral is set to max value. @gautierhattenberger can you double-check this to be sure ?
Throw error on incorrect use of min/max when the reverse attribute is used.
Issue warning to ask to switch to new way of reversing a channel.

this throw an error and stop compilation:
`<channel function="THROTTLE" min="1932" neutral="1932" max="1107" average="0" reverse="1"/>`

this issue a warning:
`<channel function="THROTTLE" min="1932" neutral="1932" max="1107" average="0"/>`

### DIFF
--- a/sw/tools/generators/gen_radio.ml
+++ b/sw/tools/generators/gen_radio.ml
@@ -120,12 +120,7 @@ let generate = fun radio xml_file out_file ->
     let (mini, maxi) = if c.reverse then (c.max, c.min) else (c.min, c.max) in
     fprintf out "#define RADIO_%s_MIN %d\n" c.cname mini;
     fprintf out "#define RADIO_%s_MAX %d\n\n" c.cname maxi;
-    if c.max < c.min then (
-      if c.reverse then
-        failwith (sprintf "Error: \"%s\" radio channel: max must be superior to min!" c.cname)
-      else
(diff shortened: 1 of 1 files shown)


sw/tools/generators/gen_radio.ml
```diff
@@ -39,7 +39,7 @@ let check_function_name = fun s ->
 
 
 let norm1_ppm = fun c ->
```
gautierhattenberger, line 41: the function `norm1_ppm2` should be updated the same way

Fabien-B: Oups, yes, done.

Fabien-B:
I edited as requested for the warnings.
I think it makes more sense to fail if min>max AND reverse="1", because it mix the old and the new way of reversing a channel, and it will most probably not do what the user expect.

Fabien-B:
Now that we are at V6.0, time to take a look at this PR?
We can keep the old behavior (min>max) for some times, and remove it completely in some time?
And I stand for throwing an error if (min>max AND reverse="1").

fvantienen:
I'm for removing the (min>max). Yes it makes people need to change their configs, but is way more consistent. We can add a warning/error

Fabien-B:
I throw an error if min>max.
I still have to check all config files.

Fabien-B:
Fixing conf files in a separe PR : #2710

gautierhattenberger (approved):

PR #2564, opened 2020-08-10
Module for transparent control of external GPIOs

tomvand:
This PR adds a new module that can redirect calls to gpio functions to other modules, providing *transparent* access to gpio pins on expansion ICs (e.g. PCA9534).

Unfortunately the current paparazzi code makes calls directly to the libopencm3 gpio functions without any HAL inbetween, so I had to use some gcc magic to wrap those functions. This is a bit hacky and will probably not port to other compilers.

I have tested the gpio_setup_output and gpio_toggle functions and these seem to work well. (Tested on Trashcan STM32F4 with softI2C to PCA9536).

I have renamed the pca95x4 peripheral to pca95xx, as it is also compatible with the pca953**6** that I am using.


tomvand:
I found some problems when the external gpio pin is used as spi slave select, due to initialization order (pin is configured before module is started). I'll post it here when I've found a solution, I will probably have to change the code in this pr.

gautierhattenberger:
ok, let me know when it is ready

tomvand:
Hi @gautierhattenberger , I fixed the initialization problem so the code should be ok now. Most of the structs are now initialized at compile time, the remainder using lazy initialization when necessary. This has also improved the API somewhat.

I've also slipped in a small fix to softi2c, which was hanging sometimes. (I suspect that sys_time_usleep hangs when interrupts are disabled by spi, but did not confirm this; adding LED code for debugging seemed to fix the issue...).

I have tested the module by letting it control the chip select pin of the pmw3901 and that seems to work correctly now :)

conf/modules/gpio_ext_pca95xx.xml
```diff
@@ -0,0 +1,27 @@
...
+
+Uses the gpio_ext_common mechanism for transparent redirecting of gpio
+calls. Multiple PCA95XX can be used, on different I2C busses.
+Configuration and writing can be performed non-blocking (reading is
+always blocking).
+    </description>
+    <section name="GPIO_EXT" prefix="GPIO_EXT_">
+      <define name="PROVIDERx" value="GPIO_EXT_PCA95XX" description="Set-up module as provider for port GPIOEXTx"/>
+      <define name="BLOCKINGx" value="TRUE|FALSE" description="Whether write operations to this port should be blocking. (Note: reading is always blocking)."/>
+      <define name="I2C_PERIPHx" value="&i2c0" description="Pointer(!) to i2c peripheral"/>
```
gautierhattenberger, line 15: to avoid using `&` for consistency with other modules, you can have in your code:
```
#ifdef I2C_PERIPHx
#define I2C_PERIHx_PTR &(I2C_PERIPHx)
#else
#define I2C_PERIHx_PTR NULL
#endif
```

sw/airborne/modules/gpio_ext/gpio_ext_common.c
```diff
@@ -0,0 +1,196 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+#include <assert.h>
+
+
+/* External GPIO providers */
+#include "gpio_ext_pca95xx.h"
+
+#define GPIO_EXT_NOT_PROVIDED NULL
+#define GPIO_EXT_PCA95XX &pca95xx_functions
```
gautierhattenberger, line 41: doesn't seems to be used, and should not be here anyway I think

sw/airborne/modules/gpio_ext/gpio_ext_common.c
```diff
@@ -0,0 +1,196 @@
...
+
+
+
+/*
+ * External GPIO implementation
+ */
+#define SAFE_CALL(_port, _fn, _args...) if(gpio_ext_impl[_port - GPIOEXT1] && gpio_ext_impl[_port - GPIOEXT1]->_fn) gpio_ext_impl[_port - GPIOEXT1]->_fn(_args);
+
+// Wrapping functions
+void __wrap_gpio_setup_output(uint32_t port, uint32_t gpios);
```
gautierhattenberger, line 132: The original gpio functions have a `gpio_port_t` type for the port (but not always...) which is arch dependent. Did you test this with the regular stm32 arch, chibios or both ?

sw/airborne/modules/gpio_ext/gpio_ext_common.h
```diff
@@ -0,0 +1,52 @@
...
+  void (*set)(uint32_t port, uint32_t gpios);
+  void (*clear)(uint32_t port, uint32_t gpios);
+  void (*toggle)(uint32_t port, uint32_t gpios);
+};
+
+
+// External GPIO configuration
+extern struct i2c_periph *gpio_ext_i2c_periph[];
+extern const uint8_t gpio_ext_i2c_addr[];
+extern const bool gpio_ext_blocking[];
```
gautierhattenberger, line 49: Actually, I don't see why the `gpio_ext_common` should depend on i2c. All these arrays and related defines should be left to the provider, so this module could be used by any type of peripherals.

gautierhattenberger (changes requested):
As a side note, where do you actually use these wrapper functions ?

sw/airborne/modules/gpio_ext/gpio_ext_common.c
tomvand: I know, I saw that the function arguments are different between architectures, but was hesitant to change anything because it might break other people's code. The current uint32_t does not match the stm32 functions either iirc, but still seems to work and is large enough for all architectures. I have tested this only on the stm32 arch.

sw/airborne/modules/gpio_ext/gpio_ext_common.c
tomvand: This define is used in the airframe file, to specify which module provides the GPIOEXTx port. I can move it to gpio_ext_pca95xx.h instead?

sw/airborne/modules/gpio_ext/gpio_ext_common.h
tomvand: I put it here because the defines *might* also be used by other modules in the future, preventing code duplication. Putting this in the provider module also makes a lot of sense, I'll give it a shot and see if it is better that way.

tomvand:
GCC replaces *all* gpio functions with the wrapped version, so technically everywhere ;)

I use the wrappers for the pmw3901 chip select pin, because my autopilot board did not have enough easily-accessible gpio pins. I did not want to change any spi code or potentially any other modules to use the external gpio, hence the wrappers.

sw/airborne/modules/gpio_ext/gpio_ext_common.c
gautierhattenberger: yes, if possible

sw/airborne/modules/gpio_ext/gpio_ext_common.c
gautierhattenberger: then I don't think this will work with chibios since we can overload functions

gautierhattenberger:
This module will most likely not work with ChibiOS. But it might change once we upgrade to a more recent version, we'll see....

sw/airborne/modules/gpio_ext/gpio_ext_common.c
tomvand: I have never used chibios before so I can't really comment, but looking at the code I think that the only problem right now is that the gpio_arch  functions are static? Apart from that I think it should work, it would just redirect the gpio calls to the existing arch functions unless things are being written to a GPIOEXT port.

What do you mean by overloading functions? It looks like normal c code to me? (arch/chibios/mcu_periph/gpio_arch.c/h)

gautierhattenberger: the ports in chibios are `ioport_t` which are pointers to structures, incompatible with uint32, that's why we have this `gpio_port_t` typedef at some places
but I wouldn't bother to much since it seems that latest chibios versions are not doing differently

sw/airborne/modules/gpio_ext/gpio_ext_common.c
tomvand: fixed

sw/airborne/modules/gpio_ext/gpio_ext_common.c
tomvand: Ok thanks, then I understand. I'll leave it like this for now.

sw/airborne/modules/gpio_ext/gpio_ext_common.h
tomvand: Fixed

conf/modules/gpio_ext_pca95xx.xml
tomvand: Thanks for the suggestion, fixed. I've also updated the module xml to generate the missing USE_I2Cx defines.

PR #2568, opened 2020-08-17
Indi rate interface

EwoudSmeur:
Rohan, one of our students, would like to pass a rate reference to the INDI controller from a module.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
@@ -35,7 +35,7 @@
 void stabilization_attitude_init(void)
 {
   // Check if the indi init is already done for rate control
-#ifndef STABILIZATION_RATE_INDI
+#ifndef USE_STABILIZATION_RATE_INDI
   stabilization_indi_init();
 #endif
 }
@@ -62,9 +62,7 @@ void stabilization_attitude_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t head
 
 void stabilization_attitude_run(bool in_flight)
 {
-  struct Int32Quat quat_sp = {1, 0, 0, 0}; // always hover
-
-  stabilization_indi_attitude_run(in_flight, quat_sp);
+  stabilization_indi_attitude_run(stab_att_sp_quat, in_flight);
 }
 
 void stabilization_attitude_read_rc(bool in_flight, bool in_carefree, bool coordinated_turn)
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
@@ -62,7 +62,9 @@ void stabilization_attitude_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t head
 
 void stabilization_attitude_run(bool in_flight)
 {
-  stabilization_indi_run(in_flight, FALSE);
+  struct Int32Quat quat_sp = {1, 0, 0, 0}; // always hover
+
+  stabilization_indi_attitude_run(in_flight, quat_sp);
 }
 
 void stabilization_attitude_read_rc(bool in_flight, bool in_carefree, bool coordinated_turn)
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -35,6 +35,9 @@
 #include "firmwares/rotorcraft/stabilization/stabilization_attitude_rc_setpoint.h"
 #include "firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.h"
 
+// Added to access variable "stabilization_rate_sp" for direct rate control
+#include "firmwares/rotorcraft/stabilization/stabilization_rate_indi.h"
+
 #include "math/pprz_algebra_float.h"
 #include "state.h"
 #include "generated/airframe.h"
@@ -350,9 +353,9 @@ static void stabilization_indi_calc_cmd(struct Int32Quat *att_err, bool rate_con
 
   struct FloatRates rate_ref;
   if (rate_control) { //Check if we are running the rate controller
-    rate_ref.p = (float)radio_control.values[RADIO_ROLL]  / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
-    rate_ref.q = (float)radio_control.values[RADIO_PITCH] / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
-    rate_ref.r = (float)radio_control.values[RADIO_YAW]   / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
+    rate_ref.p = stabilization_rate_sp.p;
+    rate_ref.q = stabilization_rate_sp.q;
+    rate_ref.r = stabilization_rate_sp.r;
   } else {
     //calculate the virtual control (reference acceleration) based on a PD controller
     rate_ref.p = reference_acceleration.err_p * QUAT1_FLOAT_OF_BFP(att_err->qx)
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -348,27 +348,8 @@ void stabilization_indi_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t heading)
  *
  * Function that calculates the INDI commands
  */
-static void stabilization_indi_calc_cmd(struct Int32Quat *att_err, bool rate_control, bool in_flight)
+static void stabilization_indi_calc_cmd(struct FloatRates rate_ref, bool in_flight)
 {
-
-  struct FloatRates rate_ref;
-  if (rate_control) { //Check if we are running the rate controller
-    rate_ref.p = stabilization_rate_sp.p;
-    rate_ref.q = stabilization_rate_sp.q;
-    rate_ref.r = stabilization_rate_sp.r;
-  } else {
-    //calculate the virtual control (reference acceleration) based on a PD controller
-    rate_ref.p = reference_acceleration.err_p * QUAT1_FLOAT_OF_BFP(att_err->qx)
-                 / reference_acceleration.rate_p;
-    rate_ref.q = reference_acceleration.err_q * QUAT1_FLOAT_OF_BFP(att_err->qy)
-                 / reference_acceleration.rate_q;
-    rate_ref.r = reference_acceleration.err_r * QUAT1_FLOAT_OF_BFP(att_err->qz)
-                 / reference_acceleration.rate_r;
-
-    // Possibly we can use some bounding here
-    /*BoundAbs(rate_ref.r, 5.0);*/
-  }
-
   struct FloatRates *body_rates = stateGetBodyRates_f();
 
   q_filt = (q_filt*3+body_rates->q)/4;
@@ -495,7 +476,7 @@ static void stabilization_indi_calc_cmd(struct Int32Quat *att_err, bool rate_con
  *
  * Function that should be called to run the INDI controller
  */
-void stabilization_indi_run(bool in_flight, bool rate_control)
+void stabilization_indi_attitude_run(bool in_flight, struct Int32Quat quat_sp)
 {
 
   /* Propagate the filter on the gyroscopes */
@@ -519,13 +500,26 @@ void stabilization_indi_run(bool in_flight, bool rate_control)
   /* attitude error                          */
   struct Int32Quat att_err;
   struct Int32Quat *att_quat = stateGetNedToBodyQuat_i();
-  int32_quat_inv_comp(&att_err, att_quat, &stab_att_sp_quat);
+  int32_quat_inv_comp(&att_err, att_quat, &quat_sp); // [quat_sp is stab_att_sp_quat -- HOW TO DISTINGUISH WHAT YOU GET FROM RC AND SET BY USER FROM ANOTHER MODULE?]
   /* wrap it in the shortest direction       */
   int32_quat_wrap_shortest(&att_err);
   int32_quat_normalize(&att_err);
 
+  struct FloatRates rate_ref; 
+
+  //calculate the virtual control (reference acceleration) based on a PD controller
+  rate_ref.p = reference_acceleration.err_p * QUAT1_FLOAT_OF_BFP(att_err->qx)
+                / reference_acceleration.rate_p;
+  rate_ref.q = reference_acceleration.err_q * QUAT1_FLOAT_OF_BFP(att_err->qy)
+                / reference_acceleration.rate_q;
+  rate_ref.r = reference_acceleration.err_r * QUAT1_FLOAT_OF_BFP(att_err->qz)
+                / reference_acceleration.rate_r;
+
+  // Possibly we can use some bounding here
+  /*BoundAbs(rate_ref.r, 5.0);*/
+
   /* compute the INDI command */
-  stabilization_indi_calc_cmd(&att_err, rate_control, in_flight);
+  stabilization_indi_calc_cmd(rate_ref, in_flight);
 
   // Set the stab_cmd to 42 to indicate that it is not used
   stabilization_cmd[COMMAND_ROLL] = 42;
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -346,18 +346,8 @@ void stabilization_indi_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t heading)
  *
  * Function that calculates the INDI commands
  */
-void stabilization_indi_calc_cmd(struct FloatRates rate_ref, bool in_flight)
+void stabilization_indi_calc_cmd(struct FloatRates angular_accel_ref, bool in_flight)
 {
-  struct FloatRates *body_rates = stateGetBodyRates_f();
-
-  q_filt = (q_filt*3+body_rates->q)/4;
-  r_filt = (r_filt*3+body_rates->r)/4;
-
-  //calculate the virtual control (reference acceleration) based on a PD controller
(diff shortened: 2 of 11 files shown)


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -23,29 +23,106 @@
...
+  }
+
+  if (YAW_RATE_DEADBAND_EXCEEDED() && !THROTTLE_STICK_DOWN()) {
+    stabilization_rate_sp.p = (float) radio_control.values[RADIO_YAW] * STABILIZATION_INDI_MAX_RATE / MAX_PPRZ;
+  } else {
+    stabilization_rate_sp.p = 0;
+  }
+}
+
+void stabilization_rate_set_rpy_setpoint_i(struct Int32Eulers *rpy)
```
EwoudSmeur, line 123: Why do you need this function?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -350,9 +353,9 @@ static void stabilization_indi_calc_cmd(struct Int32Quat *att_err, bool rate_con
 
   struct FloatRates rate_ref;
   if (rate_control) { //Check if we are running the rate controller
-    rate_ref.p = (float)radio_control.values[RADIO_ROLL]  / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
-    rate_ref.q = (float)radio_control.values[RADIO_PITCH] / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
-    rate_ref.r = (float)radio_control.values[RADIO_YAW]   / MAX_PPRZ * STABILIZATION_INDI_MAX_RATE;
+    rate_ref.p = stabilization_rate_sp.p;
```
EwoudSmeur, line 356: The stabilization_rate_sp struct is now defined in the stabilization_rate_indi c file, but if you do not build with rate control this will not be defined, leading to a compile error. The best thing would be to move that definition to this c file and the stabilization_ind_simple.c file. The declaration can remain in the stabilization_indi_header file.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
electrobusy: To set the commanded rates coming from a different module. If you check the "stabilization_attitude_*.c" files, there is also a similar function.

EwoudSmeur: Ok, but then it is best to rename it in my opinion, because you're setting rates and not roll, pitch, yaw. Also, you are passing an Int32Eulers struct pointer but assigning to a FloatRates struct.

gautierhattenberger:
In order to make the code more flexible to different control inputs, would it be possible to have two separated functions for attitude and rate that would pass the setpoint as an argument. So the prototypes of the functions would look like:
```
stabilization_indi_attitude_run(bool in_flight, struct Int32Quat quat_sp);
stabilization_indi_rate_run(bool in_flight, struct Int32Rates rates_sp);
stabilization_indi_calc_cmd(struct FloatRates rate_ref, bool in_flight);
```
It moves a bit parts of the code, but in the end the `rate_run` and `attitude_run` functions can locally choose the proper setpoint, and the functions can also be called from external modules without dependency.

EwoudSmeur:
I think that is a good idea @gautierhattenberger, thanks for the suggestion. Can you do this @electrobusy?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
electrobusy: Yes, you are right. I found in state.h file the definition for rates: "Int32Rates". I will change that and make sure that there is no type mismatch.

electrobusy:
Yes, I think I can. So basically, each `_run` function receives the respective control input, the attitude in the "stabilization_indi_simple" (stabilization_indi) file(s) and in the "stabilization_rate_indi", the same function receives the rates.

However, I there is also a "stabilization_rate" file. Should I also make changes in `_run` function there?

gautierhattenberger:
No need to change the general functions `stabilization_rate_run` functions for now, only the functions directly related to INDI.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
EwoudSmeur: Take care that Int32Rates != FloatRates

electrobusy: Yes, I made the conversion using: RATES_FLOAT_OF_BFP ;)

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -388,18 +349,53 @@ static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct I
  * @param in_flight not used
  * @param rate_control rate control enabled, otherwise attitude control
  */
-void stabilization_indi_run(bool in_flight __attribute__((unused)), bool rate_control)
+void stabilization_indi_attitude_run(bool in_flight __attribute__((unused)), struct Int32Quat quat_sp) // how does quat_sp is set when using RC?
```
gautierhattenberger, line 352: > // how does quat_sp is set when using RC?

this will come from above, depending on the mode

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -388,18 +349,53 @@ static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct I
...
 {
   /* attitude error                          */
-  struct Int32Quat att_err;
+  struct Int32Quat att_err;
```
gautierhattenberger, line 355: remove tailing whitespace

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -23,37 +23,140 @@
  *  Rate stabilization for rotorcrafts based on INDI by Ewoud Smeur.
  */
 #include "firmwares/rotorcraft/stabilization.h"
-#include "firmwares/rotorcraft/stabilization/stabilization_rate.h"
+#include "firmwares/rotorcraft/stabilization/stabilization_rate_indi.h"
```
gautierhattenberger, line 26: So, this where the magic should come from. In fact `rate_indi` must implement the `rate` interface (and in this case add extra function for external control).
I'm suggesting the following structure:
- in `stabilization_rate_indi.h`
  - include `stabilization_rate.h`
  - remove the basic `stabilization_rate_*` functions declarations, already provided by rate.h
  - keep the `stabilization_rate_indi_*` for external modules
- in `stabilization_rate_indi.c`
  - include `stabilization_rate_indi.h` as you did now
  - implement the `rate_indi_*` functions as you did now
  - implement the basic `rate_*` functions by calling the proper indi related functions, especially the `stabilization_rate_run` function
- in `stabilization_rate_indi.xml` module file:
  - change the header to include `stabilization_rate_indi.h`
  - compile only the `stabilization_rate_indi.c` as it is now

How is it going to work ?
- if you are in rate mode, the autopilot will call the basic `stabilization_rate_*` functions, setting the setpoint from RC and calling the run afterwards
- if you are in a custom mode, you have to set the setpoint and then call the `rate_indi_enter` and `rate_indi_run` functions by yourself

Does it makes sense ?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -23,37 +23,140 @@
...
-void stabilization_rate_enter(void)
+static void stabilization_indi_rate_calc_cmd(int32_t indi_commands[], struct FloatRates rate_ref, bool in_flight)
 {
-  stabilization_indi_enter();
+  /* Get the angular acceleration references */
+  indi.angular_accel_ref.p =  indi.reference_acceleration.rate_p * (rate_ref.p - body_rates->p);
+  indi.angular_accel_ref.q =  indi.reference_acceleration.rate_q * (rate_ref.q - body_rates->q);
+  indi.angular_accel_ref.r =  indi.reference_acceleration.rate_r * (rate_ref.r - body_rates->r);
+
+  stabilization_indi_calc_cmd(indi_commands)
```
gautierhattenberger, line 137: this function seems to be static in the stabilization indi code, it should be extern

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
electrobusy: may I ask why? the compiler will just pass through it, no?

gautierhattenberger: yes, indeed, it is only to follow the coding style: http://docs.paparazziuav.org/latest/stylec.html
you can also use the `./fix_code_style.sh` script at the root of the project with files to format as arguments.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
electrobusy: Yes, I think I got it. Thanks a lot for your detailed feedback! 
I got through this morning and push the code in a few minutes. Though I have been doing without compiling it, because there is something I would like to review further in this exact file.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -127,33 +129,37 @@ void stabilization_rate_set_setpoint_i(struct Int32Rates *pqr)
...
 
-static void stabilization_indi_rate_calc_cmd(int32_t indi_commands[], struct FloatRates rate_ref, bool in_flight)
+static void stabilization_indi_rate_calc_cmd(int32_t indi_commands[], bool in_flight)
 {
   /* Get the angular acceleration references */
   indi.angular_accel_ref.p =  indi.reference_acceleration.rate_p * (rate_ref.p - body_rates->p);
   indi.angular_accel_ref.q =  indi.reference_acceleration.rate_q * (rate_ref.q - body_rates->q);
   indi.angular_accel_ref.r =  indi.reference_acceleration.rate_r * (rate_ref.r - body_rates->r);
 
+// WE NEED TO CHOOSE BETWEEN STABILIZATION_INDI_SIMPLE AND STABILIZATION_INDI - NOT SURE IF I DID CORRECTLY HERE:
```
electrobusy, line 139: Could someone tell me if I did correctly here? Because we have two INDI files. One, which is the `_simple` and second which considers a control allocation method.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -288,50 +288,8 @@ static inline void finite_difference(float output[3], float new[3], float old[3]
  * @param att_err quaternion attitude error
  * @param rate_control rate control enabled, otherwise attitude control
  */
-static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct Int32Quat *att_err, bool rate_control)
+void stabilization_indi_calc_cmd(int32_t indi_commands[])
```
gautierhattenberger, line 291: Something seems wrong here. `indi_commands` is the output vector, so their is no more setpoint input to this function. It should be the same `rate_ref` than full indi, from which the acceleration setpoint is computed.
As a side note, the same function in full indi code is directly writing the commands the the `actuators_pprz` array. Maybe it is better to pass the output array as an argument like it is done here.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
gautierhattenberger: The code in this function has several errors I think.
Beside that, even if the prototype of the `calc_cmd` function is the same (rate ref as input, command array as output), it will still be necessary to use an `ifdef` as array outputs are different.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.h
```diff
@@ -26,12 +27,15 @@
...
+
 #include "math/pprz_algebra_int.h"
 
-extern void stabilization_rate_init(void);
-extern void stabilization_rate_read_rc(void);
-extern void stabilization_rate_read_rc_switched_sticks(void);
-extern void stabilization_rate_run(bool in_flight);
-extern void stabilization_rate_enter(void);
+extern void stabilization_indi_rate_set_setpoint_i(struct Int32Rates *pqr);
+extern void stabilization_rate_run(bool in_flight, struct Int32Rates rates_sp);
```
gautierhattenberger, line 36: This might conflict with the same function declared in `stabilization_rate.h`. So here it should be `stabilization_indi_rate_run`. And in the C code, you need to implement both function with one simply calling the second:
```
void stabilization_rate_run(bool in_flight)
{
  ...
  stabilization_indi_rate_run(in_flight, rate_ref);
}
```

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
electrobusy: Would you recomend having the same function in both indi codes with the same inputs?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_indi.c
```diff
@@ -62,7 +62,9 @@ void stabilization_attitude_set_earth_cmd_i(struct Int32Vect2 *cmd, int32_t head
 
 void stabilization_attitude_run(bool in_flight)
 {
-  stabilization_indi_run(in_flight, FALSE);
+  struct Int32Quat quat_sp = {1, 0, 0, 0}; // always hover
+
+  stabilization_indi_attitude_run(in_flight, quat_sp);
```
gautierhattenberger, line 67: It should be `stab_att_quat_sp` here I think.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi_simple.c
```diff
@@ -74,8 +74,7 @@
 struct Int32Eulers stab_att_sp_euler;
 struct Int32Quat   stab_att_sp_quat;
 
-static int32_t stabilization_att_indi_cmd[COMMANDS_NB];
-static inline void stabilization_indi_calc_cmd(int32_t indi_commands[], struct Int32Quat *att_err, bool rate_control);
+// static int32_t stabilization_att_indi_cmd[COMMANDS_NB];
```
gautierhattenberger, line 77: remove if not used anymore

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -22,38 +23,194 @@
...
+
+#if PERIODIC_TELEMETRY
+#include "subsystems/datalink/telemetry.h"
+
+static void send_rate(struct transport_tx *trans, struct link_device *dev)
+{
+  pprz_msg_send_RATE_LOOP(trans, dev, AC_ID,
+                          &stabilization_rate_sp.p,
+                          &stabilization_rate_sp.q,
+                          &stabilization_rate_sp.r);
```
gautierhattenberger, line 86: It seems that the `RATE_LOOP` message have more field than that: http://docs.paparazziuav.org/latest/paparazzi_messages.html#RATE_LOOP

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -22,38 +23,194 @@
...
+void stabilization_rate_set_setpoint_i(struct Int32Rates *pqr)
 {
-  stabilization_indi_enter();
+  RATES_FLOAT_OF_BFP(stabilization_rate_sp, *pqr);
 }
+*/ 
+
+/* -----
+-----> CHECK FROM HERE ONWARDS! 
+----- */
```
gautierhattenberger, line 163: you can remove commented code if not used

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.c
```diff
@@ -22,38 +23,194 @@
...
+  stabilization_indi_calc_cmd(angular_accel_ref, in_flight)
+}
+
+/**
+ * @brief Run this indi rate interface 
+ */
+void stabilization_rate_indi_run(bool in_flight, struct Int32Rates rates_sp)
+{
+  /* compute the INDI rate command */
+  stabilization_indi_rate_calc_cmd(indi_comands);
```
gautierhattenberger, line 198: `rates_sp` instead of `indi_comands` ?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_rate_indi.h
```diff
@@ -26,12 +27,15 @@
...
 #include "math/pprz_algebra_int.h"
 
-extern void stabilization_rate_init(void);
-extern void stabilization_rate_read_rc(void);
-extern void stabilization_rate_read_rc_switched_sticks(void);
-extern void stabilization_rate_run(bool in_flight);
-extern void stabilization_rate_enter(void);
+extern struct FloatRates stabilization_rate_sp;
+
+/* extern void stabilization_rate_indi_set_setpoint_i(struct Int32Rates *pqr); -- TO BE REMOVED */
```
gautierhattenberger, line 37: you can remove

gautierhattenberger (changes requested):
Almost there!
The reference to gazebo models submodule is changed, which is most likely a merge/rebase side effect. Can you fix it so that it is not modified ?

gautierhattenberger (approved):

gautierhattenberger:
If you are all good with this, I can merge

@electrobusy thanks for your perseverance in applying the requested changes

electrobusy:
I would like to thank both you and @EwoudSmeur for the feedback :)

EwoudSmeur:
Thanks @electrobusy and @gautierhattenberger!

PR #2569, opened 2020-08-18
Fix settings in log

### LINKED ISSUE #2563: Setting changes in flight log not traceable
When changing settings during a flight, this is logged in the GCS log with the 'SETTING' message. This message contains the setting ID. The .log file contains all setting definitions, but the setting ID is not in the .log file, and cannot be found by straightforward counting of dl_setting tags. The generated settings.h file contains the setting ID's, but this one is overwritten all the time. Is it possible to find the settings order somehow, or to save it in the .log file?

gautierhattenberger:
[log] add the latest generated settings if available to log file
    
    This way, the exact order of the settings used during the flight might
    be available.
    The only false case would be to start a server, rebuild and aircraft
    with different parameters or target, then flash and start the plane. In
    this case, there is no guarantee that the information is correct.
    However, this is the best we can do, since we can't know when starting
    the server with which target the aircraft will be used.
    
    close #2563

Also fix the log replay

### DIFF
--- a/sw/ground_segment/tmtc/server.ml
+++ b/sw/ground_segment/tmtc/server.ml
@@ -77,8 +77,18 @@ let expand_aicraft x =
     Xml.Element ("ignoring_aircraft",["name", ac_name],[])
   in
   try
+    (* parse aircraft *)
     let ac = Aircraft.parse_aircraft ~parse_all:true "" x in
-    if List.length ac.Aircraft.xml > 0 then Xml.Element (Xml.tag x, Xml.attribs x, ac.Aircraft.xml)
+    (* Add latest generated settings if any with a special tag as it is the only way to have this information.
+     * The settings parsed by Aircraft module will not include module settings as we don't know the target.
+     * It should be the correct settings, unless an aircraft is rebuilt with different parameters or target
+     * and with the server already running and not restarted. *)
+    let settings_xml = try
+        let xml = ExtXml.parse_file (Env.paparazzi_home // "var" // "aircrafts" // ac_name // "settings.xml") in
+        [Xml.Element ("generated_settings", [], Xml.children xml)]
+      with _ -> []
+    in
+    if List.length ac.Aircraft.xml > 0 then Xml.Element (Xml.tag x, Xml.attribs x, ac.Aircraft.xml @ settings_xml)
     else failwith "Nothing to parse"
   with
     | Failure msg -> handle_error_message "Fail with" msg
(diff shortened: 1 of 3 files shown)


EwoudSmeur:
I did a quick test and saw that the first dl_setting was listed double:
```
        <dl_settings>
          <dl_settings name="System">
            <dl_settings name="Telemetry">
              <dl_setting min="0" step="1" max="13" var="telemetry_mode_Main" shortname="Main" values="default|ppm|raw_sensors|scaled_sensors|ahrs|rate_loop|attitude_setpoint_viz|attitude_loop|vert_loop|vel_guidance|h_loop|aligner|tune_hover|RTCM3" header="generated/periodic_telemetry">
                <key_press key="h" value="10"/>
                <key_press key="q" value="9"/>
                <key_press key="v" value="8"/>
                <key_press key="a" value="7"/>
                <key_press key="v" value="6"/>
                <key_press key="d" value="0"/>
              </dl_setting>
            </dl_settings>
          </dl_settings>
        </dl_settings>
      </settings>
      <generated_settings>
        <dl_settings>
          <dl_settings name="System">
            <dl_settings name="Telemetry">
              <dl_setting min="0" step="1" max="13" var="telemetry_mode_Main" shortname="Main" values="default|ppm|raw_sensors|scaled_sensors|ahrs|rate_loop|attitude_setpoint_viz|attitude_loop|vert_loop|vel_guidance|h_loop|aligner|tune_hover|RTCM3" header="generated/periodic_telemetry">
                <key_press key="h" value="10"/>
                <key_press key="q" value="9"/>
                <key_press key="v" value="8"/>
                <key_press key="a" value="7"/>
                <key_press key="v" value="6"/>
                <key_press key="d" value="0"/>
              </dl_setting>
            </dl_settings>
```
Is that supposed to be like that?

gautierhattenberger:
Yes. The `settings` node contains the part generated by the aircraft parser, and that is incomplete in this case since it is not associated to a particular target. I didn't want to make special case for this parser, so I let the server itself add the `generated_settings` node, that is only available when the aircraft is generated. This information is only an indication since if a clean/build occurs after the start of the server, it may no longer match.
So don't change settings without restarting the server before flying again and it will be fine.

gautierhattenberger:
As a side node, the old version was always "wrong", but nobody really noticed. The new generator just doesn't provide module settings when the target is unknown.

EwoudSmeur:
Ok, so I should just look at the node generated_settings? I can live with not rebuilding with a running server :wink: 
Thanks a lot for fixing!

gautierhattenberger:
yes, `generated_settings` node should be correct

gautierhattenberger:
@EwoudSmeur is it good for you ?

EwoudSmeur:
Yes perfect, thanks! :smile:

EwoudSmeur (commented):
Great fix!

EwoudSmeur (approved):

PR #2571, opened 2020-08-21
[bebop] improve network script

gautierhattenberger:
- check that wifi join command is actually correct, with retries
- allow configuration for static IP


EwoudSmeur:
That is some useful functionality! I was actually wondering, how is the wifi key configured? Should it be set as an environment variable via telnet?

gautierhattenberger:
At the moment, you have to write it by end in the pprz.conf directly on the bebop. Maybe I can add the option to change the `AMODE` field from the bebop script.

OpenUAS:
I'll will test with 4x bebop1 in on 3.9.2020 and if working honour the PR else I'll fix it

gautierhattenberger:
just added the encryption configuration option, not tested, but the rest is working for us

EwoudSmeur:
Superb, thanks! :smile:

OpenUAS:
Worked can be honored. Came across some possible improvments, but if an issue will fix it and make PR or post an issue.

OpenUAS (approved):
Changes Worked, although if there are comments in the pprz.conf not OK but that can be fixed in another PR and issue. 

4x Bebop 1 to the router in the network all can ping eachother
working

PR #2577, opened 2020-09-10
Fix various (second batch)

gautierhattenberger:
A second set of small fixes with:
- remove compile error in dshot driver (remove wrong and unused assert)
- fix file name report for chibios SD logger
- build file binary in OBJDIR for chibios to prevent linking at each upload command
- allow upside-down initialization for the float DCM filter
- fix makefile script to get the number of CPU core
- LTO is optional in board makefiles


noether (approved):
They look good to me. Thanks for the fixes :P

PR #2578, opened 2020-09-10
Added sensor for using TFMini via I2C port

OpenUAS:
A driver for the TFMini ranger when connected via I2C

Before code
nosensor

After driver added
working_tfmini_i2c

Note that not all the possible features are implemented, but enough for the sensor to be useful , before 0% useful, now 100% useful...


sw/airborne/modules/lidar/tfmini_i2c.c
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Send measured value and status information so it can be read back in e.g. log file for debugging
+ */
+static void tfmini_i2c_send_lidar(struct transport_tx *trans, struct link_device *dev)
+{
+  pprz_msg_send_LIDAR(trans, dev, AC_ID,
+                      &tfmini_i2c.dist,
+                      &tfmini_i2c.status,
+                      (uint8_t*)tfmini_i2c.trans.status);//No guarantee on value but for telemetry of no importance
```
gautierhattenberger, line 135: `tfmini_i2c.trans.status` should not be cast to a pointer. Create a uint8 variable first and then pass its address.

sw/airborne/modules/lidar/tfmini_i2c.c
```diff
@@ -0,0 +1,252 @@
...
+      if (tfmini_i2c.compensate_rotation) {
+        float phi = stateGetNedToBodyEulers_f()->phi;
+        float theta = stateGetNedToBodyEulers_f()->theta;
+        float gain = (float)fabs((double)(cosf(phi) * cosf(theta)));
+        tfmini_i2c.dist = tfmini_i2c.dist * gain;
+      }
+
+#else // SITL
+tfmini_i2c.dist = stateGetPositionEnu_f()->z;
+#endif // SITL
```
gautierhattenberger, line 229: this doesn't look at the correct place, check that it is compiling with SITL

sw/airborne/modules/lidar/tfmini_i2c.c
OpenUAS: indeed, thx for spotting my lazyness ;)

gautierhattenberger (approved):

PR #2579, opened 2020-09-14
[Joystick] Add option to wait for an event on all axis to start.

Fabien-B:
Initial position of axis are unknown until an event occurs on this axis.
This PR adds the option "-c" on input2ivy to wait for an event to occurs on each axis before the program start its normal behavior (sending ivy messages).
This ensures that no false values are sent.

### DIFF
--- a/sw/ground_segment/joystick/input2ivy.ml
+++ b/sw/ground_segment/joystick/input2ivy.ml
@@ -40,6 +40,7 @@ open Random
 let (//) = Filename.concat
 let conf_dir = Env.paparazzi_home // "conf"
 let verbose = ref false (* Command line option *)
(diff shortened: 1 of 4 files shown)


OpenUAS:
A real life very useful addition, thx! It would be nice if the extra parameter and explanation is reflected in the Sphinx or Wiki docs. Even after the PR is honoured is fine IMHO.

Fabien-B:
You're very right about that, I am the one pushing to use sphinx more, and then I forget to use it :-)
I wrote more sphinx doc the last days so I will add that and make this big doc PR very soon.

gautierhattenberger (approved):

PR #2581, opened 2020-09-15
[fix] check FP and AP modules for correct firmware

gautierhattenberger:


noether (approved):
Now the gvf_module works again in the FP. 

Thanks!

Issue #2501, opened 2020-03-16
Increasing index telemetry message


fvantienen:
Messages with increasing index which are used in for example both the Flight Recorder and Telemetry will both increase the index. This could lead to missing information in one of the targets for the telemetry message.
Example messages: `ESC`, `UART_ERRORS`, `I2C_ERRORS` etc..

I think this should be solved with a separate counter per telemetry. What do you think @gautierhattenberger ?

gautierhattenberger:
Are you talking about this index: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/mcu_periph/uart.c#L224 ?
How would you implement this ?

fvantienen:
yes that index. For this message it's not a problem for me but especially for the `ESC` message I have this problem.

I just came up with a possible solution. Would it be possible to get the amount of `processes` in the `telemetry` file in the generator? Such that each of them ahs an unique number which they can give to the callback function inside the `periodic_telemetry_send_.......` functions?
This all should be done in `gen_periodic.ml`

fvantienen:
Then if you have the total number then inside these functions you can make a list of numbers which you can count so it would change to:
```
uint8_t uart_nb_cnt[PERIODIC_PROCESS_NB];
...
uart_nb_cnt[process_id]++;
  if (uart_nb_cnt[process_id] == 9) {
    uart_nb_cnt[process_id] = 0;
  }
```

Would you have time to change the ocaml part of this @gautierhattenberger ? Then I can take a look at the airborne code.

gautierhattenberger:
see branch `gen_telemetry_process_number`

fvantienen:
Thank you!
One smalle request can you fix this line to a correct ID instead of `0`: https://github.com/paparazzi/paparazzi/commit/ea658b7fcaef6f45e935e10b1d64d2c0ebe03687#diff-8569302ad9668c0e2c3250673a443ff3L183

fvantienen:
I see what is going wrong there probably. The `p_id` should be defined above the for loop over the processes.

gautierhattenberger:
I fixed it, thanks. I guess we are lucky nobody was using this...

OpenUAS:
If all is sane what prevents  it take to honor this pull request? If it needs more testing, we can test and then confirm OK or not.

fvantienen:
I see two options to solve this problem:
[...]

PR #2582, opened 2020-09-16
[generator] generate the number of telemetry processes

gautierhattenberger:
also increment properly process id

related to #2501


fvantienen:
Tested and works. Will work on a fix for #2501

fvantienen (approved):

PR #2586, opened 2020-09-21
[GDB] New gdb name for ARM.

Fabien-B:
In Ubuntu 18.04 or lower, ARM stuff come from the gcc-arm-embedded PPA, and GDB is invoked by `arm-none-eabi-gdb`.
In Ubuntu 20.04, ARM stuff come from official ubuntu packages with packages `gcc-arm-none-eabi` and `gdb-multiarch`. GDB is now invoked with `gdb-multiarch`.

### DIFF
--- a/conf/Makefile.arm-embedded-toolchain
+++ b/conf/Makefile.arm-embedded-toolchain
@@ -38,8 +38,12 @@ CP    = $(PREFIX)-objcopy
 DMP   = $(PREFIX)-objdump
 NM    = $(PREFIX)-nm
 SIZE  = $(PREFIX)-size
-GDB   = $(PREFIX)-gdb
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
GDB is also defined in `conf/Makefile.arm-linux-toolchain` and `conf/Makefile.arm-embedded-toolchain`, better to update them as well

Fabien-B:
Done.
`arm-linux-gnueabi[hf]-gdb` was removed since at least ubuntu 14.04, replaced with gdb-multiarch.

gautierhattenberger (approved):

OpenUAS:
Not ideal solution IMHO see issues #2589

PR #2590, opened 2020-09-24
[pprzlink] fix bug in request message callback

gautierhattenberger:
- update pprzlink to fix bug for the number of parameters in callback https://github.com/paparazzi/pprzlink/pull/127
- also split pprzlink makefile rules for update and install (it allows to install the libraries with local changes)


noether (approved):
Cool, it works as expected here.

Thanks

Issue #2594, opened 2020-10-04
--force-yes deprecated in travis.yml environment variables


matteobarbera:
The use of --force-yes in https://github.com/paparazzi/paparazzi/blob/fab125357423147bb5962d6d174ae5d4e4feb1ac/.travis.yml#L14 is deprecated. The warning suggests using the --allow options as a safer alternative:

- `--allow-downgrades`
- `--allow-remove-essential`
- `--allow-unauthenticated`

I'm not sure what the original goal of the --force-yes command, if it was intended to force the install of the paparazzi packages, in which case the `--allow-unauthenticated` option should be used instead

PR #2595, opened 2020-10-04
replaced deprecated --force-yes command

matteobarbera:
Fix for Issue #2594

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ compiler:
 before_install:
   - sudo add-apt-repository ppa:paparazzi-uav/ppa -y
   - sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa -y
-  - sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list'
-  - wget http://packages.osrfoundation.org/gazebo.key -O - | sudo apt-key add -
   - sudo apt-get update -q
 install:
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
It seems that the problem is coming from gazebo package from ROS. The version for 18.04 bionic have been update the Oct 3, 2020, but the reference to dependency packages seems to be broken (`9.15.0-1~bionic` replaced by `9.15.0-2~bionic`). Maybe it will solve by itself when they fix the problem.
Or we could use the version from ubuntu which is much older (`9.0.0+dfsg5-3ubuntu1`). Is it problem ? Since it is only for compilation tests, maybe it is good enough.

dewagter:
Since only for compilation tests, the old version is OK if it compiles.

gautierhattenberger:
@matteobarbera can you try to update your PR without adding the ROS/Gazebo repository to the travis config ?

matteobarbera:
@gautierhattenberger do you mean removing the Gazebo repository from the pre-install commands? 

https://github.com/paparazzi/paparazzi/blob/fab125357423147bb5962d6d174ae5d4e4feb1ac/.travis.yml#L10-L11

It seems that this is a recurring issue, and is being addressed https://github.com/osrf/gazebo/issues/2744 and https://github.com/ignition-tooling/release-tools/issues/307

gautierhattenberger:
yes, remove the Gazebo repository, it should use the (old) version from ubuntu

matteobarbera:
The Travis job log shows that the different Gazebo packages were taken from the much older Ubuntu versions as you expected
`Get:101 http://archive.ubuntu.com/ubuntu bionic/universe amd64 gazebo9-common all 9.0.0+dfsg5-3ubuntu1 [34.4 MB]`

gautierhattenberger (approved):
it seems fine to use old gazebo for compilation tests

PR #2596, opened 2020-10-07
[board] make the Tawaki header file more generic

gautierhattenberger:
it is fixing some wrong pin and timer assignments between v1.0 and v1.1


gautierhattenberger:
tested and working

PR #2597, opened 2020-10-07
[nav] bio-inspired navigation module

gautierhattenberger:
based on the preliminary work of an internship
integrate a bio-inspired navigation based on the behavior of fish

require the message from https://github.com/paparazzi/pprzlink/pull/130

### DIFF
--- a/conf/flight_plans/ENAC/fish_outdoor.xml
+++ b/conf/flight_plans/ENAC/fish_outdoor.xml
@@ -1,0 +1,115 @@
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+
+<flight_plan alt="190" ground_alt="185" lat0="43.462976" lon0="1.2733345" max_dist_from_home="50" name="Demo fish nav Muret outdoor" security_height="0.3">
+  <header>
+#include "autopilot.h"
+#include "subsystems/datalink/datalink.h"
+  </header>
+  <waypoints>
+    <waypoint name="HOME" x="0.0" y="0.0"/>
+    <waypoint name="STDBY" x="1.0" y="0.0"/>
+    <waypoint name="TD" x="1.0" y="0"/>
+    <waypoint name="S1" x="16.4" y="60.3"/>
+    <waypoint name="S2" x="52.5" y="35.7"/>
+    <waypoint name="S3" x="56.0" y="-19.8"/>
+    <waypoint name="S4" x="29.9" y="-52.4"/>
+    <waypoint name="S5" x="-23.8" y="-50.1"/>
+    <waypoint name="S6" x="-57.6" y="-22.4"/>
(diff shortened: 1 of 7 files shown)


conf/flight_plans/ENAC/fish_outdoor.xml
```diff
@@ -0,0 +1,115 @@
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+
+<flight_plan alt="190" ground_alt="185" lat0="43.462976" lon0="1.2733345" max_dist_from_home="50" name="Demo fish nav Muret outdoor" security_height="0.3">
+  <header>
+#include "autopilot.h"
+#include "subsystems/datalink/datalink.h"
```
noether, line 6: Just curiosity, why is this necessary here? Because of the exceptions in the flightplan?

conf/modules/nav_fish.xml
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="nav_fish" dir="nav">
+  <doc>
+    <description>
+      Bio-inspired swarm navigation
```
noether, line 6: If available, a link to HAL or similar to the algorithm description?

sw/airborne/modules/nav/nav_fish.c
```diff
@@ -0,0 +1,505 @@
...
+      &nav_fish.f_ali,
+      &nav_fish.f_att);
+}
+
+/** sign function
+ *
+ * @param number x
+ * @return sign of x
+ */
+static float sign(float x)
```
noether, line 232: I do not know whether is necessary here. 

Sometimes, this sign function implementation produces chattering (jumping between -1 and 1 all the time) if we are close to zero (also I do not know whether you had any issue in your flights). I do not say you must implement it here since I do not know how easy is to implement hysteresis. Probably, it is something worth to implement in paparazzi_math xD.

noether (approved):
I have some small comments, but the PR looks clear, and it can be taken as a good template for other guys.

conf/flight_plans/ENAC/fish_outdoor.xml
gautierhattenberger: yes, for `datalink_time` variable

sw/airborne/modules/nav/nav_fish.c
gautierhattenberger: The function using this might change anyway, so I'm not sure if we will keep it in the end. But I agree that this could be integrated someday to the math lib, as well as an hysteresis function.

conf/modules/nav_fish.xml
gautierhattenberger: done

PR #2602, opened 2020-10-15
Some minor improvements and probably the last ones.

hendrixgr:
I changed a bit the OSD busy flag checking in order to make it more strict and robust
I also changed the board_matek.xml autoload module a bit in order to suppress the warnings during compile time when the spi settings between the airframe and the imu module xml file don't agree.
(changed the <define to <configure)
Other than that nothing worth to be mentioned.
IMG_20201010_105919
IMG_20201015_161600


hendrixgr:
If you need to change file names or directory names or anything else please do.

OpenUAS:
Nice, PR honoured would be good. FYI I have a v2 of the board on my desk with MPU9250 could flash and get telemetry tested over transparent_usb after AC changes. Have a list of possible improvements, but the sooner your nice code is in master the more option I have to fix it myself...

hendrixgr:
Version 2?
Any links?
Chris

max7456.c
```diff
@@ -0,0 +1,733 @@
+/*
+ * Copyright (C) 2013 Chris
```
gautierhattenberger, line 2: this file should not have been added I guess

gautierhattenberger (changes requested):
one extra file included by mistake I guess

max7456.c
hendrixgr: If i understood you well...
The max7456.c file is needed as it has been changed,  a new osd nvram busy method is implemented that checks the osd nvram busy flag every time the osd display is written.

gautierhattenberger: I can see the changes in the file `sw/airborne/modules/display/max7456.c`, but their is also a `max7456.c` at the root directory. Are they the same or not ?

hendrixgr: What?
I don't have the slightest idea how this file got there, DELETED.
Only the files in the display directory are neede.

gautierhattenberger (approved):

OpenUAS:
:1st_place_medal:

PR #2604, opened 2020-10-25
[fix] parse correctly settings with names

gautierhattenberger:
the appear with a ~ in the settings list which should be removed before testing for modules


gautierhattenberger:
@noether could you test if this is fixing your settings problem ?

noether:
> @noether could you test if this is fixing your settings problem ?

I did not have the chance to test it today. I will do it tomorrow in a moment.

noether:
Thanks :P, this fix works as expected.

noether (approved):

PR #2605, opened 2020-10-26
SPI SD Logger updates

tomvand:
Fixes/updates for the SPI SD logger:
- Allow arbitrary devices for download. *I have only tested this with usb serial!*
- Update sdlogger_download to pprzlink v2. Hardcoded as before, so pprzlink v1 is no longer supported...
- Reduce download chunk size to 64 bytes max. I had missing packets (not sure if buffers on the drone or on my pc are overflowing), fixed by limiting the max number of bytes sent at once.
- ~~Remove warning from settings_xml_parse.py because it was interfering with the download program. Not sure if this is used anywhere else...~~
- Updated documentation.

The fixes are slightly hacky in places because I did not have a lot of time to work on this. Let me know if something needs to be fixed/cleaned-up.

Tested on stm32f4 (trashcan/crazybeef4fr) with the bitcraze usd-deck with a transcend 8GB class 10 MicroSDHC and with usb_serial as sdlogger downlink device. Compilation tested with Lisa Hex airframe with uart as downlink device.

### DIFF
--- a/sw/lib/python/settings_xml_parse.py
+++ b/sw/lib/python/settings_xml_parse.py
@@ -6,6 +6,9 @@ import os
 import sys
 from lxml import etree
 
+def printerr(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
 # if PAPARAZZI_HOME not set, then assume the tree containing this
 # file is a reasonable substitute
 PPRZ_HOME = os.getenv("PAPARAZZI_HOME", os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)),
@@ -29,7 +32,7 @@ class PaparazziACSettings:
         # extract aircraft node from conf.xml file
         ac_node = conf_tree.xpath('/conf/aircraft[@ac_id=%i]' % ac_id)
         if (len(ac_node) != 1):
-            print("Aircraft ID %i not found." % ac_id)
+            printerr("Aircraft ID %i not found." % ac_id)
             sys.exit(1)
 
         # save AC name for reference
@@ -39,7 +42,7 @@ class PaparazziACSettings:
         settings_xml_path = os.path.join(paparazzi_home, 'var/aircrafts/' + self.name + '/settings.xml')
 
         if not os.path.isfile(settings_xml_path):
-            print("Could not find %s, did you build it?" % settings_xml_path)
+            printerr("Could not find %s, did you build it?" % settings_xml_path)
             sys.exit(1)
 
         index = 0 # keep track of index/id of setting starting at 0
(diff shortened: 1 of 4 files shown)


sw/logalizer/sdlogger_download.c
```diff
@@ -323,11 +330,12 @@ void write_command(float value)
 */
 void parse_single_byte(unsigned char byte)
 {
+//  printf("parser.state %d\n", parser.state);
```
gautierhattenberger, line 333: maybe better to remove all commented print ?

sw/lib/python/settings_xml_parse.py
```diff
@@ -99,8 +99,8 @@ def __init__(self, ac_id):
                 if 'values' in the_setting.attrib:
                     settings.values = the_setting.attrib['values'].split('|')
                     count = int((settings.max_value - settings.min_value + settings.step) / settings.step)
-                    if len(settings.values) != count:
-                        print("Warning: possibly wrong number of values (%i) for %s (expected %i)" % (len(settings.values), shortname, count))
+#                     if len(settings.values) != count:
```
gautierhattenberger, line 102: why removing this warning ?

gautierhattenberger (changes requested):

sw/logalizer/sdlogger_download.c
tomvand: I'd rather leave them in place, I actually found it helpful when testing and trying to understand the code. Ok?

sw/lib/python/settings_xml_parse.py
tomvand: sdlogger_download reads the stdout of this script and fails on reading this warning message.

tomvand:
Back from holiday, sorry it took a bit longer ;)

sw/logalizer/sdlogger_download.c
gautierhattenberger: I guess we can keep that. Maybe a compile flag to enable them all at once ?

tomvand: Will do

sw/lib/python/settings_xml_parse.py
tomvand: Fixed enum settings in module xml, restored warning message

gautierhattenberger: then, what about having a `verbose` (or `silence`) option for all the prints ?

tomvand: The warning print was actually correct, so I don't mind keeping it in place now that I fixed the error in the module xml. Alternatively, I would suggest printing warnings to stderr instead of adding a verbose flag?

gautierhattenberger: ok, I'm fine with a print to stderr

tomvand: Added printing to stderr, only tested with the sd download script.

PR #2606, opened 2020-10-26
Updated osd files and a new module

hendrixgr:
Hi.
The OSD files are much improved, magnetic course indication is added,, the home direction indication is changed and a new module for recovering the UAV is added.
I have heavily tested the OSD (with PAL camera, NTSC has 13 rows instead the 15 of PAL) and lightly the UAV recovery module
which works fine so far.


hendrixgr:
I live it at your discretion what to include and what not, i just want to contribute as much as possible.
Anyway the OSD is greatly improved and tested.

gautierhattenberger:
I'll have a look at this next week hopefully...

hendrixgr:
No problem at all!
Chris

gautierhattenberger:
is this closed and replaced by #2614  ?

hendrixgr:
Yes it is the same request, i deleted the previous one when i was trying to find why the build in Semaphore only fails.
It is still failing, something to do with adrone2 opticflow.

gautierhattenberger:
Okay, then I'm closing this one and only review the other one
The CI fail is not related, although I don't know why either, most likely related to a wrong dependency in the build somewhere. If I run by hand on Semaphore, it always works....

PR #2610, opened 2020-11-04
Added extern statement for oval_status in firmwares/fixedwing/nav.h

jburns11:
Added extern statement for oval_status in firmwares/fixedwing/nav.h so that this variable can be used in flight plans. If you try to compile with conf/flight_plans/cube.xml, you can see the undefined error for this variable.


jburns11:
Also, this is somewhat related to PR 2609, as it was discovered while attempting to use cube.xml

PR #2612, opened 2020-11-05
Better OSD display, more OSD format capabilities, a PCA9685 and a UAV recovery module.

hendrixgr:
Just deleted the old request and created a new one so it should be easier to see what i modified.
The OSD is much better, more powerful and more robust, a UAV recovery module is added which help in auto landing and parachute recovery and finally a PCA9685 16 additional PWM servo controller module is added.


hendrixgr:
I will re submit it.

PR #2613, opened 2020-11-05
Improved OSD and two more modules

hendrixgr:
Improved OSD and
1) a module for the PCA9685 16 pwm servo controller 
2) a module for the UAV recovery using a parachute.


hendrixgr:
Something is causing  the Semaphore build of adrone2 opticflow to fail.

PR #2616, opened 2020-11-09
[Server] modify ac_id field of CONFIG message for replay sessions

Fabien-B:
During a replay session, when requested with CONFIG_REQ, the server is not consistent over the "ac_id" field.
I propose to fix it to what I think is the most logical.

When the replay starts, the server send the message `ground NEW_AIRCRAFT replay1`.
When requested by the `AIRCRAFTS_REQ` message, the server also reply with `replay1`.
In all later messages, the server use `replay1` as ac_id as well.
The only exception is that, when requested with the `CONFIG_REQ` message, the server replay with `1` as ac_id.
I think it should reply with `replay1`.

The only code I found to be affected by this change is in pprz_connect, which assume the id is an integer. Removing the cast to integer fix it, and makes a previously erroneous docstring true.

This could break external code, however, considering that `ac_id` fields in the `ground` message class are all of type string, I think that such code should be fixed anyway.

### DIFF
--- a/sw/ground_segment/tmtc/server.ml
+++ b/sw/ground_segment/tmtc/server.ml
@@ -749,10 +749,9 @@ let listen_intruders = fun log ->
   ignore(Ground_Pprz.message_bind "INTRUDER" (update_intruder log))
 
 let send_config = fun http _asker args ->
-  let real_id' = PprzLink.string_assoc "ac_id" args in
-  let ac_id' = real_id' in
+  let real_id = PprzLink.string_assoc "ac_id" args in
   try
-    let _is_replayed, ac_id, root_dir, conf_xml = replayed real_id' in
+    let _is_replayed, ac_id, root_dir, conf_xml = replayed real_id in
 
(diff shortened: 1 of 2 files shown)


sw/ground_segment/tmtc/server.ml
```diff
@@ -749,9 +749,10 @@ let listen_intruders = fun log ->
   ignore(Ground_Pprz.message_bind "INTRUDER" (update_intruder log))
 
 let send_config = fun http _asker args ->
-  let ac_id' = PprzLink.string_assoc "ac_id" args in
+  let real_id' = PprzLink.string_assoc "ac_id" args in
+  let ac_id' = real_id' in
```
gautierhattenberger, line 753: Use `real_id` and `ac_id`, no need for the `'` anymore.
Just replace the `ac_id'` at the end by `real_id`.

gautierhattenberger (changes requested):

sw/ground_segment/tmtc/server.ml
Fabien-B: Ok, I did it.

gautierhattenberger (approved):

Issue #1772, opened 2016-06-27
[Feature Request] Run opticflow on multiple cameras symulteaneously


kirkscheper:
With the current implementation only one opticflow module can run at a time on one camera. I am not sure how popular something like this would be but I can imagine that it may be interesting to be able to run opticflow on multiple cameras (eg bottom camera and forward looking camera) simultaneously.

flixr:
You can't really run multiple instances of the same module.
Instead one would have to do something like:
- create an array of the opticflow struct to hold the parameters for each cam
- somehow deal with the sets of configuration options (defines) for each one
- extend the module to register a callback for each camera (and probably create multiple threads for processing)

matteobarbera:
I'm trying to tackle this issue, while trying to keep duplicate code to a minimum. My current idea, however, requires changes to `cv.h` and `cv.c`:

- Add `uint8_t id` parameter to the `video_listener` struct
- Add `uint8_t id` as function parameters for the `cv_add_to_device`* functions
- Pass `listener->id` to the cv linked function in `cv_async_thread` and `cv_run_device` https://github.com/paparazzi/paparazzi/blob/5887bf9746cb52ca1d46631b8d7435dc7ce32ffb/sw/airborne/modules/computer_vision/cv.c#L166 https://github.com/paparazzi/paparazzi/blob/5887bf9746cb52ca1d46631b8d7435dc7ce32ffb/sw/airborne/modules/computer_vision/cv.c#L201

This way any computer vision function could take advantage of the id and be able to use it as an index for e.g. an array of opticflow structs. All cv files that need to deal with multiple cameras currently have to create extra functions to differentiate between the cameras, this solution could reduce unnecessary code and make many of these files easier to maintain (this isn't currently a huge issue).

Lots of files would have to be changed which might cause problems for a lot of people, but I think this sort of change is better than trying to work around the `cv.c` limitations and having to create lots of extra unnecessary code.

I cannot think of a way to avoid writing a lot of duplicate code to configure multiple cameras, as this is done from XML defines (and there are a lot of them). This limits the max number of cameras the opticflow module can use to 2 rather than an arbitrary number of cameras, which would be ideal.

PR #2618, opened 2020-11-10
[computer vision] Opticflow multiple cameras

matteobarbera:
Proposal for #1772 

~~The files currently compile, I plan to test the opticflow module on a Bebop with both cameras before finalizing the pull request.~~ 
The modified `cv_opticflow` module is able to run the computer vision functions on both front and bottom camera simultaneously. I tested it on a Bebop 1, I added the required defines to the `bebop_opticflow_indoor.xml` airframe to showcase the module usage, but perhaps making a whole new airframe might be better.

I have not found a clever solution on how to avoid writing a lot of duplicate code for the defines in `cv_opticflow.xml` (and as a consequence in the initialization of the multiple `opticflow_t` structs). Originally I intended for the module to be able to run on an arbitrary number of cameras, but due to how Paparazzi handles module defines I had to hardcode it to front and bottom cameras only in a lot of places.

There are a lot of changes that have been made already, the next step would be to change the opticflow ABI message, since (from my understanding) it's not currently possible to distinguish if the message contains results from the front or the bottom camera.


dewagter:
Thank you for this important update.
I understand that functions might need extra parameters, but doesn't ```cv_add_to_device``` already know the camera id as the first parameter?

matteobarbera:
Perhaps me calling it camera id is not the clearest name. In order to run the cv functions and know which opticflow struct to use, where to write the result etc. some form of identification needs to be passed to the `video_listener` `cv_function`. I chose to pass a `uint8_t` which is then used as an index in `opticflow_module.c` to know which struct in the opticflow_t array and opticflow_result_t array to use.

Originally I didn't want to hardcode an id for front/bottom camera (to support an arbitrary number of devices), so technically when `cv_add_to_device` is called there could be a hardcoded "id" or "index" that is then used by the `cv_function` by looking at the device name, without the need of an additional function parameter in `cv_add_to_device`. But I thought that having the "id" be set in the same file where it is then used was more transparent. 

This index/id being so free to choose could cause some confusion on how it's supposed to be used, so if linking the camera name to it is a better option I can change it. It's not the cleanest solution at the moment, as there are no checks/safeguards on this extra function parameter, so I am very much open to suggestions.

dewagter (approved):
Thanks for the extra information.

tomvand:
Hi Matteo, I'm currently going through your pull request. Why did you choose to set the `id` in cv_add_to_device instead of hardcoding it in the `struct video_config_t *device` (e.g. a sender_id for the `front_camera` and `bottom_camera` structs) and passing that value to the `cv_function`? To make it easier to allocate and find the correct optic flow struct? I'm fine with the solution, just wondering ;).

conf/modules/cv_opticflow.xml
```diff
@@ -57,51 +58,130 @@
...
+
+      <!-- Video device parameters for camera 2 -->
+      <define name="CAMERA2" value="bottom_camera|front_camera" description="The V4L2 camera device that is used for the calculations"/>
+
+      <!-- Camera 2 parameters -->
+      <define name="DEROTATION_CORRECTION_FACTOR_X_CAMERA2" value="1.0" description="Correction factor for derotation (in x direction), estimated from a fit between the gyro's rates and the resulting flow (caused by the camera not being exactly in the middle (Defaults are from an ARDrone 2)"/>
+      <define name="DEROTATION_CORRECTION_FACTOR_Y_CAMERA2" value="1.0" description="Correction factor for derotation (in y direction), estimated from a fit between the gyro's rates and the resulting flow (caused by the camera not being exactly in the middle (Defaults are from an ARDrone 2)"/>
+      <configure name="BODY_TO_CAM_PHI" value="0" description="Rotation from body frame to camera frame around x axis"/>
+      <configure name="BODY_TO_CAM_THETA" value="0" description="Rotation from body frame to camera frame around y axis"/>
+      <configure name="BODY_TO_CAM_PSI" value="0" description="Rotation from body frame to camera frame around z axis"/>
```
tomvand, line 71: `_CAMERA2`?

sw/airborne/modules/computer_vision/lib/vision/act_fast.c
```diff
@@ -22,16 +22,18 @@ All rights reserved.
...
 #include "image.h"
 #include "../../opticflow/opticflow_calculator.h"
 
+// ACT-FAST agents arrays
 // equal to the maximal number of corners defined by fast9_rsize in opticflow_calculator.c
+// TODO Currently hardcoded to two cameras
 #define MAX_AGENTS FAST9_MAX_CORNERS
-struct agent_t agents[MAX_AGENTS];
+struct agent_t agents_arr[MAX_AGENTS];
+struct agent_t *agents[2] = {agents_arr, agents_arr + FAST9_MAX_CORNERS / 2};
```
tomvand, line 36: Would it make more sense to define the max number of agents per camera? That way, adding a second camera later does not change tracking performance on the first. I.e.:
```
struct agent_t agents_arr[2*MAX_AGENTS];
struct agent_t *agents[2] = {agents_arr, agents_arr + MAX_AGENTS};
```

Also, I suggest using MAX_AGENTS/2 over FAST9_MAX_CORNERS/2 for readability.

Possible shorthand:
```
struct agent_t agents[2][MAX_AGENTS];
```

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -70,15 +81,18 @@ struct image_t *opticflow_module_calc(struct image_t *img);     ///< The main op
...
+      pprz_msg_send_OPTIC_FLOW_EST(trans, dev, AC_ID,
+                                   &opticflow_result[idx_camera].fps, &opticflow_result[idx_camera].corner_cnt,
+                                   &opticflow_result[idx_camera].tracked_cnt, &opticflow_result[idx_camera].flow_x,
+                                   &opticflow_result[idx_camera].flow_y, &opticflow_result[idx_camera].flow_der_x,
+                                   &opticflow_result[idx_camera].flow_der_y, &opticflow_result[idx_camera].vel_body.x,
+                                   &opticflow_result[idx_camera].vel_body.y, &opticflow_result[idx_camera].vel_body.z,
+                                   &opticflow_result[idx_camera].div_size,
+                                   &opticflow_result[idx_camera].surface_roughness,
+                                   &opticflow_result[idx_camera].divergence); // TODO: no noise measurement here...
+    }
```
tomvand, line 95: Could you expand this message to add the camera id to the telemetry?

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -109,27 +128,29 @@ void opticflow_module_run(void)
...
+                             opticflow_result[idx_camera].div_size);
+      //TODO Find an appropriate quality measure for the noise model in the state filter, for now it is tracked_cnt
+      if (opticflow_result[idx_camera].noise_measurement < 0.8) {
+        AbiSendMsgVELOCITY_ESTIMATE(VEL_OPTICFLOW_ID, now_ts,
+                                    opticflow_result[idx_camera].vel_body.x,
+                                    opticflow_result[idx_camera].vel_body.y,
+                                    0.0f, //opticflow_result.vel_body.z,
+                                    opticflow_result[idx_camera].noise_measurement,
+                                    opticflow_result[idx_camera].noise_measurement,
+                                    -1.0f //opticflow_result.noise_measurement // negative value disables filter updates with OF-based vertical velocity.
```
tomvand, line 149: Similarly, could you add a camera id to the ABI messages? Or use the sender_id for that?

tomvand (commented):
See comments, but looks good to me

sw/airborne/modules/computer_vision/opticflow_module.c
matteobarbera: There are many files that call these ABI messages, especially for `AbiSendMsgVELOCITY_ESTIMATE()`. Is it okay for me to change all of these to add the extra field? Or would it be better to do this in a separate pull request?

matteobarbera:
> Hi Matteo, I'm currently going through your pull request. Why did you choose to set the `id` in cv_add_to_device instead of hardcoding it in the `struct video_config_t *device` (e.g. a sender_id for the `front_camera` and `bottom_camera` structs) and passing that value to the `cv_function`?

I thought it would be better to avoid hardcoding `id` since an _if_ statement would also then have to be used when you need to check the id every time you need to use it to access/write to one of the structs in the array. Similar to why I didn't want to use the camera name as the identification method.

Another option that I thought of was to create a static variable in cv.c that would start at 0 and increment every time a new device is added, so no extra arguments would need to be passed to `cv_add_to_device`. However, this is a lot less transparent and could cause problems since you then have to check if the id of the first opticflow camera is indeed 0 so it can be correctly used as an index for the struct arrays.

conf/modules/cv_opticflow.xml
matteobarbera: Good catch, I was actually using the same derotation matrix for both cameras which should now also be fixed

sw/airborne/modules/computer_vision/lib/vision/act_fast.c
matteobarbera: I implemented your suggestion

sw/airborne/modules/computer_vision/opticflow_module.c
tomvand: Ok, on second thought it might not be necessary for VELOCITY_ESTIMATE, since that is already in body frame anyway.

For the OPTICAL_FLOW message I still think it could be useful since these results are in the camera frame (I think?); and wouldn't you need this for the stop-in-front-of-wall demo anyway? Perhaps a nice way of doing this without too much hassle, is to reserve a short range of sender ids in `subsystems/abi_sender_ids.h` and then calling AbiSendMsgOPTICAL_FLOW with `FLOW_OPTICFLOW_ID + idx_camera` as sender id? That way other modules can directly subscribe to the correct camera, and you won't need to change the function calls anywhere.

tomvand:
Agreed, let's stick with your solution then :)

sw/airborne/modules/computer_vision/opticflow_module.c
matteobarbera: I don't understand yet how the sender IDs work, for now I've just included the camera_id in the opticflow message, but I will try to implement what you suggested because it might be easier if you can directly subscribe to a single camera instead of having to check for the message you're interested in every time

sw/airborne/modules/computer_vision/opticflow_module.c
matteobarbera: In order to change the periodic telemetry I added a message.xml and recompiled paparazzi to make sure the generated files were also updated with the new message field. I don't see a messages.xml in master, how can I commit these changes without causing a build failure in semaphoreci?

sw/airborne/modules/computer_vision/opticflow_module.c
matteobarbera: The latest commit should have added what you suggested for the OPTICFLOW message

sw/airborne/modules/computer_vision/opticflow_module.c
tomvand: That's correct, there is no messages.xml in master ;). To include your new message:
1. Make a new branch from [pprzlink](https://github.com/paparazzi/pprzlink)
2. Add your new message definition to [message_definitions/v1.0/messages.xml](https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml)
3. Make an additional pull request at the pprzlink repo (and post the PR number here as a reminder ;) )
Once the pprzlink pull request has been merged, you can update the pprzlink submodule in *this* pull request to include your message into master. HTH

sw/airborne/subsystems/abi_sender_ids.h
```diff
@@ -355,8 +355,16 @@
 #define FLOW_OPTICFLOW_ID 1
 #endif
 
+#ifndef FLOW_OPTICFLOW_CAM1_ID
+#define FLOW_OPTICFLOW_CAM1_ID 1
+#endif
+
+#ifndef FLOW_OPTICFLOW_CAM1_ID
+#define FLOW_OPTICFLOW_CAM1_ID 2
+#endif
```
tomvand, line 364:
```suggestion
#ifndef FLOW_OPTICFLOW_CAM2_ID
#define FLOW_OPTICFLOW_CAM2_ID 2
#endif
```

sw/airborne/modules/computer_vision/opticflow_module.c
```diff
@@ -109,27 +129,30 @@ void opticflow_module_run(void)
...
+                             opticflow_result[idx_camera].flow_x,
+                             opticflow_result[idx_camera].flow_y,
+                             opticflow_result[idx_camera].flow_der_x,
+                             opticflow_result[idx_camera].flow_der_y,
+                             opticflow_result[idx_camera].noise_measurement,
+                             opticflow_result[idx_camera].div_size,
+                             opticflow_result[idx_camera].camera_id);
+      //TODO Find an appropriate quality measure for the noise model in the state filter, for now it is tracked_cnt
+      if (opticflow_result[idx_camera].noise_measurement < 0.8) {
+        AbiSendMsgVELOCITY_ESTIMATE(VEL_OPTICFLOW_ID, now_ts,
```
tomvand, line 145: Can I change my mind on adding the camera id to the VELOCITY_ESTIMATE ABI message? :$
In hindsight, I think it might also be valuable for ABI subscribers to filter out velocity estimates from cameras that might be poorly placed for velocity measurements (e.g. front-facing cameras).

```suggestion
        AbiSendMsgVELOCITY_ESTIMATE(VEL_OPTICFLOW_ID + idx_camera, now_ts,
```
plus reserving ids in `abi_sender_ids.h`, similarly to what you have done for the OPTICAL_FLOW message ;)

tomvand (commented):
Only two minor comments, pprzlink, and then it's ready to be merged :)

tomvand (dismissed):
Looks good to me! Could someone (@gautierhattenberger  ?) take a short look at the pprzlink changes before merging? Cheers

sw/airborne/modules/computer_vision/lib/vision/act_fast.c
```diff
@@ -22,16 +22,17 @@ All rights reserved.
...
 #include "math.h"
 #include "image.h"
 #include "../../opticflow/opticflow_calculator.h"
 
+// ACT-FAST agents arrays
 // equal to the maximal number of corners defined by fast9_rsize in opticflow_calculator.c
+// TODO Currently hardcoded to two cameras
 #define MAX_AGENTS FAST9_MAX_CORNERS
-struct agent_t agents[MAX_AGENTS];
+struct agent_t agents[2][MAX_AGENTS];
```
gautierhattenberger, line 35: Maybe make the number of cameras configurable ?

sw/airborne/modules/computer_vision/bebop_ae_awb.c
```diff
@@ -352,5 +352,5 @@ void bebop_ae_awb_init(void)
   awb_set.gain_scheduling_tolerance = BEBOP_AWB_GAIN_SCHEDULING_TOLERANCE;
   awb_set.gain_scheduling_step      = BEBOP_AWB_GAIN_SCHEDULING_STEP;
 
-  cv_add_to_device_async(&front_camera, update_ae_awb, BEBOP_AE_AWB_NICE, 0);
+  cv_add_to_device_async(&front_camera, update_ae_awb, BEBOP_AE_AWB_NICE, 0, 0);
```
gautierhattenberger, line 355: why not using the `camera_id ` ? and if not using it, it should be marked as unused attribute to prevent warning

matteobarbera: Do you mean the `update_ae_awb` function not using the camera_id or in the line where `cv_add_to_device_async` is called? 

If it's the former, I can change the function signature to `static struct image_t *update_ae_awb(struct image_t *img, uint8_t __attribute__((unused)) camera_id)` to make it clearer that the field is only there because of how `cv_run_device` behaves (it now needs to pass an additional field on top of img, camera_id, to the callback of `cv_function`). 

In that case, I will then also change all function signatures where I had added this parameter originally but not used it.

sw/airborne/modules/computer_vision/lib/vision/act_fast.c
matteobarbera: I changed it so it checks if the second camera has been defined

gautierhattenberger (approved):

PR #2622, opened 2020-11-23
Demo vto dshot fix

alex31:
Problem : 
* on a 4 engines configuration, one engine fail to start.

preliminary fix :
put stm32_dma and esc_dshot to the same level than https://github.com/alex31/chibios_enac_various_common.git master.

This will or will not fix the problem, but if the problem remains, we can search from a known state.


gautierhattenberger:
replaced by #2625

Issue #2607, opened 2020-10-29
DShot Protocol debugging


aestebanluengo:
@OpenUAS , @EwoudSmeur and I are trying to make the Dshot Protocol work on Ewoud's Cyclone hybrid MAV which mounts the Chimera board with the code created by @alex31 and maintained by @gautierhattenberger . The aim is to log the RPM of both motors for system identification purposes.

The branch in which we are working is the following: [https://github.com/aestebanluengo/paparazzi/tree/cyclone_sid](https://github.com/aestebanluengo/paparazzi/tree/cyclone_sid)

The RPM data is logged to the SD card using the following script: [https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/sw/airborne/firmwares/rotorcraft/autopilot_static.c](https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/sw/airborne/firmwares/rotorcraft/autopilot_static.c)
Check lines 150 and 151:
```
sd_buffer_i[11] = esc_telem->rpm;
sd_buffer_i[12] = index_last_telemetry;
```
`esc_telem->rpm` is the electrical RPM value and `index_last_telemetry` is the motor from which the RPM vale is received. As specified in the airframe file ([https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/conf/airframes/tudelft/cyfoam.xml](https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/conf/airframes/tudelft/cyfoam.xml))
```
<servos driver="DShot">
    <servo name="RM" no="0" min="0" neutral="100" max="1900"/>
    <servo name="LM" no="1" min="0" neutral="100" max="1900"/>
    <!--<servo name="SERVO_TEST" no="1" min="0" neutral="0" max="1900"/>-->
</servos>
```
index = 0 corresponds to the right motor and index = 1 corresponds to the left one.

Also, according to the previous logger indexes, the 12th and 13th columns of the Matlab matrices presented down below correspond to the `esc_telem->rpm` and `index_last_telemetry` values.

The function in which the motor index is specified to send the RPM value corresponds to `dshotSendFrame` in [https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/sw/airborne/arch/chibios/modules/actuators/esc_dshot.c](https://github.com/aestebanluengo/paparazzi/blob/cyclone_sid/sw/airborne/arch/chibios/modules/actuators/esc_dshot.c):

[...]

PR #2625, opened 2020-11-25
[fix] fix DSHOT driver and corresponding boards

gautierhattenberger:
Some bugs were introduced with the latest modifications on the DSHOT driver

Probably not related to  #2607, but it is worth testing


amtcvx (approved):

OpenUAS:
Thx for fixing, awesome! BTW will ask @aestebanluengo to retest #2607,  and report if these fixes also coincidentally fixed his issue with RPM feedback

PR #2628, opened 2020-12-10
[doc] Adds ABI page to sphinx doc.

Fabien-B:
This doc page mainly copy [http://wiki.paparazziuav.org/wiki/ABI](http://wiki.paparazziuav.org/wiki/ABI), with few details added, and some "implementation details" omitted (the source code).
The rendered page can be seen here:
[https://fabien-b.github.io/paparazzi/developer_guide/abi.html](https://fabien-b.github.io/paparazzi/developer_guide/abi.html)


doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+
+.. note:: This middleware was named *ABI* for *AirBorne Ivy*
+
+General Idea
+------------
+
+* Give an easy way to allow software components to exchange data with minimum delay nor execution overhead.
+* The components don't need to know each other, they only need the message format.
+* Each subscriber set a callback function that will be called when new data are sent.
+
```
gautierhattenberger, line 17: maybe add a note that the callbacks are not thread-safe (when applicable)

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Message definition
+------------------
+
+The messages are described in ``conf/abi.xml`` analoguous to other messages in paparazzi. Name are unique, IDs starts from 0. "type" field should be a real C type. Field name might be useless.
```
gautierhattenberger, line 22: maybe not needed to tell that field name is not required, as it is probably cleaner to define it ?

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+      <field name="a" type="float"/>
+      <field name="b" type="struct bla"/>
+     </message>
+     ...
+    </class>
+
+Airborne code for subscriber
+----------------------------
+
+Include header and declare an ``abi_event`` as a global variable. Write the callback function with the proper prototype, matching the message.
```
gautierhattenberger, line 37: specify that `abi_event ` variables can also be `static` (but not `const`)

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+The first parameter is the sender ID you want to receive the message from.
+
+* ``ABI_BROADCAST`` is used to receive messages from all senders.
+* ``ABI_DISABLE`` disable the callback (it will never be called).
+* Senders IDs can be found in the file ``sw/airborne/subsystems/abi_sender_ids.h``
+
+The second parameter is a pointer to the global ``abi_event`` you declared. This variable **can't** be reused for another bind. You must declare one abi_event per bind.
+
+The last parameter is your callback.
+
```
gautierhattenberger, line 65: add a note to tell that the "module creator" can help to prepare bindings when creating a new module ?

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+    The values 0 and 255 are reserved for ``ABI_DISABLE`` and ``ABI_BROADCAST`` and thus shall not be used.
+    
+    You must also avoid using an ID already used to send the same message type.
+
+
+Code generation
+---------------
+
+The generated code will be in ``var/include/abi_messages.h`` and include some structure definition from ``sw/airborne/subsystems/abi_common.h`` (``sw/airborne/subsystems/abi.h`` is a convenience header that only includes ``var/include/abi_messages.h``). Bind and Send functions are generated, as well as callback type definition. A linked list is used to store the binded callbacks for each message. The head of the list is in an array to allow a fast access. 
+
```
gautierhattenberger, line 97: for interested people, also give the link to the generator in `sw/tools/generators`

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,99 @@
...
+
+
```
gautierhattenberger, line 99: Also add a "in depth" section (or subsections of code gen) with the generated code corresponding to the example and how to enable it in a custom code (http://wiki.paparazziuav.org/wiki/ABI#Airborne_code_for_main_AP) ?

gautierhattenberger (changes requested):

doc/sphinx/source/developer_guide/abi.rst
Fabien-B: that also what I thought

doc/sphinx/source/developer_guide/abi.rst
Fabien-B: Done, but is this not the default behavior in C?

doc/sphinx/source/developer_guide/abi.rst
Fabien-B: Done

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,199 @@
...
+Include header and declare an ``abi_event`` as a global ``static`` variable (but **not** ``const``). Write the callback function with the proper prototype, matching the message.
+
+.. code-block:: C
+
+    #include "subsystems/abi.h"
+
+    static abi_event ev;
+
+    void data_cb(uint8_t sender_id, const float * a, const struct bla * b) {
```
gautierhattenberger, line 47: I checked in the code, there is a mistake here (and in the wiki as well). The fields are not necessarily pointers, unless typed like that in the XML file (this is a difference from pprzlink where all fields are turned to pointers). And they are not `const` either (all that was the original designed, but changed later).
So the callback should be:
```
void data_cb(uint8_t sender_id, float a, struct bla b)
```
Maybe it is worth adding a extra `c` field showing a pointer to something
```
<field name="c" type="uint32_t */>
void data_cb(uint8_t sender_id, float a, struct bla b, uint32_t * c)
```

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,199 @@
...
+
+    // Bind and Send for each messages
+    static inline void AbiBindMsgDATA(uint8_t sender_id, abi_event * ev, abi_callbackDATA cb) {
+      if (abi_queues[ABI_BARO_ABS_ID] == ev) return;
+      ev->id = sender_id;
+      ev->cb = (abi_callback)cb;
+      ABI_PREPEND(abi_queues[ABI_BARO_ABS_ID],ev);
+    }
+
+    static inline void AbiSendMsgDATA(uint8_t sender_id, float a, struct bla b) {
```
gautierhattenberger, line 176: this was correct :)
update if you add the `c` field.

gautierhattenberger (changes requested):

doc/sphinx/source/developer_guide/abi.rst
Fabien-B: Rhaa, I noticed it this morning, I corrected it, but this one flew under the radar :smile: 
For the pointer example, I imagine users will see the examples of other messages using pointers, and in the wort case, the compiler error will be clear enough ?

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,201 @@
...
+* The components don't need to know each other, they only need the message format.
+* Each subscriber set a callback function that will be called when new data are sent.
+
+.. warning:: This middleware is **NOT** thread safe. Threads are not widely used in Paparazzi but remember this if you use them.
+
+
+Message definition
+------------------
+
+The messages are described in ``conf/abi.xml`` analoguous to other messages in paparazzi. Name are unique, IDs starts from 0. "type" field should be a real C type.
```
gautierhattenberger, line 24:
```suggestion
The messages are described in ``conf/abi.xml`` analoguous to other messages in paparazzi. Name are unique, IDs starts from 0. "type" field should be a vald C type or pointer.
```

doc/sphinx/source/developer_guide/abi.rst
Fabien-B: Well, you are right, an example is better if we you to have a good doc. This is done.

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -0,0 +1,201 @@
...
+      ABI_PREPEND(abi_queues[ABI_BARO_ABS_ID],ev);
+    }
+
+    static inline void AbiSendMsgDATA(uint8_t sender_id, float a, struct bla b, struct foo * c) {
+     // Call all callback functions
+     abi_event* e;
+     ABI_FOREACH(abi_queues[ABI_DATA_ID],e) {
+      if (e->id == ABI_BROADCAST || e->id == sender_id) { // call function only if selected source or broadcast address
+       abi_callbackDATA cb = (abi_callbackDATA)(e->cb); // C black magic
+       cb(sender_id, a, b);
```
gautierhattenberger, line 184:
```suggestion
       cb(sender_id, a, b, c);
```

PR #2630, opened 2020-12-11
[module] fix formation flight module and load from flight plan

gautierhattenberger:
small bug fix and typo for formation flight module


fvantienen:
Seems fine to me 👍

PR #2632, opened 2020-12-14
SPI SD Logger: download over other serial link

tomvand:
With this pull request, the spi_sd_logger can download log files over a different serial link than the default downlink device. To do so, the module will also listen to SETTING messages on this download port.

### DIFF
--- a/sw/airborne/modules/loggers/sdlogger_spi_direct.c
+++ b/sw/airborne/modules/loggers/sdlogger_spi_direct.c
@@ -35,6 +35,8 @@
 #include "subsystems/datalink/telemetry.h"
 #include "led.h"
 
+#include <stdbool.h>
+
 #if SDLOGGER_ON_ARM
 #include "autopilot.h"
 #endif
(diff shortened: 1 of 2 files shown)


sw/airborne/modules/loggers/sdlogger_spi_direct.c
```diff
@@ -109,6 +161,10 @@ void sdlogger_spi_direct_init(void)
  */
 void sdlogger_spi_direct_periodic(void)
```
gautierhattenberger, line 162: While looking at the frequency of this function in the module file, it is hardcoded to 512 Hz, which is probably not wise. If it is suppose to run at the main frequency, just don't specify the `freq` attribute as it may not be the desired behavior when running at a different frequency.
Otherwise, it looks fine.

tomvand: I agree, but I'm having some trouble getting the logger to work at all today so I'll take another look at this after the christmas break. Happy holidays! ;)

tomvand: I have retested it this morning, everything works fine when freq="512" is present, but after removing it the datalink becomes unstable and the download program has trouble establishing a connection. I am not sure why this happens, but suspect an sd timing issue. I'd rather leave the freq as-is for now, as I don't have time to debug this further.

PR #2633, opened 2020-12-16
Fix pwm index

gautierhattenberger:
only concerns tawaki and nucleo144 using latest boardGen tool.


gautierhattenberger:
tested and working

PR #2636, opened 2020-12-29
Add user settable filtered more consistent airspeed readings

OpenUAS:
Behavior same if new defines are not added except AIRSPEED_ETS_NBSAMPLES_AVRG set to 20 instead of 10, the old value was already not such a good default to begin with


OpenUAS:
Needs validation

PR #2639, opened 2021-01-05
Addition of driver for MATEKSYS_PMW3901 sensor

pcampolucci:
Communication through MSP protocol was not yet enabled on paparazzi. This module therefore parses the message and sends flow + distance measurements. 

LINK sensor: http://www.mateksys.com/?portfolio=3901-l0x#tab-id-2

EDIT: sent pull request for message code in ppzlink https://github.com/paparazzi/pprzlink/pull/137


conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+    <!-- Enable UART and set baudrate -->
+    <define name="USE_$(MATEKSYS_3901_L0X_PORT_UPPER)"/>
+        <!--<define name="USE_$(MATEKSYS_3901_L0X_PORT)_TX" value="FALSE"/>-->
+    <define name="$(MATEKSYS_3901_L0X_PORT_UPPER)_BAUD" value="$(MATEKSYS_3901_L0X_BAUD)"/>
+    <define name="MATEKSYS_3901_L0X_PORT" value="$(MATEKSYS_3901_L0X_PORT_LOWER)"/>
+
+    <configure name="MATEKSYS_3901_L0X_MOTION_THRES" default="120"/>
+    <configure name="MATEKSYS_3901_L0X_DISTANCE_THRES" default="200"/>
+    <configure name="USE_MATEKSYS_3901_L0X_AGL" default="0"/>
+    <configure name="MATEKSYS_3901_L0X_COMPENSATE_ROTATION" default="0"/>
```
tomvand, line 38: Do these need to be `configure`'s? If this is for the default values, I'd suggest the following instead:
1. In the C file, use the `#ifndef X; #define X; #endif` pattern to set the default values
2. In this xml file, add the defines inside the description element

conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+  <doc>
+    <description>
+      Transfers flow X and Y and range data of the Mateksys3901-l0X sensor to the autopilot
+      This by the default serial connection not by connecting the debug pins on ther sensor
+      The Data is in MSP compatible output. If a MSP stack is availabe in the AP one could use that
+      In ths ase we just handle the byres comming from the sensor in MSP compatible handler for this sensor
+    </description>
+    <configure name="MATEKSYS_3901_L0X_PORT" value="UART1" description="Set to which UART the sensor serial stream pin is connected to"/>
+    <configure name="MATEKSYS_3901_L0X_BAUD" value="115200" description="Sets the baudrate of the UART"/>
+
```
tomvand, line 13: Please add a description for the other defines.

conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+  <!-- <periodic fun="mateksys3901l0x_downlink()"/> --> <!-- to get messages TODO-->
+  <makefile>
+
+    <!-- Configure default UART port and baudrate -->
+    <configure name="MATEKSYS_3901_L0X_PORT" default="UART3" case="upper|lower"/>
+    <configure name="MATEKSYS_3901_L0X_BAUD" default="B115200"/>
+
+    <!-- Enable UART and set baudrate -->
+    <define name="USE_$(MATEKSYS_3901_L0X_PORT_UPPER)"/>
+        <!--<define name="USE_$(MATEKSYS_3901_L0X_PORT)_TX" value="FALSE"/>-->
```
tomvand, line 31: Could you add a comment on when this define should be set to false, or mention this in the description of this define?

conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+    <define name="USE_MATEKSYS_3901_L0X_AGL" value="$(USE_MATEKSYS_3901_L0X_AGL)"/>
+    <define name="MATEKSYS_3901_L0X_COMPENSATE_ROTATION" value="$(MATEKSYS_3901_L0X_COMPENSATE_ROTATION)"/>
+
+    <file name="mateksys_3901_l0x.c" dir="modules/optical_flow"/>
+    <!--<file name="pprz_transport.c" dir="pprzlink/src"/>-->
+  </makefile>
+  <makefile target="nps|sim">
+    <define name="USE_SONAR" value="1"/><!-- in NPS use a virtual sonar to simulate lidar measurements -->
+  </makefile>
+</module>
```
tomvand, line 50: Missing newline

sw/airborne/modules/optical_flow/mateksys_3901_l0x.c
```diff
@@ -0,0 +1,316 @@
...
+ */
+void mateksys3901l0x_init(void)
+{
+  mateksys3901l0x.device = &((MATEKSYS_3901_L0X_PORT).device);
+  mateksys3901l0x.parse_crc = 0;
+	mateksys3901l0x.motion_quality = 0;                                             //TODO :is this a good choice?
+  mateksys3901l0x.motionX = 0;                                                    //TODO :is this a good choice?
+  mateksys3901l0x.motionY = 0;                                                    //TODO :is this a good choice?
+  mateksys3901l0x.distancemm_quality = 0;
+	mateksys3901l0x.distancemm = -1;                                                //TODO :is this a good choice?
```
tomvand, line 77: Are the TODO comments still up to date?

conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+    
+  </doc>
+
+
+  <header>
+    <file name="mateksys_3901_l0x.h" dir="modules/optical_flow"/>
+  </header>
+  <init fun="mateksys3901l0x_init()"/>
+  <event fun="mateksys3901l0x_event()"/>
+  <!-- <periodic fun="mateksys3901l0x_downlink()"/> --> <!-- to get messages TODO-->
```
tomvand, line 22: Is this comment still required?

conf/modules/optical_flow_mateksys_3901_l0x.xml
```diff
@@ -0,0 +1,50 @@
...
+    <configure name="MATEKSYS_3901_L0X_DISTANCE_THRES" default="200"/>
+    <configure name="USE_MATEKSYS_3901_L0X_AGL" default="0"/>
+    <configure name="MATEKSYS_3901_L0X_COMPENSATE_ROTATION" default="0"/>
+    <define name="MATEKSYS_3901_L0X_MOTION_THRES" value="$(MATEKSYS_3901_L0X_MOTION_THRES)"/>
+    <define name="MATEKSYS_3901_L0X_DISTANCE_THRES" value="$(MATEKSYS_3901_L0X_DISTANCE_THRES)"/>
+    <define name="USE_MATEKSYS_3901_L0X_AGL" value="$(USE_MATEKSYS_3901_L0X_AGL)"/>
+    <define name="MATEKSYS_3901_L0X_COMPENSATE_ROTATION" value="$(MATEKSYS_3901_L0X_COMPENSATE_ROTATION)"/>
+
+    <file name="mateksys_3901_l0x.c" dir="modules/optical_flow"/>
+    <!--<file name="pprz_transport.c" dir="pprzlink/src"/>-->
```
tomvand, line 45: Still required?

tomvand (changes requested):
Nice work, thanks for your pull request. Code looks very clean and easy to read :)

If I read the code correctly, your module does not output any results except to the telemetry, is that correct? Could you add code to send the range and flow data on the ABI bus as well? That would make this driver more useful for other projects. I see you already made a start with this at the end of `mateksys_3901_l0x.c`.

I see you have also committed `conf/messages.xml`, did you add any new messages? In that case, could you add these messages inside a fresh branch of the [pprzlink repo](https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml) and make an additional pull request there? Once the pprzlink pull request is merged, we can update the submodule in this pull request to include the new message into master.

conf/modules/optical_flow_mateksys_3901_l0x.xml
pcampolucci: Hey thank you for the feedback I am currently working on it. Regarding this comment, I wanted to be able to change the default values directly in the airframe. If I am setting the default values in the C file, will I be able to change them later? Do I have to set a value in the xml file as well?

For example I am setting this value in the C file

```
#ifndef MATEKSYS_3901_L0X_MOTION_THRES
#define MATEKSYS_3901_L0X_MOTION_THRES 200
#endif
```

then I add this to the xml

```
  <doc>
    <description>
      Transfers flow X and Y and range data of the Mateksys3901-l0X sensor to the autopilot
      This by the default serial connection not by connecting the debug pins on ther sensor
      The Data is in MSP compatible output. If a MSP stack is availabe in the AP one could use that
      In ths ase we just handle the byres comming from the sensor in MSP compatible handler for this sensor
    </description>
    <configure name="MATEKSYS_3901_L0X_PORT" value="UART1" description="Set to which UART the sensor serial stream pin is connected to"/>
    <configure name="MATEKSYS_3901_L0X_BAUD" value="115200" description="Sets the baudrate of the UART"/>
    <define name="MATEKSYS_3901_L0X_MOTION_THRES" value="120" description="Sets the minimum motion quality to accept the flow measurement"/>
  </doc>
```

Which value will be taken at exectution? 200 or 120? Should I add something to the makefile?

Thank you

conf/modules/optical_flow_mateksys_3901_l0x.xml
pcampolucci: @OpenUAS should I remove this one? It is still unchanged from the template we did at the beginning.

dewagter: You can add:
it the driver is used on a serial port where the TX pin is not available, since the TX is not needed for this sensor it can be disabled.

conf/modules/optical_flow_mateksys_3901_l0x.xml
dewagter: values that differ from one drone to another should be defines.

<configure> blocks are defines in Makefiles
<define> blocks become defines in the code

So MATEKSYS_3901_L0X_PORT_LOWER is a configure: it only exists in the Makefile: the same value has to be filled at several locations.

but MATEKSYS_3901_L0X_MOTION_THRES, MATEKSYS_3901_L0X_DISTANCE_THRES etc should be defines only. So remove lines 35-38 and keep 39-42 (as an option) AND put them in the code as @tomvand suggests.

dewagter:
After a tiny cleanup (mainly configure/define) it's OK for me. Thanks for the work!

pcampolucci:
> After a tiny cleanup (mainly configure/define) it's OK for me. Thanks for the work!

@dewagter what do you think of the formula for velocity at lines 246 and 271 of the C file? Its only for debugging telemetry and it fits the ground truth, but I am not 100% sure if that is correct.

dewagter:
> @dewagter what do you think of the formula for velocity at lines 246 and 271 of the C file? Its only for debugging telemetry and it fits the ground truth, but I am not 100% sure if that is correct.

It depends on the datasheet of the sensor.
It is in any case a linearization for small attitude angles and flat floors, but that should be OK.
And mathematically it should be a ```tan``` instead of a ```sin``` as the ground distance/sec is the opposing side in the triangle (but that should not make a visible difference in your logs).
If the sensor data ```mateksys3901l0x.motionY``` is in degrees per second, and ```mateksys3901l0x.distancemm``` is millimeter, then the destination ```mateksys3901l0x.velocityY``` should be in mm/sec and all is correct.

PS: in paparazzi use ```RadOfDeg(...)```

pcampolucci:
> > @dewagter what do you think of the formula for velocity at lines 246 and 271 of the C file? Its only for debugging telemetry and it fits the ground truth, but I am not 100% sure if that is correct.
> 
> It depends on the datasheet of the sensor.
> It is in any case a linearization for small attitude angles and flat floors, but that should be OK.
> And mathematically it should be a `tan` instead of a `sin` as the ground distance/sec is the opposing side in the triangle (but that should not make a visible difference in your logs).
> If the sensor data `mateksys3901l0x.motionY` is in degrees per second, and `mateksys3901l0x.distancemm` is millimeter, then the destination `mateksys3901l0x.velocityY` should be in mm/sec and all is correct.
> 
> PS: in paparazzi use `RadOfDeg(...)`

Perfect thank you!

conf/messages.xml
```diff
@@ -0,0 +1 @@
+my_messages.xml
```
gautierhattenberger, line 1: please remove the link to this file (and the `my_messages.xml` file)

sw/airborne/modules/optical_flow/mateksys_3901_l0x.h
```diff
@@ -0,0 +1,99 @@
...
+  MATEKSYS_3901_L0X_PARSE_MOTIONX_B3,
+  MATEKSYS_3901_L0X_PARSE_MOTIONX_B4,
+  MATEKSYS_3901_L0X_PARSE_CHECKSUM,        
+};
+
+struct Mateksys3901l0X {
+  struct link_device *device;
+  enum Mateksys3901l0XParseStatus parse_status;
+  uint8_t  sensor_id;
+	uint8_t  motion_quality;
```
gautierhattenberger, line 78: please fix indentation

gautierhattenberger (changes requested):

conf/abi.xml
```diff
@@ -204,6 +204,13 @@
       <field name="throttle"  type="int16_t">Throttle input in pprz_t [0;9600] or [-9600;9600] (for vertical speed control for instance)</field>
     </message>
 
+    <message name="OPTICAL_FLOW_INT32" id="29">
```
gautierhattenberger, line 207: There is already an OPTICAL_FLOW message, and beside some values that could be ignored by the new module, the only difference is the type of `flow_` fields.
If it is not possible to fit in a `int16_t` here, why not make it to `int32_t` for everyone ? Actually, `uint32_t` is strange, no negative values ? Is is really in deg/s ?
Making to message for the same topic is annoying because it means that data consumers will need to bind to both of them.

gautierhattenberger (changes requested):

conf/abi.xml
pcampolucci: Thank you for the comment, unfortunately there is zero documentation on the sensor so am not fully sure that the unit is that one. I will conduct more tests and try to implement it in the already existing message.

gautierhattenberger: If you can't find the units, it is not a big deal, it is probably sensor dependent anyway. The most important to me is merging the two OPTICAL_FLOW messages.

pcampolucci:
@gautierhattenberger I have adapted the old message and removed the last one, it makes much more sense I see now. Should be all good now. 

I have sent a pull request to ppzlink https://github.com/paparazzi/pprzlink/pull/137, it is required for telemetry reasons.

gautierhattenberger (approved):

sw/airborne/modules/optical_flow/mateksys_3901_l0x.c
```diff
@@ -0,0 +1,348 @@
...
+{
+  pprz_msg_send_MATEKSYS_FLOW_LIDAR(trans, dev, AC_ID,
+                                    &mateksys3901l0x.sensor_id,
+                                    &mateksys3901l0x.motion_quality,
+                                    &mateksys3901l0x.motionX_clean,
+                                    &mateksys3901l0x.motionY_clean,
+                                    &mateksys3901l0x.distancemm_quality,
+                                    &mateksys3901l0x.distance_clean,
+                                    &mateksys3901l0x.velocityX,
+                                    &mateksys3901l0x.velocityY);
```
gautierhattenberger, line 83: Actually, the only missing element now is this message from pprzlink. As mentioned in https://github.com/paparazzi/pprzlink/pull/137, I'd like to merge some optical flow message and avoid a dedicated one per sensor if possible. Once the pprzlink PR is accpeted and this PR updated accordingly (message and reference to pprzlink submodule), it is good to merge.

gautierhattenberger (changes requested):

sw/airborne/modules/optical_flow/mateksys_3901_l0x.c
pcampolucci: done. About the reference to the submodule, the new ppzlink branch is on a fork, so maybe the PR of ppzlink should be merged to master first?

gautierhattenberger: yes, once pprzlink message is in master, you can update submodule ref here

sw/airborne/modules/optical_flow/px4flow_i2c.c
```diff
@@ -267,6 +267,7 @@ void px4flow_i2c_downlink(void)
   uint8_t id = PX4FLOW_I2C_ID;
 
   float timestamp = get_sys_time_float();
+  uint8_t distance_quality = 0;
 
 #if REQUEST_INT_FRAME
   int16_t flow_x = px4flow.i2c_int_frame.pixel_flow_x_integral;
```
gautierhattenberger, line 273: since fields are passed by pointer, it is important that the types are correct, so here `flow_x` and `flow_y` must be `int32_t` as well.

gautierhattenberger (changes requested):

gautierhattenberger (approved):

gautierhattenberger:
I merged pprzlink PR, update the submodule here and it is all good, thanks

gautierhattenberger:
The build failure is due to an other PR that was pending (https://github.com/paparazzi/paparazzi/pull/2618). I merged the other one before this one, hopefully this will be enough to solve the issue here. Wait and see.

pcampolucci:
> The build failure is due to an other PR that was pending (#2618). I merged the other one before this one, hopefully this will be enough to solve the issue here. Wait and see.

If the build has failed, how do I rerun the test? I cannot find an option to do that, should I just wait?

gautierhattenberger:
Github still see a conflict with pprzlink, so even if I can rebuild by hand, not sure it will be enough. Could you merge with latest master and push again ? Be careful to select the correct pprzlink version, it is always a mess when merging when submodules are modified on both side. Sorry for the inconvenience.

pcampolucci:
@gautierhattenberger I have rebased the repository on paparazzi/paparazzi master with pprzlink latest version. Let me know if this is correct or I revert the commit. Thank you.

pcampolucci:
seems that i did a mess, should I open a new pull request? I have a backup branch

gautierhattenberger:
If you manage to clean your branch, just force push it in this PR. I will squash all commits into one and rebase when integrating to master branch anyway.

gautierhattenberger:
@pcampolucci let me know if you need help to finish this

pcampolucci:
@gautierhattenberger I am redoing the code based on the backup branch. A bit of brute forcing but I think the rebase messed up the whole thing. I am doing a new pull request in 30 mins max. i am sorry but I got confused with the forks I think.

PR #2640, opened 2021-01-06
[tracker] Start automatically. Add setting to start/stop it.

Fabien-B:
Tracker start sending data automatically while the drone is in flight.
Also add a setting to start sending data when the drone is not in flight.

### DIFF
--- a/conf/modules/e_identification_fr.xml
+++ b/conf/modules/e_identification_fr.xml
@@ -6,11 +6,18 @@
     <configure name="E_ID_PORT" value="UARTx" description="UART where the e-id device is connected to (UART1, UART2, etc"/>
     <configure name="E_ID_BAUD" value="B115200" description="UART baud rate"/>
   </doc>
+  <settings>
+    <dl_settings>
+      <dl_settings name="eID">
+        <dl_setting shortname="e_id_running" var="e_identification_started" min="0" step="1" max="1"/>
+      </dl_settings>
+    </dl_settings>
+  </settings>
   <header>
     <file name="e_identification_fr.h"/>
   </header>
   <init fun="e_identification_fr_init()"/>
-  <periodic fun="e_identification_fr_periodic()" freq="4" autorun="FALSE"/>
+  <periodic fun="e_identification_fr_periodic()" freq="4" autorun="TRUE"/>
   <makefile target="ap">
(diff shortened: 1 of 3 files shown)


sw/airborne/modules/e_identification_fr/e_identification_fr.c
```diff
@@ -138,8 +139,11 @@ void e_identification_fr_init()
 
 void e_identification_fr_periodic()
 {
-
-  if (gps.fix) {
+  if(autopilot_in_flight()) {
```
gautierhattenberger, line 142: `autopilot_in_flight` is currently only working on rotorcraft, but fixedwing can use `launch` variable instead (use ifdef `FIXEDWING_FIRMWARE` or `ROTORCRAFT_FIRMWARE`)
see #2641

conf/modules/e_identification_fr.xml
```diff
@@ -6,11 +6,18 @@
     <configure name="E_ID_PORT" value="UARTx" description="UART where the e-id device is connected to (UART1, UART2, etc"/>
     <configure name="E_ID_BAUD" value="B115200" description="UART baud rate"/>
   </doc>
+  <settings>
+    <dl_settings>
+      <dl_settings name="eID">
+        <dl_setting shortname="e_id_running" var="e_identification_started" min="0" step="1" max="1"/>
```
gautierhattenberger, line 12: add `values="STOPPED|RUNNING` attribute ?

sw/airborne/modules/e_identification_fr/e_identification_fr.c
```diff
@@ -138,8 +139,11 @@ void e_identification_fr_init()
...
+    e_identification_started = true;
```
gautierhattenberger, line 143: so no way to stop in flight ? might be better like that I guess

gautierhattenberger (changes requested):

OpenUAS:
A function to start or stop which is callable in a flightplan welcomed

gautierhattenberger (approved):

gautierhattenberger:
> A function to start or stop which is callable in a flightplan welcomed

This is possible by changing the running state of the module periodic function. As this is a legal requirement, I don't want to make it "too" easy.

PR #2641, opened 2021-01-06
[nav] in_flight detection is not implemented in fixedwing firmware

gautierhattenberger:
this cause the `nav_home` to fail in vertical control


noether:
I wonder how you found this bug. I hope no crash happened :P

noether (approved):

gautierhattenberger:
no crash, but an other PR about in_flight detection :)

PR #2644, opened 2021-01-12
Add explicit python*2* to sw/ext/Makefile mavlink target

tomvand:
Add explicit python*2* instead of python to mavlink build target in sw/ext/Makefile after bug report from @matteobarbera 

@matteobarbera : can you test this PR to see if it fixes the problem?


matteobarbera:
Tested with `make clean && make` with Ubuntu `python` command defaulting to python3.9, it compiled without errors

gautierhattenberger (approved):

PR #2652, opened 2021-01-26
[fix] allow float values for delay in module XML

gautierhattenberger:
see comments in #2553

@martinmm can you check that it is now working for you ? Some of the sensors using this option are from you I think, we can update them for the new format if you can give me the correct values.

### DIFF
--- a/conf/modules/MPPT.xml
+++ b/conf/modules/MPPT.xml
@@ -18,7 +18,7 @@
     <file name="MPPT.h"/>
   </header>
   <init fun="MPPT_init()"/>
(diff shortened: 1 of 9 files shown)


Fabien-B (commented):
Tested, it works.
But does it really need accept integer numbers ? Right now, this is silently changing from the previous behavior. It might be better to accept number only in the range [0;1[ which will force users to update their modules. (with an appropriate error message of course)

Fabien-B:
And update the doc on the wiki

Fabien-B (approved):

Fabien-B (approved):

PR #2653, opened 2021-01-27
[tests] add a compilation test node to modules

gautierhattenberger:
When adding a test node to a makefile section, with required compilation
flags, include and other options, all the files (not arch dependent
files) can be compiled with a TAP compatible program, included in the
standard tests of the CI servers.
Not all module's XML files are converted, but a large part of the most
important parts are already covered. More will be added later. The
number of tested airframes (full compilation of all targets) have been
reduced to speed the CI compile time but still covers the relevant
architecture and boards.
The main benefit is that the overall coverage is already better than
before as previous test aircraft were compiling more or less the same
part of the airborne code, while this new mechanism is more efficient to
test modules not included in any config.


gautierhattenberger:
no feedback, merging anyway

OpenUAS:
Super!

PR #2657, opened 2021-02-02
Update to ChibiOS 20.3

gautierhattenberger:
This is an update to the latest ChibiOS stable version (20.3). Related airframe files have been updated as well, except for some F1 boards for which I'm not sure if it is necessary to do the job (libopencm3 version still available). I have tested that the code is running on a Tawaki board. Only compiled the other boards.

I have olse dropped a few boards:
- discovery: there is a support of the nucleo144 with F7 for people testing on breakout boards
- VMS ECU: not sure this project is still running or using this code, @podhrmic if yes, I can add it back, no problem
- XVert: I'm not even sure it ever worked and the frame is not produced or sold


podhrmic (dismissed):
Thanks @gautierhattenberger - looks good, and the VMS ECU is no longer used.

podhrmic (approved):

PR #2658, opened 2021-02-09
[submodule] update eigen ref and version

gautierhattenberger:
now hosted on Gitlab, latest version: 3.3.9

it is also fixing some deprecated functions

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -43,7 +43,7 @@
 	url = https://github.com/tudelft/gazebo_models.git
 [submodule "sw/ext/eigen"]
 	path = sw/ext/eigen
-	url = https://github.com/eigenteam/eigen-git-mirror.git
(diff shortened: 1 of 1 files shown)


noether:
Have you checked whether the kalman filter that guys wrote works? just in case.

I will also test the GVF_parametric with this new Eigen version. If it does not work, I would say it will be during the compiling stage.

I will check it in two or three days :P. Thanks.

gautierhattenberger:
Actually I had to update eigen to make our code compile without `deprecated` warnings... and it is still working :)

noether:
Haha, good. I remember that I had to deactivate one warning for the same reason. I will check the branch asap, and modify the GVF if needed. I will let you know it.

Thanks

noether:
The gvf compiles perfectly here with this Eigen lib.

Thanks for the update :P.

noether (approved):

PR #2659, opened 2021-02-10
[ocaml] fix SRTM download path

### LINKED ISSUE #2642: HGT files not downloaded anymore
A directory server change on https://dds.cr.usgs.gov/ make current OCAML code not download SRTM anymore
continents are now in subdir e.g. https://dds.cr.usgs.gov/srtm/version2_1/SRTM3/Australia/
A way to change the code could be try to download from first in a list ,if download not possible try next of list since the list is small easier to code and take less space than a lookup table which file belon to which directory... an likly just as fast...

Africa/
Australia/
Eurasia/
Islands/
North_America/
South_America/

OCAML Fu is missing at the moment, else would have made PR, for now for the needy ones with HD space left run this oneliner for your continent change the last part..

`wget -e robots=off --cut-dirs=3 --user-agent=Mozilla/5.0 --reject="index.html*" --no-parent --recursive --relative --level=1 --no-directories https://dds.cr.usgs.gov/srtm/version2_1/SRTM3/Eurasia`

gautierhattenberger:
SRTM data are no more available for direct download, using ESA data
instead. The new SRTM data are SRTM3 GL1 (1 arcsec instead of 3
previously), so the code is now adapted to support both formats,
the most accurate available will be used first.
The only way to force the update of new files is to remove the previous
ones from data/srtm folder.

close #2642


dewagter (approved):
thank you!
Looks great to me:
 - hopefully, the download now works automatically. 
 - I see that you still load the old tiles as well if they are present or manually downloaded.

PR #2660, opened 2021-02-16
Fixed bug in 180 degree mode

hendrixgr:
Fixed a bug in 180 degree mode that prevented the correct azimuth servo travel.


hendrixgr:
I need to check it further

PR #2662, opened 2021-02-17
[python supervision] Remove deprecated function.

Fabien-B:
method time.clock() is deprecated in python 3.8.
It was not needed, so removing it.

### DIFF
--- a/sw/supervision/python/parser.py
+++ b/sw/supervision/python/parser.py
@@ -809,16 +809,13 @@ class Data(object):
 # [Data methods] Init data methods
 
     def init_data(self):
-        init = time.clock()
-
         self.load_cache()
         self.load_conf_files()
         self.load_sets_and_configs()
         self.load_devices()
         self.load_sessions_and_programs()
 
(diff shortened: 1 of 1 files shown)


sw/supervision/python/parser.py
```diff
@@ -809,17 +809,12 @@ def __init__(self, conf_path):
...
     def init_data(self):
-        init = time.clock()
-
         self.load_cache()
         self.load_conf_files()
         self.load_sets_and_configs()
         self.load_devices()
         self.load_sessions_and_programs()
 
-        LOGGER.info("All data loaded in : %s seconds.\n",
```
gautierhattenberger, line 820: keep the log info without the time ?

yannickj: And get rid of the import time at the beginning of the file ?

Fabien-B: Done, and done.

gautierhattenberger (approved):

PR #2664, opened 2021-02-20
Added log name argument to server.ml

jburns11:
I wanted to be able to name the output files of the server, so I added a cmd line argument to server.ml for this purpose. The timestamp previously used is still used if no name is provided. If a name is provided, the timestamp will be added to the string...

For example, the command...

./server -log_name ardrone2

... Will produce the files...

ardrone2__DATE__TIME.data and ardrone2__DATE__TIME.log

I am not very experienced with ocaml, so please feel free to suggest another way to do this.

Best,
Jack


gautierhattenberger (approved):
CI server failing is not related, tested y hand and it's fine

PR #2665, opened 2021-02-22
Fix python comparison.

Fabien-B:
Fix minor syntax error.


gautierhattenberger (approved):
semaphore failing is not related, so okay for merging

PR #2666, opened 2021-02-22
[fix] define _fini dummy syscall for chibios

gautierhattenberger:
bug reported when compiling ekf2 INS with chibios


fvantienen:
Also had that problem indeed! Tnx, this should fix it

fvantienen (approved):

PR #2667, opened 2021-02-23
Px4fmu v5 tim12

Dennis-Wijngaarden:
This pull request has been made for the following 2 reasons:

- Fixed the STM32F76xxl.ld to prvevent px4fmu v5.0 ap board from crashing after flashing paparazzi
- Enabled Timer 12 for the px4fmu v5.0 board to use for PWM signals on PWM ports 7 and 8


fvantienen (approved):

Fabien-B:
I am not much of an expert, but I asked to one of my colleague, and according to him this will slow down access to flash memory because AXIM is not cached.
So if this modification is necessary for Px4, it should have its own linker script.

Fabien-B:
It appears that AXIM and ITCM are interfaces to the same memory, with respective addresses 0X08000000 and 0X00200000.
In this linker script, flash0 is set to addr 0X08008000 to accommodate a bootloader at 0X08000000, but flash1 was still pointing to the same address.
Could you try to reset STM32F76xxI.ld as it was before, and just change the address of flash1 to 0X00208000 ?
If I understood correctly, it should have the same effect as your modification, but keeping the avantages of ITCM.

Dennis-Wijngaarden:
> It appears that AXIM and ITCM are interfaces to the same memory, with respective addresses 0X08000000 and 0X00200000.
> In this linker script, flash0 is set to addr 0X08008000 to accommodate a bootloader at 0X08000000, but flash1 was still pointing to the same address.
> Could you try to reset STM32F76xxI.ld as it was before, and just change the address of flash1 to 0X00208000 ?
> If I understood correctly, it should have the same effect as your modification, but keeping the avantages of ITCM.

I tried to make those changes and tested it on the px4fmuv5. Unfortunately, the autopilot is still crashing after flashing. Any suggestion is welcome to fix this problem while keeping the advantages of ITCM.

fvantienen:
Px4 basically has it own linker script here in this folder, the other autopilots use the linker script from the chibios folder.

Dennis-Wijngaarden:
> It appears that AXIM and ITCM are interfaces to the same memory, with respective addresses 0X08000000 and 0X00200000.
> In this linker script, flash0 is set to addr 0X08008000 to accommodate a bootloader at 0X08000000, but flash1 was still pointing to the same address.
> Could you try to reset STM32F76xxI.ld as it was before, and just change the address of flash1 to 0X00208000 ?
> If I understood correctly, it should have the same effect as your modification, but keeping the avantages of ITCM.

Update: it is partially working with this change, except for lines 58 and 66 in the .ld script. I cannot change flash0 to flash1 (as it is in master branch now) without crashing the autopilot.

Fabien-B:
This linker script is also used for every board that have a bootloader.
If it doesn't work, then I suggest to copy this file (and call it something like STM32F76xxI_px4V5.ld), and change accordingly `CHIBIOS_BOARD_LINKER` in the file `conf/boards/px4fmu_5.0_chibios.makefile`.

fvantienen:
But I expect it won't work for any autopilot with bootloader, not sure if you can test that? Because I expect them to have the same behaviour/problem.

Fabien-B:
I don't really get what is the problem here, why VECTOR and XTOR can be moved to 0X00208000 but not TEXT or VARIOUS. ?
We don't have any board that use a bootloader here so I can't test that, and I don't think there is currently any board featuring an F7 using a bootloader, except the Px4V5.
But I may try using a bootloader soon.

gautierhattenberger:
@fvantienen can we leave the linker script issue outside of this PR (eventually make a new one for this) so we can merge the new timer definitions ?

gautierhattenberger:
The fix on linker script is probably not optimal, but since I can't test it, let's keep a solution that works

PR #2668, opened 2021-02-26
[indi_simple] Fixed rate feedback filter and initialisation

43shawn43:
Before it did not matter if the rate filter was enabled, since every loop the body rates were copied to the rates_filt_fo in the loop itself.
Now the rates_filt_fo is defined and copied from the bodyrates in the initialisation phase as a global variable. If the filter is enabled around any of the body axis (p,q,r) a first order filter will be applied using the rates_filt_fo from the previous iteration. If the filter is disabled the rates_filt_fo will be copied form the body_rates at that timestep.


gautierhattenberger (approved):
looks better now! thanks

PR #2669, opened 2021-02-28
[modules][ocaml] implement module dependency

gautierhattenberger:
- support depends, conflicts and provides nodes, all under a 'dep' node
- autoload are kept for now and are not moved yet
- modules are updated for new dep format
- it is now required that module's name and filename are the same
- order of loaded modules are respecting the dependency relations (in particular in the init sequence)

Other modules that may require dependencies will be update eventually later. The proposed modifications should be fully compatible with current airframe files.


Fabien-B (approved):
Tested, it works !

Fabien-B (approved):

PR #2670, opened 2021-03-03
Allow rapid re-arming after kill

EwoudSmeur:
Sometimes, a drone can be accidentally killed in flight. Re-arming often takes
too long and the drone it crashes. This would be avoided if after kill, the
drone can rapidly re-arm by unkilling.

Maybe someone should shed his/her light on the robustness of this implementation. Personally, I don't like that autopilot_arming_set() could be called from anywhere to mess up the state machine due to different causes. Can this function be removed entirely?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
@@ -42,9 +42,13 @@
 #define MOTOR_ARMING_DELAY  40
 #endif
 
+// Allow re-arming from motors_on after RC kill within 10 seconds
+#define MOTOR_RE_ARM_TIME 10.0
+
 /// Motors ON check state machine states
 enum arming_state {
   STATUS_INITIALISE_RC,
+  STATUS_MOTORS_RC_KILLED,
   STATUS_MOTORS_AUTOMATICALLY_OFF,
   STATUS_MOTORS_AUTOMATICALLY_OFF_SAFETY_WAIT,
   STATUS_MOTORS_OFF,
@@ -64,9 +68,8 @@ static inline void autopilot_arming_init(void)
 {
   autopilot_motors_on_counter = 0;
   autopilot_check_motor_status = STATUS_INITIALISE_RC;
-  motor_kill_time = 0.0;
+  motor_kill_time = -MOTOR_RE_ARM_TIME;
 }
-
 
 /** Update the status of the check_motors state machine.
  */
@@ -74,9 +77,11 @@ static inline void autopilot_arming_set(bool motors_on)
 {
   if (motors_on) {
     autopilot_check_motor_status = STATUS_MOTORS_ON;
-  } else if (autopilot_check_motor_status == STATUS_MOTORS_ON) {
-    autopilot_check_motor_status = STATUS_MOTORS_AUTOMATICALLY_OFF;
+  } else if (kill_switch_is_on() && (autopilot_check_motor_status == STATUS_MOTORS_ON)) {
+    autopilot_check_motor_status = STATUS_MOTORS_RC_KILLED;
     motor_kill_time = get_sys_time_float();
+  } else {
+    autopilot_check_motor_status = STATUS_MOTORS_AUTOMATICALLY_OFF;
   }
 }
 
@@ -125,13 +130,17 @@ static inline void autopilot_arming_check_motors_on(void)
           autopilot_check_motor_status = STATUS_MOTORS_OFF;
         }
         break;
+      case STATUS_MOTORS_RC_KILLED: // Motors were killed by kill mode
+        // If the vehicle was killed accidentally, allow rapid re-arm
+        if ( (get_sys_time_float() - motor_kill_time) < MOTOR_RE_ARM_TIME) {
+          autopilot_check_motor_status = STATUS_MOTORS_ON;
+        } else {
+          autopilot_check_motor_status = STATUS_MOTORS_AUTOMATICALLY_OFF;
+          autopilot.motors_on = false;
+        }
+        break;
       case STATUS_MOTORS_AUTOMATICALLY_OFF: // Motors were disarmed externally
         //(possibly due to crash)
-        // If the vehicle was killed accidentally, allow rapid re-arm
-        if ( (get_sys_time_float() - motor_kill_time) < 10.0) {
-          autopilot.motors_on = true;
-          autopilot_check_motor_status = STATUS_MOTORS_ON;
-        }
         //wait extra delay before enabling the normal arming state machine
         autopilot.motors_on = false;
         autopilot_motors_on_counter = 0;
@@ -205,6 +214,10 @@ static inline void autopilot_arming_check_motors_on(void)
     }
   } else {
     autopilot.arming_status = AP_ARMING_STATUS_KILLED;
+    if (kill_switch_is_on()) {
+      autopilot_check_motor_status = STATUS_MOTORS_RC_KILLED;
+      autopilot.motors_on = false;
+    }
   }
 }
 
--- a/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
@@ -32,6 +32,7 @@
 #include "autopilot_arming_common.h"
 #include "autopilot_firmware.h"
 #include "autopilot.h"
+#include "mcu_periph/sys_time.h"
 
 /** Delay until motors are armed/disarmed.
  * In number of rc frames received.
@@ -56,12 +57,14 @@ enum arming_state {
 
 uint32_t autopilot_motors_on_counter;
 enum arming_state autopilot_check_motor_status;
+float motor_kill_time;
 
 
 static inline void autopilot_arming_init(void)
 {
   autopilot_motors_on_counter = 0;
   autopilot_check_motor_status = STATUS_INITIALISE_RC;
+  motor_kill_time = 0.0;
 }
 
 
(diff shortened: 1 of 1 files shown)


sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -71,8 +74,9 @@ static inline void autopilot_arming_set(bool motors_on)
 {
   if (motors_on) {
     autopilot_check_motor_status = STATUS_MOTORS_ON;
-  } else {
+  } else if (autopilot_check_motor_status == STATUS_MOTORS_ON) {
```
dewagter, line 77: this is just in case it is called repeatedly?

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -123,6 +127,11 @@ static inline void autopilot_arming_check_motors_on(void)
         break;
       case STATUS_MOTORS_AUTOMATICALLY_OFF: // Motors were disarmed externally
         //(possibly due to crash)
+        // If the vehicle was killed accidentally, allow rapid re-arm
+        if ( (get_sys_time_float() - motor_kill_time) < 10.0) {
+          autopilot.motors_on = true;
+          autopilot_check_motor_status = STATUS_MOTORS_ON;
+        }
         //wait extra delay before enabling the normal arming state machine
         autopilot.motors_on = false;
```
dewagter, line 136: motors_on seems to be reset to false even if the kill-time was < 10s.

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -56,12 +57,14 @@ enum arming_state {
...
 uint32_t autopilot_motors_on_counter;
 enum arming_state autopilot_check_motor_status;
+float motor_kill_time;
 
 
 static inline void autopilot_arming_init(void)
 {
   autopilot_motors_on_counter = 0;
   autopilot_check_motor_status = STATUS_INITIALISE_RC;
+  motor_kill_time = 0.0;
```
dewagter, line 67: this means that the first 10 seconds after booting, the kill can arm? Can it be set to -10seconds?

dewagter (dismissed):
Looks great, thanks Ewoud!
Just 2 points remaining: startup 10 sec and how comes the motors_on = false does not affect it.

gautierhattenberger:
I'm not sure I got the logic. How do you make the difference between a normal and accidental kill (even in flight). The arming set function is called by `autopilot_set_motors_on` which can be used in the flight plan, or during safety procedure.
What is suppose to happen during the 10 s after calling this function ?
If you have the RC on and waiting in any mode except KILL, can you even kill your motors from the autopilot ?
If we remove the arming_set function but still allow to set motors_on to false from set_motors_on, don't we have inconsistent states between the autopilot and arming status ?

dewagter:
> I'm not sure I got the logic. How do you make the difference between a normal and accidental kill (even in flight). The arming set function is called by `autopilot_set_motors_on` which can be used in the flight plan, or during safety procedure.
> What is suppose to happen during the 10 s after calling this function ?
> If you have the RC on and waiting in any mode except KILL, can you even kill your motors from the autopilot ?
> If we remove the arming_set function but still allow to set motors_on to false from set_motors_on, don't we have inconsistent states between the autopilot and arming status ?

The goal is:

 - kill mode always kills instantly (no change)
 - un-kill within 10 seconds of a kill should re-arm
 - un-kill at any other time un-kills but does not arm (no change)

gautierhattenberger:
what kind of "un-kill" are considering ? from autopilot, from GCS, from RC ? Does this concerns only "RC-controlled" modes, NAV mode, both ?
Currently, if you un-kill with a set_motors_on, you also call the arming_set so the arming status is ARMED. Am I wrong ?

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
EwoudSmeur: The 'difficulty' here is that from autopilot_check_motor_status==STATUS_MOTORS_AUTOMATICALLY_OFF you can now do a rapid re-arm. If you would kill from being in a disarmed state and then immediately unkill, you could otherwise arm the vehicle. This way, only if you kill from an motors_on state you can rapidly re-arm.

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
EwoudSmeur: Good point, it will now only be re-armed the next iteration when it is in STATUS_MOTORS_ON... Which also works. I can put an else case though

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
EwoudSmeur: No normally not, as you can only re-arm from "STATUS_MOTORS_AUTOMATICALLY_OFF", and you can only get there once you're armed. I'm fine with putting -10 though.

EwoudSmeur:
Thanks for the comments @dewagter and @gautierhattenberger!

We are mainly considering a kill from RC, which we would like to rapidly undo. I don't think there would be a need to have this for other sources as arming from the GCS is already instant.

Indeed it looks like I broke disarming from another source, which would lead to immediately re-arming. Would it solve the issue if we separate 'external' disarms from RC kill? In the end this file is really aimed at arming from RC anyway right?

> If we remove the arming_set function but still allow to set motors_on to false from set_motors_on, don't we have inconsistent states between the autopilot and arming status?
Yes indeed, this is another problem. What if we separate STATUS_MOTORS_AUTOMATICALLY_OFF from something like STATUS_MOTORS_KILLED? or STATUS_MOTORS_RC_KILLED?

gautierhattenberger:
I think that the best option is to consider all possibilities in the arming state machine, so as you suggest with something like STATUS_MOTORS_RC_KILLED that allow to go back to motors ON directly within 10 seconds and change to MOTORS_OFF after that delay. But going to RC_KILLED should be only allowed from MOTORS_ON status I think.
Another question, how do you configure this "kill from RC" ? Because if it is from RADIO_KILL_SWITCH, using this will enter KILL mode, thus by-passing the arming state machine.

EwoudSmeur:
I think I can do these things and make it work, but to be honest, my head is spinning at this point. There are so many functions in different files that affect the arming state, motors_on, and kill_throttle in different combinations. They are mainly in autopilot.c, autopilot_static.c and autopilot_generated files. Is it an idea to put everything in the arming files, to have all arming logic in one place?

gautierhattenberger:
The kill_throttle / motors_on issue is an annoying legacy, I agree. But I would not merge every thing into arming, those are still different things somehow.
The current version looks better to me. Did you have the time to test it ?

EwoudSmeur:
Ok yes I understand, it is complex to consider all possible applications and conditions with (un)arming/(un)killing.

I tested this with a simulation, and I could verify that with RC:
- not being armed, killing and unkilling does not arm.
- arming, killing, and unkilling within 10 seconds re-arms
- arming, killing, and unkilling after 10 seconds does not arm

I think there is no need to modify the other arming options (throttle and switch) as it should already be easy to arm (and we never use those actually).

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -56,21 +61,25 @@ enum arming_state {
...
 uint32_t autopilot_motors_on_counter;
 enum arming_state autopilot_check_motor_status;
+float motor_kill_time;
 
 
 static inline void autopilot_arming_init(void)
 {
   autopilot_motors_on_counter = 0;
   autopilot_check_motor_status = STATUS_INITIALISE_RC;
+  motor_kill_time = -MOTOR_RE_ARM_TIME;
```
gautierhattenberger, line 71: I don't think this is necessary, it can be initialized to 0

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
```diff
@@ -196,6 +214,10 @@ static inline void autopilot_arming_check_motors_on(void)
     }
   } else {
     autopilot.arming_status = AP_ARMING_STATUS_KILLED;
+    if (kill_switch_is_on()) {
+      autopilot_check_motor_status = STATUS_MOTORS_RC_KILLED;
```
gautierhattenberger, line 218: you need to set the `motor_kill_time` before entering this state

gautierhattenberger (changes requested):
I'm fine with merging like this after the two small fixes

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
EwoudSmeur: This is done in the function `autopilot_arming_set()`. Is that also ok with you?

sw/airborne/firmwares/rotorcraft/autopilot_arming_yaw.h
EwoudSmeur: Hypothetically, someone could power their drone, switch to kill mode and unkill and it could 're'-arm. Probably it is unlikely, but It also doesn't hurt to have this protection right? I added it after a suggestion from @dewagter.

gautierhattenberger: That is the point of the second comment. The only way to enter the state `STATUS_MOTORS_RC_KILLED` is from arming_set (which sets the timer) or when in kill mode + kill switch on. So if you also set the timer in the second case, you can only execute the state `RC_KILLED` with a valid time, even after startup.

EwoudSmeur: Thanks for explaining. Since `autopilot_arming_check_motors_on` is called repeatedly, setting the `motor_kill_time` there is inconvenient. I now left only one route to `STATUS_MOTORS_RC_KILLED`, via `autopilot_arming_set`. This should guarantee the same safety.

gautierhattenberger (approved):

PR #2674, opened 2021-03-19
[build] split subdirs rules

gautierhattenberger:
when compiling ground_segment target, it should not depend on misc tools
that are themselves compiling the ext libraries, including libopencm3.
this way, the command 'make ground_segment' should be enough to run a
simulation without installing a cross compiler

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,8 @@ PPRZCENTER=sw/supervision
 MISC=sw/ground_segment/misc
 LOGALIZER=sw/logalizer
 
-SUBDIRS = $(PPRZCENTER) $(MISC) $(LOGALIZER) sw/tools
+SUBDIRS = $(PPRZCENTER) $(LOGALIZER) sw/tools
+SUBDIRS_EXTRA = $(MISC)
 
 #
 # Communication protocol version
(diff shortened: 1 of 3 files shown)


Fabien-B:
I removed gcc-arm to test it, but it is no enough:
```
make -C sw/ext pprzlink.update
../../conf/Makefile.arm-embedded-toolchain:29: *** Error: arm-none-eabi-gcc cross-compiler not found! Recommended toolchain is https://launchpad.net/gcc-arm-embedded.  Stop.
make: *** [Makefile:147: libpprzlink.update] Error 2
```

Fabien-B:
I commented line 33 in `sw/ext/Makefile` : `include $(PAPARAZZI_SRC)/conf/Makefile.arm-embedded-toolchain`
It works, I can build ground_segment, build an AC with NPS and run a simulation.

gautierhattenberger:
Since we really need that to compile some of the tools in ext, I would prefer to keep it. So I see two options:
- change the error for a warning, so if you don't use it, it's not blocking
- split the makefile in ext to have at least pprzlink separated from the rest

Fabien-B (approved):
Works for me

PR #2676, opened 2021-03-24
[Flight plan] Add support for LTP waypoints.

Fabien-B:
This adds an optional attribute wp_frame to the root node flight_plan, that default to "utm".
If this attribute is set to "LTP", waypoints frame is then the LTP frame (ENU).

### DIFF
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -66,6 +66,7 @@ max_dist_from_home CDATA #REQUIRED
 ground_alt CDATA #REQUIRED
 security_height CDATA #REQUIRED
 alt CDATA #REQUIRED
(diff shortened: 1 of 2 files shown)


Fabien-B:
I had a bug for the altitude but it should works now.
Tested with default_alt, alt and height, it works:
`<waypoint name="p1" x="2" y="4"/>`
`<waypoint name="p1" x="2" y="4" alt="149"/>`
`<waypoint name="p1" x="2" y="4" height="5/>`

gautierhattenberger (approved):

OpenUAS:
Awesomeness, an unexpected, for me at least, gift. ! Thx

gautierhattenberger:
We are also considering to drop UTM coordinates for fixedwing at some point, but that is another story

PR #2682, opened 2021-03-27
keep eigen on a valid existing tag, 3.3.9 in this case

gautierhattenberger:
Some builds are failing on semaphore with and error `fatal: remote error: upload-pack: not our ref`. This might be due to the fact it is currently pointing to an ID that is neither a tag or a branch.
With this we stick to the closest release tag, 3.3.9, so only 3 commits back. Changes after that are related to sparse matrix that we don't use.
Of course, this may not be the source of error, but still worth trying it I think.


noether (approved):
Yeah, 

I do not think anybody in Pprz is using these extra commits from Eigen.

Please, go ahead

PR #2684, opened 2021-04-07
[build] output flags in orginal order

### LINKED ISSUE #2673: LDFLAGS order reversed
Hi, the current version of paparazzi seems to reverse the order of the LDFLAGS, which are order-dependent and this causes problems when linking multiple dependencies such as in the [cv_opencvdemo module](https://github.com/paparazzi/paparazzi/blob/master/conf/modules/cv_opencvdemo.xml). This problem did not occur during last year's tu delft pprz course, so apparently something has changed in the meantime.

gautierhattenberger:
it does matter for LDFLAGS in particular

close #2673


dewagter (approved):
Confirmed to solve the problem. Elegant fix.

PR #2687, opened 2021-04-08
Update from the TUDelft Course 2021

dewagter:
- Files for PDF exercise (https://github.com/tudelft/paparazzi/pull/67)
 - Explicit call to python2 for systems that default to python3
 - Description and comments fixes
 - [flight_plan] remove warning about dummy WP
 - Added configs for ps4 gamepads, fixed typos (https://github.com/tudelft/paparazzi/pull/68)
 - Update: for flight testing in the CyberZoo
 - Fix color gain in mav_course_exercise.xml airframe (https://github.com/tudelft/paparazzi/pull/72)
 - [flight_plan] Re-measured the autoland border. It is not safe to exit the red square so it auto-lands to prevent damage. (https://github.com/tudelft/paparazzi/pull/73)
 - [Camera] Vision runs faster is unnecessary frames are not grabbed. (https://github.com/tudelft/paparazzi/pull/74)

### DIFF
--- a/sw/airborne/modules/mav_course_exercise/mav_exercise.c
+++ b/sw/airborne/modules/mav_course_exercise/mav_exercise.c
@@ -1,0 +1,176 @@
+/*
+ * Copyright (C) 2021 Matteo Barbera <matteo.barbera97@gmail.com>
+ *
+ * This file is part of Paparazzi.
+ *
+ * Paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * Paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


sw/airborne/modules/mav_course_exercise/mav_exercise.c
```diff
@@ -0,0 +1,176 @@
...
+
+#include "mav_exercise.h"
+#include "subsystems/abi.h"
+#include "firmwares/rotorcraft/navigation.h"
+#include "state.h"
+#include "autopilot_static.h"
+#include <stdio.h>
+
+#define NAV_C // needed to get the nav functions like Inside...
+#include "generated/flight_plan.h"
```
gautierhattenberger, line 29: Maybe these functions should be made public ? Including the flight_plan with `NAV_C` flag several times can be a problem when using variables.

dewagter: Sound like a good idea. But could this be a new issue/PR?

gautierhattenberger: yes, better not change this before a release

gautierhattenberger (approved):

PR #2688, opened 2021-04-08
Reorganize INDI simple for better takeoff

EwoudSmeur:
With INDI simple, there was still the possibility to turn off the control just by giving a low thrust command. With these changes, you also have to be not in_flight and instead the increments are not accumulated, so you end up with a proportional controller.

I reorganized some of the code to implement this


gautierhattenberger (approved):

gautierhattenberger:
Semaphore is lost, but it looks fine

PR #2694, opened 2021-04-12
Place dummy waypoint inside max_dist_from_home. Fix #2693.

### LINKED ISSUE #2693: dummy waypoint at 59 meters from HOME gives compile warnings in indoor flightplans (max 20m from HOME)

Fabien-B:


gautierhattenberger:
What about placing this dummy WP at (0, 0) ? It is not supposed to be used anyway

gautierhattenberger:
This is indeed a good question, thank you for asking.

This dummy was added March 9th 2005 by Pascal with commit 27b5a45c55d5e5573136b4e6a7b19186191891ee. It allows to use the `go` instruction with a WP id and if not found, to a (x, y) position with the `x` and `y` attributes, that are then stored in the dummy WP aforementioned.

However, on August 12th 2005, some indigent person named `rad <USER@DOMAIN>` removed the `x` and `y` attributes from the flight plan DTD on the reason that `*** empty log message ***` (see commit d69696215ae712a44d6270cc524b384e0f70de23), so the functionality still exists, but can't be used.

Of course, the temptation to curse this "rad" is great, but better keep the cleaning for after the coming release (so the day after tomorrow).

gautierhattenberger (approved):

PR #2695, opened 2021-04-14
[cleaning] drop support of LPC21 boards and related conf

gautierhattenberger:
Spring cleaning is coming


dewagter (approved):
-you can remove the complete mag_micromag module as only lpc had the _hw files.

gautierhattenberger:
I will make a PR soon to remove old, unused or deprecated code, so I will include this one

PR #2696, opened 2021-04-15
[cleaning] removing some old stuff

gautierhattenberger:
- infrared sensors and related ahrs
- mkk and asctec actuators
- old carto modules
- fix bebop actuators for NPS
- clean airframe and conf

If you think that some things should really be kept, I can add them back, but keeping these old i2c ESC in particular is really not the best I think.
On the other hand, if you have suggestions to remove more things, I will gladly do that!


dewagter:
Well, this PR is a bit painful. Together with the removal of LPC, it means that none of my best-tested most flown airframes are supported anymore in the latest paparazzi. I guess this is OK, but I was still using asctec actuators in several airframes (MavTec).

gautierhattenberger:
We also have some LPC and Asctec frames lying around, but don't use them anymore, and it is hard to tell what is being used or not in the end. My goal is of course not to ground old stuff, but to reduce the maintenance effort. If you tell me that you still need these actuators, I can revert that from this PR and only remove / fix the rest.

dewagter:
Will you also drop all STM32F103 and F105 boards (without chibios)?

gautierhattenberger:
This was suggested by Freek. I think it is a longer term option and I won't do it soon. However I might remove some board like LisaL that is probably not used anymore.
If you have suggestions for outdated or broken things, let me know. So far I see:
- the beagle bone black (not sure it ever worked)
- ardrone2 that might be broken now, if someone could check that

gautierhattenberger:
@dewagter I have added back the actuators `asctec_v2` that is most used in the tudelft config (v1 doesn't seem to be used anywhere) and mkk are mainly with LisaL and "untested" airframes. Is it good enough for you ?

dewagter (approved):
Thank you very much. I'll still plan to upgrade these airframes, but again, it's a real pitty to remove the best-tested platform when there is so much other unused stuff.

gautierhattenberger:
This is part of my problem, I don't know what is used or not, so I have started with what we have at the lab here.

The reason I'm doing this is that for a project I need to make many changes to the code that have some impacts on the module files in particular. And this is a lot of work for me if `sed` is not applicable...
Still, thank you for accepting this one.

gautierhattenberger:
@dewagter I just realized that when removing the `cartography` module (one of mine that was never really used or even finished), I removed the module `photogrammetry_calculator` that is one of yours (same folder). Is it still used ? Should I restore it ?

dewagter:
@gautierhattenberger the goal for us is to not need fixed-wing anymore but be able to do everything with hybrid code. Once that code can do search patterns the photo_calc might be adapted to hybrids. So the fixedwing only version can go.

On another note: the ```healt_tools.py```allows to generate lists of used modules. If there is a way (github?) to get a list of all the active forks since 2021, I could automatically make a list of all modules that have not been compiled for more than a year.

gautierhattenberger:
@dewagter Then I leave it like that. We can bring it back any time is needed.

One of my goal with this work is actually to reach a common architecture for fixed-wing and rotorcraft, or at least something with much more consistency that what we have now.

Checking and testing modules is something pretty complex. The "old" way by compiling example airframes is not efficient. In my case I have a lot of active things in branches or untracked files that are not public. This is the main reason for introducing the `test` node in modules (#2653). At least the arch independent code of modules can be better tested, even when no one is using it in the main repository.

PR #2699, opened 2021-04-21
[uavcan] Fix NPS

fvantienen:
This should fix the NPS simulator for the uavcan actuators


gautierhattenberger:
This might fix the compilation issue, but can you actually fly in simulation with that ? For bebop and ardrone, I have made an `actuators_nps` module that is autoloaded (so replace the actuator for nps target). This is just a dummy compilation of pwm driver to avoid undefined ref. But maybe this will not make a difference for your case.

fvantienen:
It doesn't fly yet in NPS, but this should work. I still need to merge some of the control and other parts to be able to fly this in master.

gautierhattenberger:
ok, I'm merging this fix for now then

gautierhattenberger (approved):

Issue #2677, opened 2021-03-24
Remove static sectors in V6.0


Fabien-B:
I propose to remove static sectors in the next major release (V6.0)
IMO, the added complexity does not worth the computing time saved, especially on "recent" microcontrollers like STM32.

I propose to make all sectors dynamic, and to remove the `type` attribute in the `sector` node of the flight plans.

Does anyone wants to keep static sectors, and if so, can you explain why ?

fvantienen:
Since #2701 is merged this can be closed

PR #2701, opened 2021-04-29
Remove static sectors

Fabien-B:
As suggested in #2677, I propose here to remove static sectors. All sector will be dynamic ones.

### DIFF
--- a/conf/flight_plans/ENAC/crazyflie_multi_ranger_test.xml
+++ b/conf/flight_plans/ENAC/crazyflie_multi_ranger_test.xml
@@ -42,7 +42,7 @@ static inline void obstacle_detection_cb(uint8_t sender_id __attribute__((unused
       <corner name="_N3"/>
       <corner name="_N4"/>
     </sector>
-    <sector name="Survey" color="green" type="dynamic">
+    <sector name="Survey" color="green">
       <corner name="S1"/>
       <corner name="S2"/>
       <corner name="S3"/>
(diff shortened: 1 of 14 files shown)


fvantienen (approved):
I fully agree, this also removes the problem with the correct orientation.

gautierhattenberger:
I also agree, but maybe one little thing: we should keep the DTD unchanged for some time and print a warning with a deprecated message when using `type`. People with personal flight plans should have some time to fix them if needed.

Fabien-B:
I suggest making it an error, so people will take care of the problem before the DTD change.
Running now in `Error: attribute "type" on tag sector is deprecated and must be removed` is better than not paying attention to a warning, and running in a few weeks in a n obscure `DTD error`.

fvantienen:
I agree with the error, this forces people to update their files.

On a side note, shouldn't we list those deprecations somewhere with a specific enddate and a link to the specific code to remove/update. This reminds both the user and maintainers to update/remove this code. So my advice is to do the following:
- Remove all code that is labeled 'deprecated' for more than 1 year or add them to the list
- For new deprecations like this, add them to a list in a file somewhere with a specific enddate

Fabien-B:
Maybe we can use this github action, together with issues for that?
https://github.com/marketplace/actions/due-dates
And we can also set a "deprecation" tag to these issues so people can find them easily.

fvantienen:
For me that seems fine 👍  What do you think @gautierhattenberger ?

gautierhattenberger:
we can give a try until github have a proper "due date" option for pull requests

Here, we have also moved the definition of sectors outside the `NAV_C` part, so you can use that anywhere in the code and don't need the (dangerous) workaround to define NAV_C twice.

fvantienen (approved):
Seems fine 👍

PR #2703, opened 2021-05-04
[rtcm2ivy] convert printf to printf_debug

jim-sharples:
replaced printf by printf_debug to avoid freezing the paparazzi centre


fvantienen (approved):
Seems fine with me

PR #2704, opened 2021-05-04
[ins] EKF2 fix localization of global waypoints

fvantienen:
Fix localization of global waypoints when EKF2 resets the local origin.


gautierhattenberger (approved):

gautierhattenberger:
this issue should probably be factorized, as all INS have that problem probably

PR #2706, opened 2021-05-04
Add github action to set dates on issues.

Fabien-B:
See https://github.com/marketplace/actions/due-dates


fvantienen (approved):
I'm ok with trying this out for now!

PR #2709, opened 2021-05-04
[rotorcraft] Add GPS loss bypass with datalink timeout

fvantienen:
This adds an option to disable GPS loss whenever there is a valid datalink connection.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/main_ap.c
+++ b/sw/airborne/firmwares/rotorcraft/main_ap.c
@@ -318,6 +318,9 @@ void failsafe_check(void)
 #if NO_GPS_LOST_WITH_RC_VALID
       radio_control.status != RC_OK &&
 #endif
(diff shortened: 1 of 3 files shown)


sw/airborne/firmwares/rotorcraft/main_ap.c
```diff
@@ -317,6 +317,9 @@ void failsafe_check(void)
       autopilot_get_motors_on() &&
 #if NO_GPS_LOST_WITH_RC_VALID
       radio_control.status != RC_OK &&
+#endif
+#if NO_GPS_LOST_WITH_DATALINK_TIME
```
gautierhattenberger, line 321: it is more correct to use `#ifdef` for non-boolean values, although it makes no real difference here

gautierhattenberger (approved):

PR #2710, opened 2021-05-05
Fix radio confs.

Fabien-B:
Ill-formed comments in radio files cause lxml to crash.

+ if min > max, invert them and set reverse="1"
+ remove obsolete "ctl" attribute
+ remove ill-formed comments ( "--" in the comment)

The ill-formed comment has been removed by hand, the rest has been edited via a script.


fvantienen:
There are a lot of old radio configurations. Can we maybe just remove those which are not linked to in the `conf` files?

gautierhattenberger:
I'm not sure it is essential to remove radio conf. It is not like they can be really deprecated, and it is not easy to tell which one is used. Even for use, we are use some of them that are not in the "official" enac conf.

gautierhattenberger (approved):

PR #2716, opened 2021-05-07
Feature accurate cpu load

alex31:
We used CH_DBG_THREADS_PROFILING to get cpu load information, but this chibios module is deprecated, not accurate, and forbid use of scheduler tickless mode.

This PR introduce the usage of CH_DBG_STATISTICS which is : 
° CPU cycle precise instead of SYSTICK precise : better measure of threads that are waken often but use a small quantum of CPU time
° compatible with tickless mode
° take into account time spent in ISR

This PR also enable tickless scheduler mode


gautierhattenberger:
It seems that some types are not compatible with the version of ChibiOS for paparazzi: https://semaphoreci.com/paparazziuav/paparazzi/branches/pull-request-2716/builds/3

gautierhattenberger:
maybe something is broken with the docker image, I will check that

gautierhattenberger (approved):

PR #2717, opened 2021-05-07
° fix dshot telemetry query

alex31:
° make dshot speed possibly dynamic when speed has to be chosen at execution

Only C files has been fixed, still need to make an airframe which use it, which involve I need to understand how 
° to enable SD5 which is the default telemetry usart rx on the servoB connector
° use the abstract object uart5
° extract SD5 pointer from  abstract object uart5 where it should be used

° for a future usage where dshot command will be issued from two timers, permit the specification of telemetry uart for each timers


fvantienen:
Is this ready to be reviewed/merged, or is it still a work in progress?

gautierhattenberger:
still in progress, I want to test it and improve the configuration

OpenUAS:
A hint on "improve the configuration" would be welcomed, then maybe we could find and put some development resources to get the PR fixed and honored before release of v6

OpenUAS:
OK maybe a nice issue to fix for @pcampolucci as the next one, will discuss ASAP

Fabien-B:
Handled in #3253.

PR #2718, opened 2021-05-11
Filter rates indi

EwoudSmeur:
This PR makes the filtering of body rates in the INDI controller use the first order filter in the filters folder. Now only the cutoff frequency needs to be specified to activate the filter. I updated all airframe files that I could see were using the filtering.

### DIFF
--- a/conf/airframes/ENAC/hoops_gen_ap.xml
+++ b/conf/airframes/ENAC/hoops_gen_ap.xml
@@ -207,8 +207,8 @@
     <define name="G2_R" value="0.20"/>
 
     <!--unit: Hz-->
-    <define name="STABILIZATION_INDI_FILT_CUTOFF_P" value="20.0"/>
-    <define name="STABILIZATION_INDI_FILT_CUTOFF_Q" value="20.0"/>
+    <define name="FILT_CUTOFF_P" value="20.0"/>
+    <define name="FILT_CUTOFF_Q" value="20.0"/>
 
     <!-- reference acceleration for attitude control -->
     <define name="REF_ERR_P" value="170.0"/>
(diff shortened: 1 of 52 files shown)


conf/airframes/ENAC/hoops_gen_ap.xml
```diff
@@ -206,9 +206,9 @@
     <define name="G1_R" value="0.0022"/>
     <define name="G2_R" value="0.20"/>
 
-    <define name="FILTER_ROLL_RATE" value="TRUE"/>
-    <define name="FILTER_PITCH_RATE" value="TRUE"/>
-    <define name="FILTER_YAW_RATE" value="FALSE"/>
+    <!--unit: rad/s-->
+    <define name="STABILIZATION_INDI_FILT_CUTOFF_P" value="100.0"/>
```
fvantienen, line 210: I think these are incorrect

gautierhattenberger:
Why sometimes the `FILT_CUTOFF_R` is removed and sometimes replaced by `FILT_CUTOFF_RDOT` ?

conf/airframes/ENAC/hoops_gen_ap.xml
EwoudSmeur: :sweat_smile:

EwoudSmeur:
> Why sometimes the `FILT_CUTOFF_R` is removed and sometimes replaced by `FILT_CUTOFF_RDOT` ?

The naming before was confusing/incorrect. There are two filters that are used: one for the angular acceleration and one for the angular rates, because of different noise levels due to the derivative. Before we only had the filter for the angular acceleration and then R referred to angular acceleration around body Z.

gautierhattenberger (approved):

fvantienen (approved):

fvantienen:
I can't merge it because of semaphore failing, but for me it is fine to merge

PR #2719, opened 2021-05-12
[fix] take the reverse into account in macro as well

gautierhattenberger:
the reverse flag is directly applied by the parser, so correct values are use for all macros


fvantienen (approved):
This looks good to me!

PR #2721, opened 2021-05-14
Autopilot reorg

gautierhattenberger:
This PR is reorganizing the core autopilot `main_ap` around better defined "tasks" and harmonized between the different firmwares.

The new structure is made to guarantee a certain sequence in the init phase and between the tasks' blocks. In particular, a fixed time between some tasks can fixed: between sensor periodic and estimation periodic, between estimation and control, etc.

As a result, all modules have been updated to fit into their dedicated tasks. In addition to that, the dependencies between module in terms of actual modules or functionalities have been explicitly specified. The module generator is therefore checking that all required components are loaded. New modules automatically loaded according the current target/firmware/board have been introduced and may replace the makefiles at some point.

Some minor changes are included here like a few airframe update, unused code cleaning, fixes for warnings,...

Some work is still needed, especially concerning the actuators and radio control (due to FBW splitting) or baro (stupid "baro_board" subsystem). This can be done in a later PR, this one is already big enough...

It would be nice to do more test on real hardware. My own test are working for CI compilation, simulation, and a few real frames.


Fabien-B:
I did not fly with it, but I flashed a fixedwing with apogee with ChibiOs, all seems to work.

gautierhattenberger:
@dewagter do you have comments or remarks about those modifications ?

noether:
I will look at it in detail on Friday.

dewagter (approved):
The new dependencies and tasks look good to me. If @fvantienen could check the Nederdrone then it looks good to me.

fvantienen:
Will try today on a Nederdrone and test fly it indoors

fvantienen:
The Nederdrone can't align because the barometer does not work. Isn't `px4fmu_5.0_chibios` missing some board xml file?

noether:
> The Nederdrone can't align because the barometer does not work. Isn't `px4fmu_5.0_chibios` missing some board xml file?

Did you try

#ifndef USE_BARO_BOARD
#define USE_BARO_BOARD 0
#endif
?

gautierhattenberger:
probably, I have not made a board xml for all boards
the baro module is really a problem with the BARO_BOARD automagic
in fact I would prefer that people include explicitly the sensors they are using, but I can make it work for some boards

noether (dismissed):
It looks quite good to me. Thanks Gautier.

fvantienen:
Isn't the idea that in the board xml file you can load for example a specific barometer with setting the correct parameters? And do the same for the IMU?
This can then still be overwritten (maybe also disabled) by the airframe, but would be extremely convenient for most people.

gautierhattenberger:
@fvantienen this is exactly what I want to do. However, it is not possible at the moment to disable a module from airframe file by loading another one. I would like to add more dependency options like `recommend` or `suggest` to have a module loaded if no other module provide the same functionality. So for now, the plan is to keep the old magic in a `makefile` section of a baro board module loaded for all with `ap` target.
I'm currently fixing the details, and also add some changes to fixedwing firmware that I forgot and will let you know when you can test again.

gautierhattenberger:
@fvantienen you can try again with your navy drone, it should work now for the barometer

gautierhattenberger:
@fvantienen Could you check again that the baro problem is solved now ? Thanks

fvantienen:
Sorry have been a bit busy, but I will try on Thursday/Friday is that is ok with you?

gautierhattenberger:
sure, thank you for testing

gautierhattenberger:
@fvantienen @noether I have updated the timing and group and task for this new AP organization.
See in particular https://github.com/paparazzi/paparazzi/pull/2721/commits/869adb279480539a3bb8106bf4e2a4b63d23c33a and https://github.com/paparazzi/paparazzi/pull/2721/commits/81a695e55063c3f5a706bd7b27637814ae65af4c

- ChibiOS implementation behave the same way than the other arch.
- get_sys_time functions for milli and micor seconds are also more accurate.
- some tasks are grouped in a gnc group as they should be called in sequence: estimation, control, default, actuators
- the timing between the sensors and gnc is half of the main period, it correspond to the default resolution of the sys_time virtual timers
- the timing of AP threads is improved with a calculated wakeup time instead of a fixed sleep time
- the monitor task for fixedwing is moved in the firmware specific autopilot part and called by a module
- all timings have been accurately tested with high speed log and a dedicated profiler for ChibiOS

It would be nice if you can give a try :)

noether:
> @fvantienen @noether I have updated the timing and group and task for this new AP organization.
> See in particular [869adb2](https://github.com/paparazzi/paparazzi/commit/869adb279480539a3bb8106bf4e2a4b63d23c33a) and [81a695e](https://github.com/paparazzi/paparazzi/commit/81a695e55063c3f5a706bd7b27637814ae65af4c)
> 
>     * ChibiOS implementation behave the same way than the other arch.
> 
>     * get_sys_time functions for milli and micor seconds are also more accurate.
> 
>     * some tasks are grouped in a gnc group as they should be called in sequence: estimation, control, default, actuators
> 
>     * the timing between the sensors and gnc is half of the main period, it correspond to the default resolution of the sys_time virtual timers
> 
>     * the timing of AP threads is improved with a calculated wakeup time instead of a fixed sleep time
> 
>     * the monitor task for fixedwing is moved in the firmware specific autopilot part and called by a module
> 
>     * all timings have been accurately tested with high speed log and a dedicated profiler for ChibiOS
> 
> 
> It would be nice if you can give a try :)

Sure, now I am flying some experiments with a three-month-old master xD. Once I finish the experiments, and push a commit, I will synch my master, and try the code. However, I do not think it will be possible for me until the end of August :(.

Fabien-B (approved):
Tested with a rotorcraft, everything was ok.

PR #2722, opened 2021-05-15
[generators] add missing close files

### LINKED ISSUE #2720: [generators] Creating empty files
Somehow in certain cases certain generators will generate empty files for the aircrafts. After a lot of debugging I figured out that adding `close_out out` to `gen_airframe.ml`, `gen_modules.ml` and `gen_settings.ml` at the end will solve the problem.

To replicate this problem for example change the airspeed sensor in the `tudelft/nederdrone4.xml` file to `ets`.

Not sure why this is happening and why closing the file properly(should have been done at the first place anyways) helps. @gautierhattenberger do you have any idea and are you ok with adding the `close_out`? You might also know some other places where this should be added, but the above 3 files solved my problem.

gautierhattenberger:
Fix small bug as reported by @fvantienen 

close  #2720


fvantienen (approved):
Tnx for this fix!

PR #2727, opened 2021-05-19
[Maps] Use correct address for google tiles.

Fabien-B:
Use correct address to get google tiles, and remove all hackery that was needed with the previous address.


dewagter (approved):
Great simplification. Thanks

PR #2735, opened 2021-06-02
[guidance_indi_hybryd] Fixed wrong syntax in define definitions

Dennis-Wijngaarden:
Fixes two syntax definition errors in indi_guidance_hybrid code


EwoudSmeur (approved):
Thanks for finding this mistake!

PR #2737, opened 2021-06-03
Train

tmldeponti:
Added flight plan to train novel pilots in the cyberzoo. More info in the wiki https://github.com/tudelft/mavlab/wiki/Quadrotor-safe-Cyberzoo-pilot-training


EwoudSmeur:
@tmldeponti I think you made a mistake with your branches... :wink:

PR #2740, opened 2021-06-15
Fix flight plans

Fabien-B:
Remove most of the illegal characters in flight plans.

### DIFF
--- a/conf/flight_plans/AGGIEAIR/rotorcraft_opticlow_test.xml
+++ b/conf/flight_plans/AGGIEAIR/rotorcraft_opticlow_test.xml
@@ -37,7 +37,7 @@
       <attitude pitch="0" roll="0" throttle="0" until="FALSE" vmode="throttle"/>
     </block>
     <block name="Takeoff" strip_button="Takeoff" strip_icon="takeoff.png">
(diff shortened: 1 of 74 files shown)


gautierhattenberger (commented):
what about the deref operator `->` ? It is possible to use the keyword `@DEREF`, but in many cases, it is probably even better to use the macros like `GetPosX`.

Fabien-B:
I Replaced it for most cases I think.
What it seen very often is `stateGetPositionEnu_f()->z`.
The closest macro I found is for rotorcrafts: `#define GetPosAlt() (stateGetPositionEnu_f()->z+state.ned_origin_f.hmsl)`.
Should these occurrences be replaced by this macro ? Won't it change the behavior ? What is `state.ned_origin_f.hmsl` ?

gautierhattenberger:
It is a bad way to have the reference altitude, aka ground alt. So it is probably a good idea to add a macro `GetPosHeight` that juste returns `stateGetPositionEnu_f()->z`

Fabien-B:
I agree, I did that.

gautierhattenberger (approved):

PR #2742, opened 2021-06-28
[driver] support for RM3100 magneto over I2C

gautierhattenberger:
tested and working in the lab


fvantienen (approved):
Seems good to me :)

PR #2750, opened 2021-07-13
[combined] Small fixes

fvantienen:
This pull request has some small fixes for the following:
- [ekf2] Possibility to work with optitrack
- [hybrid] Initialization reset correctly on enter
- [conf] Add Nederdrone with trailing edge motors
- [ground] Add herelink RSSI tool

### DIFF
--- a/conf/airframes/tudelft/neddrone4_tem.xml
+++ b/conf/airframes/tudelft/neddrone4_tem.xml
@@ -71,8 +71,9 @@
       <configure name="UAVCAN_USE_CAN2" value="TRUE"/>
     </module>
     <module name="imu"           type="mpu6000"/>
-    <module name="gps"           type="ublox"/>
-    <module name="gps"           type="ubx_ucenter"/>
+    <module name="gps"           type="datalink"/>
+    <!--module name="gps"           type="ublox"/>
+    <module name="gps"           type="ubx_ucenter"/-->
     <module name="stabilization" type="indi_simple"/>
     <module name="stabilization" type="rate_indi"/>
     <module name="ins"           type="ekf2" />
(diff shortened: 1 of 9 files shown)


gautierhattenberger (approved):
It looks fine, but can you explain the idea behind the `sv_mul` parameter ?

fvantienen:
Yes it is a bit stupid, but the EKF2 needs a minimum amount of satellites in order to initialize. So if you have too less markers it won't start. Maybe I can disable that, I will quickly take a look.

gautierhattenberger:
Is it only for init or is it also used later while running ? Because, if you have some glitches, in the markers detection, you can end up with strong variations of the number of SV. It would be better to disable all checks related to that when using motion capture I guess.

fvantienen:
I is only for init and I saw I can disable those checks, so will do that :)

fvantienen:
This should do the trick, also added some documentation to the module.

gautierhattenberger (approved):

gautierhattenberger:
@fvantienen it seems that the build is really failing because of the missing MASK_* defines (https://semaphoreci.com/paparazziuav/paparazzi/branches/master/builds/3863). With the current version of this filter, they are only available in a cpp file, not in a header. Did you forget to update the library ? Otherwise, you have to make a local copy of the defines in `sw/ext/ecl/EKF/gps_checks.cpp`.

fvantienen:
Ahh shit I only tested the cyberzoo one where I needed to overwrite them, will fix it like they do in their own `common.h` just hardcode the decimal value for now. I think that is the easiest

fvantienen:
I created a new pull request, but still stupid that semphore is failing for other reasons at pull requests a lot

gautierhattenberger:
I agree, but I can't find a proper solution. It seems to be failing because of eigen, which is our only gitlab submodule. Maybe we can try to make our own github fork to make this work, but this is annoying.

PR #2751, opened 2021-07-15
[boards] Add PPM input to Pixhawk 4 (px4fmuv5)

fvantienen:
This adds the PPM input port from the pixhawk 4. Tested and works.


Fabien-B (approved):
Looks good to me

PR #2754, opened 2021-07-22
Replay settings

Fabien-B:
In a replay session, settings file is not generated.
This PR generates it.
The settings updates well in the GCS.
Obviously, changing a setting won't affect it...

It seems I have reverted some 10 years old changes. Is it still relevant or not ?
https://github.com/paparazzi/paparazzi/commit/f0dcd7eceb52bb0defe5746799e9bdbd62fb9475


gautierhattenberger:
The old commit message says that it is to prevent warnings. Is it still the case ?

Fabien-B:
I haven't saw any, but I should test again to be sure.

Fabien-B:
I ran a full replay, no warnings.

gautierhattenberger (approved):
let's see if it works

OpenUAS:
Does not work, please rollback PR. Tested on 2x different systems, when rolled back it work again.

OpenUAS:
New fix makes it work.

PR #2756, opened 2021-08-03
Nederdrone name written fully

EwoudSmeur:
This commit changes the name of the 'neddrone' airframe files to 'nederdrone', to avoid confusion as that is the official name.


dewagter (approved):
OK

PR #2759, opened 2021-08-10
Fix MPU6000 spikes in certain chipbatches

OpenUAS:
On certain Flight-controller, either bad batch of MPU6000, overheat, Non clean power and whatnot max value spikes can occur. This completely messes up the IMU attitude state. those boards are unique, to expensive or not feasible to replace chip.To still be able to fly these flight-controllers an 3value median filter is added to the RAW output. Default MPU module behaviour is still the same but by adding` <define name="IMU_MPU_USE_MEDIAN_FILTER" value="TRUE"/>` to the mpu6000 module in airframe the issue is resolved with this fix, see screenshot of before and after... BTW mind the Scale, no it is not more noise... ;-)

An airframe was added to PR be able to validate and testfly the fix.

### **Before**
BeforeFix

### **After**
AfterFix

### DIFF
--- a/sw/airborne/subsystems/imu/imu_mpu6000.c
+++ b/sw/airborne/subsystems/imu/imu_mpu6000.c
@@ -28,11 +28,13 @@
 #include "subsystems/imu.h"
 #include "subsystems/abi.h"
 #include "mcu_periph/spi.h"
+#ifdef IMU_MPU_USE_MEDIAN_FILTER
+#include "filters/median_filter.h"
+#endif
 
 /* SPI defaults set in subsystem makefile, can be configured from airframe file */
 PRINT_CONFIG_VAR(IMU_MPU_SPI_SLAVE_IDX)
 PRINT_CONFIG_VAR(IMU_MPU_SPI_DEV)
-
 
 /* MPU60x0 gyro/accel internal lowpass frequency */
 #if !defined IMU_MPU_LOWPASS_FILTER && !defined  IMU_MPU_SMPLRT_DIV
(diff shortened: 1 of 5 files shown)


OpenUAS:
BTW the airframe was added to be able to validate and testfly the fix.

sw/airborne/subsystems/imu/imu_mpu6000.c
```diff
@@ -28,12 +28,14 @@
 #include "subsystems/imu.h"
 #include "subsystems/abi.h"
 #include "mcu_periph/spi.h"
+#ifdef IMU_MPU_USE_MEDIAN_FILTER
```
gautierhattenberger, line 31: please replace the `#ifdef` by `#if` in this case, same for the other occurrences

sw/airborne/subsystems/imu/imu_mpu6000.c
```diff
@@ -102,11 +104,21 @@ PRINT_CONFIG_VAR(IMU_MPU_Y_SIGN)
 #endif
 PRINT_CONFIG_VAR(IMU_MPU_Z_SIGN)
 
-
 struct ImuMpu6000 imu_mpu_spi;
 
+#ifdef IMU_MPU_USE_MEDIAN_FILTER
+struct MedianFilter3Int medianfilter_accel;
+struct MedianFilter3Int medianfilter_rates;
```
gautierhattenberger, line 111: can be marked `static`

gautierhattenberger (changes requested):

sw/airborne/subsystems/imu/imu_mpu6000.c
OpenUAS: Thanks, will make changes ASAP

gautierhattenberger (approved):

PR #2761, opened 2021-08-16
[fix] correct ADC buffer offset for ChibiOS

gautierhattenberger:
- with ChibiOS 20.3 the half buffer starting position is computed by
  hand and was wrong in this case: it should be the product of the half
  buffer size by the number of channels
- some other minor fixes:
  - using && instead of & in boolean comparison
  - use an intermediate const variable for the number of channels to
    suppress a warning (comparison always true due to limited data range)
    when the number of channels is zero (no ADC in use)


fvantienen (approved):
Tested and fixes the problem!

fvantienen:
I can't merge it, but is ok to be merged

PR #2762, opened 2021-08-17
Pixhawk ADC 5 and ADC 6

fvantienen:
This adds the external ADC ports on the pixhawk 4. Needs to be merged after #2761 in order to work.


conf/modules/actuators_uavcan.xml
```diff
@@ -15,6 +15,7 @@
   </header>
   <makefile target="ap">
     <file name="actuators_uavcan.c" dir="subsystems/actuators"/>
+    <file name="pprz_random.c" dir="math"/>
```
gautierhattenberger, line 18: with https://github.com/paparazzi/paparazzi/pull/2721 it should be specified as an explicit dependency to the math module

sw/airborne/subsystems/actuators/actuators_uavcan.c
```diff
@@ -102,7 +103,11 @@ static void actuators_uavcan_send_esc(struct transport_tx *trans, struct link_de
   float energy = telem[i].energy;
   pprz_msg_send_ESC(trans, dev, AC_ID, &telem[i].current, &electrical.vsupply, &power,
                     &rpm, &telem[i].voltage, &energy, &esc_idx);
-  esc_idx++;
+  
+  // Randomness added for multiple  transport devices
+  if (rand_uniform() > 0.05) {
+    esc_idx++;
+  }
```
gautierhattenberger, line 110: If I get it correctly, it is changing esc randomly instead of having a fixed interval ? I agree that changing at every step is not nice, but why not changing every X run of the function to have fixed length of data in the telemetry ?

fvantienen: The problem is actually when you have multiple transport devices such as a `datalink` and `flight_recorder`. Having it fixed it would depend on the frequency defined in the telemetry file to be able to receive all the motors. This randomness prevents the configuring of 2 frequencies which aren't interfering.

gautierhattenberger (approved):

PR #2763, opened 2021-08-17
[ins] add support for gps course as heading (for indoor use)

gautierhattenberger:
Copied from ahrs_int_cmpl_quat


mrtbrnz:
Tested in ENAC flight arena. Working nicely !

PR #2764, opened 2021-08-19
FIX simulation mag saturation

EwoudSmeur:
We were using the default for the Nederdrone, but apparently saturating the mag sensor in simulation causing heading issues. I think it is not really helpful to have this saturation limit here (one would need to make a new nps_sensors h file for every sensor setup) while most of the time we're not interested in simulating this limit anyway. Would it make sense to just set it to a large value in the default?

### DIFF
--- a/conf/simulator/nps/nps_sensors_params_default.h
+++ b/conf/simulator/nps/nps_sensors_params_default.h
@@ -93,9 +93,9 @@
 /*
  *  Magnetometer
  */
- /* HMC5843 has 12 bit resolution */
-#define NPS_MAG_MIN -2047
-#define NPS_MAG_MAX  2047
+ /* Large number to avoid saturation */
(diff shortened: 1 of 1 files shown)


gautierhattenberger (approved):
What is the resolution of your mag for this drone ?
Still I agree that for the default setting in NPS, it is probably not necessary to set a low limit, this type of sensor is unlikely to be saturated in real life.

EwoudSmeur:
Thanks! We use the LIS3MDL, which has a range of 4 to 16 gauss and a resolution of 16 bit, so I think it should be fine in flight.

EwoudSmeur:
Actually, we may also make the limits for the accelerometer and gyroscope very big for the default, just in case. If someone would like to simulate sensor limits they can still make a nps sensor parameter file. (idea of @fvantienen)

PR #2767, opened 2021-08-23
[replay] Use generated settings.

Fabien-B:
Use "generated_settings" node instead of "settings" for replay settings.
Generate an empty node if not found.
According to #2569, "settings" node is "wrong", so its better if its not generated for replay.

Can you try it @OpenUAS to be sure it still works for you?


gautierhattenberger (approved):
I forgot the last changes about settings in log file, better not use the old `settings` node, since it can be wrong even with older logs

gautierhattenberger:
tested and working

PR #2769, opened 2021-08-24
[nps] clean sensors params files

gautierhattenberger:
- try to set the sensor frequency automatically at the correct value
- make more parameters configurable
- set higher bounds for all sensors to avoid saturation
- remove old unused files
- update airframe files

This is following #2764. It turns out that sensor params where completely wrong for saturation and freq, which was make simulation run wild in many situations. With this, the default file should adapt to most cases, with nearly perfect sensors. A noisy file example is kept for testing.


conf/simulator/nps/nps_sensors_params_default.h
```diff
@@ -30,12 +30,27 @@
 #define NPS_BODY_TO_IMU_THETA  IMU_BODY_TO_IMU_THETA
 #define NPS_BODY_TO_IMU_PSI    IMU_BODY_TO_IMU_PSI
 
+// try to determine propagate frequency
```
EwoudSmeur, line 33: Should this not depend on the IMU instead of the AHRS/INS? I think some INS/AHRS (ekf2?) may implement a callback for a gyro sensor update, but propagate the whole filter a slower rate.

EwoudSmeur (commented):
This is great Gautier! I'm just wondering about the sensor frequency.

conf/simulator/nps/nps_sensors_params_default.h
gautierhattenberger: The exact IMU freq is not known, it depends of many sensor-dependent settings. So using the propagation frequency is the best guess I think. I made it configurable, so it can be changed with `NPS_XXXX_DT` if needed.

EwoudSmeur (approved):
Ok, thanks!

PR #2777, opened 2021-09-08
Better link status

Fabien-B:
Improve TELEMETRY_STATUS when multiple (redundant) links are used.
Previously, there was no way to tell if one of the links was down.
Now, it should behave as expected:
- if one of the link is down, it goes red in the Link tab. It goes orange in the strip (with correct number of links up).
- if link combiner is down, all links are shown as down (link may still be up, but no message are received), and the indicator in the strip goes red.
- if all links are lost, they all goes red, and the indicator in the strip goes red.


Fabien-B:
It should be ok now.

gautierhattenberger (approved):

PR #2779, opened 2021-09-11
EKF2 Optical Flow Interface 

pcampolucci:
This pull request mainly concerns the extension of the EKF2 state estimation capabilities through optical flow input. The branch has been extensively tested and validated successfully. During the development also other modules have been improved. In total, this PR brings improvements on:

- ins_ekf2 module: new optical flow interface
- mateksys_3901_l0x: added rotation compensation for height + calibration strategy
- gps_datalink: used to get fake gps, now gives also fake magnetometer if selected by the user
- Nederquad: new airframe for quadrotor used at TUDelft

Added wiki on flow sensor: https://wiki.paparazziuav.org/wiki/Module/Mateksys_3901_l0x
Added wiki on ins_ekf2: https://wiki.paparazziuav.org/wiki/Subsystem/ins#ekf2 (to be finished)


gautierhattenberger (changes requested):
It seems that this branch is not based on the latest master and are changing the references on pprzlink and eigen while it shouldn't be the case.
Please rebase or merge with master first.

pcampolucci:
Dear @gautierhattenberger . Thank you a lot for the precious comments, I have implemented all of them but the IMU telemetry update, on which I asked you some more info about. 

Regarding the merge, I did the following:

1. set upstream paparazzi/paparazzi
2. git fetch upstream 
3. git merge upstream/master

Updates were available and I therefore merged, but that didnt seem to solve the issue. 

I tried git rebase on a backup branch but it was cancelling most of my updates. 

Any idea on what I might be doing wrong here with submodules? (only pprzlink has been actually modified)

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -168,8 +192,14 @@ static void parse_gps_datalink(uint8_t numsv, int32_t ecef_x, int32_t ecef_y, in
   gps_datalink.last_3dfix_ticks = sys_time.nb_sec_rem;
   gps_datalink.last_3dfix_time = sys_time.nb_sec;
 
-  // publish new GPS data
   uint32_t now_ts = get_sys_time_usec();
+
+  // if selected, publish magnetometer data
+  if (GPS_DATALINK_USE_MAG) {
```
gautierhattenberger, line 198: can be replaced by `#if GPS_DATALINK_USE_MAG` since it doesn't change at runtime, same below

gautierhattenberger (changes requested):
Sometimes merging submodules is a bit complicated. I only way I know is that you should manually reset each submodules to the correct revision. For example, enter the folder `sw/ext/pprzlink`, checkout the same revision than in master branch, go back to the root of the project and commit the submodule.

Regarding imu data. The normal way would be to copy the raw data in `imu.mag_unscaled` then call `imu_scale_mag` function (see https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/sensors/mag_hmc58xx.c#L110).
In this case, raw data are always scaled, so you can directly copy the datalink mag values to `imu.mag_unscaled` and `imu.mag`.

sw/airborne/subsystems/gps/gps_datalink.c
```diff
@@ -77,6 +78,10 @@ static void send_magnetometer(int32_t course, uint32_t now_ts)
   mag_real.z = 0;
   MAGS_BFP_OF_REAL(mag, mag_real);
 
+  // update IMU information
+  VECT3_COPY(imu.mag_unscaled, mag);
+  imu_scale_mag(&imu);
```
gautierhattenberger, line 83: If you call the `imu_scale_mag`, the coefficients from the airframe will be applied. So in this case, you should have a sensitivity of 1 and an offset of 0. If you are fine with that I can validate this code. Otherwise, you can make a direct copy to `imu.mag` as I mentioned before, so airframe calibration has no effect. It is up to you, I don't have a strong preference for this.

pcampolucci:
@gautierhattenberger everything should be now ok. I made changes in pprzlink for the message of optical flow and the one of INS_EKF2. I am making a pull request on pprzlink now to have that edit merged in master.

gautierhattenberger:
@pcampolucci Î have merged the PR in pprzlink, so you can update the reference here and fix the small comment in px4flow modules. Beside that it looks okay to me.

gautierhattenberger (approved):

pcampolucci:
@gautierhattenberger everything should be now in place. On my machine compiles without warnings, not sure why it doesn't pass the test.

gautierhattenberger (approved):
looks fine, CI build fail is not related

PR #2782, opened 2021-09-22
ADD tilt twist control (twist around z)

EwoudSmeur:
This commit implements tilt twist control for vehicles that rotate slow
in the z axis compared to the other axes (like tailsitters).

The attitude error is decomposed in a tilt and a twist part, and the
tilt error is compensated assuming that the rotation around z is zero
(very slow in comparison).

I can add the test to the other algebra tests if needed.


sw/airborne/math/pprz_algebra_float.c
```diff
@@ -618,6 +618,39 @@ void float_quat_of_rmat(struct FloatQuat *q, struct FloatRMat *rm)
   }
 }
 
+#include <stdio.h>
```
gautierhattenberger, line 621: to be removed

sw/airborne/math/pprz_algebra_float.c
```diff
@@ -618,6 +618,39 @@ void float_quat_of_rmat(struct FloatQuat *q, struct FloatRMat *rm)
...
+/**
+ * @brief Tilt twist decomposition of a quaternion (z axis)
+ *
```
gautierhattenberger, line 624: maybe provide a bit of mathematical details ? it is not so obvious what this function does

sw/airborne/math/pprz_algebra_float.h
```diff
@@ -481,6 +481,9 @@ extern void float_quat_of_orientation_vect(struct FloatQuat *q, const struct Flo
 /// Quaternion from rotation matrix.
 extern void float_quat_of_rmat(struct FloatQuat *q, struct FloatRMat *rm);
 
+/// Tilt twist decomposition of quaternion
+void float_quat_tilt_twist(struct FloatQuat *tilt, struct FloatQuat *twist, struct FloatQuat *quat);
```
gautierhattenberger, line 485: add `extern`, for consistency, it is technically not needed

gautierhattenberger (changes requested):
the test binary file should be removed

EwoudSmeur:
@gautierhattenberger thanks for your comments, sorry I guess I was a bit hasty pushing this! It is tested though 😇

gautierhattenberger (approved):
thanks

PR #2635, opened 2020-12-28
Better sonar range readings on ADC sonar with noisy output

OpenUAS:
To make an analog sonar usable by adding and option to filter the often noisy output. Normal sane advice could be to use either PWM or I2C version of the sensor but sometimes the lack of input ports,... there is now a usable analog alternative. Tested in Pixracer ADC port on AC Minimag.

PR #2785, opened 2021-09-27
Added ADC4 and synced ADC baremetal FC names with Chibios FC names

OpenUAS:
As a preparation for finalizing PR #2635 added a ADC4 to the Chibios boardfile. And still to be able to go from bar metal board to Chibios without changing airframe config the px4fmu_4.0 baremetal board has now the same ADC namings as the Chibios version.

Since moving forward to PPRZ v6 all boardfiles of old baremetal are set to use Chibios. 

All real airframes (Hardware) are owned, available and managed by PR requestor, including 3rd party ones.
A possible negative impact on 3rd party is thus zero.

Validated and real life tested on Pixracer v1 flightcontroller The V,C and also ADC workings by means of Analog sonar input, everything works as it should.

### DIFF
--- a/conf/airframes/OPENUAS/openuas_durafly_g2.xml
+++ b/conf/airframes/OPENUAS/openuas_durafly_g2.xml
@@ -29,7 +29,7 @@
   <firmware name="fixedwing">
 
     <!-- ********************** For in the real flying hardware ********************-->
-    <target name="ap" board="px4fmu_4.0">
+    <target name="ap" board="px4fmu_4.0_chibios">
       <define name="FLASH_MODE" value="PX4_BOOTLOADER"/>
 
       <configure name="CPU_LED" value="1"/>
--- a/conf/airframes/OPENUAS/openuas_durafly_goblin.xml
+++ b/conf/airframes/OPENUAS/openuas_durafly_goblin.xml
@@ -30,7 +30,7 @@ NOTES:
 
   </description>
   <firmware name="fixedwing">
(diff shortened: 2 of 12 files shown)


fvantienen:
I don't think the ADC's are fully correct with chibios now. Because the VDD_5V is something different than the voltage sensor. This is an internal measurement of the 5V regulator, so that way you can't measure the battery voltage. These ADC's are mainly to check the correct functioning of the board and overcurrent lowering the voltage on the busses. 
(`ADC_CHANNEL_VSUPPLY ADC_1` should stay and ADC3 can be disabled by default unless someone writes something useful for it)

My advise is to not change the current configuration from the chibios version.

OpenUAS:
The current working version of Chibios is NOT changed one the barmetal has now the same adc naming convention as the Chibios version. The bare metal version, self created ;) was IMHO better naming since I made it compatible with Lisa MX so moving over you arframe was simple. But in the event of v6 Chibios only , I did not wan to disrupt chibios at all so therefor choose to use the Chibios version as a ase. If there arestill errors in the Chibios version they should be fixed. Before the PR I've teste same Pixeracer on Barmetal, 100% OK voltageand current reading and tested on Chibios also V/C reading, gave the same correct result.

gautierhattenberger:
please rebase and fix issues if you want to include this for next release

OpenUAS:
Dropping Pull request since the dawn of v6.x and compatibility with baremetal config is not worth the effort. Will redo test for Chibios ADC in separate one.

PR #2786, opened 2021-10-04
Drop old compat

gautierhattenberger:
remove compatibility code and makefile parts for Ubuntu16.04, both pprzlink and main paparazzi


OpenUAS:
Fine, already adjusted this page to reflect the change https://wiki.paparazziuav.org/wiki/Installation#Quickstart_for_Ubuntu_users

PR #2792, opened 2021-10-31
Server json stream

gautierhattenberger:
can be used by external plotter like PlotJuggler (default port)
always enabled by default
merge https://github.com/paparazzi/pprzlink/pull/154 and update pprzlink ref before accepting this PR

### DIFF
--- a/sw/ground_segment/tmtc/server.ml
+++ b/sw/ground_segment/tmtc/server.ml
@@ -141,15 +141,15 @@ let logger = fun () ->
   close_out f;
   open_out (logs_path // data_name)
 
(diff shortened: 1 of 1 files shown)


Fabien-B:
Pprzlink ref updated, it would be better to name the aircraft by its name rather than its ID, and it might be good to wrap the message in a message with the "timestamp" key like in this example: https://github.com/facontidavide/PlotJuggler/blob/main/plotjuggler_plugins/DataStreamUDP/udp_client.py

Fabien-B (approved):

PR #2797, opened 2021-11-12
Move subsystems to modules

gautierhattenberger:
Move all files in subsystems to modules directories, since everything is now really processed as a module.


noether:
I will check it next week. I have to do some airframe files tests, so I will do them with this patch :P.

noether:
It works well here, and I do not see anything weird in the code.

noether (approved):

PR #2799, opened 2021-11-19
check geo conversions

EwoudSmeur:
We had some issues trying to accurately land with respect to a ground-based GPS, and figured there must be something wrong with some of the / a conversion(s) somewhere. I tried this first one and it seems off by a factor of roughly 2.5. Not sure if I'm doing something wrong here. I'm planning on making a few more tests in the coming days.

### DIFF
--- a/sw/airborne/test/Makefile
+++ b/sw/airborne/test/Makefile
@@ -23,6 +23,9 @@ test_algebra: test_algebra.c ../math/pprz_trig_int.c ../math/pprz_algebra_int.c 
 test_bla: test_bla.c ../math/pprz_trig_int.c ../math/pprz_algebra_int.c ../math/pprz_algebra_float.c ../math/pprz_algebra_double.c
 	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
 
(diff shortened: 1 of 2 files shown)


dewagter:
Thanks for this test program. Great to check conversion accuracy!

2.5 sounds awfully close to the POS_FRAC of 8 bits (math.h) combined with the GPS in centimeters (gps-message) which would be 2.56.

I think the problem is in how you fill the coordinates in the ```_i``` structs.

gautierhattenberger:
see something like: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/multi/traffic_info.c#L326

EwoudSmeur:
Thanks for the tips, indeed I didn't pay attention that the output of ned_of_lla_point_i is in cm instead of POS_FRAC. Is there a reason for this? Wouldn't it make sense to use POS_FRAC here as well? Now it does match close enough though!

gautierhattenberger:
cm was the native resolution of GPS position when we wrote the math lib...

gautierhattenberger (approved):

PR #2801, opened 2021-11-22
fix guidance module comment

EwoudSmeur:
I came across this comment, which I thought was incorrect, looking at autopilot_static.c


noether (approved):

noether:
Yes, it is the same for fixed-wing actually. I would wait for Gautier to confirm it.

gautierhattenberger:
I confirm

PR #2803, opened 2021-11-29
[fix] correct pseudo_mag projection

gautierhattenberger:
The projection of the magnetometer was not correct. The invariant INS require to have the mag correctly projected in 3D and not only in the horizontal plane.

The PR #2779 is also doing the wrong projection when sending Abi MAG message. It would be better to merge the different ways to produce virtual MAG data from GPS course, or maybe create a dedicated message REMOTE_MAG to send the correct measurement from the ground independently from the REMOTE_GPS message.

### DIFF
--- a/sw/airborne/modules/ins/ins_float_invariant.c
+++ b/sw/airborne/modules/ins/ins_float_invariant.c
@@ -478,6 +478,8 @@ void ins_float_invariant_update_gps(struct GpsState *gps_s)
   }
 
 #if !USE_MAGNETOMETER
+  struct FloatVect3 pseudo_mag = { 0.f, 0.f, 0.f };
+  bool update_mag = false;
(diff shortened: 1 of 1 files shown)


EwoudSmeur (approved):
It looks good to me, ok with merging.
However, it looks like the final comparison with the expected field is in 2D, so it may be possible to simplify this a bit. However, this is more complete. 

I agree with your idea of the remote mag message. I think it would be a neater way of doing it. Having this same computation in every INS seems redundant.

gautierhattenberger:
Actually, the "virtual" mag vector is in 3D (even if the virtual field is along the x/north axis only). The comparison (computation of `Eb` in error function) is in full 3D and requires a correct projection (so not only turning measure vector around psi).
I will open en issue as a reminder to make a correct remote mag sensor.

PR #2806, opened 2021-12-01
Update chibios fatfs lfn2 dmastack

alex31:
proposal to resolve garbage in sdLog filename and content.


gautierhattenberger:
still need testing on F4 and F7 boards

gautierhattenberger:
tested and working

Issue #2134, opened 2017-10-24
NPS: AHRS and INS are not bypassed during sensor callbacks, producing incorrect results


tomvand:
The current implementation of the AHRS and INS bypass results in incorrect state information in several sensor ABI message callbacks, among which the IMU_ACCEL message.

In [`nps_autopilot_rotorcraft.c:nps_autopilot_run_step(...)`](https://github.com/paparazzi/paparazzi/blob/master/sw/simulator/nps/nps_autopilot_rotorcraft.c#L93) the sensors are updated ([line 105-148](https://github.com/paparazzi/paparazzi/blob/master/sw/simulator/nps/nps_autopilot_rotorcraft.c#L105)) before the AHRS and INS are overwritten in line [150-156](https://github.com/paparazzi/paparazzi/blob/master/sw/simulator/nps/nps_autopilot_rotorcraft.c#L150).
The sensor updates send out the corresponding ABI messages, which causes the AHRS (and likely INS, not confirmed) to update its estimate and overwrite the vehicle state in state.h, even though it is disabled! As a result, the stateGetXXX functions return incorrect values when called from the sensor callbacks, as the vehicle state is only overwritten by its true value after all ABI callbacks have been completed.

I have not found a clean solution to this problem:
* Overwriting the vehicle state *beforehand* does not fix the problem as it does not overwrite the AHRS' internal state. Even if the AHRS' state could be overwritten, this would not prevent it from updating the vehicle state before the next subscribers are called.
* Both the AHRS and other modules subscribe to the sensor messages, which makes it difficult to update the modules but skip the AHRS.
* Setting the AHRS' sender ID's to ABI_DISABLE would prevent it from subscribing to the sensor messages, but 1) the bypass of the AHRS can be changed at runtime and 2) the sender ID define names differ per AHRS implementation.
* Sensor message callbacks are not shared between the AHRS implementations, so there is no central place to check the bypass flag when new sensor values arrive at the AHRS. Checking this flag in every AHRS implementation would solve this problem, but results in duplicate code and could be forgotten in new implementations. Similarly, there is no shared output function that overwrites the vehicle state in state.h that could check this flag.
* It looks like [ahrs.c](https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/subsystems/ahrs.c#L101) *can* actually enable and disable AHRS. Perhaps this can be (ab)used to disable the AHRS? I could not find a similar mechanism for the INS though.

[...]

PR #2808, opened 2021-12-10
Fix #2807 inconsistent coordinate transform in nps_fdm_gazebo

### LINKED ISSUE #2807: Bebop Orange Avoider (fdm gazebo) position flicker in GCS
While testing the `bebop_orange_avoider` and `bebop_orange_avoider_guided` in simulation I noticed that the drone position was not being displayed correctly in the GCS

 

This bug seems to have been introduced during the autopilot reorg pull request #2721.

After some attempted debugging with @OpenUAS we found out a couple of possible sources for this issue. The position flicker is reflected in the ROTORCRAFT_FP message, and it seems to be caused by the `NPS_BYPASS_INS` variable, defined in file [conf/simulator/gazebo/airframes/bebop.xml](https://github.com/paparazzi/paparazzi/blob/master/conf/simulator/gazebo/airframes/bebop.xml).

While setting `NPS_BYPASS_INS` to false stops the flicker, there is then a mismatch between the position of the drone in the GCS and in Gazebo, which leads us to believe there is a conflicting value in the initialization.

We attempted to compare  these airframes to the `ardrone2_gazebo` airframe, which also uses
`<module name="fdm" type="gazebo"/>` but does not exhibit the same behavior. We saw that for the ardrone2 the `<section name="SIMULATOR">` block there is a dedicated file for the sensor parameters
https://github.com/paparazzi/paparazzi/blob/6e6e2f11ce7c18c8eabf5a2233fb80d378d4e102/conf/airframes/examples/ardrone2_gazebo.xml#L194
which hardcodes different variables related to sensor timings while for the bebop this is set to the magical 512Hz

https://github.com/paparazzi/paparazzi/blob/6e6e2f11ce7c18c8eabf5a2233fb80d378d4e102/conf/simulator/nps/nps_sensors_params_default.h#L34-L42

so possibly there are multiple issues with the orange_avoider airframes. We found a very similar sounding issue posted by @tomvand #2134, maybe the autopilot reorg reintroduced it?

tomvand:
Work in progress, see issue #2807. Final PR may take until January because of holidays.

### DIFF
--- a/conf/airframes/tudelft/bebop_course_orangeavoid.xml
+++ b/conf/airframes/tudelft/bebop_course_orangeavoid.xml
@@ -23,9 +23,6 @@
       <module name="fdm" type="gazebo"/>
       <define name="VIDEO_CAPTURE_PATH" value="/tmp/paparazzi/images"/>
       <define name="FILE_LOGGER_PATH" value="/tmp/paparazzi/log"/>
-
-      <!--<define name="NPS_NO_GPS" value="TRUE"/>
-      <define name="NO_GPS_NEEDED_FOR_NAV" value="TRUE"/>-->
       
       <!-- Detect orange -->
       <define name="COLOR_OBJECT_DETECTOR_LUM_MIN1" value="41"/>
@@ -250,7 +247,7 @@
   <section name="SIMULATOR" prefix="NPS_">
     <define name="ACTUATOR_NAMES" value="nw_motor, ne_motor, se_motor, sw_motor" type="string[]"/>
     <define name="JSBSIM_MODEL" value="bebop" type="string"/>
-    <define name="SENSORS_PARAMS" value="nps_sensors_params_ardrone2.h" type="string"/>
+    <define name="SENSORS_PARAMS" value="nps_sensors_params_default.h" type="string"/>
     <define name="MT9F002_SENSOR_RES_DIVIDER" value="4"/>
   </section>
   <include href="conf/simulator/gazebo/airframes/bebop.xml"/>
--- a/conf/airframes/tudelft/bebop_course_orangeavoid.xml
+++ b/conf/airframes/tudelft/bebop_course_orangeavoid.xml
@@ -23,6 +23,9 @@
       <module name="fdm" type="gazebo"/>
       <define name="VIDEO_CAPTURE_PATH" value="/tmp/paparazzi/images"/>
       <define name="FILE_LOGGER_PATH" value="/tmp/paparazzi/log"/>
+
+      <!--<define name="NPS_NO_GPS" value="TRUE"/>
+      <define name="NO_GPS_NEEDED_FOR_NAV" value="TRUE"/>-->
       
       <!-- Detect orange -->
       <define name="COLOR_OBJECT_DETECTOR_LUM_MIN1" value="41"/>
@@ -247,7 +250,7 @@
   <section name="SIMULATOR" prefix="NPS_">
     <define name="ACTUATOR_NAMES" value="nw_motor, ne_motor, se_motor, sw_motor" type="string[]"/>
     <define name="JSBSIM_MODEL" value="bebop" type="string"/>
-    <define name="SENSORS_PARAMS" value="nps_sensors_params_default.h" type="string"/>
+    <define name="SENSORS_PARAMS" value="nps_sensors_params_ardrone2.h" type="string"/>
     <define name="MT9F002_SENSOR_RES_DIVIDER" value="4"/>
   </section>
   <include href="conf/simulator/gazebo/airframes/bebop.xml"/>
--- a/sw/simulator/nps/nps_fdm_gazebo.cpp
+++ b/sw/simulator/nps/nps_fdm_gazebo.cpp
@@ -59,6 +59,7 @@ extern "C" {
 #include "modules/actuators/motor_mixing_types.h"
 
 #include "math/pprz_geodetic_float.h"
+#include "math/pprz_geodetic_double.h"
 #include "state.h"
 }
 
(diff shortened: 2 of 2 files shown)


matteobarbera:
The position jitter is still present when I run it in session Simulation, although the jump is smaller 

`fdm` uses `EcefCoor_d` while `state` position is stored in `float`

tomvand:
Ok, yes I'm seeing the same now.

It's possible that it's indeed the use of floats. The header warns to use double because of accuracy problems, but I could only find a float ltp definition and no conversion function to double.

It could also be #2134 though, but I haven't thought of a way to see which is causing this problem yet. Disabling the INS bypass would show it but causes unstable (vertical) behavior on my pc...

matteobarbera:
Turning BYPASS_INS off fixes the jumping for `east` and reduces jitter in `north` and `up`, showing the same behavior tested and shown in #2807 (`north` and `up` also move a bit after ins bypass disabled)

gautierhattenberger:
An error of 2 cm between "perfect" position and estimated from internal INS is probably acceptable, but the jumps are clearly coming from the fact that the bypass is not overwriting the state at every steps, which I can't explain at the moment.

gautierhattenberger:
@tomvand do you have more modifications, or can we add this PR to the next release ?

pcampolucci:
Hello @gautierhattenberger and @tomvand. I recently gave a look with @matteobarbera at the issue and this is what I found. 

The variable `fdm.ltpprz_pos` obtained from Gazebo is used to overwrite the state in the function `sim_overwrite_ins` in the file `nps_autopilot_rotorcraft.c` through `stateSetPositionNed_f(&ltp_pos);` I have observed that the state override does not introduce flickering in the state that is then used in the simulation, but both the telemetry messages and then the visuals in the GCS show it still. I then went to the function for the publication of the message `ROTORCRAFT_FP` and simply added the line `int enu_x = stateGetPositionEnu_i()->x;` in the function as follows

```
static void send_fp(struct transport_tx *trans, struct link_device *dev)
{
  int32_t carrot_up = -guidance_v_z_sp;
  int32_t carrot_heading = ANGLE_BFP_OF_REAL(guidance_h.sp.heading);
  int32_t thrust = (int32_t)autopilot.throttle;
#if GUIDANCE_INDI_HYBRID
  struct FloatEulers eulers_zxy;
  float_eulers_of_quat_zxy(&eulers_zxy, stateGetNedToBodyQuat_f());
  int32_t state_psi = ANGLE_BFP_OF_REAL(eulers_zxy.psi);
#else
  int32_t state_psi = stateGetNedToBodyEulers_i()->psi;
#endif
  int enu_x = stateGetPositionEnu_i()->x;
  pprz_msg_send_ROTORCRAFT_FP(trans, dev, AC_ID,
                              &(stateGetPositionEnu_i()->x),
                              &(stateGetPositionEnu_i()->y),
                              &(stateGetPositionEnu_i()->z),
                              &(stateGetSpeedEnu_i()->x),
                              &(stateGetSpeedEnu_i()->y),
                              &(stateGetSpeedEnu_i()->z),
                              &(stateGetNedToBodyEulers_i()->phi),
                              &(stateGetNedToBodyEulers_i()->theta),
                              &state_psi,
                              &guidance_h.sp.pos.y,
                              &guidance_h.sp.pos.x,
                              &carrot_up,
                              &carrot_heading,
                              &thrust,
                              &autopilot.flight_time);
}
```

and fixed the flickering completely. So in brief I think that the state is not polluted by this flickering but the telemetry message and the visuals are and that can be removing by calling `stateGetPositionEnu_i` before sending the message, which is weird and I cannot give myself an explanation about this. 

In addition. 

We have noticed that once the flickering is removed some of the state have somewhat a small, constant drifting behavior. 

Peek 2022-01-15 23-03

I have backtraced and it looks like that this drifting happens within the `pose` variable specified in `nps_fdm_gazebo.cpp` as follows

```ignition::math::Pose3d pose = model->WorldPose(); // In LOCAL xyz frame```

So that goes to deep for me to find the cause for now. 

What do you think is the cause of the flickering then given the above?

tomvand:
Hey I'm back, but still have a hard time wrapping my head around this issue :confounded: 

@pcampolucci , that's an interesting find. I know that the state is converted from NED to ENU when the stateGetPosition function is called, but I would expect the same conversion to take place within the pprz_msg_send_xx call. I don't know why your extra line would fix things, I have not had time to verify it here yet. Does it also work when you replace it with `stateCalcPositionEnu_i` (no return value)?

@gautierhattenberger I had hoped this PR would solve the issue completely, but it seems there is more going on. Nevertheless this PR is already an improvement over the current code. Before merging, I think it would be good to apply the same fix to the other vectors like velocity and acceleration (I'm not sure if/how the same fix should be applied to the attitude and angular rates...); right now only the position is corrected.

@pcampolucci regarding the drift: does this happen in flight or when the drone is landed? Since the movement is very slow, it might be an accuracy error in Gazebo's collision handling with the ground?

tomvand:
I've changed the code to perform the calculations in double (@gautierhattenberger is there an LtpDef_d struct in state.h that I missed? For now I'm casting it every frame). I've also applied the same spherical transform fix to the other vector transformations.

@matteobarbera , @pcampolucci Can you check if the latest commit works, and whether it improves things? On my pc the result in the GCS at least looks ok:
Screenshot from 2022-01-18 10-14-23

If there are no obvious errors in the latest commit, then this PR can be merged. I'll mark it as ready for review now.

gautierhattenberger:
@tomvand `LtpDef_d` is not used in state.h at the moment, but it does exists in the math lib, and it used locally in modules/gps/gps_sim_nps.c

tomvand:
Ok thanks, good to know.

gautierhattenberger:
I just looked at your code. I'm fine with making the copy each time. If it is only for gazebo, I don't think it is necessary to include it in main state.h at the moment.
I'll wait for the feedback of @pcampolucci and merge if it is okay

matteobarbera:
@tomvand Just tested the new commits, the flicker is still present to some degree but like yours in the GCS iit's clear the drone path is already smoother

tomvand:
For my reference, is your gif with or without the ins bypass?

It looks good enough to be merged as far as I'm concerned. Hope this helps somewhat :)

pcampolucci:
@tomvand I have checked as well and the flicker is still present but to a more limited extent respect to previously. As before, the line removes it, but again the reason is unknown. This flickering is doable for the course. 
Peek 2022-01-18 10-55
Also about the drifting, it is still present and it occurs also when the drone is in flight.

pcampolucci (approved):

matteobarbera:
@tomvand I tested it as is, with INS_BYPASS on. I also noticed that while the flickering was reduced, the drift in the `ROTORCRAFT_FP:north` happened still. Very small but it never stops decreasing the value

gautierhattenberger:
My summary:
- it works much better but still not perfect
- the real problem might not be solved, but if a part of it comes from Gazebo, there is not much we can do I guess
- there is something wrong with the stateGet function, but I really don't get it: can it be related to compiler optimization ? what if sim code is compiled without optim ?
Things can be better, but I think it is good enough to merge for next release.  So feel free to open a new PR for remaining issues if you have a solution.

gautierhattenberger (approved):

PR #2813, opened 2022-01-07
[nav] add adaptive navigation patterns

gautierhattenberger:
These patterns have been developped during an atmospheric science
project to track clouds in real-time based on sensor data.
The modifications includes:
- adding adaptive nav modules (lace, rosette, trinity, spiral_3D)
- new or improved meteo sensor modules (cloud sensor, meteostick)
- improvements to mission interface (UPDATE message, waiting patterns,
unique ID check)
- compact telemetry message MINIMAL_COM to reduce bandwidth (message
supported by server and logplotter)
- basic cloud simulation module (circle or polygon sector) and demo
flight plan

See for details:
Titouan Verdu, Gautier Hattenberger, Simon Lacroix. Flight patterns
  for clouds exploration with a fleet of UAVs. 2019 International
  Conference on Unmanned Aircraft Systems (ICUAS 2019), Jul 2019, Atlanta,
  United States. https://hal-enac.archives-ouvertes.fr/hal-02137839

### DIFF
--- a/sw/airborne/autopilot.c
+++ b/sw/airborne/autopilot.c
@@ -87,6 +87,20 @@ static void send_actuators(struct transport_tx *trans, struct link_device *dev)
 }
 #endif
 
+static void send_minimal_com(struct transport_tx *trans, struct link_device *dev)
+{
+  float lat = DegOfRad(stateGetPositionLla_f()->lat);
+  float lon = DegOfRad(stateGetPositionLla_f()->lon);
+  float hmsl = stateGetPositionUtm_f()->alt;
+  float gspeed = stateGetHorizontalSpeedNorm_f();
+  float course = stateGetHorizontalSpeedDir_f();
+  float climb = stateGetSpeedEnu_f()->z;
+  uint8_t throttle = (uint8_t)(100 * autopilot.throttle / MAX_PPRZ);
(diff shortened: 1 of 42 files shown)


noether:
Looks like a nice feature!

I will look at it after a deadline that I have next week.

sw/airborne/autopilot.c
```diff
@@ -87,6 +87,20 @@ static void send_actuators(struct transport_tx *trans, struct link_device *dev)
 }
 #endif
 
+static void send_minimal_com(struct transport_tx *trans, struct link_device *dev)
```
noether, line 90: Why not to have only a telemetry profile for these data?

I do not see why this is necessary for autopilot.c.

noether (approved):
It is a lot of work with nice improvements.

The only point that I do not get is why the "min comms" when one can define a "minimal" telemetry profile. Why is that?

sw/airborne/autopilot.c
gautierhattenberger: This message is reducing the bandwidth merging information from several messages. Indeed, we were using a specific telemetry file with adapted messages frequency. We wanted to avoid saturation on the modems since it was flying multiple aircraft out of line-of-sight, but still keep a good refresh rate (better than minimal at least).
I think there is no harm to keep it, but redefining the complete set of messages is still in the todo list.

PR #2814, opened 2022-01-13
Ro bust morphing usb pwm debug

alex31:
make some cleanup and reorganize how things are ordered when usbStorage thread detects vbus and should connect sdio and usb together.


gautierhattenberger:
This branch is not properly rebased on master, it needs to be cleaned up before merging

PR #2817, opened 2022-01-27
Motors on indi

EwoudSmeur:
When using INDI, the control allocation is done in the controller, and commands for each motor are directly calculated. However, that leaves it up to the user to put protections in the airframe command laws to not spin the motors before the motors are armed. 

This commit puts a protection already in the controller. This should work for most actuators, though I'm not sure about motors that can spin in either direction?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -491,7 +491,12 @@ void stabilization_indi_rate_run(struct FloatRates rate_sp, bool in_flight)
     if (act_is_servo[i]) {
       BoundAbs(indi_u[i], MAX_PPRZ);
     } else {
-      Bound(indi_u[i], 0, MAX_PPRZ);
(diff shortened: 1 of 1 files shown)


gautierhattenberger (approved):
Someday, we'll need a better interface for actuators than an array of uint16...
I'm approving this nonetheless, but I'd like to wait before merging, I plan a release by the end of the week. After that we can start breaking things again.

gautierhattenberger:
On a second thought, this might be a problem for non-motors actuators (e.g. servos) that are controlled by INDI. Not sure this is the best option.

EwoudSmeur:
Haha at least it is simple 🤣 But indeed I agree, an interface that includes safety protections for the actuators would be great.

Regarding the servos: one has to define an array anyway 'act_is_servo' in the airframe file, because those commands should be able to go below zero. Here I'm checking the same array and only the motors are put to -MAX_PPRZ. Still, maybe there are yet different actuators possible for which it would not work well? One can still put a line in the command laws, at least now this will be safe for the default systems.

gautierhattenberger (approved):

Issue #2824, opened 2022-02-15
speed up execution time of cv_opticflow.xml


pcampolucci:
A recent look at the Paparazzi optical flow module `cv_opticflow.xml` highlighted several bottlenecks that were proven to slow down dramatically the processing speed of the module. These are:

1. the `image_to_grayscale()` operation in the `calc_fast9_lukas_kanade()` function from `opticalflow_calculator.c`
2. the convolution operation to generate the pyramid in `pyramid_next_level()` from the `image.c` script

We have already tried (together with @matteobarbera), change the kernel for the convolution and separate the operation in a vertical and horizontal convolution, but given the size of the kernel this did not speed up enough. The possible solutions would be:

1. remove the `image_to_grayscale()` operation by adapting the optical flow functions (such as `pyramid_next_level()` and `image_subpixel_window()`), such as the pyramid one, to select only the Y channel of the original image. 
2. find a better convolution method (maybe FFT?) 
3. In the case use for Bebop, modify the MT9 camera driver to send directly grayscale images to the optical flow module
4. In the case of the Bebop, use a compiler designed for the Bebop's processor

optical flow wiki page: https://wiki.paparazziuav.org/wiki/Optical_flow
optical flow method literature: http://robots.stanford.edu/cs223b04/algo_tracking.pdf

OpenUAS:
Since #2825 is merged this item contains advice, but not something to fix anymore. Not that the 2,3,4 hints are not useful, they are, but they are not actionable for this issue ATM, therefore closed so to focus on other issue.

PR #2825, opened 2022-02-15
Opticflow pyramid build FPS improvements

matteobarbera:
@pcampolucci and I have been looking at the `cv_opticflow` module to try and improve the FPS performance. We identified a bottleneck in [`pyramid_next_level`](https://github.com/paparazzi/paparazzi/blob/742600628e684d070e4ecbe57de1dc1a66394908/sw/airborne/modules/computer_vision/lib/vision/image.c#L406), where we optimized the convolution by using a 3x3 kernel instead of 5x5, separated into two 1D kernels to improve performance. If both cameras are set to pyramid level 2 this leads to a small improvement in performance.

Currently, in master, we observed the FPS of the front camera frequently drop to ~2 FPS. This was somewhat resolved by limiting the target FPS of the `MT9F002` and `MT9V117`. The default value is the max available, which slows down the autopilot and should perhaps be modified.

With this graph we show the FPS with current master and our code, both with 
```xml
<define name="MT9F002_TARGET_FPS" value="30"/> <!-- Front cam -->
<define name="MT9V117_TARGET_FPS" value="60"/> <!-- Bottom cam -->
```
defined in both to show the effects of only the kernel improvements.
fps_opticflow_PR

Ideally we were hoping to achieve a speedup to at least ~20 FPS for both cameras. We have some suggestions on what could be attempted in #2824. We saw that if the function [`image_to_grayscale`](https://github.com/paparazzi/paparazzi/blob/742600628e684d070e4ecbe57de1dc1a66394908/sw/airborne/modules/computer_vision/lib/vision/image.c#L131) is removed the FPS increases significantly, but it's still unclear whether the compiler -O3 flag optimizes the opticflow calculation out, since no corner is then detected.

### DIFF
--- a/conf/airframes/tudelft/bebop_opticflow_indoor.xml
+++ b/conf/airframes/tudelft/bebop_opticflow_indoor.xml
@@ -26,22 +26,28 @@
     <module name="ins" type="hff_extended"/>
 
     <module name="pose_history"/>
-    <module name="bebop_cam"/>
+    <module name="bebop_cam">
+      <!-- IMPORTANT to limit these or FPS drops significantly -->
+      <define name="MT9F002_TARGET_FPS" value="30"/> <!-- Front cam -->
+      <define name="MT9V117_TARGET_FPS" value="60"/> <!-- Bottom cam -->
+    </module>
     <module name="cv_opticflow">
       <define name="OPTICFLOW_CAMERA" value="bottom_camera"/>
       <define name="OPTICFLOW_DEROTATION_CORRECTION_FACTOR_X" value="0.8"/> <!--Obtained from a linefit-->
       <define name="OPTICFLOW_DEROTATION_CORRECTION_FACTOR_Y" value="0.85"/> <!--Obtained from a linefit-->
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
I'm not really, is this a draft waiting for more improvements, or are you willing to merge this already ? Otherwise, I'm fine with the modifications, the code looks okay, but I have no experience with the optical flow algorithm.

pcampolucci:
We would like to merge this already, and another pull request will follow. Would that be possible?

gautierhattenberger (approved):

Fabien-B:
Hi @pcampolucci , I saw that you edited the wiki and that's great!
For the next time, could you consider writing the doc in the sphinx doc instead ?
It's located in doc/sphinx. Do a `make html` to build it locally for tests.
Then when its merged, the doc at https://paparazzi-uav.readthedocs.io is automatically updated.
It's good to keep the doc synchronized with the code, and don't need any maintenance work like the wiki need.

pcampolucci:
@Fabien-B good call, I will definitely do it next time!

PR #2826, opened 2022-02-17
Add script to get PprzGCS, and make it default.

Fabien-B:
To push adoption of the new GCS I propose to make PprzGCS the default choice.
The script will make a link to the "pprzgcs" command if it exists.
If the command do not exist, it will download the latest AppImage, and make a link to it.
What do you think?


fvantienen:
@Fabien-B I think this one can be closed because of #2858 ?

Fabien-B:
You're right, I'm closing it.

PR #2828, opened 2022-02-21
[autopilot] merge AP and FBW in normal operation

gautierhattenberger:
There is no longer a separated FBW in normal operation for fixedwing,
only a single AP task/thread.
The code for main_ap is now factorized for all firmwares.
Dual MCU support (separated AP and FBW) is still possible, based on the
rotorcraft implementation and intermcu pprzlink messages. Relevant code
is also factorized and available in all firmwares.
The recovery mode (in case of hard fault) is still available for
fixedwing with ChibiOS implementation.
Only intermcu_uart is supported at the moment. Other options (spi/can)
might be added back in the future if really used.

Some other evolution are made:
- in the radio_control and actuators modules, now integrated to AP and
using ABI messages
- the message parsing on datalink event (defined in module) that is now
properly filtering the messages based on their class and allowing to
have multiple callbacks for the same message
- cleaning the part of code that were using the old intermcu fixedwing
interface


gautierhattenberger:
I'm also adding that it has been tested in most configuration (Normal, separated AP/FBW, recovery), architecture (stm32, chibios) and firmware (fixedwing, rotorcraft, rover).
This is indeed a huge modification, barely impossible to do step-by-step, but with this, the main AP is now the same for all firmware and event FBW can be used the same way in any configuration. I think this is a great step for a cleaner organization of the autopilot code.
I'm also attaching some schematics explaining the new architecture and timing.

PPRZ_Airborne_architecture_single_AP_2022

gautierhattenberger:
And the main loop timing:

PPRZ_Main_ap_loop

gautierhattenberger:
@fvantienen @dewagter @noether Any comment on this one ? I would like to merge this soon to avoid divergence and conflicts since it is modifying many files.

noether (approved):

conf/modules/actuators_dummy.xml
```diff
@@ -10,6 +10,7 @@
     </description>
   </doc>
   <dep>
+    <!--depends>actuators</depends-->
```
fvantienen, line 13: Can this be removed?

fvantienen (approved):
Quickly scrolled through it, but looks good! I like the idea

conf/modules/actuators_dummy.xml
gautierhattenberger: I'm planing to change the actuators interface, so I'd rather clean this later

PR #2831, opened 2022-02-25
[gps] Update ublox RTK support and ground tool

fvantienen:
Add RTK tool for ublox GPS and update the RTK messages

### DIFF
--- a/sw/airborne/modules/gps/gps_ubx.c
+++ b/sw/airborne/modules/gps/gps_ubx.c
@@ -60,7 +60,7 @@
 #define GOT_CHECKSUM1 8
 
 #define RXM_RTCM_VERSION        0x02
-#define NAV_RELPOSNED_VERSION   0x00
+#define NAV_RELPOSNED_VERSION   0x01
 /* last error type */
 #define GPS_UBX_ERR_NONE         0
 #define GPS_UBX_ERR_OVERRUN      1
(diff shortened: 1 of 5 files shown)


gautierhattenberger (commented):
Why not including the header `var/include/ubx_protocol.h` that is already defining the macros ? And if not, their is no reason to keep them in the header file. The header file itself is not really needed at all I would say.

fvantienen:
I did not include `ubx_protocol.h` because that one has other includes and dependencies on functions like `ubx_send_bytes`. That would add a lot of unneeded defines and bypasses to make that work.

I chose to use a separate header file, to make the `c` file more readable and keep those defines separated. It is not needed, but makes it easier to read

sw/airborne/modules/gps/gps_ubx.c
```diff
@@ -60,7 +60,7 @@
 #define GOT_CHECKSUM1 8
 
 #define RXM_RTCM_VERSION        0x02
-#define NAV_RELPOSNED_VERSION   0x00
+#define NAV_RELPOSNED_VERSION   0x01
```
gautierhattenberger, line 63: what is the implication of this modification ?

gautierhattenberger: yes, if you can keep compatibility it is better I think

gautierhattenberger (approved):

PR #2833, opened 2022-03-02
[ins] Fix EKF2 update rate and disable baro filtering

fvantienen:
- Fix EKF2 update rate problems when running control loops at higher frequencies than 100Hz.
- Disable MS5611 baro filtering, to make it behave similar to other baro sensors.


fvantienen:
I chose to remove the median filter completely, because other barometric sensors don't have the filter. This makes it way more consistent

gautierhattenberger (approved):

PR #2835, opened 2022-03-09
[fix] protect ivy operations with mutex in nps

gautierhattenberger:
It prevents NPS to crash with Ivy messages, especially when simulating several drones.
Also checking the argc attribute and avoiding an unnecessary memcpy.


noether (approved):
Cool, I had (and I suppose I still have) this problem when simulating several drones simultaneously. Sometimes, the simulator crashed. Maybe here is the reason why.

Hopefully, these mutexes will help. 

Thanks

PR #2829, opened 2022-02-24
[ins] EKF2 update to latest version

fvantienen:
Several fixes for the EKF2

PR #2837, opened 2022-03-09
Improve message header handling.

Fabien-B:
Set the destination id to the correct value.
Depends on https://github.com/paparazzi/pprzlink/pull/158.


fvantienen:
After the EKF2 is merged we can merge this #2829 since it is depending on an updated `pprzlink`

gautierhattenberger (approved):

PR #2838, opened 2022-03-10
[datalink] move all hardcoded message parsing to modules datalink events

gautierhattenberger:
all incoming messages in autopilot are processed with modules callbacks to make the code cleaner and less coupled


fvantienen (changes requested):
Looks good only maybe fix the error in the tests

conf/modules/air_data.xml
```diff
@@ -36,6 +36,7 @@
   </header>
   <init fun="air_data_init()"/>
   <periodic fun="air_data_periodic()" freq="10"/>
+  <datalink message="WIND_INFO" fun="air_data_parse_WIND_INFO(dev,trans,buf)"/>
```
fvantienen, line 39: Since you set `dev` and `trans` to unused, you can also just remove them here instead.

sw/airborne/modules/core/settings.c
```diff
@@ -88,3 +88,49 @@ int32_t settings_clear(void)
   settings_clear_flag = false;
   return -1;
 }
+
+void settings_parse_msg_SETTING(struct link_device *dev, struct transport_tx *trans, uint8_t *buf)
+{
+#ifndef INTERMCU_FBW
+  if (DL_SETTING_ac_id(buf) != AC_ID) { return; }
```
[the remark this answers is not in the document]
gautierhattenberger: fixed, it was a missing include

conf/modules/air_data.xml
gautierhattenberger: they are used only when `WIND_INFO_RET` is defined, so I need to keep them with the unused flag to avoid warnings in default case

fvantienen (approved):
Looks good now!

PR #2839, opened 2022-03-11
[gps] Fix week number calculation

fvantienen:
Not 100% sure why it needs to be `-1` week.
I first thought it had to be `-6` days, because the GPS week number started on `06-01-1980`. But after analysing the result it seems to be needed to subtract `7` days to come to the correct week. Maybe because the weeks start on sundays?


fvantienen:
Maybe for the future we can remove the week number fully and do the opposite. We communicate the day, month and year?

gautierhattenberger:
I think the error is in `gps_day_number` just above. It gives the number of days in Jan 1st 1980 instead of the 6th. If you subtract 6 days, it seems to work correctly.

fvantienen:
Though 6 is not enough, it should be 7. As compared with https://www.ngs.noaa.gov/CORS/Gpscal.shtml and some manual calculations.
That function above indeed does from the first of January, but comes from https://www.u-blox.com/sites/default/files/GPS-WeekNumber-Rollover-Workaround_AppNote_%28UBX-19016936%29.pdf. So I didn't want to change that.

fvantienen:
Oh wait my mistake I was counting the wrong one with a leap year, it indeed needs to be `-6`. Will correct that

gautierhattenberger (approved):

Issue #2730, opened 2021-05-20
Guidance_h ref acceleration jumps


EwoudSmeur:
When simulating tudelft/bebop_indi_actuators.xml, trying to include the reference acceleration from the horizontal reference as a feed forward, I noticed that it was jumping to zero and back (jittering) all the time when going to a new reference position and the speed is saturated. I traced the problem to the function gh_saturate_ref_speed (https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c#L144). Or it could be the function that calculates the max speed based on the heading angle the drone (https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c#L141).

I think this affects any controller that uses the reference acceleration. I think I can solve this bug, but I'm wondering if we are happy with the current reference model or if we want to change it (3D, floating point?), before I spend time on it.

OpenUAS:
Thanks for reporting and offer to fix. @gautierhattenberger any opinion on this welcomed.

gautierhattenberger:
If we can have a smooth acceleration setpoint, I think it is better than the current situation even if it "works". I also don't mind that much if the internal computation is done in floating point, and even better if it is a more generic library.

PR #2842, opened 2022-03-11
Ref model float

EwoudSmeur:
I had an issue (#2730) with the horizontal reference model, and thought it was actually a good idea to rewrite it in floating point in order to simplify. I kept the reference position in int64 in order to have no problem with range and integration of velocity. Tested in simulation and in our cyberzoo.


sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.h
```diff
@@ -54,24 +54,19 @@ extern float gh_max_speed;
 #define GH_FREQ_FRAC 9
```
gautierhattenberger, line 54: interesting, it seems that the integration frequency of 512 is implicitly encoded here...
this is not used anymore (you have added a proper `dt`) and should be removed I think

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c
```diff
@@ -42,229 +40,156 @@ static const int32_t gh_max_accel = BFP_OF_REAL(GUIDANCE_H_REF_MAX_ACCEL, GH_ACC
...
 {
   gh_ref.omega = GUIDANCE_H_REF_OMEGA;
   gh_ref.zeta = GUIDANCE_H_REF_ZETA;
-  gh_ref.zeta_omega = BFP_OF_REAL((GUIDANCE_H_REF_ZETA * GUIDANCE_H_REF_OMEGA), GH_ZETA_OMEGA_FRAC);
-  gh_ref.omega_2 = BFP_OF_REAL((GUIDANCE_H_REF_OMEGA * GUIDANCE_H_REF_OMEGA), GH_OMEGA_2_FRAC);
+  gh_ref.zeta_omega = GUIDANCE_H_REF_ZETA * GUIDANCE_H_REF_OMEGA;
+  gh_ref.omega_2 = GUIDANCE_H_REF_OMEGA * GUIDANCE_H_REF_OMEGA;
   gh_set_tau(GUIDANCE_H_REF_TAU);
   gh_set_max_speed(GUIDANCE_H_REF_MAX_SPEED);
+  gh_ref.dt = (1.0/PERIODIC_FREQUENCY);
```
gautierhattenberger, line 59: `1.0f`
also add the float specifier whenever needed later

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c
```diff
@@ -42,229 +40,156 @@ static const int32_t gh_max_accel = BFP_OF_REAL(GUIDANCE_H_REF_MAX_ACCEL, GH_ACC
...
 }
 
 
 float gh_set_max_speed(float max_speed)
 {
   /* limit to 100m/s as int version would overflow at  2^14 = 128 m/s */
   gh_ref.max_speed = Min(fabs(max_speed), 100.0f);
```
gautierhattenberger, line 66: should be `fabsf` here

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c
```diff
@@ -42,229 +40,156 @@ static const int32_t gh_max_accel = BFP_OF_REAL(GUIDANCE_H_REF_MAX_ACCEL, GH_ACC
...
-  BoundAbs(gh_ref.accel.y, gh_ref.max_accel.y);
+  // Speed squared
+  float v_norm2 = VECT2_NORM2(*speed_sp);
+
+  // Apply saturation if above max speed
+  if (v_norm2 > (gh_ref.max_speed * gh_ref.max_speed)) {
+    // speed_sp/sqrt(v_norm2)*vmax
+    float factor = gh_ref.max_speed / sqrtf(v_norm2);
+    VECT2_SMUL(*speed_sp, *speed_sp, factor);
+  }
```
gautierhattenberger, line 180: if speed is saturated, shouldn't the acceleration be saturated as well (in the "direction" of speed saturation) ?

gautierhattenberger (changes requested):
someday, a full float version, might be a good rewrite... later

EwoudSmeur:
Thanks for the comments @gautierhattenberger!

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -75,7 +75,7 @@ PRINT_CONFIG_VAR(NAVIGATION_FREQUENCY)
 #endif
 
 #define CLOSE_TO_WAYPOINT (15 << INT32_POS_FRAC)
-#define CARROT_DIST (12 << INT32_POS_FRAC)
+#define CARROT_DIST (200 << INT32_POS_FRAC)
```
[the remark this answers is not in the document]
EwoudSmeur: Maybe not, but it is an issue that this distance is quite small (and not configurable). If the reference model is not very aggressive, it will seriously limit the speed and acceleration towards waypoints that are far away. It is also not configurable...

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.c
EwoudSmeur: I don't think so, why would that be the case? The 'desired speed' (vel_sp) may be very large for a waypoint far away, but the saturated vel_sp will have a realistic value. From that the acceleration setpoint is calculated, which may or may not be saturated, Depending on the current speed of the reference model, the desired acceleration vector may have a different direction than the desired velocity vector. The acceleration should be saturated such that its direction is preserved. I hope I understood your question correctly though 😄

sw/airborne/firmwares/rotorcraft/navigation.c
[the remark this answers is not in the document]
gautierhattenberger: I agree that it should be configurable, but not for changing the previous behavior with such a large value. It will make the carrot jump on waypoints in a range of 200 m if I'm not wrong (and if ref traj is not used, which is not usually the case I agree).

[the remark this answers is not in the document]
EwoudSmeur: Ok I agree, I'll make it configurable

sw/airborne/firmwares/rotorcraft/navigation.c
```diff
@@ -74,8 +74,12 @@ PRINT_CONFIG_VAR(NAVIGATION_FREQUENCY)
 #define FAILSAFE_MODE_DISTANCE (1.5*MAX_DIST_FROM_HOME)
 #endif
 
+#ifndef NAV_CARROT_DIST
+#define NAV_CARROT_DIST 12
+#endif
+
 #define CLOSE_TO_WAYPOINT (15 << INT32_POS_FRAC)
-#define CARROT_DIST (12 << INT32_POS_FRAC)
+#define CARROT_DIST ( (int32_t) NAV_CARROT_DIST << INT32_POS_FRAC)
```
gautierhattenberger, line 82: with this, you "loose" the sub-metric precision (NAV_CARROT_DIST is cast to int before shifting). I suggest:
```suggestion
#define CARROT_DIST ((int32_t) POS_BFP_OF_REAL(NAV_CARROT_DIST))
```

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.h
```diff
@@ -49,29 +50,20 @@ extern float gh_max_speed;
...
 
-/** Update frequency
- */
-#define GH_FREQ_FRAC 9
-#define GH_FREQ (1<<GH_FREQ_FRAC)
 
-#define GH_ACCEL_REF_FRAC 8
-#define GH_SPEED_REF_FRAC (GH_ACCEL_REF_FRAC + GH_FREQ_FRAC)
-#define GH_POS_REF_FRAC (GH_SPEED_REF_FRAC + GH_FREQ_FRAC)
+#define GH_POS_REF_FRAC 26
```
gautierhattenberger, line 54: I think you should add a comment to explain this value. Also, as I mentioned before, the old `GH_FREQ_FRAC` of 9 is probably an implicit factor corresponding to the 512Hz, which is maybe not relevant anymore since you have introduced the real `dt`. Do we have a risk of overflow now, if the frequency is not 512 ?

gautierhattenberger:
You should check the log of semaphore, it seems that there is an error of type in `modules/gps/gps_sim_hitl.c`

sw/airborne/firmwares/rotorcraft/navigation.c
EwoudSmeur: Ah yes of course, good one!

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.h
EwoudSmeur: Thanks, I added a comment with the range and resolution, taking into account the frequency. I think we're pretty safe on both sides. Actually, the range is a bit overkill, more than the circumference of the earth... will update for better resolution!

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.h
```diff
@@ -49,29 +50,23 @@ extern float gh_max_speed;
...
-#define GH_FREQ_FRAC 9
-#define GH_FREQ (1<<GH_FREQ_FRAC)
-
-#define GH_ACCEL_REF_FRAC 8
-#define GH_SPEED_REF_FRAC (GH_ACCEL_REF_FRAC + GH_FREQ_FRAC)
-#define GH_POS_REF_FRAC (GH_SPEED_REF_FRAC + GH_FREQ_FRAC)
+/** fixedpoint representation: Q26.37 will give a range of
+ * 67e3 km and a resolution of 1.5e-11 m. At a rate of 500Hz,
+ * a ref speed of 7.3e-9 m/s could still update the position.
+ * /
```
gautierhattenberger, line 56: extra space not closing the comment properly
```suggestion
 */
```

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/guidance/guidance_h_ref.h
EwoudSmeur: Oops! 😅

gautierhattenberger (approved):

PR #2845, opened 2022-03-20
Rover steering wheel

jesusBV20:
This PR adds support to rovers with two actuators; namely, thrust (both directions) and steering wheel.

The four major (independent) points are:

- New firmware for the vehicle based on the existing firmware for rovers.
- We have added a NPS simulator for the rover. It behaves very similarly to a rover traveling at around 20km/h on a flat world.
- We have modified the "classic" GVF. Now you can call it from different firmware (fixed-wing and rover steering wheel).
- We have added UCM directory in _flight_plans_, _airframefiles_ and _radios_ for our configurations.
- We have added a STM32 main loop in firmware rover for boards without ChibiOS architecture (copy&paste from rotorcraft). 

The main functionalities related with our rover steering wheel guidance module are:

- A PID (+ feedforward) controller for the control of the ground speed.
- A course controller supported by GVF.
- Three autopilot modes: 1. Manual, 2. Cruise (it sets as desired velocity the last one measured) but the course is still manual, 3. Auto (course and speed are both autonomous).


jesusBV20:
¿Is there a way to merge conf/modules/board_matek_f765_wing.xml and conf/modules/board_matek_f765_car.xml? We just want to avoid the make of conf/modules/osd_max7456.xml.

sw/airborne/firmwares/rover/guidance/rover_guidance_steering.c
```diff
@@ -0,0 +1,176 @@
...
+  // - Updating PID
+  guidance_control.speed_error = (guidance_control.cmd.speed - guidance_control.state_speed);
+  update_pid_f(&rover_pid, guidance_control.speed_error, time_step);
+
+  guidance_control.throttle = BoundThrottle(guidance_control.kf*guidance_control.cmd.speed + get_pid_f(&rover_pid));
+}
+
+
+/** PERIODIC function **/
+void rover_guidance_steering_periodic(void)
```
gautierhattenberger, line 108: The code of this function is not really in the "spirit" of the autopilot code generation, since the modes here must match the one in the autopilot XML file.
So the different parts of this code should be directly called by the XML file. Just provide some extern functions like `rover_guidance_steering_enter`, `rover_guidance_steering_speed_ctrl`, `rover_guidance_steering_heading_ctrl`, etc.

sw/airborne/firmwares/rover/guidance/rover_guidance_steering.h
```diff
@@ -0,0 +1,133 @@
+/*  rover_guidance_steering.h */
```
gautierhattenberger, line 1: add license header

sw/airborne/firmwares/rover/main.c
```diff
@@ -0,0 +1,77 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/**
+ * @file firmwares/rotorcraft/main.c
```
gautierhattenberger, line 23: The bare_metal main file is now in `sw/airborne/main_bare.c`, please remove this main (.c and .h)

sw/airborne/modules/guidance/gvf/gvf.c
```diff
@@ -457,3 +496,259 @@ bool gvf_sin_wp_alpha(uint8_t wp, float alpha, float w, float off, float A)
   return true;
 }
 
+
+#elif defined(ROVER_FIRMWARE) // ROVER TRAJECTORIES
```
gautierhattenberger, line 500: if more firmware specific code should be added, consider making a separated C file for each of them

gautierhattenberger (changes requested):

sw/airborne/modules/guidance/gvf/gvf.c
jesusBV20: Okay, we'll keep these conditionals for now, but we'll consider your suggestion in a next future.

jesusBV20:
Any suggestion to fix this "ap_status -> UKN" in GCS? We are receiving "ap_mode" from "ROVER_STATUS" telemetry msg but I think that there is something wrong with the labels.

image

sw/airborne/firmwares/rover/main.c
jesusBV20: Oooh, cool. I didn't see that change.

jesusBV20:
I have to test these changes out of the simulator. I'll tell you if everything works like before.

gautierhattenberger (approved):

gautierhattenberger:
It looks fine so far, I'll wait for the results of your tests before merging.
The display error is maybe not related to this PR, so I'll check that later.

jesusBV20:
Okey @gautierhattenberger, everything looks fine. The control of the rover works as it has to.

PR #2849, opened 2022-03-25
[baro] Fix Bebop barometer configuration

fvantienen:
This fixes the barometer for both the Bebop 1, Bebop 2 and the Disco.
Small question @gautierhattenberger why is the `baro_board` module in the `ap.xml`, is that still needed?


gautierhattenberger (approved):

gautierhattenberger:
the `baro_board` is a real s*** that I struggle to remove. The old makefile had a lot of magic to make the baro available on supported boards but still configurable. For now, my option to avoid changing all the airframes failing to compile if I remove this, is to load this module with ap (so all boards) and keep some raw makefile magic to keep the previous behavor.
Long term options are:
- explicit include of baro in airframe files
- make board modules depend on the relevant baro: it means it will be loaded for this board even if you don't want to, you can select the source with ABI filtering for the state filters
- add a new option to dependency algo to "suggest" a module that is only loaded if the functionalities provided by the module are not specifically chosen by the user
- any other idea ?

fvantienen:
In my opinion the best idea would be to add the baro to the board file. Then maybe add options in the airframe to 'remove' automatically added modules if needed. This way we can also add other features which are by default included in the board, like the IMU.

It would clean up a lot of the airframe files and you either would have the option to filter through ABI or to add a specific remove option in the airframe. (For example `<unload name="ms5611_...."/>`) Since most airframes use the sensors on the board it would clean up a lot!

An idea I had as well was to add an `instance` or specific name to the loading of a module(with a struct), such that they can be loaded multiple times. This way you could filter and disable only specific instances. I think this should be added either way anyways in the future, to make it possible to use similar sensors with multiple instances.

gautierhattenberger:
I agree that adding an `unload` option would be the easiest way. Before doing that, I would like to consider the option of a `suggest` node for dependencies, because it might also solve the "problem" of auto-loaded module, which was a workaround to load modules for simulation.
Your second point, with `instance` that modules are more "objects" than actual "modules". I don't see how to make it in C without a lot of code generation, macros, etc. I feel that try to do that without the support of an adapted language (C++ would be the most obvious solution). Re-writing Paparazzi in an other language is not something we can afford with the number of developers involved at the moment.

fvantienen:
Ok that seems like a good idea.

For the second point it would be really easy if you can give and instance struct, for example `mpu6000_t`. Then code generation should make an object with the instance name for example: `struct mpu600_t board_imu;` (possibly with options to initialize variables inside this struct to configure it). Then the functions will get called with a pointer to that instance struct as argument, similar like we are doing without the instance struct. So in total it would look something like this:
```
struct mpu6000_t board_imu = {
  .spi_dev = &spi1,
  .accel_range = 16G,
  ...
};
...
mpu6000_init(&board_imu);
...
mpu600_event(&board_imu);
....
```

And in the XML it would look something like this:
```
<module name="mpu6000" instance="board_imu">
  <conf name="spi_dev" value="spi1"/>
  <conf name="accel_range" value="16G"/>
</module>
```

The only thing that needs to be changed is the module generator which should create that struct and pass it to the appropriate functions. This should not be really hard and gives to the option to load modules multiple times, with different configurations(such as different ABI id's).

gautierhattenberger:
sure, but don't forget that you also need to deal with the settings and messages (probably other stuff I , so as I said, not impossible, but requires a lot of code generation.
it's not a priority for me at the moment at least

PR #2852, opened 2022-03-30
[makefile] Remove ffast-math compiler flags

fvantienen:
The `ffast-math` compiler flag does extremely hard optimalization outside IEEE specification. It assumes all math is finite, which is not the case for the ekf2, as it makes use of `NAN`. See https://stackoverflow.com/questions/7420665/what-does-gccs-ffast-math-actually-do for more information. Since this optimalization can cause mistakes in floating point calculations, I would like to remove it fully.
```
Sets the options -fno-math-errno, -funsafe-math-optimizations, -ffinite-math-only, -fno-rounding-math, -fno-signaling-nans, -fcx-limited-range and -fexcess-precision=fast
```

This was first introduced with the old omap architecture in [f708d4a](https://github.com/paparazzi/paparazzi/commit/f708d4ae7b8b3456337136f1e886bfbf8d276d9a).


fvantienen:
If you look closely these are only `CXXFLAGS`. Only for arm linux, like a bebop it was also configured for c code if hard float was defined. So this was almost never used except if you load cpp files like the ekf2.
This should not change the load of other platforms, because we almost never load cpp files.

gautierhattenberger:
Ok, it's true that regular C code should not be affected. However, the few parts using C++ are EKF2 and another INS using Eigen, with a lot of float computation. Even if the mekf INS is probably not used, I would like to test this first for the CPU load, if it is acceptable or not to apply that for all.

gautierhattenberger:
Actually, it seems that only stm32 (so old boards using libopencm3) and linux are concerned here ? It seems the flag is not set for ChibiOS anyway. If this is the case, then I don't mind merging this.

fvantienen:
Yes indeed chibios is not changed at all(it was never set) and stm32 only the c++ code. That is why I wanted to make it more universal and this seemed the correct way to do it.

gautierhattenberger (approved):

PR #2853, opened 2022-04-01
[flight plan] improve operator politeness when using the 'call' statements

gautierhattenberger:
`call` and `call_once` might be considered a bit rude to the autopilot, so
the operator should ask politely, but not too often, excessive
politeness might not be sincere

see https://en.wikipedia.org/wiki/INTERCAL for details concerning this new feature


fvantienen:
Maybe asking for this pull request a bit more polite would help?

Fabien-B:
Looks good to me.

PR #2856, opened 2022-04-07
[ocaml] move the GCS-only related code out of xpprz lib

gautierhattenberger:
Most of the pprz.xlib is used by the legacy GCS only, using in particular the deprecated canvans.
This PR is isolating the relevant part to the GCS folder, so it will be easier in the future to avoid the compilation of this part by default.


gautierhattenberger:
Updating PR, with the correction for ocaml warnings in ubuntu 22.04

Fabien-B (dismissed):

PR #2858, opened 2022-04-08
[Sessions] Use GCS wrapper as default GCS.

Fabien-B:


fvantienen:
Maybe also add the `pprzgcs` package in the simple installation on the wiki? https://wiki.paparazziuav.org/wiki/Installation

gautierhattenberger (approved):

PR #2859, opened 2022-04-11
[chibios] Add STM32H7

fvantienen:
This adds support for the STM32H7 and updates ChibiOS to enable this support.
- Adds support for STM32H7 (cube orange)
- Updates ChibiOS to latest trunk with minor changes in own branch
- Cleans up chibios board configurations and removes a lot of double code
- Cleans up peripheral drivers for chibios by removing double code
- VTOR (interrupt table) is automatically set by chibios
- Cleaned up PWM configuration for chibios
- Settings function without actual variable declaration in code
- Removed unused Navstik autopilot

Please check #[59db5b1](https://github.com/paparazzi/paparazzi/pull/2859/commits/ea071ab75e3615acb5aad626c52b2f42459db5b1) more extensive for numbering errors.
Also check if your chibios boards are still functioning properly with the latest ChibiOS, since I might have missed some configurations and I don't have all the hardware to test.


fvantienen:
I fixed some last small mistakes and tested it again on the Cube Orange. Everything seems to work

sw/tools/generators/gen_settings.ml
```diff
@@ -90,13 +90,16 @@ let print_dl_settings = fun out settings settings_xml ->
         try
           let h = ExtXml.attrib s "handler" in
           begin
+            let t = ExtXml.attrib_or_default s "type" "" in
```
gautierhattenberger, line 93: I a got it right, you want to call a function without a variable associated. If yes, this is something I wanted to do for a long time, however this implementation might be limited. It is still necessary to associate a variable. Also, the `type` attribute seems related to persistent settings, is it breaking this "feature" (not sure if it is really working, but still). Also, I wanted to make an "automatic" variable to get the return value of the function, what do you think ?

sw/airborne/autopilot.c
```diff
@@ -300,20 +291,6 @@ bool autopilot_in_flight(void)
   return autopilot.in_flight;
 }
 
-/** set power switch
- */
-void autopilot_set_power_switch(bool power_switch)
```
gautierhattenberger, line 305: Please, can you make a separated module from this instead of completely removing it ?

gautierhattenberger (changes requested):

.gitmodules
```diff
@@ -6,10 +6,10 @@
 	url = https://github.com/paparazzi/luftboot.git
 [submodule "sw/ext/fatfs"]
 	path = sw/ext/fatfs
-	url = https://github.com/enacuavlab/fatfs.git
+	url = https://github.com/paparazzi/fatfs.git
 [submodule "sw/ext/chibios"]
 	path = sw/ext/chibios
-	url = https://github.com/ChibiOS/ChibiOS.git
+	url = https://github.com/paparazzi/ChibiOS.git
```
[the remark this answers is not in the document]
fvantienen: There are only 2 minor commits compared to the ChibiOS official SVN. For the FDCAN I have supplied a patch, but that seems to be going to get fixed soonish. For the SPIv2 which is causing problems I currently don't have a fix and switched back all processors to SPIv1(I also don't know what is going wrong or why).
Both should eventually be fixed in the ChibiOS official SVN, this is only sort of needed temporary because I prefer to merge this rather sooner than later.

sw/airborne/arch/chibios/mcu_periph/adc_arch.c
```diff
@@ -66,22 +66,16 @@
 ///#endif
 
 
-// architecture dependent settings
-#if defined(__STM32F10x_H) || defined(__STM32F105xC_H) || defined (__STM32F107xC_H)
-// STM32F1xx
-#define ADC_SAMPLE_RATE ADC_SAMPLE_41P5
-#define ADC_CR2_CFG ADC_CR2_TSVREFE
```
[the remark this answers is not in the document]
fvantienen: This is now all set in the configuration function `adc_configure` below https://github.com/paparazzi/paparazzi/pull/2859/files/9419301763072d0c76a3b8e42dae0c2509cb7bc2#diff-6e83a2506a4d4e913810170ca39337dd235cca7f8b80ef047ff66fec81e3325aR175

sw/airborne/autopilot.h
```diff
@@ -173,15 +172,6 @@ extern bool autopilot_in_flight(void);
...
 
-/** Set power switch state
- *  This will actually enable the switch if POWER_SWITCH_GPIO is defined
- *  Also provide macro for dl_setting backward compatibility
- *
- * @param[in] power_switch true to enable, false to disable
- */
-extern void autopilot_set_power_switch(bool power_switch);
-#define autopilot_SetPowerSwitch(_ps) autopilot_set_power_switch(_ps)
-
```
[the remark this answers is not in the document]
fvantienen: I will look into creating a module for this functionality

sw/airborne/mcu_periph/uart.h
```diff
@@ -33,18 +33,20 @@
 #include "std.h"
 
 #ifndef UART_RX_BUFFER_SIZE
-#if defined STM32F4 || defined STM32F7 //the F4 and F7 have enough memory
-#define UART_RX_BUFFER_SIZE 256
-#else
+// Only for the STM32F1 less buffer
+#if defined(STM32F1)
```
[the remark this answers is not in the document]
fvantienen: Here only `STM32F1` works(for the stm32 arch), since we can't include `hal.h` and else sometimes the wrong buffer size is set causing it to crash. I was doubting to remove the if statement here completely, but wasn't sure if this works for every board having a bigger buffer size. (CI should fail with boards exceeding the available RAM)

sw/airborne/boards/apogee/chibios/v1.0/board.h
```diff
@@ -343,102 +341,10 @@
     { PWM_SERVO_6_ACTIVE, NULL }, \
   }, \
   0, \
-  0 \
-}
-
-/**
- * DSHOT
```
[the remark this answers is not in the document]
fvantienen: I didn't notie this, thanks! I will fix it

sw/tools/generators/gen_settings.ml
fvantienen: Yes you got it totally correct. This method actually works perfectly fine and no variable needs to be created, the `var` tag is still needed for naming and other purposes. But in the generated code it isn't used at all anymore because of the `type="fun"`.
The type "fun" doesn't work together with setting it persistent as well, but that should generate an error by `gen_settings`. The other persistent settings just work fine.

For the option to get the return value of a function it could be added really easily as well, currently it generates something like this:
```
#define DlSetting(_idx, _value) { \
  switch (_idx) { \
...
    case 10: mcu_reboot( _value ); break;\
...
}

#define PeriodicSendDlValue(_trans, _dev) { \
  static uint8_t i;\
  float var;\
  if (i >= 88) i = 0;\
  switch (i) { \
...
case 10: var = 0; break;\
...
}

static inline float settings_get_value(uint8_t i) {
  switch (i) {
...
    case 10: return 0;
...
}
```
Which could of course also be something like this:
```
case 10: _value = mcu_reboot( _value ); break;\
...
case 10: var = some_other_func(); break;\
...
case 10: return some_other_func();
```

I'm not that great at ocaml and the generators etc. so if you have better suggestions that also work, please let me know. This was just an easy fix for something that was annoying me for years.

See for example: https://github.com/paparazzi/paparazzi/pull/2859/files/9419301763072d0c76a3b8e42dae0c2509cb7bc2#diff-b0b28a5c9934bc204875c10ce4060dd75bcc3b0956e89142f3cd8e5bd878a1afR10

sw/airborne/autopilot.c
fvantienen: Will do that

sw/airborne/autopilot.h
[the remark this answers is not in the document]
fvantienen: Done!

sw/airborne/autopilot.c
fvantienen: Done!

fvantienen:
I fixed all the needed changes

fvantienen:
I now also made the PWM configuration cleaner and more generic. This is now almost at the point that the board file can be fully generated from the cfg and no additional header file is needed.
This is all depending on the correct names inside the `board.cfg` file, see the cube orange as example. If we set the last defines like `SDLOG_BAT_ADC`,`SDLOG_BAT_CHAN`,`DefaultVoltageOfAdc`,`DefaultMilliAmpereOfAdc`,`ADC_CHANNEL_VSUPPLY`,`ADC_CHANNEL_CURRENT` in the header part of the cfg it becomes fully generic. What do you guys think about the pin names from the cube orange? @gautierhattenberger @alex31 (They collide with some defines that is why `--no-pp-pin` is disabled) Do you guys have suggestions or changes before we can try to go for fully generated files?

fvantienen:
Since I was so close on generic board files, I wanted to show what I meant by giving some example. See the px4fmu_v5.0_chibios and the cube_orange. They both make use of the `common_board.h`.

sw/airborne/boards/apogee/chibios/v1.0/board.h
```diff
@@ -205,146 +202,92 @@
...
-#define PWM_SERVO_0_GPIO GPIOB
-#define PWM_SERVO_0_PIN GPIO0
-#define PWM_SERVO_0_AF GPIO_AF2
-#define PWM_SERVO_0_DRIVER PWMD3
-#define PWM_SERVO_0_CHANNEL 2
-#define PWM_SERVO_0_ACTIVE PWM_OUTPUT_ACTIVE_HIGH
-#else
-#define PWM_SERVO_0_ACTIVE PWM_OUTPUT_DISABLED
-#endif
-
```
gautierhattenberger, line 222: It seems that this part is not matching between cfg and header. CFG seems correct with definition of servo from 0 (without index shift), but header is starting from 1 with shifted index. On Apogee the servos are starting at index 0.

fvantienen: Somehow I mess up the apogee always and forgot to run make in the folder. Fixed now!

fvantienen:
Ok the splitting changes some more things than I wanted. Will fix your apogee board again.

sw/airborne/boards/apogee/chibios/v1.0/board.cfg
```diff
@@ -188,146 +188,185 @@ HEADER
 /*
  * PWM defines
  */
+#if defined(LINE_SERVO0)
```
gautierhattenberger, line 191: Only for the Apogee that is using a single h file, since the header from CFG is placed before ChibiOS defines, you can't add these lines otherwise PWM outputs are not defined. You shouldn't loose time to properly split the files for this board, so just remove the `#if defined(LINE_SERVOX)` protection and it should be fine.
It seems to work correctly for the Tawaki/F7 board.

gautierhattenberger (changes requested):

sw/airborne/boards/apogee/chibios/v1.0/board.cfg
fvantienen: I think the apogee board was just cursed to fail every commit. But I corrected the mistake and it should work now

PR #2860, opened 2022-04-13
[chibios] Make constants consistent with different architectures

fvantienen:
This fixes the semaphore build and makes disables `single-precision-constant` fully on chibios. This is consistent with bare stm32

### DIFF
--- a/conf/boards/holybro_kakute_f7.makefile
+++ b/conf/boards/holybro_kakute_f7.makefile
@@ -19,7 +19,7 @@ MCU=cortex-m7
 
 ## FPU on F7
 USE_FPU=hard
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
We can change the default parameter for ChibiOS, but for the specific Kakute board, I remember have some issues with the FPU, crashing with double precision, so it should be tested first

fvantienen:
Weird that it crashes for the Kakute, I can leave that out for now or do you have the opportunity to test it out quickly? Or someone else, we don't have the board available.
I thought because it was an F7 it shouldn't have been a problem at all for that one as well.

gautierhattenberger:
I did a test with KakuteF7 board and it seems that the single precision flag is not required (but of course selecting the correct FPU is important).

gautierhattenberger (approved):

PR #2862, opened 2022-04-16
[sphinx] Adds a documentation section on Modules

matteobarbera:
In an effort to improve the documentation of Paparazzi I ported over some documentation and combined it with some of the Paparazzi Wiki pages to start populating the sphinx docs.

Feedback is welcome, I saw some of the Wiki pages have outdated info, I double-checked as much as possible but perhaps some old info made it in here.


doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -6,4 +6,462 @@ Modules
...
+Modules are stand-alone programs, and hence do not have a `main` function in their source code.
+Instead, the autopilot will regularly call other functions that are part of your module, such as a module periodic
+function. Which functions are called is defined by the module xml file described earlier.
+
+The section `Module XML`_ lists the types of functions you can register in the module xml: ``init``, ``periodic``,
+``event`` and ``datalink``, of which init and periodic are the most common.
+The ``init`` function is called once at startup. You can use this function to initialize important variables of your
+module, or memory intensive structures such as large arrays, or for instance to subscribe to new video frames.
+Once the autopilot is fully initialized, it will enter an infinite loop in which it will continuously read new sensor
+data, feed this to the guidance and stabilization controllers, and send new commands to the Bebop's motors.
```
Fabien-B, line 328: That's oddly specific :smile: 
I think that now, "all is modules", so the infinite really just call all "event" functions, then some "periodic" functions.

Fabien-B (changes requested):
Thank you for this great doc!
Modules are such a big stuff in paparazzi, I think it may be good to split it in multiple pages.
There was a lot of changes recently on modules. @gautierhattenberger take a look?
I think that when doc is ported (and updated) here, we should remove it from the wiki, with a link redirecting to this doc.

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -6,4 +6,462 @@ Modules
...
+ </modules>
+
+The main_freq parameter (in Hz) is optional. If present, it allows to specify the frequency of the main loop.
+Default is the main periodic frequency of the autopilot. If the frequency does not needed to be specified,
+then the first line is just: ``<modules>``.
+
+The children ``define`` generate compilation defines for all the targets of a module and can be used with or without an
+associated value.
+It allows to keep the module description more generic add to place the airframe dependent parameters in the airframe xml file.
+The child ``configure`` generates a makefile variable.
```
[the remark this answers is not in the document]
gautierhattenberger: yes, the `modules` node is deprecated and should not be documented here. All modules should be in `firmware` sections.

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -6,4 +6,462 @@ Modules
...
+        <define name="DEMO_MODULE_LED" value="LED_X" description="LED Selection"/>
+      </doc>
+
+      <settings>
+        <dl_settings name="bla">
+          <dl_setting min="0" max="5" step="1" var="bla_bla" shortname="bb"/>
+        </dl_settings>
+      <settings>
+
+      <depends>foo.xml,bar|baz</depends>
```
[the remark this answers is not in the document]
gautierhattenberger: The new structure for dependencies is now
```
<dep>
  <depends>module1,module2|module3,@func1</depends>
  <provides>func2</provides>
  <conflicts>module4,@func3</conflicts>
</dep>
```
where `@` refers to functionality and the rest to modules name. By the way, the module `name` attribute *must* match the modules file name. It is therefore not necessary to have the `.xml` extension.

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -6,4 +6,462 @@ Modules
...
+|               |               |   | "include" or "I" (ex: ``name="$(ARCH_SRC)" type="include"`` ->      |
+|               |               |   | ``target.CFLAGS += -I$(ARCH_SRC)`` default is "define"              |
+|               +---------------+-------------------------------------------------------------------------+
+|               | file          | - | name : the name of the c file (located in                           |
+|               |               |   | ``sw/airborne/modules/<dir_name>``) to add in the Makefile          |
+|               |               |   | (ex: ``name="demo_module.c"`` ->                                    |
+|               |               |   | ``target.srcs += modules/<dir_name>/demo_module.c)``                |
+|               |               |                                                                         |
+|               |               | - | dir : select a directory for this file only                         |
+|               |               |   | (overrides thedefault directory)                                    |
```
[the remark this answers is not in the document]
gautierhattenberger: The `cond` attribute allows to make a conditional compilation of functions, header or compilation instructions.
In the case of elements related to C code, `cond="COND"` will produce:
```
#if COND
...
#endif
```
In the case of elements related to makefile, `cond="COND"` will produce:
```
COND
...
endif
```
Note that in the case of makefile, the `COND` should include the "if" statement (like `ifdef`, `ifeq`, ....)

gautierhattenberger:
Some extra info. It is possible to have third party modules included with the `PAPARAZZI_MODULES_PATH` environment variable. See https://github.com/paparazzi/paparazzi/blob/master/sw/lib/ocaml/env.ml#L49
This was experimental, I have not checked for a long time if it is really working. But I think this should be documented (and fixed if needed).

Also, I made a picture to explain the general idea of modules, maybe you can add that at the beginning if you think it helps to clarify
PPRZ_Modules
.

matteobarbera:
@Fabien-B @gautierhattenberger Thanks for all the feedback! I tried to address all your comments in my commits. There are a couple of changes that I wasn't 100% sure of:

- dep and children: what is meant by functionalities? Name of a function within the module? Should the '@' be added only in the `depends` and `conflicts` nodes and not `provides`?
- Regarding the third party modules, where should the `PAPARAZZI_MODULES_PATH` be specified? Within a file or in the terminal `extern PAPARAZZI_MODULES_PATH=...` etc?

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -4,6 +4,535 @@
...
+certain variables, or changing the value of more variables at once) a ``handler`` attribute can be added to specify
+a macro to be called whenever the setting is changed. This macro is associated with a module and **must be named**
+``module-name_handler-name()``.
+
+As an example, take a look at an excerpt from ``conf/modules/digital_cam.xml``:
+
+.. code-block:: xml
+
+  <dl_settings name="dc">
+    <dl_setting max="255" min="0" step="1" module="digital_cam/dc" var="0" handler="send_command" shortname="Shutter">
```
Fabien-B, line 492: Add closing tag.
```suggestion
  <dl_settings name="dc">
    <dl_setting max="255" min="0" step="1" module="digital_cam/dc" var="0" handler="send_command" shortname="Shutter">
  </dl_settings>
```

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -4,6 +4,535 @@
...
+.. code-block:: C
+
+	#ifndef MY_DEFINE
+	#define MY_DEFINE 0
+	#endif
+	int my_setting = MY_DEFINE;
+
+In this example, ``MY_DEFINE`` provides the initial value of ``my_setting``. ``MY_DEFINE`` can be set from
+the airframe file, but if it is not defined there this code will give it a default value of 0. The actual parameter
+is stored in ``my\_setting``, for which a ``<dl_setting>`` element is included in the module's XML file.
```
Fabien-B, line 514:
```suggestion
is stored in ``my_setting``, for which a ``<dl_setting>`` element is included in the module's XML file.
```

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -4,6 +4,535 @@
...
+    │  │  │  ├── module1.xml
+    │  │  │  ├── module1.h
+    │  │  │  └── module1.c
+    │  │  ├── module2
+    │  │  │  ├── module2.xml
+    │  │  │  ├── module2.h
+    │  │  │  └── module2.c
+
+.. warning::
+  TODO Where should ``PAPARAZZI_MODULES_PATH`` be specified?
```
Fabien-B, line 538: Its an environment variable.

Fabien-B (changes requested):
A few more fixes, and I will merge it soon!

gautierhattenberger:
"functionalities" are keywords specified in the `provides` nodes. So, as you said, a `@` is only needed for the other nodes `requires` and `conflicts`.

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -4,6 +4,535 @@
...
+|               | provides      | Advertises the functionality that the module provides (e.g. actuators,  |
+|               |               | imu)                                                                    |
+|               +---------------+-------------------------------------------------------------------------+
+|               | conflicts      | Comma separated list of conflicting modules                            |
+|               |               |                                                                         |
+|               |               | The elements can be a module name (as set in the module XML ``name``    |
+|               |               | node) or a functionality, which has to be preceded by @                 |
++---------------+---------------+-------------------------------------------------------------------------+
+| | autoload    | name          | The name of the module which should also be automatically loaded        |
+| | (0 or 1)    |               |                                                                         |
```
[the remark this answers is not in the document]
gautierhattenberger: Not exactly. At the moment we don't have a nice way to load alternate modules like gps or imu for simulation. So you can't always skip this autoload thing (it will try to load them even if the parent module is not selected). But in general, for "normal" dependency it should not be used.

doc/sphinx/source/developer_guide/modules.rst
```diff
@@ -4,6 +4,538 @@
...
++---------------+---------------+-------------------------------------------------------------------------+
+| | event       | fun           | Event function name called in each cycle of the main AP loop            |
+| | (0 or more) |               |                                                                         |
++---------------+---------------+-------------------------------------------------------------------------+
+| | datalink    | message       | Name of the datalink (uplink) message to be parsed                      |
+| | (0 or more) +---------------+-------------------------------------------------------------------------+
+|               | fun           | Name of the function called when a message arrived                      |
++---------------+---------------+-------------------------------------------------------------------------+
+| | makefile    | target        | A list of build targets separated with pipes                            |
+| | (0 or more) |               | (ex: ``<makefile target="tunnel\|foo">``)                               |
```
Fabien-B, line 282:
```suggestion
| | (0 or more) |               | (ex: ``<makefile target="tunnel|foo">``)                                |
```

Fabien-B (changes requested):
I spotted one last typo :smiley:

doc/sphinx/source/developer_guide/modules.rst
matteobarbera: Nice catch :)

Fabien-B (approved):

PR #2863, opened 2022-04-19
New python supervision

Fabien-B:
Complete rewrite of the Python paparazzi center.
Please test it and report any bugs.


sw/supervision/python/.gitignore
```diff
@@ -0,0 +1 @@
+generated/
```
gautierhattenberger, line 1: maybe it is easier to include the correct generated files (but still keep the .ui files for developers) so standard user can use this without installing development libraries and running the makefile

sw/supervision/python/Makefile
```diff
@@ -0,0 +1,15 @@
+
+all:
```
gautierhattenberger, line 2: add a `clean` target ?

sw/supervision/python/app_settings.py
```diff
@@ -0,0 +1,26 @@
+from PyQt5.QtWidgets import *
```
gautierhattenberger, line 1: add license headers to all python files ?

gautierhattenberger (changes requested):

sw/supervision/python/.gitignore
Fabien-B: ok, done.

sw/supervision/python/Makefile
Fabien-B: done

sw/supervision/python/app_settings.py
Fabien-B: I added a short 2 line notice. Not a big fan of huge license headers.

gautierhattenberger (approved):

PR #2871, opened 2022-04-22
[ModuleCreator] Increase maximum periodic frequency.

Fabien-B:


fvantienen:
Why limit to 999? What if we want to run even faster like a couple of kHz?

Fabien-B:
You're right, increased to 10Khz, which is the resolution of the RTOS.

fvantienen (approved):
Thanks!

PR #2873, opened 2022-04-28
[indi] only calculate command if in flight

gautierhattenberger:
In particular, don't run `wls_alloc` if not needed, it has a good chance to fail while on ground anyway.
The CPU usage is rising when this optimization is called for nothing, which can be avoided.


EwoudSmeur:
LGTM. What do you mean that it has a good chance to fail? I hope not 😅

gautierhattenberger:
Actually, I have not checked the return status to know if it stops after too many loops or not, but for sure the computation time is at least doubled while on ground !

PR #2876, opened 2022-05-03
[sphinx] Improvements to the ABI section

matteobarbera:
I added a little more information to some of the sections as well as rearranging the contents to something that I think reads more logically. I also added a small section detailing all the steps that must be taken to create a new message.

### DIFF
--- a/doc/sphinx/source/developer_guide/abi.rst
+++ b/doc/sphinx/source/developer_guide/abi.rst
@@ -21,7 +21,11 @@ General Idea
 Message definition
 ------------------
 
-The messages are described in ``conf/abi.xml`` analoguous to other messages in paparazzi. Name are unique, IDs starts from 0. "type" field should be a valid C type or pointer.
+The messages are described in ``conf/abi.xml`` analoguous to other messages in paparazzi. Names are unique, IDs starts from 0.
+``"type"`` field should be a valid C type or pointer.
+When adding a new message, make sure that the ``"id"`` is not used by any other message. After changing the XML file, you need to
+run ``make clean`` and ``make`` in the ``paparazzi`` folder to generate the corresponding .c and .h files (for more information
+on the code generation have a look at the `In depth`_ section).
 
 .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -49,58 +82,68 @@ Include header and declare an ``abi_event`` as a global ``static`` variable (but
      // do something here
     }
 
+.. warning:: Make sure the callback doesn't take too much time (process the received data elsewhere), as ABI callbacks are not thread safe.
```
gautierhattenberger, line 85: It is true that ABI callback are not thread safe, however making it short or long to process is not related:
- not thread safe: don't use it for inter-thread communication unless you have implement yourself a proper protection for the shared memory
- the ABI callback is called synchronously to the data send function and thus should be short (and obviously non-blocking) to avoid delaying the execution of the rest of the main loop.

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -49,58 +82,68 @@ Include header and declare an ``abi_event`` as a global ``static`` variable (but
...
+
 In the initialization function (or later) call the binding function for the message you want to receive.
 
 .. code-block:: C
 
     AbiBindMsgDATA(ABI_BROADCAST, &ev, data_cb);
 
-
 The first parameter is the sender ID you want to receive the message from.
+Senders IDs can be found in the file ``sw/airborne/modules/core/abi_sender_ids.h``
```
gautierhattenberger, line 94: I think we should keep a note here to recall the special binding IDs `ABI_DISABLE` and `ABI_BROADCAST`

doc/sphinx/source/developer_guide/abi.rst
```diff
@@ -49,58 +82,68 @@ Include header and declare an ``abi_event`` as a global ``static`` variable (but
...
-    You must also avoid using an ID already used to send the same message type.
+#. In ``sw/airborne/modules/core/abi_sender_ids.h`` define a ``SENDER_ID`` for your message as described in `Message definition`_
+
+#. Run ``make clean`` and ``make`` in the ``paparazzi`` folder
+
+#. In your module .c file where the new message should be sent, include the header ``"modules/core/abi.h"``, and call
+the function ``AbiSendMsg<YOUR_MESSAGE_NAME>(<YOUR_SENDER_ID>, ...)`` with the appropriate function parameters corresponding
+to the message fields, as shown in `Airborne code for publisher`_.
+
+#. To receive this message in another module .c file, follow the steps described in `Airborne code for subscriber`_
```
gautierhattenberger, line 137: When creating a new module with the GUI tool (from the menu Tools -> Module Creator), you can specify the ABI bindings that you will need, and it will generate the callbacks with the correct prototypes. I'm not really sure we should mention this, since this tool is not really documented yet, but it is still a nice feature when starting a new module.

gautierhattenberger (changes requested):

gautierhattenberger (approved):

PR #2878, opened 2022-05-09
[chibios] usb serial

Fabien-B:
Serial over USB for ChibiOS.


conf/modules/shell.xml
```diff
@@ -28,5 +28,6 @@
     <define name="$(SHELL_PORT_UPPER)_BAUD" value="$(SHELL_BAUD)"/>
     <define name="SHELL_DEV" value="$(SHELL_PORT_LOWER)"/>
     <define name="SHELL_DYNAMIC_ENTRIES_NUMBER" value="$(SHELL_DYNAMIC_ENTRIES_NUMBER)"/>
+    <file_arch name="usb_ser_hw.c" dir="."/>
```
Fabien-B, line 31: How to build this file only if SHELL_DEV is `usb_serial` ?

fvantienen:
Really nice that you added this driver! I wanted to add it as well.
Shouldn't the USB serial be a separate module or just always be compiled? That could solve also your file_arch problem and makes it possible to also use it for telemetry or other UART's. Maybe implement it like normal UART's such that if `USE_...` is defined it is loaded.
This doesn't work in combination with the MSD driver I guess now? That should be possible and relatively easy to implement and is something I want as well.

sw/airborne/mcu_periph/usb_serial.h
```diff
@@ -32,7 +32,16 @@
 #include "std.h"
 #include "pprzlink/pprzlink_device.h"
 
+// we really need just 1 byte
+#define USB_RX_BUFFER_SIZE 10
+
 struct usb_serial_periph {
+  /** Receive buffer */
+  uint8_t rx_buf[USB_RX_BUFFER_SIZE];
```
fvantienen, line 40: Maybe make this the same size as the UART buffers?

Fabien-B:
Good idea, just including `usb_ser_hw.c` in the uart module works nicely.
For the `transparent_usb` module, and all VCOM functions, I'm not sure why it there, it has something to do with the baremetal arch.
I could make the MSD and serial work together, but for now the autopilot is shut down before MSD is started, to avoid logging while MSD is running, and also for safety reasons (so the motors won't spin...)

fvantienen:
It fails because the nps doesn't have the file, so a condition should be added:
```
make[2]: *** No rule to make target '/home/pprz/paparazzi/var/aircrafts/quad_mavlink/nps/arch/sim/usb_ser_hw.c', needed by '/home/pprz/paparazzi/var/aircrafts/quad_mavlink/nps/arch/sim/usb_ser_hw.o'.  Stop.
```

Btw it can then also be removed from all the other modules that already depend on the `uart` module.

sw/airborne/mcu_periph/usb_serial.h
gautierhattenberger: I'm not found of including uart header here, but why not. At least USB buffer should be configurable independently of uart buffer.

conf/modules/uart.xml
```diff
@@ -17,6 +17,7 @@
     <file name="uart.c" dir="mcu_periph"/>
     <file_arch name="uart_arch.c" dir="mcu_periph"/>
     <file name="serial_port.c" dir="arch/linux" cond="ifeq ($(ARCH), linux)"/>
+    <file_arch name="usb_ser_hw.c" dir="."/>
```
gautierhattenberger, line 20: Should we have a condition for this file ?

Fabien-B:
I added a setting, so that when `HAL_USE_SERIAL_USB` is defined, the logger wait for this setting to be enabled for mounting the USB storage.
If `HAL_USE_SERIAL_USB` is not defined, the usual behavior apply: the USB storage is mounted when USB is plugged.

sw/airborne/mcu_periph/usb_serial.h
Fabien-B: Its configurable now. But its really just an buffer around the real buffer to be compliant with the `char_available`, `get_char` interface.

OpenUAS:
Thanks, really great if all works out :1st_place_medal:

fvantienen:
If you fix the linux firmware it is good to merge:
```
make[2]: *** No rule to make target '/home/pprz/paparazzi/var/aircrafts/ardrone2_gazebo/ap/arch/linux/usb_ser_hw.c', needed by '/home/pprz/paparazzi/var/aircrafts/ardrone2_gazebo/ap/arch/linux/usb_ser_hw.o'.  Stop.
make[2]: *** Waiting for unfinished jobs....
Makefile.ac:212: recipe for target 'ap.compile' failed
make[1]: *** [ap.compile] Error 2
```

I guess the best option would be to maybe add a condition for linux or a dummy?

fvantienen (approved):

fvantienen:
For me this would be ok to squash and merge

gautierhattenberger (approved):

PR #2882, opened 2022-05-19
[sphinx] Adds Logging section in Developer Guide

matteobarbera:
I've written a page to detail some of the logging options that are available for Paparazzi. Not everything that I saw on the Paparazzi Wiki made it here because from my understanding most of the other options will soon become deprecated. 

I wasn't 100% sure about the information on the Pprzlog format so comments on that section are welcome.

### DIFF
--- a/doc/sphinx/source/developer_guide/logging.rst
+++ b/doc/sphinx/source/developer_guide/logging.rst
@@ -1,0 +1,7 @@
+.. developer_guide logging
+
+=============
+Logging
+=============
+
+TBD
--- a/doc/sphinx/source/developer_guide/logging.rst
+++ b/doc/sphinx/source/developer_guide/logging.rst
@@ -138,7 +138,7 @@ airframe in the ``conf.xml``. For example, ``conf/userconf/tudelft/conf.xml`` sp
      ac_id="20"
      airframe="airframes/examples/bebop.xml"
      radio="radios/dummy.xml"
-     telemetry="telemetry/default_rotorcraft.xml"  <!-- The telemetry file -->
+     telemetry="telemetry/default_rotorcraft.xml"
      flight_plan="flight_plans/rotorcraft_basic.xml"
      settings="settings/rotorcraft_basic.xml settings/control/rotorcraft_speed.xml"
      settings_modules="modules/ahrs_float_mlkf.xml modules/air_data.xml modules/bebop_ae_awb.xml modules/bebop_cam.xml modules/geo_mag.xml modules/gps.xml modules/guidance_rotorcraft.xml modules/imu_common.xml modules/ins_extended.xml modules/nav_basic_rotorcraft.xml modules/stabilization_int_quat.xml modules/video_rtp_stream.xml"
@@ -196,7 +196,7 @@ The name of the files associated to a specific log is the same, and is generated
 The lines of the ``data`` file are formatted according to the message description listed in the ``conf/messages.xml``
 file. For example:
 
-.. code-block::
+.. code-block:: text
 
   30.5941 186 ATTITUDE 0.036228 0.018550 0.021443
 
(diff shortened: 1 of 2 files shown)


doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+.. code-block:: C
+
+  #define PRINT(string,...) fprintf(stderr, "[orange_avoider->%s()] " string,__FUNCTION__ , ##__VA_ARGS__)
+
+This allows you to only have to write ``PRINT("<your debug message>")`` in your code and it will automatically
+prepend ``"[orange_avoider->%s()] "`` to your message, which can help debugging if multiple modules
+are printing information simultaneously.
+
+When running in simulation ``fprintf`` automatically prints the message in the Paparazzi Center. You can also
+read these messages on the real drone using the following steps:
```
gautierhattenberger, line 43: I suggest to specify that this example is only valid for Bebop 1 et 2.

doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+Logging on ChibiOS systems can be achieved using the ``flight_recorder`` module, which depends on a
+few other modules that also need to be included in your airframe.
+
+.. code-block:: xml
+
+  <firmware name="rotorcraft">
+    <module name="tlsf"/>
+    <module name="pprzlog"/>
+    <module name="logger" type="sd_chibios"/>
+    <module name="flight_recorder"/>
```
gautierhattenberger, line 106: With the new dependency solver, it is only necessary to load `flight_recorder` module.

Also, when using `sd_chibios` only, a log file is opened and can be used to log any type of data in ASCII (or even binary if needed) like a regular file. When loading the flight recorder module, two separated files are opened. Have a look at the `meteo_stick` module for an example of using the different options depending of the module's configuration.

doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+.. tip::
+
+  You can add the ``export`` lines above to your ``.bashrc`` file to automatically set the environment variables
+  every time a new Terminal is opened.
+
+Once the environment variables are set you can run ``sd2log`` in your terminal:
+
+.. code-block:: bash
+
+  ~/paparazzi/sw/logalizer/sd2log fr_XXXX.LOG
```
gautierhattenberger, line 171: `~/paparazzi/sw/logalizer/sd2log fr_XXXX.LOG output_dir_path` to extract the log in an other folder than default log path

doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+
+Pprzlog format
+-----------------
+
+The Pprzlog format creates multiple files that can be used to analyse flight data and replay the flight. A log is
+split into the following files:
+
+- A ``.log`` file, an XML file, which contains a copy of the whole configuration (airframes, flight plans, ...)
+- A ``.data`` file, an ascii file, which contains the list of the received messages. Each message is time-stamped in
+  milliseconds since the creation of the file and marked with the ID of the sending aircraft
```
gautierhattenberger, line 192: actually a floating point value in seconds with 3 digits

doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+- A ``.tlm`` file, ???
```
gautierhattenberger, line 193: the `tlm` file is a copy of the original `.LOG` file, with the same naming `data_time.*`. It makes it easier to associate with decoded log files.
By the way, a recommended use is to directly call `sd2log` on the file still on the SD card mounted as a mass storage, it is not really needed to copy it first on the computer.

doc/sphinx/source/developer_guide/logging.rst
```diff
@@ -0,0 +1,264 @@
...
+
+  <message name="ATTITUDE" id="6">
+    <field name="phi"   type="float" unit="rad" alt_unit="deg"/>
+    <field name="psi"   type="float" unit="rad" alt_unit="deg"/>
+    <field name="theta" type="float" unit="rad" alt_unit="deg"/>
+  </message>
+
+In this case, at the time the message was logged, the attitude of the drone corresponded to :math:`{\phi} = 0.036228`,
+:math:`{\theta} = 0.018550`, and :math:`{\psi} = 0.021443` radians. Note that the appropriate ``messages.xml``
+description, i.e. the one which has been used while th elog was created, is itself stored in the associated ``.log``
```
gautierhattenberger, line 216: `the log`

gautierhattenberger (changes requested):

doc/sphinx/source/developer_guide/logging.rst
matteobarbera: Is this not possible with every Linux based OS that has WiFi and specifies a host for a telnet connection? E.g. the Disco, although during flight you wouldn't be able to maintain WiFi connection

matteobarbera:
I've addressed the changes (except for the Bebop 1/2 comment, I thought it worked with every Linux based OS that has WiFi and that specifies a host) and added a section on SD ChibiOS logger covering how to create custom log functions for binary and ASCII logs. 

I used `meteo_stick.c` as a reference as suggested, but as I've never attempted to do something like this it's very possible that something is incorrect in my explanation  :sweat_smile:

doc/sphinx/source/developer_guide/logging.rst
gautierhattenberger: It's true that it works for Disco which is sharing the same hardware than Bebop more or less. The only Linux-based system that we support are Parrot drones, so I guess it is fine even if there are so things specific to these platforms, like default IP or directory tree.

gautierhattenberger (approved):

Issue #2787, opened 2021-10-04
CPPM input stopped working in v6.0_dirty


OpenUAS:
On a Lisa MXS flightcontroller I've flashed git object ID 0b64c15879fd9659c85c5c03b2041551f871e8fa
This is the closest  ID I could find before the massive changes of building the firmware where pulled in.

There CPPM input still works, tested with Airframe openuas_wltech_xk_a160.xml

On **exactly the same hardware**
This Git ID does not work anymore... 23d99579550630b60c51500ab4502e5c6215efd1

This makes continuous test-flying the 6.0_Dirty branch on Lisa M(X)(S) (and other AP's also) quite impossible.
Even more so due to e.g. multiple https://flapper-drones.com/wp/  with very tiny receivers with only CPPM output exist replacing all RX is not feasible on those.

OpenUAS:
Maybe relevant, same everything with the only difference when compiling

 <target name="ap" board="lisa_mx_2.1_chibios">
to
 <target name="ap" board="lisa_mx_2.1">

Gives:

**Chibios:**
mcu_periph/sys_time.c:39:1: note: #pragma message: Config: SYS_TIME_FREQUENCY = 10000
   39 | PRINT_CONFIG_VAR(SYS_TIME_FREQUENCY)

**Baremetal:**
mcu_periph/sys_time.c:39:1: note: #pragma message: Config: SYS_TIME_FREQUENCY = (2 * 512)
   39 | PRINT_CONFIG_VAR(SYS_TIME_FREQUENCY)

gautierhattenberger:
chibios and baremetal used to have a different timing for the polling of event and main freq. However, I have harmonized that part with the recent code reorg with #2721

OpenUAS:
@gautierhattenberger Want to debug to help fixing, a hint what has the highest chance of being broken would be great. 

BTW already browsing through relevant sourccode, and plan getting BMP2 debug setup on the LisaMX2, but there is just so much to do  and this make testflying for other thing to fix a bit problematic. Any help in a timely fix to try out welcomed.

gautierhattenberger:
just to be sure, we are talking about the `radio_control_ppm` not working with ChibiOS, right ?

OpenUAS:
[...]

PR #2887, opened 2022-07-05
fixed CPPM reception issue for lisa_m and lisa_mx ap

pcampolucci:
bug fix from issue: https://github.com/paparazzi/paparazzi/issues/2787


OpenUAS (dismissed):
Look sane to me, thx

pcampolucci:
> A lot of airframe are now failing due to the `#error` message. We should either:
> 
> * change the error for a warning
> * fix the airframes
> * disable the PWM by default for these boards

I would keep the PWM enabled, so I added a warning. 

Also, @OpenUAS this fix applies also to chibios boards.

gautierhattenberger (approved):

PR #2888, opened 2022-07-15
[fix] update the autopilot throttle in manual for fixedwing

gautierhattenberger:
this is only fixing the display and the ENERGY message, flight behavior is not affected


OpenUAS:
Before the fix :confused:

before_throttle_fix

After the fix :relieved:

after_throttle_fix

Thanks a lot, no more confusion between GCS operator and Pilot ;-)

OpenUAS (approved):
It works again, great :1st_place_medal:

PR #2889, opened 2022-07-20
Add QMC5883 Magnetometer Driver

pcampolucci:
Added driver to run the magnetometer here https://nettigo.pl/attachments/440 through I2C communication protocol. 

@OpenUAS I modified quite a bit at a low lever, but copied the postprocessing of the raw signals obtained. 

@gautierhattenberger and @OpenUAS please let me know if there is some additional features you want or some explaination needed. 

Cheers


sw/airborne/modules/sensors/mag_qmc5883.c
```diff
@@ -0,0 +1,136 @@
...
+#endif
+#if MODULE_QMC5883_UPDATE_AHRS ||  MODULE_QMC5883_SYNC_SEND
+    mag_qmc5883.data_available = false;
+#endif
+  }
+}
+
+void mag_qmc5883_report(void)
+{
+  //debuggy = (int32_t)(mag_qmc5883.init_status);
```
OpenUAS: leftovers can be removed ;)

sw/airborne/modules/sensors/mag_qmc5883.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2021 Paparazzi Team
```
OpenUAS, line 2: Make it 2022, I've never released the old branch

sw/airborne/modules/sensors/mag_qmc5883.c
```diff
@@ -0,0 +1,288 @@
...
+#endif
+
+/**
+ * Initialization Function
+ */
+void mag_qmc5883_init(void)
+{
+  qmc.trans.status = I2CTransDone;
+  qmc.waddr = QMC5883_WRITE_ADDR;
+  qmc.raddr = QMC5883_READ_ADDR;
```
gautierhattenberger, line 113: you don't need 2 different addresses, only one on 8 bits, without the terminal 0/1. So in you case the `QMC5883_WRITE_ADDR` the R/W bit is added by the driver if needed.

sw/airborne/modules/sensors/mag_qmc5883.c
```diff
@@ -0,0 +1,288 @@
...
+      // get raw data for debug
+      qmc.raw_mag_data[0] = (uint8_t)(qmc.trans.buf[0]);
+      qmc.raw_mag_data[1] = (uint8_t)(qmc.trans.buf[1]);
+      qmc.raw_mag_data[2] = (uint8_t)(qmc.trans.buf[2]);
+      qmc.raw_mag_data[3] = (uint8_t)(qmc.trans.buf[3]);
+      qmc.raw_mag_data[4] = (uint8_t)(qmc.trans.buf[4]);
+      qmc.raw_mag_data[5] = (uint8_t)(qmc.trans.buf[5]);
+      qmc.status_register.DRDY = ((uint8_t)(qmc.trans.buf[6]) >> 0) & 1;
+      qmc.status_register.OVL = ((uint8_t)(qmc.trans.buf[6]) >> 1) & 1;
+      qmc.status_register.DOR = ((uint8_t)(qmc.trans.buf[6]) >> 2) & 1;
```
gautierhattenberger, line 244: I don't think this is the right way to do this. Here, you need 3 calls of the periodic function to process a new data, so from 100Hz you go down to 33Hz. Also you use two steps to set the read register and then read the data, while it can be done in one step with `i2c_transceive` function.
So my suggestion here is:
- in the periodic function, at a merged state REQUEST_ACQUIRE, call i2c_transceive
```
qmc.trans.buf[0] = QMC5883_REG_DATXL;
i2c_transceive(&MAG_QMC5883_I2C_DEV, &qmc.trans, qmc.addr, 1, 7);
```
- in the event function, if state is PARSE and transaction is Success, parse data and send ABI event message. Don't forget to process the transaction failures or success in other sensor states.

sw/airborne/modules/sensors/mag_qmc5883_regs.h
```diff
@@ -0,0 +1,79 @@
...
+ * @file modules/sensors/mag_qmc5883_regs.h
+ * Register defenitions for the QMC5883 magnetometer.
+ */
+
+#ifndef QMC5883_REGS_H
+#define QMC5883_REGS_H
+
+/* read and write I2C addresses based on default 0x0D */
+#define QMC5883_READ_ADDR  0x1B
+#define QMC5883_WRITE_ADDR 0x1A
```
gautierhattenberger, line 32: only `#define QMC5883_ADDR 0x1A` is needed

gautierhattenberger (changes requested):

OpenUAS:
A complete rewrite see #2953

PR #2893, opened 2022-08-16
GVF firmware adjusted to work with GVF Viewer

jesusBV20:
GVF firmware was adjusted to work with GVF Viewer. This new PprzGCS tool allows us to paint parametric trajectories and the associated vector fields.

- The bug that doesn't allow us to compile fixed wing airframes has been fixed.
- Minor changes in GVF flight plans.

- With the GVF_OCAML_GCS switcher we can activate or deactivate the classic GCS painter.
- Classic and parametric GVF telemetry adjusted to send even more information using the parameters field of the GVF/GVF_PARAMETRIC message.

Changes in PprzGCS: https://github.com/paparazzi/PprzGCS/pull/15


noether (approved):
For me it looks okei. Lets see whether Gautier has any questions.

sw/airborne/firmwares/rover/guidance/rover_guidance_steering.c
```diff
@@ -71,7 +72,7 @@ void rover_guidance_steering_init(void)
 void rover_guidance_steering_heading_ctrl(void)
 {
   float delta = 0.0;
-  float omega = guidance_control.gvf_omega; //GVF give us this omega
```
gautierhattenberger, line 74: It seems that the only dependency of this guidance control to GVF is this variable. Maybe it is worth passing this value as an argument of the function so it can be reused ?
The function itself is probably called from the autopilot XML file. It from here that you can decide to use GVF or another strategy.

gautierhattenberger (changes requested):
I only have a small remark regarding guidance steering. If I'm wrong or if it is too complicated to change, maybe it can stay like this for now.

gautierhattenberger (commented):
Modes are coming from XML definition, this should be extracted automatically. It might be broken, but can you confirm that the modes are not displayed in the new GCS ? And what about the old one ?

noether:
> Modes are coming from XML definition, this should be extracted automatically. It might be broken, but can you confirm that the modes are not displayed in the new GCS ? And what about the old one ?

It happened the same in the old one (only with the rover firmware). We haven't figured out why (possibly until now).

sw/ground_segment/tmtc/server_globals.ml
```diff
@@ -8,6 +8,7 @@ let port = ref 8889
 (** FIXME: Should be read from messages.xml *)
```
jesusBV20, line 8: I think that the ap modes should be read from messages.xml, but It isn't implemented right now.

PR #2898, opened 2022-09-13
[chibios] Fix i2c error handling

fvantienen:
This makes the i2c error handling more robust and prevents the autopilot from hanging when the I2C line is detected busy. Previously it would hang the autopilot for 50ms, now it cleans the i2c bus in 20 * 10 microseconds = 0.2ms (busy hanging autopilot).

For the long run we should fix the priorities of the autopilot threads. Similar to ardupilot we could push a very high priority to the main control loop (~500Hz) on interrupts, to prevent the low level threads (like i2c, uart etc.) from hanging the main control loop. See https://forum.chibios.org/viewtopic.php?f=24&t=4631&sid=d32de529b5b1cd156c6fa22d1b53627a&start=10 for changing priority on the fly.


fvantienen:
This should be merged after https://github.com/paparazzi/ChibiOS/pull/1 is merged

alex31 (commented):
Seems clean to me. There is a choice to get dynamically  pin alternate configuration, via palReadLineMode(). There is another way of doing it, one can retrieve alternate function from board.cfg and store the information in the i2c_init struct. 

My only concern is with the use of chibios patches from ardupilot, the sooner theses patches are upstream, the better.

Did no go into detail in the chibios patches, but since ardupilot use it, and has accumulated hours of flight with theses patches, one can assume they are safe.

PR #2899, opened 2022-09-13
[link] Make sure that -nouplink also doesn't send ping

fvantienen:


gautierhattenberger (approved):
I agree to disable ping with `nouplink` option. We'll see if it is necessary in the future to add an option to "force" the ping even with no uplink to get ping time (and avoid datalink lost on AP side) while preventing to send commands.

Issue #2788, opened 2021-10-05
RM3100 magnetometer module frequency doesn't benefit from sensor's high data rate


matteobarbera:
Together with @OpenUAS, while investigating #2781, we saw that the rm3100 module only polls the sensor at the 60Hz periodic frequency defined in the XML https://github.com/paparazzi/paparazzi/blob/da5b5277958126862a5bc089e224dfef77514ac4/conf/modules/mag_rm3100.xml#L33

One of the advantages of the RM3100 is that its data rate goes up to 600Hz, and the data rate can be defined in the airframe with the DATA_RATE setting, but unless the user is aware that they have to change the module frequency too, all the benefits of increasing the sensor frequency are lost.

I thought at the beginning that setting it by default to a high periodic frequency would improve the situation, perhaps even not specifying a frequency in the module XML and letting it run at max frequency. However, we saw that when the data rate is lower than the module frequency there are spikes in the data coming from the sensor, so if a user sets it to a lower data rate than they would also encounter problems.

I would propose either:

- more detailed documentation in the module files
- find a way to programmatically change the module frequency with the data rate
- issue a warning to the user (during compilation time?) if the settings could cause issues. 

With the last two points I don't know if it is possible to obtain the periodic module frequency as a variable, and if it is at what stage of compilation.

OpenUAS:
@gautierhattenberger Thanks for the elaborate remarks, very nice! As soon as @matteobarbera is available again will fix and testfly together

As stideremark @matteobarbera  IMHO rather go for prescaling computed at compile time for more deterministic behavior.

PR #2901, opened 2022-09-19
[mag] Make RM3100 module frequency variable

matteobarbera:
This is a pull request addressing issue #2788. The module frequency is turned into a variable to allow the user to change it as they see fit to make use of the high polling frequency the RM3100 mag sensor is capable of.

With this change, the variable `MODULE_RM3100_RATE` can be defined in the airframe to change the default module frequency of 60 Hz.
```xml
<module name="mag" type="rm3100">
      <configure name="MODULE_RM3100_RATE" value="150"/>
</module>
```

I tested the PR by defining `MODULE_RM3100_RATE` and setting it to 60, 150, 300 to check that the frequency change works as intended.

rm3100_freq_pr

### DIFF
--- a/conf/modules/mag_rm3100.xml
+++ b/conf/modules/mag_rm3100.xml
@@ -7,7 +7,6 @@
       I2C interface.
     </description>
     <configure name="MAG_RM3100_I2C_DEV" value="i2cX" description="I2C device to use"/>
-    <configure name="MODULE_RM3100_RATE" value="60" description="Sensor polling rate (module frequency in Hz)"/>
     <define name="MODULE_RM3100_SYNC_SEND" value="TRUE|FALSE" description="Send IMU_RAW message with each new measurement (default: FALSE)"/>
     <define name="MODULE_RM3100_UPDATE_AHRS" value="TRUE|FALSE" description="Copy measurements to imu and send as ABI message (default: FALSE)"/>
     <define name="RM3100_CHAN_X_SIGN" value="+|-" description="Reverse polarity of x axis (default: +)"/>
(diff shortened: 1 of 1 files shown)


matteobarbera:
I wanted to also add a check for frequencies less than 1 in [mag_rm3100.c](https://github.com/tudelft/paparazzi/blob/5e2a082a83a64515a3933dcba79290159c20defa/sw/airborne/modules/sensors/mag_rm3100.c), but for some reason this does not cause an error in compilation.

```c
#if (MODULE_RM3100_RATE < 1)
#error "Module frequency cannot be less than 1 Hz"
#endif
```

fvantienen:
The plot of 60Hz shows clearly that you read the same sensor measurement twice. I haven't looked at the code, butthis seems to me like there is an error in the implementation of the driver.

conf/modules/mag_rm3100.xml
```diff
@@ -44,6 +45,7 @@
     <configure name="MAG_RM3100_I2C_DEV" case="upper|lower"/>
     <define name="USE_$(MAG_RM3100_I2C_DEV_UPPER)"/>
     <define name="MAG_RM3100_I2C_DEV" value="$(MAG_RM3100_I2C_DEV_LOWER)"/>
+    <define name="MODULE_RM3100_RATE" value="60"/>
```
fvantienen, line 48: Why this define here if you do nothing with it?

conf/modules/mag_rm3100.xml
```diff
@@ -7,6 +7,7 @@
       I2C interface.
     </description>
     <configure name="MAG_RM3100_I2C_DEV" value="i2cX" description="I2C device to use"/>
+    <define name="MODULE_RM3100_RATE" value="60" description="Sensor polling rate (module frequency in Hz)"/>
```
fvantienen, line 10: Shouldn't this be a configure?

fvantienen (changes requested):

matteobarbera:
> The plot of 60Hz shows clearly that you read the same sensor measurement twice. I haven't looked at the code, but this seems to me like there is an error in the implementation of the driver.

The graph is based on data from the sdlogger logging at 500Hz, could also be because of this?

conf/modules/mag_rm3100.xml
matteobarbera: I added it there too because otherwise if the user does not specify the variable in the airframe XML the compilation throws an error. This way it at least defaults to the old value of 60 Hz

gautierhattenberger:
Check this module to see how to make a configurable period with default value: https://github.com/paparazzi/paparazzi/blob/master/conf/modules/baro_ms5611_i2c.xml#L23

gautierhattenberger (approved):

PR #2903, opened 2022-09-20
[gps][ins] improve gps support for ins filters

gautierhattenberger:
- use conveniance functions to access GPS position and speed
- try to improve support of UBX fix status

### DIFF
--- a/sw/airborne/modules/gps/gps_ubx.c
+++ b/sw/airborne/modules/gps/gps_ubx.c
@@ -102,6 +102,7 @@ void gps_ubx_init(void)
   gps_ubx.msg_available = false;
   gps_ubx.error_cnt = 0;
   gps_ubx.error_last = GPS_UBX_ERR_NONE;
+  gps_ubx.pacc_valid = false;
 
   gps_ubx.state.comp_id = GPS_UBX_ID;
 }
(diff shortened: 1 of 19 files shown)


sw/airborne/modules/gps/gps_ubx.c
```diff
@@ -194,14 +194,20 @@ static void gps_ubx_parse_nav_pvt(void)
   gps_ubx.state.hacc        = UBX_NAV_PVT_hAcc(gps_ubx.msg_buf) / 10;
   gps_ubx.state.vacc        = UBX_NAV_PVT_vAcc(gps_ubx.msg_buf) / 10;
   gps_ubx.state.sacc        = UBX_NAV_PVT_sAcc(gps_ubx.msg_buf) / 10;
+
+  //FIXME stupid workaround for PVT only
+  gps_ubx.state.pacc = gps_ubx.state.hacc; // report horizontal accuracy
```
fvantienen, line 199: Doesn't this overwrite a pacc if we read a valid one from another message as well?

sw/airborne/modules/gps/gps_ubx.c
```diff
@@ -380,6 +388,7 @@ static void gps_ubx_parse_nav_relposned(void)
       } else{
         gps_ubx.state.fix = 0;
       }
+    if(relPosValid) {
```
fvantienen, line 391: Not sure if the fix state is valid in this case, but I guess it is ok. Though indentation seems off.

fvantienen (approved):
Looks ok, but maybe run the code styler on some files to ix indents and other small sttyle mistakes.

sw/airborne/modules/gps/gps_ubx.c
gautierhattenberger: I don't have a good solution for now. Reconstructing `pacc` from `hacc` and `vacc` is not specified by ublox. With this, it is just not advised to use PVT with an other message providing `pacc`. I would prefer not to add another config flag. Any suggestion ? Using a bool to tell if `pacc` is already updated somewhere else ?

sw/airborne/modules/gps/gps_ubx.c
gautierhattenberger: Maybe I should revert this, it was a quick fix that I didn't use in the end. So if there is a chance that this is wrong, better remove it.

fvantienen (approved):

PR #2913, opened 2022-10-03
[nav] add a guided instruction for the flight plan

gautierhattenberger:
It allows (for rotorcraft) to use all the possibilities of the guided mode from a single instruction in the flight plan, but without leaving the NAV mode as it was done before. Thus all capabilities of the FP are still accessible.


fvantienen (approved):
If this is tested it is ok for me to merge

gautierhattenberger:
I still have a few things to merge and test, but it is starting to work

gautierhattenberger:
close and replace with #2931

PR #2915, opened 2022-10-06
Fix imu conf

gautierhattenberger:
- fix signs corrections
- fix binding of ahrs aligner

### DIFF
--- a/sw/airborne/modules/ahrs/ahrs_aligner.c
+++ b/sw/airborne/modules/ahrs/ahrs_aligner.c
@@ -121,7 +121,7 @@ void ahrs_aligner_run(void)
   // Could not find all sensors
   if(gyro == NULL || accel == NULL)
     return;
-  
+
   RATES_ADD(gyro_sum, gyro->scaled);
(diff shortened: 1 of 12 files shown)


fvantienen:
The `IMU_..._SIGN` seem to be interesting but I see a problem when you load a new type of calibration, then the signs aren't applied. With the current PR state you have this:
- Signs apply to defaults from sensors
- Signs apply to old calibration
- Signs **don't** apply to new calibration

This makes it a bit weird if you have the signs defined and go from default to new calibration. What do you think?

gautierhattenberger:
I'm considering that the two calibrations are not compatible, you either choose one or another. With the new calibration, the correct way is to provide a rotation matrix. It would be useful to provide a set of macros with precomputed matrix, are at least to make configuration easy. But old config should be only for single IMU, without mixing with new one.
Also, all airframes with some negative signs are broken in the current master...

fvantienen:
Ok then maybe if we add a warning when you have `IMU_..._SIGN` defined and a new calibration that it is incompatible would be good to add

fvantienen (approved):

PR #2918, opened 2022-10-07
[nps] fix imu neutrals

### LINKED ISSUE #2917: new IMU calibration causes wrong sensor data from NPS
Running the NPS simulation in example airframes (airframes/examples/ardrone2.xml to be exact), I came across a problem related to **IMU ACCEL calibration values** in this airframe:

https://github.com/paparazzi/paparazzi/blob/46055fc8f8607d72954277704bb90e0b535a0523/conf/airframes/examples/ardrone2.xml#L75-L77

These numbers seem to make trouble in scaled IMU values when using NPS, making **wrong attitude estimation in NPS** simulation.

The problem arises from here:
https://github.com/paparazzi/paparazzi/blob/46055fc8f8607d72954277704bb90e0b535a0523/conf/simulator/nps/nps_sensors_params_default.h#L69-L71

which previously was:
https://github.com/paparazzi/paparazzi/blob/b991dcc8c264c18d11631992308c2276a49a4421/conf/simulator/nps/nps_sensors_params_default.h#L60-L62

As a result, **NPS doesn't compensate neutral values** defined for ACCEL in the airframe file but **IMU module in airborne code scales NPS output based on those neutral values**, causing wrong scaled values in attitude estimation.

One way to solve this is using new calibration method instead of those defines in the airframe file:
`<define name="ACCEL_CALIB" value="{{.abi_id=1, .calibrated={.neutral=true, .scale=false}, .neutral={2048,2048,2048}}}"/>`

But this requires that every airframe with old IMU calibration method and NPS simulation be modified. Is there a better solution?

gautierhattenberger:
setting correct neutrals if needed, not required with new IMU calibration

fix #2917


celarco (commented):
Thank you @gautierhattenberger. This seems to solve the issue.

gautierhattenberger:
@fvantienen is it ok for you ?

fvantienen (approved):
This is fine. Sorry for all the small mistakes.

PR #2919, opened 2022-10-10
[ground] Add heading to ublox2ivy

fvantienen:
Should be merged after https://github.com/paparazzi/pprzlink/pull/164

### DIFF
--- a/sw/ground_segment/misc/ublox2ivy.c
+++ b/sw/ground_segment/misc/ublox2ivy.c
@@ -33,6 +33,7 @@
 #include <netinet/in.h>
 #include <pthread.h>
 #include <string.h>
+#include <math.h>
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
It seems that `sw/ground_segment/tmtc/gpsd2ivy.c` is also sending the TARGET_POS message and it should be updated as well

fvantienen:
Updated gpsd2ivy as well, since it doesn't have heading information it sends NAN. That way the autopilot knows that it doesn't have heading information.

gautierhattenberger:
you can update the pprzlink ref now, after that it can be merged

fvantienen:
Updated the reference

gautierhattenberger (approved):

PR #2920, opened 2022-10-10
fix and update mavlink

Fabien-B:
Replace `python2` by `python` as python2 is deprecated, and it make paparazzi build fail on Ubuntu 22.04.

Also take this opportunity to update mavlink, and link to the original repository as the changes in the paparazzi fork are included upstream.

This PR needs testing, can someone using mavlink test it ?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 	url = https://github.com/swift-nav/libsbp.git
 [submodule "sw/ext/mavlink"]
 	path = sw/ext/mavlink
-	url = https://github.com/paparazzi/mavlink.git
+	url = https://github.com/mavlink/mavlink.git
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
It seems that there is a dependency issue with the `future` module from python

Fabien-B:
Then I propose to remove mavlink from the `all` Makefile target.
It is still possible to use it if you care to install the `future` module.

fvantienen:
Somehow it still got build in semaphore?

gautierhattenberger:
problem is coming from here: https://github.com/paparazzi/paparazzi/blob/master/Makefile#L109

Fabien-B:
Now it should work, when the docker image will be updated.

gautierhattenberger (approved):

Issue #2895, opened 2022-09-02
Installation issues Ubuntu 22 (libSDL.so)


tomvand:
Hi,

Due to laptop problems I have to reinstall paparazzi (commit 9098b7c1d) on a new system with Ubuntu 22, and here I'm running into some issues. I have manually installed git, build-essential and the likes, and then follow the one-liner instructions on the wiki here: https://wiki.paparazziuav.org/wiki/Installation (the Ubuntu 20 and above version). Unfortunately, running `make` seems to have two new issues that did not occur on my older Ubuntu 18 setup:

1. OCaml compilation gives a labels omitted warning: `Warning 6 [labels-omitted]: label callback was omitted in the application of this function.`
2. OCaml compilation of `input2ivy` fails because `libSDL.so` can't be found: 

```
File "_none_", line 1:
Error: I/O error: libSDL.so: No such file or directory
make[1]: *** [Makefile:78: input2ivy] Error 2
make: *** [Makefile:168: joystick] Error 2
```

Since this is a completely fresh install of Ubuntu 22, I suspect there is a dependency missing in the installation one-liner/documentation, or something has changed or moved in Ubuntu 22 or perhaps its ocaml packages.

I've uploaded the full ouput of the one-liner here: https://gist.github.com/tomvand/01b379e2ddc650575f814c66289f70f0 

libSDL.so is present on the system:
```
libSDL.so: /usr/lib/x86_64-linux-gnu/libSDL.so
```

I've verified that the following packages have been installed:
```
libsdl2-dev
libsdl1.2-dev
libsdl-ocaml
libsdl-ocaml-dev
```

I've also found the gitter discussion of June 14th this year and verified the additional dependencies mentioned there. If Ubuntu 22 is not supported yet, then this would be a good feature for the next release ;).

Cheers,
Tom

tomvand:
Ok, I managed to get it working. I've included the steps below. I don't know if every step was absolutely necessary, so consider these only for reference:

---

Remove apt's OCaml installation to prevent version conflicts with the manual install:
```
sudo apt remove *ocaml*
```

Install opam package manager for OCaml:
```
sudo apt install opam
opam init
eval $(opam env)
```

Switch to the latest OCaml version
```
opam switch create 4.14.0
eval $(opam env --switch=4.14.0)
```

Install dependencies except ivy
```
[...]

PR #2922, opened 2022-10-10
Fix gcc prefix and update libopencm3.

Fabien-B:
A fix was needed in libopencm3 (see #2895), it may be a good opportunity to update libopencm3.
Some fixes are needed to build with the latest libopencm3.

See:
https://github.com/libopencm3/libopencm3/commit/7be50a5e75ed2d163d38a6759347c5e778ac02ab
https://github.com/libopencm3/libopencm3/commit/034dbf20ff8c54dcbee5238390b37a0d35180f44

This PR needs to be tested, anyone using stm32 baremetal ?

Luftboot must be updated before first : https://github.com/paparazzi/luftboot/pull/10


Fabien-B:
There are some issue left with the definition of `RST_ADC1`.
It is defined for some targets, but defined as `RST_ADC` for the F4.

noether:
Sure, I will take a look at it

OpenUAS:
If you can finalize, then I can test on LisaM(X)(S) Pixracer and some others via baremetal. and adding AHB_CLK for 100Mhz for Crazybee Pro F4 boards would be welcomed. This is likely the last update for libopencm3 since PPRZ going to Chibios for all defaults, so make it worthwhile as best as possible :)

Fabien-B:
Should be ok now, could someone test it on a "stm32" arch ? (using libopencm3)

gautierhattenberger (approved):

PR #2926, opened 2022-10-11
[nps] fix imu sensitivity (#2925)

### LINKED ISSUE #2925: sensor calibration sensitivity needs to be fixed for NPS
This issue is similar to #2917 which was partly fixed with #2918. The problem is that **the sensitivities** also need to be fixed alongside neutrals. I will open a pull request for this. 

I am sorry I didn't mention that earlier. I didn't think this can cause trouble until I tested #2918 further.

celarco:
setting correct IMU sensitivity for NPS if needed

fix #2925


celarco:
Hi @gautierhattenberger. Would please review this PR?

gautierhattenberger (approved):

PR #2931, opened 2022-10-15
Tag tracking speed integration

gautierhattenberger:
- add a guided mode instruction in flight plan
- a speed control to the tag tracking module
- change guidance API to a more generic name (not related to GUIDED mode)
- add a demo flight plan

replace #2913

### DIFF
--- a/conf/airframes/ENAC/rover_ostrich.xml
+++ b/conf/airframes/ENAC/rover_ostrich.xml
@@ -72,17 +72,20 @@
 
   <section name="IMU" prefix="IMU_">
   
-    <define name="GYRO_P_SIGN" value="1"/>
+    <!--define name="GYRO_P_SIGN" value="1"/>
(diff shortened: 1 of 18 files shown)


gautierhattenberger:
@noether I also want to reorganize the guidance/nav for rotorcraft, so if you have your work in a public branch, let me know, at least to avoid double work and faster convergence (I have not started yet).

noether:
Ah cool! We did not start it yet (we started first with the guidance of the rover). Once we start, I will let you know it.

What I had in mind for the GVF is the following. GVF generates the desired 3D velocity or 3D acceleration, then sends such information to low-level controllers like INDI for the accelerations (this should work very well, we have done something similar with Ewoud when we did the collaborative transportation with rotorcraft). For the 3D velocity, it can always be split into vertical and horizontal, but the result maybe will not be the best.

PR #2934, opened 2022-10-20
Fix #2911

### LINKED ISSUE #2911: Test build fails

podhrmic:
This should fix #2911

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ before_install:
   - sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa -y
   - sudo apt-get update -q
 install:
-  - sudo apt-get install -y --allow-unauthenticated paparazzi-dev paparazzi-jsbsim gcc-arm-embedded gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libipc-run-perl rustc cargo libgazebo9-dev libgps-dev python3-lxml
+  - sudo apt-get install -y --allow-unauthenticated paparazzi-dev paparazzi-jsbsim gcc-arm-embedded gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libipc-run-perl libgazebo9-dev libgps-dev python3-lxml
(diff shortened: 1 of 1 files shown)


podhrmic:
@fvantienen not sure if all the checks have passed, but this is at least showing a possible solution

gautierhattenberger:
@podhrmic Travis is not used anymore (no longer free), so we need to include that to the docker image used in Semaphore. But thank you anyway for the command that we should run.

podhrmic:
@gautierhattenberger thanks for the info! I looked at the docker image, and it looks like it is already calling the right command (see [here](https://github.com/paparazzi/paparazzi/blob/master/docker/ci/Dockerfile#L4)). Perhaps the image need to be rebuilt?

gautierhattenberger:
It seems that docker hub is no longer doing automated builds with free plans, so the docker image is one year old now. Maybe that is the problem. I'm not that familiar with docker, so I'll need some time to find how to make the image and push it manually...

podhrmic:
I don't seem to have permissions to push to the [paparazziuav](https://hub.docker.com/u/paparazziuav) docker hub, but rebuilding the image from scratch *should* fix the problem.

gautierhattenberger:
The docker image is now updated with a github action.

PR #2943, opened 2022-10-27
[supervision] Make python supervision the default one.

Fabien-B:
I propose to make the python paparazzi center the default one.
The old one can still be use by launching `./paparazzi -legacy`.

### DIFF
--- a/paparazzi
+++ b/paparazzi
@@ -10,10 +10,11 @@ env = os.environ
 env['PAPARAZZI_HOME'] = PAPARAZZI_HOME
 env['PAPARAZZI_SRC'] = PAPARAZZI_SRC
 
-if len(sys.argv) > 1 and sys.argv[1] == "-python":
-    path = os.path.normpath(os.path.join(dirname, 'sw', 'supervision', 'python', 'paparazzicenter.py'))
+if len(sys.argv) > 1 and sys.argv[1] == "-legacy":
+    del sys.argv[1]
+    path = os.path.normpath(os.path.join(dirname, 'sw', 'supervision', 'paparazzicenter'))
     os.execve(path, sys.argv, env)
 else:
-    path = os.path.normpath(os.path.join(dirname, 'sw', 'supervision', 'paparazzicenter'))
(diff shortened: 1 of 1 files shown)


fvantienen (approved):

gautierhattenberger:
I think there is only one little thing missing to use the python center by default. The selection of the settings is only based on the modules' settings and the user settings that are already present in the conf xml file. It is not possible to add or remove user file like before (or by editing the conf file by hand).
So we have 2 options here:
- modify the interface to add the missing buttons
- get rid of user settings and only use modules' settings

On the long term, I would prefer the second one. Maybe it is not so difficult now that we have firmware specific modules. We can post-pone a bit this change until we know if old settings can be removed or not.

Fabien-B:
#2959 added the requested feature, should be ok to merge now?

dewagter:
Since this merge I get the following error: any idea?

Traceback (most recent call last):
  File "/home/mavlab/paparazzi/sw/supervision/python/paparazzicenter.py", line 8, in <module>
    from configuration_panel import ConfigurationPanel
  File "/home/mavlab/paparazzi/sw/supervision/python/configuration_panel.py", line 5, in <module>
    from generated.ui_configuration_panel import Ui_ConfigurationPanel
  File "/home/mavlab/paparazzi/sw/supervision/python/generated/ui_configuration_panel.py", line 75, in <module>
    from conf_widget import ConfWidget
  File "/home/mavlab/paparazzi/sw/supervision/python/conf_widget.py", line 7, in <module>
    from conf_settings_widget import ConfSettingsWidget
  File "/home/mavlab/paparazzi/sw/supervision/python/conf_settings_widget.py", line 3, in <module>
    from generated.ui_conf_settings import Ui_SettingsConf
ModuleNotFoundError: No module named 'generated.ui_conf_settings'

gautierhattenberger:
@Fabien-B the files `ui/conf_settings.ui` and `generated/ui_conf_settings.py` seem to be missing

Fabien-B:
Oops, here they are in #2961.

PR #2946, opened 2022-11-03
Convergence rebase

EwoudSmeur:
This is the code that was used for the convergence paper at IMAV2021, where we flew fixedwing, rotorcraft and hybrid all within the rotorcraft firmware: http://www.imavs.org/papers/2021/1.pdf

I am not particularly sure about the change in conf/modules/actuators_disco.xml, maybe @Dennis-Wijngaarden knows if that would still be needed on the current master.

### DIFF
--- a/conf/flight_plans/tudelft/delft_bebop_convergence.xml
+++ b/conf/flight_plans/tudelft/delft_bebop_convergence.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
 
-<flight_plan alt="10" ground_alt="0" lat0="51 59 27.6" lon0="4 22 42.0" max_dist_from_home="1000" name="Delft Basic" security_height="2">
+<flight_plan alt="5" ground_alt="0" lat0="51 59 27.6" lon0="4 22 42.0" max_dist_from_home="200" name="Delft Basic" security_height="2">
   <header>
     #include "autopilot.h"
     #include "subsystems/datalink/datalink.h"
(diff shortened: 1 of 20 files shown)


conf/flight_plans/tudelft/delft_bebop_convergence.xml
```diff
@@ -0,0 +1,144 @@
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+
+<flight_plan alt="5" ground_alt="0" lat0="51 59 27.6" lon0="4 22 42.0" max_dist_from_home="200" name="Delft Basic" security_height="2">
+  <header>
+    #include "autopilot.h"
+    #include "subsystems/datalink/datalink.h"
+    #include "subsystems/electrical.h"
+    #include "subsystems/radio_control.h"
```
gautierhattenberger, line 8: subsystems don't exist anymore

conf/flight_plans/tudelft/delft_bebop_convergence.xml
```diff
@@ -0,0 +1,144 @@
...
+    <sector color="yellow" name="Soft_Geo">
+      <corner name="SG1"/>
+      <corner name="SG2"/>
+      <corner name="SG3"/>
+      <corner name="SG4"/>
+    </sector>
+  </sectors>
+  <exceptions>
+    <!--<exception cond="!InsideFlight_Area(GetPosX(), GetPosY())" deroute="Holding point"/>-->
+    <exception cond="Or(!InsideFlight_Area(GetPosX(), GetPosY()), GetPosAlt() > GetAltRef() + 20) &&
```
gautierhattenberger, line 45: if possible, replace special characters with keywords (e.g `@AND`, `@OR`, `@GT`, `@LT`, ...)

conf/flight_plans/tudelft/disco_convergence.xml
```diff
@@ -0,0 +1,108 @@
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+
+<flight_plan alt="70" ground_alt="0" lat0="52.1682089" lon0="4.4128499" max_dist_from_home="500" name="Altura fp Valkenburg" security_height="30">
+  <header>
+    #include "autopilot.h"
+    #include "subsystems/datalink/datalink.h"
+    #include "subsystems/radio_control.h"
```
gautierhattenberger, line 7: remove subsystems

conf/flight_plans/tudelft/disco_convergence.xml
```diff
@@ -0,0 +1,108 @@
...
+        <go from="p2" hmode="route" wp="p3"/>
+        <go from="p3" hmode="route" wp="p4"/>
+        <go from="p4" hmode="route" wp="p1"/>
+      </while>
+    </block>
+    <block name="route_run">
+      <set value="true" var="force_forward"/>
+      <call_once fun="NavResurrect()"/>
+      <while cond="true">
+        <go from="p1" hmode="route" wp="p2"/>
```
gautierhattenberger, line 86: you can use the `path` instruction to make this easier to write

conf/modules/logger_file.xml
```diff
@@ -12,7 +12,7 @@
 	<file name="file_logger.h" />
   </header>
   <periodic fun="file_logger_periodic()" start="file_logger_start()"
-		    stop="file_logger_stop()" autorun="FALSE" period="0.01" />
+		    stop="file_logger_stop()" autorun="FALSE" period="0.002" />
```
gautierhattenberger, line 15: You should make the period configurable

conf/telemetry/default_rotorcraft.xml
```diff
@@ -36,6 +36,7 @@
       <message name="LIDAR"                    period="1.2"/>
       <message name="INS_EKF2"                 period=".25"/>
       <message name="WIND_INFO_RET"            period="1."/>
+      <message name="AHRS_REF_QUAT"            period="0.1"/>
```
gautierhattenberger, line 39: is it really necessary to have this in the default mode as well ?

gautierhattenberger (changes requested):

conf/telemetry/default_rotorcraft.xml
EwoudSmeur: It can be very useful for detecting attitude control/estimation problems or when investigating a crash. If you think it is not worth it we can keep it out, but we tend to add it for every AC...

EwoudSmeur:
Thanks for the review @gautierhattenberger! I checked the path instruction, but then it seems to me that the line NavSegment routine is broken: if I change the starting waypoint while the drone flies the segment, instead of going along the new line it approaches a point next to the 'end' waypoint, after which it does fly to the end waypoint. As long as I don't move the waypoints during the flying of a segment it seems to work fine. Path or line have the same behavior. I'll make an issue for this.

conf/modules/logger_file.xml
```diff
@@ -12,8 +12,10 @@
 	<file name="file_logger.h" />
   </header>
   <periodic fun="file_logger_periodic()" start="file_logger_start()"
-		    stop="file_logger_stop()" autorun="FALSE" period="0.002" />
+		    stop="file_logger_stop()" autorun="FALSE" freq="FILE_LOGGER_FREQUENCY" />
   <makefile>
 	<file name="file_logger.c"/>
+  <configure name="FILE_LOGGER_FREQUENCY" default="PERIODIC_FREQUENCY"/>
```
gautierhattenberger, line 18: check indentation and add to `doc` section as well ?

conf/flight_plans/tudelft/delft_bebop_convergence.xml
```diff
@@ -3,9 +3,9 @@
 <flight_plan alt="5" ground_alt="0" lat0="51 59 27.6" lon0="4 22 42.0" max_dist_from_home="200" name="Delft Basic" security_height="2">
   <header>
     #include "autopilot.h"
-    #include "subsystems/datalink/datalink.h"
-    #include "subsystems/electrical.h"
-    #include "subsystems/radio_control.h"
+    #include "modules/datalink/datalink.h"
```
gautierhattenberger, line 6: there is no problem to keep them, but all modules are visible from flight plan, so it is not even necessary to keep the manual includes (same for autopilot.h I think)

conf/flight_plans/tudelft/delft_bebop_convergence.xml
```diff
@@ -41,35 +41,34 @@
     </sector>
   </sectors>
   <exceptions>
-    <!--<exception cond="!InsideFlight_Area(GetPosX(), GetPosY())" deroute="Holding point"/>-->
-    <exception cond="Or(!InsideFlight_Area(GetPosX(), GetPosY()), GetPosAlt() > GetAltRef() + 20) &&
-      !(IndexOfBlock('Takeoff') >= nav_block) &&
-      !(nav_block >= IndexOfBlock('landed'))" deroute="descend att 0"/>
+    <exception cond="Or(!InsideFlight_Area(GetPosX(), GetPosY()), GetPosAlt() @GT GetAltRef() + 20) &&
```
gautierhattenberger, line 44: I don't mind to keep the `&&` but it is still better to use `@AND` (and `@OR` instead of `||`) to avoid issues with XML parsers. The one with Ocaml code generator is very permissive, but that might not be the case for the one with the Qt/C++ GCS if you want to use it in the future.

conf/telemetry/default_rotorcraft.xml
gautierhattenberger: Then I would make it slower. It is even faster than the position and attitude message at the moment. You could still have a fast message in the flight recorder section.

gautierhattenberger:
`path` element is supposed to be converted to a sequence of `go from/to` segment, so it is strange to have a different behavior. But it shouldn't be related to this PR, so of course you can open an issue for that.

EwoudSmeur:
Thanks again @gautierhattenberger for your comments, you were right pprzGCS was crashing because of the &&

gautierhattenberger:
@EwoudSmeur I can see that you have see corrections for this branch in your repository that are not pushed to the tudelft repo on which this PR is based. If you update the correct branch, I think we can merge this.

gautierhattenberger (approved):

PR #2947, opened 2022-11-03
More intuitive axes systems for indoor test areas with Optitrack positioning systems

tblaha:
In earlier versions of PPRZ, the natnet2ivy had conversions hardcoded that align the axis system with the Earth ENU at the TU Delft Cyberzoo. Newer versions remove this to remain general, but created a new axis system that is not very intuitive (x to the viewer) or very understandable from the code.

This PR provides 2 new axes systems, with the goal of minimizing setup effort in Optitrack Motive, while providing intuitive axes that are easy to write flight plans for and easy to understand on the GCS. They are described more in depth in natnet2ivy.py

Cyberzoo axes: y/North away from the viewer, x/East to the right, z/Up upwards
Geographic axes: y/North to the actual north, x/East to the actual East, z/Up upwards (an argument needs to be provided to specify the angle offset between the two frames)

### DIFF
--- a/sw/ground_segment/python/natnet3.x/natnet2ivy.py
+++ b/sw/ground_segment/python/natnet3.x/natnet2ivy.py
@@ -195,12 +195,12 @@ else:
 
 q_x_correction = Quat(
     axis=[0., 0., 1.],
-    angle=np.deg2rad(x_angle)
+    angle=np.deg2rad(-x_angle)
 )
 q_edge_near_to_x_left = Quat(axis=[0., 1., 1.], angle=np.pi)
     
 q_total = q_x_correction * q_edge_near_to_x_left * q_edge_correction
-nose_correction = {'left': 0., 'far': -90., 'right': 180., 'near': 90.}
+nose_correction = {'left': -90., 'far': 180., 'right': 90., 'near': 0.}
 q_nose_correction = Quat(
     axis=[0., 0., 1.],
     angle=np.deg2rad(nose_correction[args.ac_nose] - x_angle)
@@ -265,13 +265,13 @@ def receiveRigidBodyList( rigidBodyList, stamp ):
         vel = compute_velocity(i)
 
         # Rotate position and velocity according to the quaternions found above
-        pos = q_total.rotate([pos[0], pos[1], pos[2]]).tolist()
-        vel = q_total.rotate([vel[0], vel[1], vel[2]]).tolist()
+        pos = q_total.rotate([pos[0], pos[1], pos[2]])
+        vel = q_total.rotate([vel[0], vel[1], vel[2]])
 
         # Rotate the attitude delta to the new frame
         quat = Quat(real=quat[3], imaginary=[quat[0], quat[1], quat[2]])
         quat = q_nose_correction * (q_total * quat * q_total.inverse)
-        quat = quat.elements[[3, 0, 1, 2]].tolist()
+        quat = quat.elements[[1, 2, 3, 0]]
 
         # Check which message to send
         if args.rgl_msg:
--- a/sw/ground_segment/python/natnet3.x/natnet2ivy.py
+++ b/sw/ground_segment/python/natnet3.x/natnet2ivy.py
@@ -59,36 +59,41 @@ left │                          │ right
 
 To forward correct East-North-Up (ENU) and Attitude Quaternion data to the 
 drone, we need to know:
-1. The orientation of the long edge of the Optitrack ground-plane tool during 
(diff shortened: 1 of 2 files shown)


sw/ground_segment/python/natnet3.x/natnet2ivy.py
```diff
@@ -73,7 +158,8 @@
 parser.add_argument('-rg', '--remote_gps', dest='rgl_msg', action='store_true', help="use the old REMOTE_GPS_LOCAL message")
 parser.add_argument('-sm', '--small', dest='small_msg', action='store_true', help="enable the EXTERNAL_POSE_SMALL message instead of the full")
 parser.add_argument('-o', '--old_natnet', dest='old_natnet', action='store_true', help="Change the NatNet version to 2.9")
-parser.add_argument('-zf', '--z_forward', dest='z_forward', action='store_true', help="Z-axis as forward")
+# parser.add_argument('-zf', '--z_forward', dest='z_forward', action='store_true', help="Z-axis as forward")
```
EwoudSmeur, line 161: You can remove this instead of commenting

sw/ground_segment/python/natnet3.x/natnet2ivy.py
```diff
@@ -91,6 +177,11 @@
 # initial track per AC
 track = dict([(ac_id, deque()) for ac_id in id_dict.keys()])
 
+# axes rotation definitions
+q_axes = Quat(axis=[1., 1., 1.], angle=2.*np.pi/3.)
+q_north = Quat(axis=[0., 0., 1.], angle=np.deg2rad(-args.north_angle))
```
EwoudSmeur, line 182: Will this work if north_angle is not provided? It will be None right?

EwoudSmeur (commented):
Very neat! I think we should change the cyberzoo sector waypoints in the flight plan rotorcraft_optitrack.xml as a generic example, which should now be easy in the NED frame +-5m.

gautierhattenberger:
This tool have been designed according to the setup at Enac not TUDelft, where we have at the moment the long bar of the calibration tool pointing to the observer. Which is explaining the unusual configuration for you. I'm fine with making an easier configuration of this tool, however, at the moment it is just breaking our setup and it can't match magically all possible configurations.
My recommendation would be that you have to configure where the X axis of Optitrack is pointing at, either with a direction from observer (front, back, right, left) or with some angle. And if not specified, the program should fail to start. This way, you have no risk to start a flight with a wrong setup (or it will be your own mistake).

EwoudSmeur:
Thanks for your input Gautier, I think your suggestion is a good one.

tblaha:
Thanks Gautier! I'll get on that.

I'd need some clarification though in order to not break your setup. Can you confirm that the optitrack axes for you are like this (with long edge of the calibration tool pointing towards the observers)? I'm asking because there also seems to be a switch in motive that changes the z axis to be up. We are not using it, as far as I've seen.

If that is agreed, then we can have a well defined system by letting the user choose where the x axis is pointing at.
```
+──────────────────────────+
│                          │
│                          │
│                          │
│                          │
│           y ⊙──►x        │
│             │            │
│             ▼            │
│             z            │
│                          │
+──────────────────────────+
│        Observers         │
└──────────────────────────┘
```

gautierhattenberger:
Thanks for the changes, will be tested next week when I'm back from vacations

tblaha:
I have refactored a bit with your comments. The --help message should be very informative (albeit a bit verbose) in how it should be configured. Could you check if my ENAC assumptions are correct? (I added `(?)` where I wasn't sure)

gautierhattenberger:
@tblaha I have pushed a branch in paparazzi/new_cyberzoo_axes with my corrections for Enac in particular, but also more generally to take into account the motive option to stream with Z up rather than Y up (we are using this option). As a result, to sequence is now to first change axis to Z up (if not done by Motive), then turn to correct ground config, then set the X-East.
I have checked most of the configurations, but I'm not 100% sure about the manual angle definition. If you can double check it.

tblaha:
Nice! I will check it, probably at the end of the week.

tblaha:
@gautierhattenberger I have 
1. Changed the reference for the x-angle to right (now that I know we both use that by default)
2. fixed the sign mistake in those calculations
3. tested everything, but couldn't use the z_up option in Motive (we were running an older version of Motive today)

gautierhattenberger (approved):

gautierhattenberger:
We have tested Yup and Zup, it should be fine

PR #2950, opened 2022-11-11
[datalink] add a datalink init function

gautierhattenberger:
and get rid of the DATALINK_C trick

also it should fix the bug of multiple definition with sim target and recent GCC


Fabien-B (dismissed):
Its better but there is still an error remaining for sim: 
/usr/bin/ld: /home/pprz/paparazzi/var/aircrafts/Bixler/sim/main_ap.o:/home/pprz/paparazzi/var/aircrafts/Bixler/sim/generated/modules.h:92: multiple definition of `gps_sim_gps_sim_periodic_check_status'; 
/home/pprz/paparazzi/var/aircrafts/Bixler/sim/arch/sim/sim_ap.o:/home/pprz/paparazzi/var/aircrafts/Bixler/sim/generated/modules.h:92: first defined here

Fabien-B:
It's probably a bad way to solves it, but if I defined `EXTERN_MODULES` as `WEAK` in the `MODULES_C` case in the generated `modules.h`, it works.
```
#ifdef MODULES_C
#define EXTERN_MODULES WEAK
#else
#define EXTERN_MODULES extern
#endif
```

gautierhattenberger:
It was due to a wrong definition of `MODULES_C` in `sim_ap.c`. It should be fine now.

Fabien-B (approved):

PR #2951, opened 2022-11-14
[checks] Add preflight checks

fvantienen:
This prevents the user from arming the motors incase a preflight check fails. This can be bypassed currently using the settings and is only preformed when the `preflight_checks` modules is loaded. I think it is advised to always have this module in your airframe as it can prevent a lot of accidental takeoffs. Takeoffs are prevented also while taking of with RC.

Default checks added:
- Attitude is valid (moved from autopilot code)
- GPS fix is valid (if you have a GPS)
- Battery level is above TAKEOFF_BAT_LEVEL (if not defined above LOW_BAT_LEVEL)
- SDLogger chibios is running (if loaded), by default a warning
- AGL distance within limits when limits are specifically defined
- MS45XX airspeed sensor offset is calibrated

Errors, warnings or success is shown in the message bubble in the new GCS. Maybe we should make this more clear to the user later.


sw/airborne/modules/checks/preflight_checks.c
```diff
@@ -0,0 +1,62 @@
...
+  bool success = true;
+
+  // Go through all the checks
+  struct preflight_check_t *check = preflight_head;
+  while(check != NULL) {
+    checks_cnt++;
+
+    // Peform the check and register fails
+    if(!check->func(error_msg)) {
+      printf("Preflight fail: %s", error_msg);
```
dewagter, line 52: Can the first error not propagate to telemetry so it also works on microcontrollers?

fvantienen: That is the idea to even send all of them next to logging them on the sd card.

sw/airborne/firmwares/rotorcraft/autopilot_arming_throttle.h
```diff
@@ -117,14 +117,14 @@ static inline void autopilot_arming_check_motors_on(void)
         }
         break;
       case STATE_MOTORS_OFF_READY:
-        autopilot.motors_on = false;
+        autopilot_set_motors_on(false, false);
```
Fabien-B, line 120: I think you made a typo here ? (2 parameters ?)

fvantienen: Aha yes, I had that first when I wanted the second one to forcefully bypass the motors on.
This PR is just a simple showcase (not 100% tested and finished) on how I might wanted to implement this feature.

gautierhattenberger:
I agree this is the good direction.
One idea: at the moment, the check and error sending is done when starting the motor, we might also want to have a regular "silent" check when on ground, with a possible alarm on the GCS (state filter status message ?) if at least one is failing.

dewagter:
@fvantienen rebased.

fvantienen:
This module is now updated and can be improved later, but I would like to add this as soon for testing purposes.
We can maybe make this module auto loading later when fully tested?

dewagter (dismissed):
Great improvement!

dewagter:
Might even consider having it autoload.

fvantienen:
I already noticed some mistake in the arming. You can't arm it with the RC. I have a fix which I will push soon.
Also the messages are spammed if you arm it through the RC, so I want to put a limiter in the messages

fvantienen:
Next to the message limiter there also needs to be a fix to bypass the quick re-arm when kill within 10 seconds. Now it might fail on pre-flight checks which I think is not the best behaviour.

fvantienen:
- Automatic re-arm incase of accidental kill is fixed
- Message limiter for the datalink messages is added
@gautierhattenberger @dewagter

dewagter:
@fvantienen will test inflight this afternoon.

sw/airborne/modules/checks/preflight_checks.c
```diff
@@ -0,0 +1,171 @@
...
+  check->next = next;
+}
+
+/**
+ * @brief Perform all the preflight checks
+ * 
+ * @return true When all preflight checks are successful
+ * @return false When one or more preflight checks fail
+ */
+bool preflight_check(void) {
```
gautierhattenberger, line 69: is this function called periodically somewhere on or only when set_motors_on is called ? I'm not sure

sw/airborne/modules/loggers/sdlog_chibios.c
```diff
@@ -106,6 +106,20 @@ static enum {
 static char chibios_sdlog_filenames[68];
 static char NO_FILE_NAME[] = "none";
 
+#if PREFLIGHT_CHECKS
+/* Preflight checks */
+#include "modules/checks/preflight_checks.h"
+static struct preflight_check_t sdlog_pfc;
+
+static void sdlog_preflight(struct preflight_result_t *result) {
+  if(chibios_sdlog_status != SDLOG_RUNNING) {
```
gautierhattenberger, line 115: I suggest to set to error only if status is `SDLOG_ERROR`. With the startup delay, it can block you for tuning on the motors. If you don't have an SD card, I'm not even sure what is happening, but your checks will always fail then.
An other option is to have a "warning" check. Something that reports a message but doesn't block the fail check ?

sw/airborne/firmwares/rotorcraft/autopilot_static.c
```diff
@@ -317,7 +311,7 @@ void autopilot_static_set_mode(uint8_t new_autopilot_mode)
 
 void autopilot_static_set_motors_on(bool motors_on)
 {
-  if (autopilot.mode != AP_MODE_KILL && ap_ahrs_is_aligned() && motors_on) {
+  if (autopilot.mode != AP_MODE_KILL && motors_on) {
```
gautierhattenberger, line 314: I'm still not sure about the implication here if "check" module is not used and someone is trying to start motors before ahrs is aligned.

sw/airborne/modules/loggers/sdlog_chibios.c
fvantienen: We currently prefer to not takeoff when we are not logging, but I can make this check optional. Also could add a warning option, such that you still get a message in the GCS.
Since most of our platforms are research platforms you don't want to take off if the SDLogger is not running for some reason, even with the startup delay you just have to wait. Since most tests/flights are useless if something goes wrong if you don't have an onboard log. If you still want to take off without logging you can use the setting to bypass the checks, but this would require a user decision and active action.

But to make this work for most people I'm also happy to add some of these checks behind defines/warnings by default, such that you can choose yourself what you prefer.

sw/airborne/firmwares/rotorcraft/autopilot_static.c
fvantienen: Yes that is indeed removed now by default. My idea was to automatically load the preflight_check module, such that this is re-added to all platforms by default. Loading the preflight_check module by default might also prevent a lot of failures/crashes of inexperienced users when we add more checks.
Is it possible to unload an automatically added module?

sw/airborne/modules/loggers/sdlog_chibios.c
fvantienen: If you don't have an SD card the logger status goes to `ERROR`, but also then we prefer not to take off. Since most of the time this is a user error forgetting to place the SD card back in the UAV. If you don't want to fly with the logger you should just remove the module from your airframe, or bypass the checks(if only temporarily).

dewagter (approved):
Checked several days.
 - does not _ARM_ when preflight fails as intended!
 - re-arm in-air after kill works (tested!)

Old GCS gives errors on INFO messages, but probably OK.

fvantienen:
There is one thing that needs to be solved: If you arm and the checks fail the propellors won't spin, but it keeps the arming state armed for the arming state machine. Then if you bypass the checks or the checks pass it will directly arm the motors.

gautierhattenberger:
Is it sufficient to call `autopilot_arming_check_motors_on()` only when all checks are validated ?

sw/airborne/modules/checks/preflight_checks.h
```diff
@@ -0,0 +1,54 @@
...
+struct preflight_check_t {
+  preflight_check_f func;
+  struct preflight_check_t *next;
+};
+
+extern bool preflight_bypass;
+extern void preflight_check_register(struct preflight_check_t *check, preflight_check_f func);
+extern bool preflight_check(void);
+extern void preflight_error(struct preflight_result_t *result, const char *fmt, ...);
+extern void preflight_success(struct preflight_result_t *result, const char *fmt, ...);
```
gautierhattenberger, line 52: I think `preflight_warning` is missing in header file

gautierhattenberger (commented):
Have you found a solution for the bypass/arming issue ?

sw/airborne/modules/checks/preflight_checks.h
fvantienen: Yes I'm still busy implementing that. Currently also fixing the bypass/arming issue

sw/airborne/modules/loggers/sdlog_chibios.c
fvantienen: This now changed to a warning by default

fvantienen:
I still have to test this version with the arming, but I think I incorporated most features. The only thing missing is the `ap_ahrs_is_aligned` is now not enabled by default, but is it ok to just have the checks modules included by default?

fvantienen:
Another thing I saw was the `autopilot.kill_throttle`, but not sure what to do with that. Because for me that seems like exactly the same as `~autopilot.motors_on`.

fvantienen:
> I still have to test this version with the arming, but I think I incorporated most features. The only thing missing is the `ap_ahrs_is_aligned` is now not enabled by default, but is it ok to just have the checks modules included by default?

@gautierhattenberger How would you suggest to automatically load the preflight module? Because I can add a depends in `system_core.xml`, but not sure if that is the best solution and what we want. Or do you want to manually load it in every airframe?

gautierhattenberger:
`kill_throttle` and `motors_on` are simple opposite for the rotorcraft firmware, but are different for fixedwing (`kill_throttle` prevents the AP to use throttle in NAV mode, but pilot can still use it in direct control, so with `motors_on`).
If the dependency is with `system_core`, it will be for all firmwares. Have you tested fixedwing for side effects ? Maybe it can be tested for some time on a few aircraft before adding it as a default for everyone ?

dewagter:
@gautierhattenberger : we do not have fixedwings with fixedwing code anymore. Is there any chance you guys still have one to test?

gautierhattenberger:
We'll try that this week.

dewagter:
Thank you very much! We just successfully re-tried the latest code in rotorcraft.

dewagter (approved):

dewagter (approved):
rebased, fixed merge conflict and approved

gautierhattenberger:
I have added a handler on the `kill_throttle` variable so preflight checks are at least called if calling resurrect from the GCS.
I have noted a couple of logic issues due to the fact that `kill_throttle` and `motors_on` have different meanings and different logic between the firmwares. It should not have real consequences since in fact `motors_on` is not used with fixedwing. But at some point it should be cleaned.
I have also run the code style checker.
Since it is not mandatory at the moment, I'm fine with merging this and see if any issues arise.
It is also nice that the "chat" widget in the GCS is used now :)

gautierhattenberger (approved):

dewagter:
Great! Thanks a lot for checking!

gautierhattenberger (approved):

PR #2953, opened 2022-11-16
Added driver for QMC5883L Magnetometer

OpenUAS:
Added a driver for the QMC5883L Magnetometer, tested on Baremetal Lisa MXS and Chibios Lisa MXS

working_qmc5883l_pprz

### DIFF
--- a/conf/modules/mag_qmc5883l.xml
+++ b/conf/modules/mag_qmc5883l.xml
@@ -54,7 +54,7 @@
         $(error mag_qmc5883l module error: please configure MAG_QMC5883L_I2C_DEV)
       endif
     </raw>
-    <configure name="MAG_QMC5883L_PERIODIC_FREQUENCY" default="60"/>
+    <configure name="MAG_QMC5883L_PERIODIC_FREQUENCY" default="50"/>
     <configure name="MAG_QMC5883L_I2C_DEV" case="upper|lower"/>
     <define name="MAG_QMC5883L_PERIODIC_FREQUENCY" value="$(MAG_QMC5883L_PERIODIC_FREQUENCY)"/>
     <define name="USE_$(MAG_QMC5883L_I2C_DEV_UPPER)"/>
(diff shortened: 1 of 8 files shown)


sw/airborne/modules/sensors/mag_qmc5883l.c
```diff
@@ -0,0 +1,136 @@
...
+ * <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file modules/sensors/mag_qmc5883l.c
+ *
+ * Module for QST QMC5883L magnetometer, the DB version
+ */
+
+#include "modules/sensors/mag_rm3100.h"
```
gautierhattenberger, line 27: wrong include

conf/modules/mag_qmc5883l.xml
```diff
@@ -0,0 +1,70 @@
...
+  <event fun="mag_qmc5883l_module_event()"/>
+  <makefile target="ap">
+    <file name="mag_qmc5883l.c"/>
+    <file name="qmc5883l.c" dir="peripherals"/>
+    <raw>
+      ifeq ($(MAG_QMC5883L_I2C_DEV),)
+        $(error mag_qmc5883l module error: please configure MAG_QMC5883L_I2C_DEV)
+      endif
+    </raw>
+    <configure name="MAG_QMC5883L_PERIODIC_FREQUENCY" default="60"/>
```
gautierhattenberger, line 57: I would not recommend a frequency of 60Hz by default. I don't know the specs of this sensor, but if possible it is better to have periodic functions as divisors of the main frequency which is by default 500 Hz (or sometimes 512) for rotorcraft, which is where this sensor will be mostly used I guess. So 50 Hz is maybe a better choice.

gautierhattenberger (changes requested):

OpenUAS:
Changed, a new review appreciated, with PR merge honoring would be best :)

PR #2954, opened 2022-11-23
INDI hybrid updates

EwoudSmeur:
We made some code for our Nederdrone project. I'm breaking it down into manageable commits, this one contains:

- Two interpolation segments during transition
- Quadratic lift effectiveness with airspeed
- Velocity setpoint input possible from module via ABI message

Unfortunately I couldn't keep the original commits as there were 150, it was way too much hassle to do pick out the changes without squashing...


gautierhattenberger:
compilation is failing when `GUIDANCE_INDI_LIFTD_ASQ` is not defined I think. And you have a conflict with the previous pull request.
It seems like what you are doing is to control the speed while flying in FP navigation ? Or something close. The control inputs for indi (and especially for hybrid) are really a mess now I think... I okay to merge something like this as a temporary solution, but we should make something clean and consistent at someday.

EwoudSmeur:
You're right, forgot to add the airframe file changes. I skipped the Nederdrone ones as they have many more changes, would be easier to do that in one go I believe. There is now also a default.

Regarding the inputs/setpoint/ref, you're right unfortunately... Somehow different sources of a position/velocity/attitude reference (NAV, modules, predefined 4D trajectories, etc.) need to be chosen/prioritized/combined.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -64,6 +64,18 @@
 #ifndef GUIDANCE_INDI_MIN_PITCH
 #define GUIDANCE_INDI_MIN_PITCH -120
 #define GUIDANCE_INDI_MAX_PITCH 25
+
+#ifndef
```
gautierhattenberger, line 68:
```suggestion
#ifndef GUIDANCE_INDI_LIFTD_ASQ
```

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -64,6 +64,18 @@
 #ifndef GUIDANCE_INDI_MIN_PITCH
 #define GUIDANCE_INDI_MIN_PITCH -120
 #define GUIDANCE_INDI_MAX_PITCH 25
```
gautierhattenberger, line 66: there is a `#endif` missing here

EwoudSmeur:
Sorry not sure how that happened! Fixed, compiles and works in simulation 👌

gautierhattenberger (approved):

PR #2955, opened 2022-11-25
Fix for adaptive INDI

EwoudSmeur:
G1+G2 was not updated with WLS before, but the new G2 was still used leading to unstable behavior.

And:
- better scaling value for pseudo inverse
- Unify file_logger module name. OCD maybe but ok :)


gautierhattenberger:
Regarding the logger module, I think it is much better to change the source code file names rather than the xml file.
- all other logger modules are named `logger_xxx`
- changing modules' file name is breaking a lot of airframe

And if possible, I would really appreciate two separated pull requests for this :)

EwoudSmeur:
Ok, all airframe files that contained the module so nothing should be broken. I get your point though and will change

gautierhattenberger (approved):

PR #2964, opened 2023-01-11
Nav rotorcraft rework

gautierhattenberger:
This is a complete rework of the navigation for rotorcraft with the key features:
- all nav functions and interface in float
- nav function are registered (decoupling between nav API and implementation of standard pattern)
- submodes to specify if setpoint is a position (legacy), speed or accel (better integration of algorithms like GVF)
- guidance (H and V) is reorganized with default function to implement (run_pos, run_speed, run_accel)
- guidance control (the old default PID) is separated from the guidance logic

Still some questions ahead:
- hybrid and INDI are still tightly coupled with guidance (with compile flag) while it should be done with the new interface
- the stabilization is also tightly coupled with guidance, it should be separated
- the new nav API is inspired from rover nav, and some code could be merged in the future

@EwoudSmeur @fvantienen What is your opinion about this approach ? Do you see a proper way to integrate the indi/hybrid in a with this ?
I would like to split the stabilization from guidance, but maybe it is a bit to much for a single pull request ?


gautierhattenberger:
Screenshot_GCS_hybrid
It is starting to work as expected for hybrid nav.
@EwoudSmeur @fvantienen @mrtbrnz Any thoughts about this possible solution for nav/guidance implementation ?

The coupling horizontal/vertical/stabilization with INDI is still a bit confusing, not sure how to improve that.

OpenUAS:
Thx, nice development. @gautierhattenberger @Fabien-B  Small remark, It would be great if the  **testdrone color** is not the same green as the expected track, can you make it Pink, Yellow or Reddish, or Cyan please.

EwoudSmeur (commented):
Nice work Gautier! I think it was a great idea to bring a bit more structure to the guidance. I had a few questions looking through the code, but from what I could see it all made sense. It's hard to keep track of all the changes, as some things were just moved. I'll try to do some simulation tests soon!

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -480,8 +349,8 @@ void guidance_indi_run(float *heading_sp) {
 // For a hybrid it is important to reduce the sideslip, which is done by changing the heading.
 // For experiments, it is possible to fix the heading to a different value.
 #ifndef KNIFE_EDGE_TEST
-  if(take_heading_control) {
-    *heading_sp = ANGLE_FLOAT_OF_BFP(nav_heading);
+  if (take_heading_control) {
+    *heading_sp = ANGLE_FLOAT_OF_BFP(nav.heading); // FIXME remove dependency to nav, take from guidance sp ?
```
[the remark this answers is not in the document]
gautierhattenberger: I just removed this dependency, it shouldn't be necessary

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -442,15 +306,20 @@ void guidance_indi_run(float *heading_sp) {
...
-  //feedforward estimate angular rotation omega = g*tan(phi)/v
+  // Coordinated turn
+  // feedforward estimate angular rotation omega = g*tan(phi)/v
   float omega;
-  const float max_phi = RadOfDeg(60.0);
-  float airspeed_turn = airspeed;
-  //We are dividing by the airspeed, so a lower bound is important
-  Bound(airspeed_turn,10.0,30.0);
+  const float max_phi = RadOfDeg(60.0f);
+#if GUIDANCE_INDI_ZERO_AIRSPEED
```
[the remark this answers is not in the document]
gautierhattenberger: This part is up to you. I think it could be modified later as it is a "functional" change, not related to the reorganization of the code.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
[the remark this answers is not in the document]
EwoudSmeur: ok!

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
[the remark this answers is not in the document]
EwoudSmeur: Ah this was already like this, then ok haha 😅

EwoudSmeur (dismissed):

gautierhattenberger:
@EwoudSmeur after testing again in simulation, I realized that the `take_heading_control` was not working as expected. So I found a fix that is also removing the `HEADING_IS_FREE` flag. Now the heading setpoint is always set by nav (in NAV mode), but the (hybrid) guidance may use it or not for the final stab setpoint (for which an internal variable is used for the "free" heading).

conf/modules/guidance_hybrid.xml
```diff
@@ -12,16 +12,16 @@
         <dl_setting var="max_airspeed" MIN="4" STEP="1" MAX="160" module="guidance/guidance_hybrid"/>
         <dl_setting var="wind_gain" MIN="1" STEP="1" MAX="256"/>
         <dl_setting var="horizontal_speed_gain" MIN="1" STEP="1" MAX="15" shortname="hor_speed_div"/>
-        <dl_setting var="max_turn_bank" MIN="10.0" STEP="1.0" MAX="60.0" shortname="max_turn_bank"/>
-        <dl_setting var="turn_bank_gain" MIN="0.1" STEP="0.01" MAX="3.0" shortname="turn_bank_gain"/>
+        <dl_setting var="max_turn_bank" MIN="10.0" STEP="1.0" MAX="60.0"/>
+        <dl_setting var="turn_bank_gain" MIN="0.1" STEP="0.01" MAX="3.0"/>
 
-        <dl_setting var="cruise_throttle" MIN="0" STEP="1" MAX="9600" shortname="cruis_throttle"/>
```
OpenUAS: Cruise Throttle

gautierhattenberger: I have removed the `shortname` instead of correcting the spelling since it was the same than `name` (so it makes no difference in the end)

fvantienen:
After maybe @EwoudSmeur has checked it we can merge it.
This PR should close #2938

conf/modules/guidance_indi_hybrid.xml
```diff
@@ -9,7 +9,7 @@
   <settings>
     <dl_settings>
       <dl_settings NAME="guidance_indi_hybrid">
-        <dl_setting var="gih_params.liftd_p50" min="0.1" step="0.1" max="10.0" shortname="liftd_p50" param="GUIDANCE_INDI_LIFTD_P50" persistent="true"/>
+        <dl_setting var="gih_params.liftd_p50" min="0.1" step="0.1" max="10.0" shortname="liftd_p50" param="GUIDANCE_INDI_LIFTD_P50" persistent="true" module="guidance/guidance_indi_hybrid"/>
```
fvantienen, line 12: Why is there a module added here?

conf/modules/guidance_pid_rotorcraft.xml
```diff
@@ -0,0 +1,51 @@
...
+      <define name="HOVER_KD" value="82" description="feedback vertical control D gain"/>
+      <define name="HOVER_KI" value="13" description="feedback vertical control I gain"/>
+      <define name="NOMINAL_HOVER_THROTTLE" value="0.655" description="nominal throttle at hover (between 0 and 1)"/>
+      <define name="ADAPT_THROTTLE_ENABLED" value="FALSE|TRUE" description="enable adaptive control by default"/>
+    </section>
+  </doc>
+  <settings target="ap|nps">
+    <dl_settings>
+      <dl_settings NAME="Vert PID">
+        <dl_setting var="guidance_pid.v_kp" min="0" step="1" max="600" shortname="kp" param="GUIDANCE_V_HOVER_KP" persistent="true" module="guidance/guidance_pid"/>
```
fvantienen, line 25: Similar question here when to add the module argument or not?

sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
```diff
@@ -128,9 +129,9 @@ static void send_fp(struct transport_tx *trans, struct link_device *dev)
   struct Int32Eulers att = *stateGetNedToBodyEulers_i();
 #endif
   pprz_msg_send_ROTORCRAFT_FP(trans, dev, AC_ID,
-                              &(stateGetPositionEnu_i()->x),
-                              &(stateGetPositionEnu_i()->y),
-                              &(stateGetPositionEnu_i()->z),
+                              &pos->x,
```
fvantienen, line 132: If our navigation is in floats does it still make sense to send it down in integers? (Maybe something for later to update)

fvantienen (approved):

conf/modules/guidance_indi_hybrid.xml
gautierhattenberger: The `module` attribute is a legacy prior to the "modules" we have now. It can be replaced by `header` and it was used to specified the header file for settings.
We will clean that someday...

sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
gautierhattenberger: yes, indeed, but it is more related to redefining most messages, so it is an other job

PR #2966, opened 2023-01-24
Usb add serial

Fabien-B:
Add a second USB serial.
The use case is during development, to use one of the serial as a telemetry port, and the other as a debug port (e.g. a shell, or just some prints).
The interfaces are named `Paparazzi UAV telemetry Port` and `Paparazzi UAV debug Port` (although you can use them as you want), and the udev rules file symlink theses to `/dev/pprz_tlm` and `/dev/pprz_dbg`.

### DIFF
--- a/conf/modules/telemetry_transparent.xml
+++ b/conf/modules/telemetry_transparent.xml
@@ -3,11 +3,13 @@
 <module name="telemetry_transparent" dir="datalink" task="datalink">
   <doc>
     <description>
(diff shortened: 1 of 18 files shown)


Fabien-B:
That should be ready !

gautierhattenberger:
It is all good for me, but can you update the description of the transparent_usb module to explain how to activate the second usb ?

gautierhattenberger (approved):

PR #2983, opened 2023-02-07
Supervision small fixes

Fabien-B:
Update targets only: this reduce the lag at each AC change.

Please tell me if you find some bugs in the paparazzi center, I will include the fixes with this PR.


dewagter (approved):
Perfect

PR #2986, opened 2023-02-10
Ubuntu22 fixes

dewagter:
fix linking errors and silent compile warnings.


dewagter:
Also apply to release6.2 (release should at least compile with latest ubuntu)

gautierhattenberger (approved):

gautierhattenberger:
We will likely make a service release for 6.2, but within a month, so we can find some more bugs...

dewagter:
This would also need to go to v6.2.

PR #2991, opened 2023-02-12
[install] Ease installation with a GUI

dewagter:
that automatically adapts to the OS


dewagter:
image
Tested in Ubuntu 18.04, 20.04 and 22.04

dewagter:
Ready to merge. This would be great to also have in v6.2.

install.py
```diff
@@ -0,0 +1,141 @@
...
+    print(cmd)
+    print('READY (verify above)')
+    print('Check documentation to resolve problems:', docs)
+    print('=================================================')
+
+def view(link):
+    webbrowser.open(link)
+
+def cmd_repo():
+    execute('sudo add-apt-repository -y ppa:paparazzi-uav/ppa && sudo apt-get update')
```
gautierhattenberger, line 38: How do you enter the sudo password ? In the terminal running the program ?

install.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
+import tkinter as tk
```
gautierhattenberger, line 2: We have dependencies with wx, gtk2, gtk3, qt and now tk ? It's not a big deal and it's not really mandatory, but still...

install.py
Fabien-B: This could be done with `pkexec`

dewagter: When running the ./install.sh script, it asks for a password and then all install clicks are executed without further prompts. If you have a user-prompt (e.g. yes/no, pw, etc, ... ) you can type it in the console that started the script.

dewagter: I didn't use `pkexec` because then I got into trouble with the $DISPLAY environment.

dewagter: I just checked again: it says: 774kb of memory will be used for installing python-tk

sw/simulator/nps/nps_sensors_utils.c
```diff
@@ -8,7 +8,7 @@ void UpdateSensorLatency(double time, gpointer cur_reading, GSList **history, do
   /* add new reading */
   struct BoozDatedSensor *cur_read = g_new(struct BoozDatedSensor, 1);
   cur_read->time = time;
-  cur_read->value = (struct DoubleVect3 *) g_memdup(cur_reading, sizeof(struct DoubleVect3));
+  cur_read->value = (struct DoubleVect3 *) g_memdup2(cur_reading, sizeof(struct DoubleVect3));
```
gautierhattenberger, line 11: this is not working for Semaphore build

install.py
gautierhattenberger: I agree it is light, and we may get rid of gtk someday...

sw/simulator/nps/nps_sensors_utils.c
Fabien-B: I tried that, this function does not exist in Ubuntu 20.04.

dewagter:
Ok, in the interest of removing the number of libraries, qt5 seems to work in ubuntu 18, 20, 22 and windows... so here is the qt5 version.

dewagter:
- [x] replaced Tk with Qt5 as this is the new standard
 - [x] moved the python file to the tools folder

gautierhattenberger:
Thank you for the effort to replace tk. I think it is all good. Just see my question about ivy python.

sw/tools/install.py
```diff
@@ -0,0 +1,167 @@
...
+    execute('git submodule init && git submodule sync && git submodule update ./sw/ext/opencv_bebop')
+    execute('sudo apt-get -f -y install cmake libjpeg-turbo8-dev libpng-dev libtiff-dev zlib1g-dev libdc1394-22-dev')
+
+def cmd_vlc():
+    execute('sudo apt-get -f -y install ffmpeg vlc jstest-gtk default-jre')
+    execute('sudo apt-get install python3-pip')
+    if float(release['RELEASE']) <= 20.04:
+        execute('pip3 install pyquaternion ivy-python') # Required for NatNat
+    else:
+        execute('pip3 install pyquaternion ivy') # Required for NatNat
```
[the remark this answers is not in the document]
dewagter: Yes. This was reported and I tested it in a new VM myself. Without these extra packages `natnet` from master/release6.2 does not start. in 18.04 pip install ivy does not work as it needs python3.8 so I install `ivy-python` instead and finally `natnet` works. For future `paparazzi-dev` packages, it would be nice to add these missing python3 packages or instead make a `requirements.txt` file such that people with more versions and `conda` have fewer problems.

dewagter:
rebased on master

sw/tools/install.py
```diff
@@ -0,0 +1,178 @@
...
+        self.execute('git submodule init && git submodule sync && git submodule update ./sw/ext/opencv_bebop')
+        self.execute('sudo apt-get -f -y install cmake libjpeg-turbo8-dev libpng-dev libtiff-dev zlib1g-dev libdc1394-22-dev')
+
+    def cmd_vlc(self):
+        self.execute('sudo apt-get -f -y install ffmpeg vlc jstest-gtk default-jre')
+        self.execute('sudo apt-get install python3-pip')
+        if float(release['RELEASE']) <= 20.04:
+            self.execute('pip3 install pyquaternion ivy-python') # Required for NatNat
+        else:
+            self.execute('pip3 install pyquaternion ivy') # Required for NatNat
```
Fabien-B, line 80:
```suggestion
        self.execute('pip3 install pyquaternion ivy-python') # Required for NatNat
```
`ivy` is something completly different: https://pypi.org/project/ivy/

dewagter: Ok, indeed, ivy was not needed in 22.04 and `ivy-python` is part of `paparazzi-dev`. I did need `pyquaternion` and `ivy-python` in 18.04 for sure. On the others, I should re-check (reinstall fresh ubuntu). If the package was already installed nothing happens so for now we can leave it.

dewagter:
I finally added some event handling during long installs as Qt5 gave popups since the install is synchronous in a button callback (was not a problem with tk).

dewagter:
execute can not launch a QThread now since there are several consecutive executes. Then the button function would need to call the thread, but that gives quite some extra code. I actually tried the Timer, but it worked less well and this was the shortest code that performed well (except Tk which was much shorter).

The goal was to ease installation, but I did not like the former 1-line install for 4 reasons:
 - I can't remember it by heart (now it's `git clone paparazzi` and `./install.sh` like in other projects)
 - Not everyone needs everything, so it is split
 - When it fails, in the one-liner, it is harder to find out which part (especially for our students)
 - It does not adapt to the OS (someone wants to add code for Debian or mac/windows?)

But for maintenance reasons, I'd suggest keeping this install code as simple and small as possible. This should IMHO be an install script turned into buttons.

dewagter:
PS: the branch is on the paparazzi github, so feel free to push any suggestion to the branch.

Fabien-B:
What do you think about the last 3 commits I suggest?
- Adding `-E` option to sudo to export environment variables. Useful if you are behind a proxy.
- Add `-y` option to some `apt-get install` (it was alreday on some of them, but not all.
- Use the "new" way to invoke pip (the other is advised against, as the result is unpredictable if you have multiple versions of python installed)

dewagter:
@Fabien-B additions look great, thanks a lot!

Fabien-B (approved):

PR #2995, opened 2023-02-15
Log message, and forward it to link.

Fabien-B:
Send INFO_MSG_GROUND to log an INFO_MSG in the drone, in either the flight recorder or the pprzlog (or both).
INFO_MSG_UP is logged by the server.

pprzlink needs to be updated.


fvantienen (approved):

fvantienen:
Very nice, thank you!

Fabien-B:
And if you build from sources the new GCS, there is a new chat widget! (don't forget to update the submodules)

PR #2997, opened 2023-02-15
[plotter] implementing a realtime plotter in python

gautierhattenberger:
based on pyqtgraph
it should implement most of the functionnality of the legacy ocaml plotter
it is replacing the previous implementation in python


gautierhattenberger:
It's all good, thanks for the update Fabien

PR #3003, opened 2023-02-24
fix state update equation

EwoudSmeur:
This is seems like wrong integration. I also don't see how it would be trapezoidal, so I removed the comment.


gautierhattenberger (approved):
agreed...

PR #3005, opened 2023-03-04
Modules recommends and suggests

gautierhattenberger:
Add two new items in module's dependency:
- `recommends`: a recommended module tells the sorting algo that if the module is found, it should be sorted accordingly. It is useful for optional dependencies, like `shell` or `mission` in some modules
- `suggests`: if a functionality is not provided by the user, a module can suggest a list of modules that can provide them. It is a convenient way to have "default" modules.

As a result the former `autoload` node is removed and replaced by suggested modules.


Fabien-B (approved):
Tested, working.
It could be nice to print the recommended modules not used after the loaded modules. Something like:
`'shell' recommended by 'sys_mon'`

PR #3008, opened 2023-03-07
[doc] Parse module description as rst.

Fabien-B:
The `description` node of modules xml files are parsed as reStructuredText (sphinx syntax), allowing to build a rich and beautiful doc right from this file!
All sphinx features should works: code blocks, images, hyperlinks, ...

### DIFF
--- a/sw/tools/modules_doc/modules_parser.py
+++ b/sw/tools/modules_doc/modules_parser.py
@@ -6,6 +6,8 @@ from sphinx.parsers import RSTParser
 from docutils import nodes
 from lxml import etree as ET
 from dataclasses import dataclass
+from docutils.frontend import OptionParser
+from docutils.utils import new_document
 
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
Can we have a problem with the doxygen doc of modules due to special characters or other strange formatting ?

Fabien-B:
Doxygen formatted descriptions still builds, but are not very nice-looking in the sphinx doc.
I wasn't even aware that there was such feature!

OpenUAS:
@Fabien-B This is nice low hanging fruit IMHO. Would it be possible to test, as so to merge with master. If impossible else feel free to assign the issue to me so. If this is working the motivation to get good documentation in the description of the modules. Or maybe even sanitize the Docs beforehand and remove non pure text and other weirdness,

PR #3013, opened 2023-03-21
[nav] implement nav_circle for hybrids

gautierhattenberger:
The flight speed is adapted to the circle radius according to the maximum bank angle allowed.
The oval implemented in base rotorcraft nav is also compatible with the hybrid primitives registered in nav.

This solution is not perfect, in particular the side acceleration in coordinated turn could be added as a pre-command to the INDI controller. But at least it seems good enough to fly circles of very different radius. Only tested in simulation so far.

### DIFF
--- a/sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
+++ b/sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
@@ -197,9 +197,9 @@ static void nav_hybrid_circle(struct EnuCoor_f *wp_center, float radius)
     NormRadAngle(trigo_diff);
     nav_rotorcraft_base.circle.radians += trigo_diff;
     // progress angle
-    float progess_angle = GUIDANCE_INDI_NAV_CIRCLE_DIST / abs_radius;
-    Bound(progess_angle, M_PI / 16.f, M_PI / 4.f);
-    float alpha = nav_rotorcraft_base.circle.qdr - sign_radius * progess_angle;
+    float progress_angle = GUIDANCE_INDI_NAV_CIRCLE_DIST / abs_radius;
+    Bound(progress_angle, M_PI / 16.f, M_PI / 4.f);
+    float alpha = nav_rotorcraft_base.circle.qdr - sign_radius * progress_angle;
     // final target position, should be on the circle, for display
     nav.target.x = wp_center->x + cosf(alpha) * abs_radius;
     nav.target.y = wp_center->y + sinf(alpha) * abs_radius;


sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -171,21 +176,65 @@ static bool nav_hybrid_approaching(struct EnuCoor_f *wp, struct EnuCoor_f *from,
...
+    desired_speed = nav_max_speed;
+  } else {
+    float radius_diff = fabsf(float_vect2_norm(&pos_diff) - abs_radius);
+    if (radius_diff > GUIDANCE_INDI_NAV_CIRCLE_DIST) {
+      // far from circle, speed proportional to diff
+      desired_speed = radius_diff * gih_params.pos_gain;
+    } else {
+      // close to circle, speed function of radius for a feasible turn
+      // MAX_BANK / 2 gives some margins for the turns
+      desired_speed = sqrtf(PPRZ_ISA_GRAVITY * abs_radius * tanf(GUIDANCE_H_MAX_BANK / 2.f));
```
EwoudSmeur, line 222: Nice idea, but maybe some safety margin should be taken here?

EwoudSmeur (changes requested):
Nice improvement Gautier! I think it's a great idea to include a feed forward centripetal acceleration based on the ground speed, but we can do that in a separate PR.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
gautierhattenberger: the max_bank divided by 2 is suppose to be the margin here

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -171,21 +176,65 @@ static bool nav_hybrid_approaching(struct EnuCoor_f *wp, struct EnuCoor_f *from,
...
+    // store last qdr
+    float last_qdr = nav_rotorcraft_base.circle.qdr;
+    // compute qdr
+    nav_rotorcraft_base.circle.qdr = atan2f(pos_diff.y, pos_diff.x);
+    // increment circle radians
+    float trigo_diff = nav_rotorcraft_base.circle.qdr - last_qdr;
+    NormRadAngle(trigo_diff);
+    nav_rotorcraft_base.circle.radians += trigo_diff;
+    // progress angle
+    float progess_angle = GUIDANCE_INDI_NAV_CIRCLE_DIST / abs_radius;
```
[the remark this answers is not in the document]
gautierhattenberger: done

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
EwoudSmeur: Ah now I see, ok!

PR #3015, opened 2023-03-24
[chibios] Small fixes

fvantienen:
Small fixes for chibios.
- No GDB symbols by default, which saves a lot of space
- By default wait for autopilot on PX4 uploader
- Fix SD logger mistakes, there is still a bug when the autopilot is powered with USB. It tries to close all the logs and hangs on https://github.com/paparazzi/paparazzi/blob/a02ec170cf9f489fbc31d527b0504a862f988d68/sw/airborne/modules/loggers/sdlog_chibios/sdLog.c#L334. I couldn't figure out why, maybe someone else can @gautierhattenberger ?
- Disable the USB debug endpoint when not in use

### DIFF
--- a/conf/Makefile.chibios
+++ b/conf/Makefile.chibios
@@ -107,7 +107,7 @@ ifeq ($(USE_OPT),)
 ifeq (,$(findstring $(RTOS_DEBUG),0 FALSE))
   CH_OPT ?= 0 -g -ggdb3 -fno-inline
 else
-  CH_OPT ?= 2 -ggdb
+  CH_OPT ?= 2
 endif
   USE_OPT = -O$(CH_OPT) \
 	-falign-functions=16 -fomit-frame-pointer \
--- a/sw/airborne/mcu_periph/usb_serial.h
+++ b/sw/airborne/mcu_periph/usb_serial.h
@@ -39,10 +39,10 @@
 #endif 
 
 #ifdef USB_MAX_ENDPOINTS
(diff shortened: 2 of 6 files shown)


Fabien-B:
If the serial USB is used the autopilot waits for the *USB Storage* setting to stop and appear as USB mass storage.
If USB serial is **not** used, it stops and appear as USB mass storage as soon as the USB is plugged

https://github.com/paparazzi/paparazzi/blob/f0a8ed30502529d03e5b7a2e39ba40104915cd24/sw/airborne/modules/loggers/sdlog_chibios/usbStorage.c#L113-L122

fvantienen:
It is not a problem with the actual storage device or usb serial devices. The problem is because of the shutting down of the logger directly at the start, because no voltage is detected.

So if you power only through usb then after the defined start time (default 30sec), it will start up the logger and the power watching thread as well. The power watching thread gets triggered directly and tries to close the log and then hangs on the above mentioned `f_close`. This is the intented behavior, but somehow it can't close the files and fully hangs the autopilot. I could remove this from happening by preventing the logger to start if no power is detected, but this just hides the problem of not being able to close the files properly.

This doesn't matter for me now, but is just a strange bug which should be addressed maybe in a different PR and we merge this ine. I could just not figure out why this was happening, so I thought maybe someone else would have some insight.

gautierhattenberger:
Ok, then maybe you can open an issue with the description of the close bug. I'll ask Alex to investigate, but it won't be a top priority.

PR #3017, opened 2023-03-24
[boards] Add support for the CubePilot - Cube Orange+

fvantienen:
This adds support for the Cube Orange+. Currently the third IMU doesn't work because of the Invensense V3 driver is not implemented yet.

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -212,17 +212,19 @@
     </message>
 
     <message name="IMU_GYRO_RAW" id="31">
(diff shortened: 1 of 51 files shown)


OpenUAS (dismissed):
For the in the makefile I would like to see a bit less extreme wildcard, more something like "*CubeOrange*-BL*"

fvantienen:
This wildcard is needed because of multiple reasons. This is the same wildcards others use and it doesn't matter as the script checks the correct board after this anyways.

Fabien-B (approved):

PR #3018, opened 2023-03-27
[fix] apply correctly horizontal and vertical setpoints

gautierhattenberger:
Horizontal and vertical can have setpoint of different types (pos, speed or accel) independently, which is wrong in many cases.
It has been introduced recently with nav and guidance reorg (#2964 ).


fvantienen (approved):
Thanks we had some similar problems indeed and were trying to fix it also.

PR #3021, opened 2023-04-04
Survey hybrid

gautierhattenberger:
Survey based on the OSAM algo for rotorcraft. Instead of going from one  sweep line to the next with a line, a circle aligns the aircraft with the next line.
Circles radius at the start of lines can be either automatic (sweep distance), fixed or direct line (equivalent to original rotorcraft).

The code itself is reorganized to use a global structure instead of many variables. Only tested in simulation for now. If working correctly, it can be used for hybrids, rotorcraft and even fixedwing using the rotorcraft firmware.

### DIFF
--- a/conf/modules/digital_cam.xml
+++ b/conf/modules/digital_cam.xml
@@ -53,6 +53,7 @@
 
   <dep>
     <conflicts>digital_cam_i2c,digital_cam_servo,digital_cam_uart,digital_cam_video</conflicts>
+    <provides>digital_cam</provides>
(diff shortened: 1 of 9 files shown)


gautierhattenberger:
image

fvantienen (approved):
I checked the general code, not for the functionality or logic but it seems good 👍 
Only thing is missing is sort of an example on how to use it.

gautierhattenberger:
I might add examples later when the related project is finished

PR #3026, opened 2023-04-26
[state] fix speed convert functions

gautierhattenberger:
It was making a mess with float/int format and even `pos_status`. Only visible with certain combinations of get/set speed/pos.
It is still a miracle that it "worked" for so long...


noether (approved):
Unbelievable

fvantienen:
How did you even find this?

gautierhattenberger:
Thanks to hybrid nav ;)
I'm lucky that it was only visible as a side effect in the display of the GCS and not in the control itself...

PR #3029, opened 2023-05-03
Changes to fix video streams not working in Ubuntu 22.04

guidoAI:
I could not see the video streams of the Bebop 2 with VLC player. Also the Python rtp viewer was not working.

The issue I had is also mentioned here for openCV: https://github.com/opencv/opencv/issues/10328

So:
- I added two shell scripts to view the streams with ffmpeg player.
- I changed the python script to solve the above problem.

For me the Python script still throws an error with the changes, since a file in var/lib/python/pprzlink does not direct to the right messages.xml file. To make it work, I add 'sw/ext/pprzlink' to the directory name there, but it is autogenerated and on paparazzi's ignore list. Don't know how to fix that, so suggestions are welcome there.


dewagter (approved):
Great. Thanks for the fix!
It is tested and is backward compatible.

PR #3031, opened 2023-05-10
[nps] Fix NPS compile warnings

fvantienen:
Fixing some warnings on arm64 compilation.

The stringop-overflow comes from `stabilization_indi.c` at line 831 (```float_mat_inv_4d(g1g2inv[0], g1g2_trans_mult[0]);```). This converts a 2d array into a 1d array and leads to warnings not finding the correct memory sizes. I could not pragma push this warning so that is why I had to disable them fully. I'm not 100% happy about the solution but I don't know a better one, so if someone has suggestions?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -828,7 +828,7 @@ void calc_g1g2_pseudo_inv(void)
   float_vect_scale(g1g2_trans_mult[0], 1000.0, INDI_OUTPUTS * INDI_OUTPUTS);
 
   //inverse of 4x4 matrix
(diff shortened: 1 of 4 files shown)


gautierhattenberger:
I also think it would be better to keep the warning enabled. I can see 2 options at least:
- memcpy the 2d array to a 1 array and back (it is barely nothing compare to when we use the wls alloc instead of pseudo inverse)
- change the input arguments of the inv_4d to `float**` and use the `MAKE_MATRIX_PTR` macro

fvantienen:
I now changed it fully to just a `m[4][4]` this way it can still check the memory as well. Though if someone can check if I updated the math correctly that would be nice @EwoudSmeur or @gautierhattenberger ?
This is indeed the better solution and since it was only used in `stabilization_indi` it was very easy to change.

gautierhattenberger (approved):
Even better like this. It looks correct for the indices.

PR #3032, opened 2023-05-16
Fix indi wu initializer

EwoudSmeur:
We found out that the initialization of indi_Wu was not working, as it was setting only the first element of the array. This lead to problems with overactuated vehicles that did not explicitly set the indi_Wu matrix.

I also reduced the tolerance on the WLS solution.


gautierhattenberger (approved):

gautierhattenberger:
Great, that must explain why I had different results when setting Wu by hand last week...

fvantienen:
Apperently thhis initializer `{0}` only works with `0` to set the whole array(according to the internet as it is hard to verify). If you use a different number than `0` it doesn't work and only sets the first value, so it is something to keep in mind.

PR #3033, opened 2023-05-16
[dshot] fix dshot telemetry and provide pprz message support

gautierhattenberger:
- Each dshot driver supports it's own telemetry uart
- Pprz message ESC can be sent over telemetry or flight recorder
- Configuration is done by giving the uart number

Example for Tawaki on default dshot connector:
`<define name="DSHOT_TIM4_TELEMETRY_NUM" value="DSHOT_TLM_RX"/>`

### DIFF
--- a/conf/telemetry/default_rotorcraft.xml
+++ b/conf/telemetry/default_rotorcraft.xml
@@ -202,6 +202,7 @@
       <message name="IMU_ACCEL_SCALED"         period=".075"/>
       <message name="IMU_MAG_SCALED"           period=".2"/>
       <message name="LIDAR"                    period="0.05"/>
+      <message name="ESC"                      period="0.05"/>
     </mode>
   </process>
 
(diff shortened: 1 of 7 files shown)


fvantienen (approved):
Nice looks good!

Since we now have multiple modules transmitting the RPM message I actually was looking at changing this ABI call a bit. Because now it is hard to relate it exactly to which motor it is coming from. We have 2 sorts of indexes generated for the actuators, 1 locally for each driver and 1 gobally for all actuators combined. Ideally the ABI call has an array of structs where these ID's are also communicated. This way it becomes way easier to use and to identify which motor RPM you got, since you can index it with the generated airframe.h

gautierhattenberger:
I would like to change the abstract actuator driver to have an array of structure holding the values in pprz_t and "real" actuator value + a bunch of info like rpm or position when available. This way, the ABI message for RPM wouldn't be necessary as you'll be able to access the data from actuator's interface with the global ID. Some day...

PR #3034, opened 2023-05-24
Only do one INDI init call

EwoudSmeur:
Before, the INDI init function was called from the stabilization code as well as the module init, leading to double binding to ABI messages and such. Now it is only the module init.


gautierhattenberger:
@EwoudSmeur for your information, I'm trying to reorganize the stabilization layer to split it from guidance, but also avoid using the Abi binding to push the throttle increment and a few other things. I still have some corner cases to solve, but if you have "requirements" regarding the way guidance and stabilization are working together, let me know.

EwoudSmeur:
Thanks for the heads-up Gautier! I think it is a good idea.
What I liked about hte ABI messages was that the stabilization could use certain references if they were set by the guidance, and ignore them if they were not set, which made it quite flexible. If we can keep the interaction between lower level and higher level control flexible in some way that would be great.

PR #3039, opened 2023-06-29
predict touchdown in case of kill

EwoudSmeur:
This is an idea for a module that predicts where a drone is going to touchdown if it would be turned off. This could be used in combination with geofencing to better guarantee that the drone doesn't fall on some dangerous location, because of ballistically passing the geofence.

At the moment the module just calculates the position and sends a message. It could be cool to display this location in the GCS for situational awareness/verification purposes.

Tested in NPS simulator.

### DIFF
--- a/conf/modules/ballistic_touchdown.xml
+++ b/conf/modules/ballistic_touchdown.xml
@@ -3,7 +3,10 @@
 <module name="ballistic_touchdown" dir="nav">
   <doc>
     <description>
-      Module that predicts the crash location if the motors would be turned off, assuming a ballistic trajectory based on the current velocity.
+      Module that predicts the crash location if the motors would be
+      turned off, assuming a ballistic trajectory based on the current
+      velocity. Suggested use: trigger exceptions in the flight plan with
+      the ballistic_pos variable.
     </description>
   </doc>
   <dep>
--- a/sw/airborne/modules/nav/ballistic_touchdown.c
+++ b/sw/airborne/modules/nav/ballistic_touchdown.c
@@ -24,7 +24,7 @@
 
 float g = -9.81f;
 
-struct FloatVect2 crash_pos;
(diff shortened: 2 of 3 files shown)


dewagter:
Great.
It should at least trigger an audio alarm when the point is outside the geofence.

gautierhattenberger:
I'm not really sure this is worth adding this as an airborne module (and use a message for that). Since the computation is simple, can be done from existing telemetry data and needs adaptation of the GCS anyway, why not adding this as a functionality of the GCS itself ? An option that you could show or hide as needed.

EwoudSmeur:
@gautierhattenberger indeed you're right for this functionality. For sure the extra message is not needed, as the required information is already sent in other messages. 

What I was thinking of eventually, would be to trigger geofencing behavior (kill) based on this touchdown point. For that the module should run onboard. By triggering the geofence on the predicted crash location, you can better place the geofence, as you don't need to worry about keeping a large margin because you don't know the speed in advance when the drone crosses the geofence.

Maybe its best to have a module onboard and in the GCS, to avoid the communication overhead?

gautierhattenberger:
Yes, I think the two things can be separated and used independently:
- an onboard module to trigger safety procedures by taking into account the ballistic trajectory in case of failure
- a visualization option integrated to the GCS to display predicted crash location.

gautierhattenberger (approved):

EwoudSmeur:
Thanks for the feedback. I would suggest merging now, we can add the GCS updates in a different PR.

dewagter (approved):
Great. Important for safety.

Fabien-B:
@EwoudSmeur To display the predicted crash site in the GCS, you might look at this commit: https://github.com/paparazzi/PprzGCS/commit/b0e1641670fddca0e3a8ce3f43ab100ba6f39a11

PR #3040, opened 2023-07-10
[imu] Add Invensense V3 driver

Fabien-B:
Driver for ICM42605 and similar IMUs.
I started from code written by @fvantienen and @jesusBV20.
It works in FIFO and polling mode.

Tested on a Matek H743-slim v1.5.

TODO:
- [ ] handle the IMU timestamps correctly, with a fixed dt.
- [ ] some cleanup

### DIFF
--- a/sw/airborne/modules/imu/imu_icm42688.c
+++ b/sw/airborne/modules/imu/imu_icm42688.c
@@ -41,8 +41,6 @@
 
 static struct invensense3_t imu_icm42688;
 
-static int samples_from_odr(int odr);
-
 void imu_icm42688_init(void)
 {
   imu_icm42688.abi_id = IMU_ICM42688_ID;
@@ -57,8 +55,7 @@ void imu_icm42688_init(void)
   imu_icm42688.gyro_range = INVENSENSE3_GYRO_RANGE;
   imu_icm42688.accel_odr = INVENSENSE3_ACCEL_ODR;
   imu_icm42688.accel_range = INVENSENSE3_ACCEL_RANGE;
-  imu_icm42688.sample_numbers = samples_from_odr(INVENSENSE3_GYRO_ODR);
-  
+    
   // The AAF freq needs to be high enough to avoid group delay and low enough to minimise noise and clipping
   imu_icm42688.gyro_aaf = INVENSENSE3_GYRO_AAF; // ~ODR/4
   imu_icm42688.accel_aaf = INVENSENSE3_ACCEL_AAF; // Fixed
@@ -77,18 +74,4 @@ void imu_icm42688_event(void)
 }
 
 
-static int samples_from_odr(int odr) {
-   float freq;
-   if(odr < INVENSENSE3_GYRO_ODR_200HZ) {
-    freq = 32000 / pow(2, odr-INVENSENSE3_GYRO_ODR_32KHZ);
-   }
-   else if(odr < INVENSENSE3_GYRO_ODR_500HZ) {
-    freq = 200 / pow(2, odr-INVENSENSE3_GYRO_ODR_200HZ);
-   }
-   else if(odr == INVENSENSE3_GYRO_ODR_500HZ) {
-    freq = 500;
-   } else {
-    // error
-   }
-  return ceilf(freq/PERIODIC_FREQUENCY);
-}
+
--- a/sw/airborne/modules/imu/imu_icm42688.c
+++ b/sw/airborne/modules/imu/imu_icm42688.c
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2022 Freek van tieen <freek.v.tienen@gmail.com>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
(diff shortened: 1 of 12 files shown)


sw/airborne/peripherals/invensense3.c
```diff
@@ -0,0 +1,875 @@
...
+
+  gyro[0].p =  (int16_t)((uint16_t)data[8] | data[9]  << 8 );
+  gyro[0].q =  (int16_t)((uint16_t)data[10]| data[11] << 8 );
+  gyro[0].r =  (int16_t)((uint16_t)data[12]| data[13] << 8 );
+
+  // Send the scaled values over ABI
+  uint32_t now_ts = get_sys_time_usec();
+  AbiSendMsgIMU_GYRO_RAW(inv->abi_id, now_ts, gyro, 1, temp_f);
+  AbiSendMsgIMU_ACCEL_RAW(inv->abi_id, now_ts, accel, 1, temp_f);
+  AbiSendMsgTEMPERATURE(inv->abi_id, temp_f);
```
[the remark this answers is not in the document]
fvantienen: I would prefer to keep it in since we have multiple sensors using the same driver on the same device. Else that makes it difficult and a lot of the same code duplicated everywhere. That makes it in my opinion worse. If you look at all the IMU's they are like 99% the same code over and over again, while now if you use this peripheral you don't have that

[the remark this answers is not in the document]
gautierhattenberger: Ok, let's keep it like this. We'll see later if it needs modifications.

sw/airborne/modules/imu/imu_icm42688.c
```diff
@@ -0,0 +1,94 @@
...
+  invensense3_periodic(&imu_icm42688);
+}
+
+void imu_icm42688_event(void)
+{
+  invensense3_event(&imu_icm42688);
+}
+
+
+static int samples_from_odr(int odr) {
```
fvantienen, line 80: I think this should be done in the driver? I haven't looked where this number is used for but I think that should be better

sw/airborne/peripherals/invensense3.c
```diff
@@ -0,0 +1,875 @@
...
+}
+
+/**
+ * @brief Parse the FIFO buffer data
+ * 
+ * @param inv The invensense v3 instance
+ * @param data The FIFO buffer data to parse
+ * @param len The length of the FIFO buffer
+ */
+static void invensense3_parse_fifo_data(struct invensense3_t *inv, uint8_t *data, uint16_t samples) { //TODO: test changes
```
fvantienen, line 455: Is this TODO done?

sw/airborne/modules/imu/imu_icm42688.c
fvantienen: Maybe just set the PERIODIC calling frequency as a parameter for the peripheral.

fvantienen (commented):
Looks good, but tested it with the cube orange and couldn't get it working. Maybe I configured something wrong but will take a look later.

sw/airborne/modules/imu/imu_icm42688.c
Fabien-B: I agree it should be done in the driver, I will do it soon

Fabien-B:
I'm curious why it didn't worked for you.
Is the IMU connected with SPI or I2C ?
I only tested it with SPI, on a Matek H743-SLIM v1.5.

fvantienen:
It is connected with SPI but will take a look later and possibly with a logic analyzer. It is the Cube Orange+

fvantienen:
Btw what are you meaning with 'handle the IMU timestamps correctly, with a fixed dt.'. I think it would be nice if next to the `sys_get_timestamp` it also forwards the sensor 'dt', but I think both are interesting.

Fabien-B:
Currently, only the latest sample from each FIFO read is used, like in the classic poll mode.
I saw that the EKF2 can use all samples if INTEGRATION is defined, but I haven't tested it.

In the ABI message, the timestamp is the time at which the FIFO was read, so if the INS take this time as the sample time, it will be wrong (all samples would be at the same time).
So be cause we know the sample rate, using it as a fixed sample rate makes sense.

fvantienen:
Aha ok that explains, we use the ekf2 with `INTEGRATION` for quite some time now and it works really well.

Fabien-B:
Is the MAG necessary for using ekf2 ? I tried using it yesterday but it complained about not having a magnetometer.
I also have questions for you, lets try the github discussions for that : https://github.com/paparazzi/paparazzi/discussions/3047

fvantienen:
Yes the magnetometer is necessary, currently I didn't implement the internal magnetometer

dewagter:
https://github.com/paparazzi/paparazzi/pull/3076 is included in this PR.

fvantienen:
I have tested it on the cube orange+ and it works. I also made an update/fix for the sample rate, please see:
https://github.com/paparazzi/paparazzi/compare/master...fvantienen:paparazzi:feature-cube_orangeplus

fvantienen:
Since I can't push to this branch all the changes are in https://github.com/paparazzi/paparazzi/pull/3017
I tested it on the Cube Orange+ and it works, also added the sample rate for the integration. This can later also be used by other AHRS/INS then the EKF2. This should increase stability of the measurements and the AHRS/INS

Fabien-B:
Ok, I'm closing this PR in favor of #3017, which I will test tomorrow.

PR #3043, opened 2023-07-20
Fix guidance hybrid indi enter

EwoudSmeur:
Upon entering while in forward flight, the reference angles need to be reset based on the ZXY rotation order for tailsitters.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
@@ -405,15 +405,6 @@ void guidance_h_nav_enter(void)
   guidance_h_set_pos(nav.carrot.y, nav.carrot.x);
   reset_guidance_reference_from_current_position();
 
-#if HYBRID_NAVIGATION
-  /*Obtain eulers with zxy rotation order*/
-  struct FloatEulers eulers_zxy;
-  float_eulers_of_quat_zxy(&eulers_zxy, stateGetNedToBodyQuat_f());
-  nav_heading = eulers_zxy.psi;
-#else
-  /* set nav_heading to current heading */
-  nav.heading = stateGetNedToBodyEulers_f()->psi;
-#endif
   guidance_h_set_heading(nav.heading);
   /* call specific implementation */
   guidance_h_run_enter();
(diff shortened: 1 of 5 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -404,8 +404,16 @@ void guidance_h_nav_enter(void)
   /* horizontal position setpoint from navigation/flightplan */
   guidance_h_set_pos(nav.carrot.y, nav.carrot.x);
   reset_guidance_reference_from_current_position();
+
+#if HYBRID_NAVIGATION
+  /*Obtain eulers with zxy rotation order*/
+  struct FloatEulers eulers_zxy;
+  float_eulers_of_quat_zxy(&eulers_zxy, stateGetNedToBodyQuat_f());
+  nav_heading = eulers_zxy.psi;
+#else
```
gautierhattenberger, line 413: This `HYBRID_NAVIGATION` flag should have been removed from the code in latest master. I'd rather not adding it back.
Since this function ends with a call `guidance_h_run_enter` which is implemented by the underlying guidance, you can reset the nav heading with the proper value in the function `guidance_indi_enter`.
Also, I'm wondering if this is really necessary, since the the rotations starts around the z axis in both cases, it should correspond to the same psi angle, or am I wrong ?

EwoudSmeur: Thanks for your comment, will change! Unfortunately one of our drones is flying on an old branch, as the AP crashes on the latest master due to something happening in the control allocation, which I still have to figure out.

Unfortunately it is necessary, it is true that the first rotation is around Z, but the first rotation has to be different in both cases. Imagine a pitch up of 91 degrees in ZXY. With ZYX you need 180 degrees psi, 89 degrees pitch up and 180 degrees roll.

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -404,8 +404,7 @@ void guidance_h_nav_enter(void)
   /* horizontal position setpoint from navigation/flightplan */
   guidance_h_set_pos(nav.carrot.y, nav.carrot.x);
   reset_guidance_reference_from_current_position();
-  /* set nav_heading to current heading */
-  nav.heading = stateGetNedToBodyEulers_f()->psi;
+
   guidance_h_set_heading(nav.heading);
   /* call specific implementation */
   guidance_h_run_enter();
```
gautierhattenberger, line 410: if the `nav.heading` is now set from the `guidance_h_run_enter()` function, maybe the last two lines should be swapped.

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
EwoudSmeur: Good point, should be fixed now!

gautierhattenberger (approved):

PR #3044, opened 2023-07-21
fix airspeed update bug from ms45xx

EwoudSmeur:


EwoudSmeur:
missed the scaling factor, I thought the calculation was wrong... Need to find another explanation for my scaled airspeed then!

PR #3045, opened 2023-07-21
fix origin altitude ekf2

EwoudSmeur:
HMSL was set to be altitude...

### DIFF
--- a/sw/airborne/modules/ins/ins_ekf2.cpp
+++ b/sw/airborne/modules/ins/ins_ekf2.cpp
@@ -769,7 +769,7 @@ static void baro_cb(uint8_t __attribute__((unused)) sender_id, uint32_t stamp, f
   ekf.set_air_density(rho);
 
   // Calculate the height above mean sea level based on pressure
-  sample.hgt = pprz_isa_height_of_pressure_full(pressure, ekf2.qnh * 100.0f); 
+  sample.hgt = pprz_isa_height_of_pressure_full(pressure, ekf2.qnh * 100.0f);
   ekf.setBaroData(sample);
 }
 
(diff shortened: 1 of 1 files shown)


fvantienen (approved):

gautierhattenberger:
@EwoudSmeur @fvantienen Are you sure about this ? In the header file `https://github.com/PX4/PX4-ECL/blob/master/EKF/common.h#L67` it says that it should be altitude above MSL in millimeters.

EwoudSmeur:
I'm sure the altitude of my absolute waypoints was off by 40 meters, and the message ins_ref reported the origin at 0m alt, which should be 40m at our flight field. It doesn't make sense to me to write hmsl to an alt value as well.

Perhaps we're not interpreting the ekf2 struct correctly when setting the origin?

gautierhattenberger:
It is not all clear, but I think the problem comes from the initial altitude used as filter's origin.
If you look at https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/ekf2/EKF2.cpp#L1171, the absolute `alt` is the filter z value + the origin, while `alt_ellipsoid` is explicitly computed from `alt`. You can also see in https://github.com/PX4/PX4-Autopilot/blob/main/msg/SensorGps.msg#L10 that `altitude` is the MSL value for them.
If me have an offset in paparazzi, I think it might come from the init value in https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/ins/ins_ekf2.cpp#L525. We are using `NAV_ALT0 + NAV_MSL0` as initial altitude. However `NAV_ALT0` is the alt AMSL, and `NAV_MSL0` is the difference between geoid and ellipsoid. So I think we should initialize with `NAV_ALT0` only and update the filter with `hmsl` like before to be consistent with the filter's "documentation".

fvantienen:
Indeed I think it would be better to change everything to AMSL if the ekf2 expects that. So changing the initialization would be best and also the conversion from ekf2 origin to pprz origin.

dewagter:
Continued in https://github.com/paparazzi/paparazzi/pull/3106

PR #3046, opened 2023-07-31
[chibios] Fix ADC reading sample rate error on H7

fvantienen:
This fixes the sample rate setting on H7 processors, which was previously not set correctly. This resulted in really bad readings on the ADC with a lot of crosstalk.

For all H7 users, please recalibrate your `VoltageOfAdc` and `CurrentOfAdc`!

### DIFF
--- a/sw/airborne/arch/chibios/mcu_periph/adc_arch.c
+++ b/sw/airborne/arch/chibios/mcu_periph/adc_arch.c
@@ -201,7 +201,7 @@ static void adc_configure(ADCConversionGroup *cfg, uint8_t num_channels, const u
 
 #if defined(STM32H7XX) || defined(STM32F3XX) || defined(STM32G4XX) || defined(STM32L4XX)
     cfg->pcsel |= (1 << chan);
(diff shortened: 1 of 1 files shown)


gautierhattenberger (commented):
I have not checked the datasheet, but it looks much more correct like this indeed. At the moment, we have only two boards with H7. Would it be possible to change the default values for these boards ? I think they have the same PX4 "standards", so their is a good chance that it is the same value.

fvantienen:
Will take a quick look for the supplied BEC of the Cube what the default value is. I don't have the Matek H743-slim but maybe @Fabien-B knows what the default value is there?

Fabien-B:
The default value was `(3.3f/65536.0f) * 13.38`, I changed it to `(3.3f/65536.0f) * 11.05`.
Was that the range you expected ?

fvantienen:
The previous value didn't say anything, but the `11` seems to be a reasonable divider.

Fabien-B:
I just changed from 13 to 11, so this is not a very big change...

Fabien-B (approved):

PR #3053, opened 2023-08-30
[ins] Fix EKF2 working with HMSL

fvantienen:
I think this will fix the EKF2 origin since now HMSL and altitude are both used and results in wrong calculations of the INS reference. (Especially line 525 and 624 were conflicting)

Now:
- The EKF2 works with HMSL (Following the standard and descriptions of PX4)
- The PPRZ origin is in WGS84 mm above the reference ellipsoid
- The origin in the EKF2 is with HMSL
- The GPS inputs the HMSL in the EKF2
- The origin from the EKF2 is converted from HMSL to WGS84 with `+ wgs84_ellipsoid_to_geoid_i(lla_ref.lat, lla_ref.lon)` for the PPRZ origin

### DIFF
--- a/sw/airborne/modules/ins/ins_ekf2.cpp
+++ b/sw/airborne/modules/ins/ins_ekf2.cpp
@@ -522,13 +522,13 @@ void ins_ekf2_init(void)
 
   /* Initialize the origin from flight plan */
 #if USE_INS_NAV_INIT
-  if(ekf.setEkfGlobalOrigin(NAV_LAT0*1e-7, NAV_LON0*1e-7, (NAV_ALT0 + NAV_MSL0)*1e-3))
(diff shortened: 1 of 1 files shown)


fvantienen:
This reverts part of https://github.com/paparazzi/paparazzi/pull/2930 and because of https://github.com/paparazzi/paparazzi/pull/2930 this is happening every time.

fvantienen:
And I also looked a bit in the EKF2 but I think it resets the origin even if you set it from the start. Because `_NED_origin_initialised` is not set to true by the `setEkfGlobalOrigin` and then inside the `collect_gps` it will overwrite the origin. I need to verify this, but this might cause some problems and was also what was causing my altitude issues.

EwoudSmeur:
I think the problem occurs when the origin is not at sea level. Great fix!

EwoudSmeur (approved):

gautierhattenberger (approved):

PR #3054, opened 2023-08-30
[indi] Feed forward rates from guidance

fvantienen:


gautierhattenberger:
Sorry, but I really don't like the idea to make a strong dependency in stabilization with the indi hybrid guidance. This way, you will have undefined references if you don't use this particular guidance.
You have the possibility to move the feedforward variables to stab and update them if needed from guidance (which I don't really like), or change the Abi THRUST message to include this feedforward term (and maybe change the message name). I don't like much more the second option, but since it is already in place, I guess it is acceptable.
I'll keep in mind to include rates setpoints when rewriting the stabilization layer, I promise.

EwoudSmeur:
Haha sorry this wasn't intended to go to master directly! But at least it works and we can think of a nice solution to add this. I think they should be settable from (any) guidance, and if they are set they can be used, otherwise they should be zero. That's why I enquired during our discussion in the other issue about this, but I didn't have time to implement this in the suggested way yet.

PR #3056, opened 2023-09-05
fix log time overflow

EwoudSmeur:
During very long flights, the time in the log would overflow and start back at zero. This complicates the parsing of log files.

It was caused because the function supplying the time in usec was overflowing after 70 minutes. By adding a function which is scaled by a factor 1/100 this should be solved.


fvantienen (dismissed):

gautierhattenberger (dismissed):
Looks good. Can be merged after updating pprzlink.

EwoudSmeur:
Thanks!

dewagter (approved):
Great fix. Tested.

PR #3059, opened 2023-09-08
[python] fuel-cell endurance estimators and status fixes

dewagter:


dewagter:
Depends on https://github.com/paparazzi/pprzlink/pull/172 : can I please merge it so all these pull requests can be fixed?

PR #3060, opened 2023-09-08
[Modules] Generic UART Sensor

dewagter:


fvantienen (approved):

sw/airborne/modules/sensors/generic_uart.c
```diff
@@ -0,0 +1,68 @@
...
+  // Transmit buffer
+  static uint8_t msg_buf_snd[128];
+  static uint8_t msg_cnt_snd = 0;
+
+  // Look for data on serial port and save it in the buffer
+  while (gen_uart_dev->char_available(gen_uart_dev->periph)) { 
+    gen_msg_buf[gen_msg_cnt++] = gen_uart_dev->get_byte(gen_uart_dev->periph);
+    
+    if(gen_msg_buf[gen_msg_cnt-1] == GENERIC_UART_ENDCHAR)
+      break;
```
gautierhattenberger, line 54: Here, it seems that there is no protection for overflow if there is no endchar in the frame. It would be safer to have that.

sw/airborne/modules/sensors/generic_uart.c
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Forward the message to the GCS
+  if(gen_msg_buf[gen_msg_cnt-1] == GENERIC_UART_ENDCHAR || gen_msg_cnt > 50) {   
+    msg_cnt_snd = gen_msg_cnt;
+    for(uint8_t i = 0; i < msg_cnt_snd; ++i)
+      msg_buf_snd[i] = gen_msg_buf[i];
```
gautierhattenberger, line 61: I'm not sure is it needed to make a copie of the buffer, the send message function will copy it to uart buffer anyway.

gautierhattenberger (changes requested):

fvantienen:
All comments should be resolved in the new PR

PR #3067, opened 2023-09-09
[modules] Approach moving target filter

dewagter:
- high filter values make the filter unstable

### DIFF
--- a/sw/airborne/modules/ctrl/approach_moving_target.c
+++ b/sw/airborne/modules/ctrl/approach_moving_target.c
@@ -27,10 +27,17 @@
 
 #include "generated/modules.h"
 #include "modules/core/abi.h"
+#include "filters/low_pass_filter.h"
 
 float amt_err_slowdown_gain = AMT_ERR_SLOWDOWN_GAIN;
 
 float approach_moving_target_angle_deg;
+
+float cutoff_freq_filters_hz = 0.5; // Hz
+
+Butterworth2LowPass target_pos_filt[3];
+Butterworth2LowPass target_vel_filt[3];
+Butterworth2LowPass target_heading_filt;
 
 #define DEBUG_AMT TRUE
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


EwoudSmeur:
This should be investigated. Perhaps a first order filter could work better with very low cutoff frequency

sw/airborne/modules/ctrl/approach_moving_target.c
```diff
@@ -75,11 +82,25 @@ static void send_approach_moving_target(struct transport_tx *trans, struct link_
...
 #if PERIODIC_TELEMETRY
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_APPROACH_MOVING_TARGET, send_approach_moving_target);
 #endif
 }
 
+void approach_moving_target_set_low_pass_freq(float filter_freq) {
+  cutoff_freq_filters_hz = filter_freq;
+  // tau = 1/(2*pi*Fc)
+  float tau = 1.0 / (2.0 * M_PI * cutoff_freq_filters_hz);
+  float sample_time = 1.0 / PERIODIC_FREQUENCY;
```
fvantienen, line 95: PERIODIC_FREQUENCY is wrong here, this should be the frequency of the `follow_diagonal_approach` (which is 100Hz) you can get that define out of the generated code.

sw/airborne/modules/ctrl/approach_moving_target.c
```diff
@@ -27,11 +27,18 @@
 
 #include "generated/modules.h"
 #include "modules/core/abi.h"
+#include "filters/low_pass_filter.h"
 
 float amt_err_slowdown_gain = AMT_ERR_SLOWDOWN_GAIN;
 
 float approach_moving_target_angle_deg;
 
+float cutoff_freq_filters_hz = 0.5; // Hz
```
fvantienen, line 36: Maybe add a define for setting the default value

fvantienen (changes requested):

fvantienen:
> tigated. Perhaps a first order filter could work better with very low cutoff

See my comments, you were filtering with a define which said 500Hz but it was actually running at 100Hz

dewagter (commented):
All requests implemented.

fvantienen (approved):

PR #3068, opened 2023-09-09
Feed forward guidance

dewagter:
@EwoudSmeur should this be implemented differently?

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
@@ -351,6 +351,16 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
   omega -= accely_filt.o[0]*FWD_SIDESLIP_GAIN;
 #endif
 
+  // We can pre-compute the required rates to achieve this turn rate:
+  // NOTE: there *should* not be any problems possible with Euler singularities here
+  struct FloatEulers *euler_zyx = stateGetNedToBodyEulers_f();
+
+  struct FloatRates ff_rates;
+
+  ff_rates.p = -sinf(euler_zyx->theta) * omega;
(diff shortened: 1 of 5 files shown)


EwoudSmeur:
Haha see #3054

dewagter:
Sorry, I missed that this had been proposed before.

Can this become an issue then?

dewagter:
Wiil fix this PR.

dewagter:
@gautierhattenberger is this proposition better, or even worse?

In the long run, passing extra information to innerloops should be the way to go: the guidance already knows things the feedback controller in the inner loop can not know yet: e.g. when following trajectories like in https://doi.org/10.1109/ICRA.2011.5980409

For the generic future version of ```stabilization.h```, besides rates_ff, the desired angular acceleration would also be nice to pass.

dewagter:
Legend:
--------
 - ```ATT``` setpoint and ```RATE=0``` -> normal attitude feedback
 - ```RATE``` setpoint and no ```ATT``` -> normal rate feedback
 - ```ATT``` and ```RATE```, then -> ATT feedback with RATE Feed-Forward

*When using stabilization with a reference model, the FF is probably overruled, as the reference lags behind the setpoint.*

EwoudSmeur:
Right now we have the reference model, which can perform a similar function: calculate feed forward references for e.g. rates or velocities, in order to track the reference. But this is a function that could also be taken by other methods, e.g. a 4D trajectory, vector field, or some other algorithm that can compute feed forward commands. I think it would be better to make one structure with which these different methods can interact with the reference.

The references that are not set during the control loop should be set to zero, such that the controller doesn't use them. I think this would make it much easier to implement and switch between different tracking methods.

gautierhattenberger (dismissed):
I'm fine with this implementation until the new stabilization structure is finished.

gautierhattenberger (approved):

gautierhattenberger (approved):

PR #3070, opened 2023-09-09
[tools] Visualizer Rotating Wing

dewagter:
Rotwing Monitor


fvantienen (approved):
For now this is ok, but this is no generic at all. This should be rewritten into something similar as https://github.com/tudelft/paparazzi_log_parsing/tree/master/matlab/3d_animation

There you can see there is a configuration for each rotorcraft: https://github.com/tudelft/paparazzi_log_parsing/blob/master/matlab/3d_animation/import_stl_model/RotatingWingV3/convert_model.m
https://github.com/tudelft/paparazzi_log_parsing/blob/master/matlab/3d_animation/import_stl_model/Nederdrone5/convert_model.m

Also having large stl/ply files in the repo is in general not a good idea, @gautierhattenberger what do you think? As this slows down git and makes the repo huge

dewagter:
Maybe I can already move the STL to the gazebo simulator as it would be useful there too.

sw/ground_segment/python/rot_wing_mon/rot_wing_mon.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2018 TUDelft
```
dewagter, line 3:
```suggestion
# Copyright (C) 2023 TUDelft
```

sw/ground_segment/python/rot_wing_mon/rot_wing_viewer.py
```diff
@@ -0,0 +1,284 @@
+#
+# Copyright (C) 2018 TUDelft
```
dewagter, line 2:
```suggestion
# Copyright (C) 2023 TUDelft
```

fvantienen (approved):

PR #3071, opened 2023-09-14
[modules] Generic UART sensor cleanup

fvantienen:
Fixup for the comments from https://github.com/paparazzi/paparazzi/pull/3060


fvantienen:
I was not 100% sure if the buffer got copied but I checked pprzlink and it should, so I removed also the send buffer.

dewagter (approved):
Important safety fix.

gautierhattenberger (approved):

PR #3072, opened 2023-09-15
[conf] Fix compile warnings

fvantienen:
This fixes some compile warnings and also updates mavlink to V2

### DIFF
--- a/conf/conf_example.xml
+++ b/conf/conf_example.xml
@@ -7,7 +7,7 @@
    telemetry="telemetry/default_fixedwing_imu.xml"
    flight_plan="flight_plans/versatile.xml"
    settings="settings/fixedwing_basic.xml settings/nps.xml"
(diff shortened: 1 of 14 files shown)


fvantienen:
Maybe someone knows how to fix the SRTM and geofence warning from the CubeOrange. Can we disable these in the `gen_flight_plan.ml` for test builds?

gautierhattenberger (approved):

gautierhattenberger:
Disabling these warnings only for certain cases would be a bit painful, so if we can live with that, I would appreciate.

fvantienen:
I'm ok with how it is now, but it was more if it was easy to change/add it would have been nice. Though having warnings in our test builds is never good and makes it harder to search for real problems

fvantienen:
```
Warning: AIRCRAFT=ardrone2_gazebo target=nps compiled sucessfully but had warnings:03:57
math/pprz_matrix_decomp_float.c: In function ‘pprz_svd_float’:03:57
math/pprz_matrix_decomp_float.c:274:19: warning: ‘l’ may be used uninitialized in this function [-Wmaybe-uninitialized]03:57
  274 |           if (a[i][l] != 0) {03:57
      |                   
```
Should be fixable and is the last warning I see next to the `flight_plan_gen` ones.

PR #3073, opened 2023-09-18
[module] faulhaber

dewagter:


fvantienen:
This one is not really needed anymore and doesn't implement it correctly

PR #3081, opened 2023-09-20
Do not kill a quadplane inflight

dewagter:
A quadplane that flies very stably in forward flight with hover motors off will set the ```inflight``` to false, effectively killing the drone.

New conditions are needed to detect the end of ```inflight``` for quadplanes.

Proposed: only let the flightplan kill in the flare block by setting ```motors_on``` to ```false```

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
@@ -72,9 +72,24 @@ static uint32_t autopilot_in_flight_counter;
 #define AUTOPILOT_IN_FLIGHT_MIN_THRUST 500
 #endif
 
+/** Z-acceleration threshold to detect ground in m/s^2 */
+#ifndef THRESHOLD_GROUND_DETECT
+#define THRESHOLD_GROUND_DETECT 25.0
+#endif
 
-/** Default ground detection estimation based on thrust and speed */
+/** Default ground-detection estimation based on accelerometer shock */
 __attribute__((weak)) bool autopilot_ground_detection(void) {
+  struct NedCoor_f *accel = stateGetAccelNed_f();
+  if (accel->z < -THRESHOLD_GROUND_DETECT ||
+      accel->z > THRESHOLD_GROUND_DETECT) {
+    return true;
+  }
+  return false;
+}
+
+
+/** Default in-flight detection estimation based on thrust and speed */
+__attribute__((weak)) bool autopilot_in_flight_detection(void) {
   if (autopilot_in_flight_counter > 0) {
     /* probably in_flight if thrust, speed and accel above IN_FLIGHT_MIN thresholds */
     if ((stabilization_cmd[COMMAND_THRUST] <= AUTOPILOT_IN_FLIGHT_MIN_THRUST) &&
@@ -267,7 +282,7 @@ void autopilot_reset_in_flight_counter(void)
 void autopilot_check_in_flight(bool motors_on)
 {
   if (autopilot.in_flight) {
-    if (autopilot_ground_detection()) {
+    if (autopilot_in_flight_detection()) {
       autopilot.in_flight = false;
       autopilot_in_flight_counter == 0;
     }
--- a/sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
+++ b/sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
@@ -78,7 +78,7 @@ static uint32_t autopilot_in_flight_counter;
 #endif
 
 /** Default ground-detection estimation based on accelerometer shock */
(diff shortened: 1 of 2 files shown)


gautierhattenberger:
I agree the "in flight" detection is not really good, however it seems difficult to have both motors off, no accel and no speed (unless you go a perfect hovering with slope soaring). I'm fine if we have an option to disable this check, but I'd rather keep it by default.

fvantienen:
I think the main problems comes from `stabilization_cmd[COMMAND_THRUST]` not being reliable because of INDI(especially in VTOL). That in combination with a nice flying UAV it is being triggered while flying, which we have had already multiple times now.
Another temporary solution could also be to set the `stabilization_cmd[COMMAND_THRUST]` for VTOL to a correct value @EwoudSmeur ?

dewagter:
@fvantienen : The ```COMMAND_THRUST``` is set to zero when the hover props are off. For quadplanes, there is however also a ```COMMAND_THROTTLE``` (sometimes called ```COMMAND_PUSHER``` or even ```COMMAND_BX`` for body-X direction thrust`).

 1) which name should become the standard for pusher props in quadplanes
 2) is it OK to solve this issue with an extra ```ifdef COMMAND_PUSHER then PUSHER also < 100```
 3) is there a cleaner way to implement: ```if all motor commands are zero``` and not just the hover motors

If the hover and pusher props are all off, then if VZ is zero this can be seen as landing (except for the updraft soaring drone)

 4) or should the airframe file be able to define an extra condition?

Thanks for you input.

EwoudSmeur:
I think the goal of the in_flight boolean does not need to be to accurately detect a landing, we have different modules for that, that can be controlled from flight plan blocks. So I think whenever motors_on is true, we keep in_flight to true. If the pusher propeller is on, motors_on has to be true. I think that would solve the issue, no? It also removes any need for naming of additional thrusters.

@fvantienen I don't think this issue is related to INDI.

gautierhattenberger:
in_flight is not only for the landing but also for the takeoff phase. When you turn on the motors, you don't what to have some parts of the control running (integrators, increment). You could be in_flight with no motors if you are not in a steady state (should we consider airspeed for soaring drones ?). But if a pusher motor is on, then motors_on should be true as well as Ewoud said. What is happening with the hybrids with pusher props ? Can you have the motors_on and a motor running ?

dewagter:
@all thanks for the replies!

@gautierhattenberger 
 - this PR is inside the ```if (autopilot.in_flight)``` and should only change the landing. The inflight detection on take-off remains the same.
 - quadplanes keep ```motors_on = true``` when the pusher is on, but the ```stabilization_cmd[COMMAND_THRUST]``` becomes zero and currently, there is no pusher command but the ```actuators[PUSHER]``` is written directly (will solve later once we agree on a name: ```COMMAND_THROTTLE```?)

@fvantienen if you mean that the ```stabilization_cmd[COMMAND_THRUST]``` should reflect the total *3D* thrust (=hover+pusher) then I could agree this would also solve the problem for quadplanes (not for soaring drones)
@EwoudSmeur under manual control, the code without this PR will also stop INDI/I-gain integration upon landing, right? It would be nice to keep this.

The only thing is: in my view, ```in_flight``` detection should not run all the time, but only for instance under RC, failsafe landing, in the navigation flare block etc, ... This code has already crashed quite some drones over the years.

Proposition: aircraft with special needs or with ground-detection sensors can override the ```autopilot_ground_detect``` function but the default remains the same?

sw/airborne/firmwares/rotorcraft/autopilot_firmware.c
```diff
@@ -234,9 +248,7 @@ void autopilot_event(void)
       || autopilot.mode == AP_MODE_FAILSAFE
 #endif
      ) {
-    struct NedCoor_f *accel = stateGetAccelNed_f();
-    if (accel->z < -THRESHOLD_GROUND_DETECT ||
-        accel->z > THRESHOLD_GROUND_DETECT) {
+    if (autopilot_ground_detection()) {
```
dewagter, line 251: Why does the ground detection in ```NAV``` need to be different from the ground detection to stop ```in_flight```? PS: Detecting ground only using ```accel->z``` is super risky.

gautierhattenberger: Here, the idea was to detect the shock when hitting the ground from spike on the vertical acceleration. It allows a "faster" detection of the ground than waiting for multiple factors to be valid.
It is super risky indeed, but not activated by default. This is something you should activate only for the final flare, when already close to the ground, and if you have a really use for it.
In any case, it should either stay the same than before, or completely removed if not desired anymore.

gautierhattenberger (commented):
I'm in favor of the weak function to customize ground detection. However the shock detector is something different.

gautierhattenberger (approved):

PR #3084, opened 2023-09-20
[imu] Enable the sending of specific IMU sensors

fvantienen:


dewagter (dismissed):
Important with mulit-IMU boards.

sw/airborne/modules/imu/imu.c
```diff
@@ -164,6 +182,8 @@ static void send_accel_raw(struct transport_tx *trans, struct link_device *dev)
   static uint8_t id = 0;
   pprz_msg_send_IMU_ACCEL_RAW(trans, dev, AC_ID, &imu.accels[id].abi_id, &imu.accels[id].temperature,
                               &imu.accels[id].unscaled.x, &imu.accels[id].unscaled.y, &imu.accels[id].unscaled.z);
+  if(imu.accel_abi_send_id != ABI_DISABLE && imu.accel_abi_send_id != ABI_BROADCAST && imu.accels[id].abi_id == imu.accel_abi_send_id)
```
gautierhattenberger, line 185: Here it seems that DISABLE and BROADCAST have the same effect (send data in a rolling way). What about using the DISABLE to actually disable the sending of messages ? Does it makes sense ?

conf/modules/imu_common.xml
```diff
@@ -7,14 +7,24 @@
 
       This takes the IMU_GYRO_RAW, IMU_ACCEL_RAW and IMU_MAG_RAW ABI messages as input.
     </description>
+    <define name="IMU_INTEGRATION" value="FALSE" description="Enable gyro/accel integration calculations (enabled for the ekf2)"/>
+    <define name="IMU_GYRO_CALIB" value="{}" description="Gyroscope calibration structures list (see struct imu_gyro_t)"/>
+    <define name="IMU_ACCEL_CALIB" value="{}" description="Accelerometer calibration structures list (see struct imu_accel_t)"/>
+    <define name="IMU_MAG_CALIB" value="{}" description="Magnetometer calibration structures list (see struct imu_mag_t)"/>
+    <define name="IMU_GYRO_ABI_SEND_ID" value="ABI_BROADCAST" description="The gyro ABI ID which is send over telemetry/logging"/>
+    <define name="IMU_ACCEL_ABI_SEND_ID" value="ABI_BROADCAST" description="The accel ABI ID which is send over telemetry/logging"/>
+    <define name="IMU_MAG_ABI_SEND_ID" value="ABI_BROADCAST" description="The mag ABI ID which is send over telemetry/logging"/>
```
gautierhattenberger, line 16: I would prefer to have the structure
```
<section name="IMU" prefix="IMU_">
  <define name="INTEGRATION" ...
</section>
```
It tells the user to place the define outside the firmware section, which is much better in this section, especially for the calibration.

gautierhattenberger (changes requested):

sw/airborne/modules/imu/imu.c
fvantienen: Yes I thought about that as well will implement it like that

conf/modules/imu_common.xml
fvantienen: Will change it

sw/airborne/modules/imu/imu.c
```diff
@@ -155,115 +155,173 @@ PRINT_CONFIG_VAR(IMU_BODY_TO_IMU_PHI)
...
 
 static void send_accel_raw(struct transport_tx *trans, struct link_device *dev)
 {
+  if(imu.accel_abi_send_id == ABI_DISABLE)
+    return;
   static uint8_t id = 0;
   pprz_msg_send_IMU_ACCEL_RAW(trans, dev, AC_ID, &imu.accels[id].abi_id, &imu.accels[id].temperature,
                               &imu.accels[id].unscaled.x, &imu.accels[id].unscaled.y, &imu.accels[id].unscaled.z);
+  if(imu.accel_abi_send_id != ABI_BROADCAST && imu.accels[id].abi_id == imu.accel_abi_send_id)
+    return;
```
gautierhattenberger, line 188: Is this really correct ? If a specific ID is selected, which is not the first in the list, at least a first message from other sensors will be sent.

fvantienen: That is indeed what is happening and the intention, because it is the easiest way to implement it. This searches for the specific ID, but while searching it might send other ID's. Else a full search must be done at every message send, which requires more computing effort, and receiving a couple message of wrong sensors during search isn't that horrible.

dewagter: The full search should only be done when the first message is sent right? Not every time? Then searching the id once sounds OK to me. The alternative would be to not send anything while searching. But sending data from the wrong source might lead to confusing situations, don't you think?

dewagter (approved):

sw/airborne/modules/imu/imu.c
fvantienen: It should be done everytime or some more complex code which I don't find useful in this case to write. Since the id is send with the msg you can filter out wrong values easily.
If you want to do more complex stuff you can add it yourself, but note that everything is run at high frequencies and each of these function can also run at the same time.

In my opinion the current implementation is perfectly fine(compact, fast, not much code duplication) for the intention and usage.

PR #3085, opened 2023-09-20
[modules] Fix flight recorder and MAVLink frequency to TELEMETRY_FREQ…

fvantienen:
Because in `gen_periodic.ml` everything is calculated with the `TELEMETRY_FREQUENCY`, the best option is to run everything in the `datalink` task. This is chosen such that we can run the `TELEMETRY_FREQUENCY` higher than the `PERIODIC_FREQUENCY` if needed, for example to send multiple IMU measurements from different sensors.


dewagter (approved):
Agree

gautierhattenberger (approved):

fvantienen:
Before merging I need to fix something else also. Since the configure for `TELEMETRY_FREQUENCY` doesn't work, as it never is defined.

Makefile.ac
```diff
@@ -111,12 +111,10 @@ ifeq ($(MAKECMDGOALS),all_ac_h)
...
 ifdef PERIODIC_FREQUENCY
-# telemetry and module periodic frequency default to PERIODIC_FREQUENCY
-TELEMETRY_FREQUENCY ?= $(PERIODIC_FREQUENCY)
+# module periodic frequency default to PERIODIC_FREQUENCY
 DEFAULT_MODULES_FREQUENCY = $(PERIODIC_FREQUENCY)
 else
-$(warning Info: PERIODIC_FREQUENCY not configured, defaulting to 60Hz for modules and telemetry)
-TELEMETRY_FREQUENCY ?= 60
+$(warning Info: PERIODIC_FREQUENCY not configured, defaulting to 60Hz for modules)
 DEFAULT_MODULES_FREQUENCY = 60
```
gautierhattenberger, line 118: I'm not even sure this `DEFAULT_MODULES_FREQUENCY` is really used anymore

fvantienen:
I fixed the frequencies now, also fixed some minor bug in misconfiguration not showing up.

Makefile.ac
fvantienen: Searched for it and is not used anymore, so will remove

gautierhattenberger (approved):

PR #3098, opened 2023-09-22
[fix] Airspeed data different from airspeed state

dewagter:
do not use air density ```rho``` to calibrate airspeed sensor but force user to use ```pressure_scale```
 - otherwise, air_data and state are different

### DIFF
--- a/conf/modules/airspeed_ms45xx_i2c.xml
+++ b/conf/modules/airspeed_ms45xx_i2c.xml
@@ -6,6 +6,8 @@
       MS45XX differential or gauge type of pressure sensor for e.g measuring airspeed.
       This sensor module is using the MS45xxDO digital pressure sensor with readings via I2C.
       Needs to be a versions with 14bit pressure and 11bit temperature data output.
+
+      If the airspeed reading is badly calibrated, use the MS45XX_PRESSURE_SCALE to compensate.
     </description>
     <define name="MS45XX_I2C_DEV" value="i2cX" description="set i2c peripheral (default: i2c2)"/>
     <define name="MS45XX_I2C_ADDR" value="0x50" description="i2c slave address (default: 0x50)"/>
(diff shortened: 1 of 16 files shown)


sw/airborne/modules/sensors/airspeed_ms45xx_i2c.c
```diff
@@ -286,7 +287,7 @@ void ms45xx_i2c_event(void)
       float temp = ms45xx.temperature / 10.0f;
       AbiSendMsgTEMPERATURE(MS45XX_SENDER_ID, temp);
       // Compute airspeed
-      ms45xx.airspeed = sqrtf(Max(ms45xx.pressure * ms45xx.airspeed_scale, 0));
+      ms45xx.airspeed = sqrtf(Max(ms45xx.pressure, 0)) * MS45XX_AIRSPEED_SCALE;
 
 #if USE_AIRSPEED_MS45XX
       stateSetAirspeed_f(ms45xx.airspeed);
```
dewagter, line 293: Is it a good thing that drivers push to the state directly or should we enforce pushing through air_data?

dewagter:
@Dennis-Wijngaarden 
V3B: ```<define NAME="MS45XX_PRESSURE_SCALE" value="1.907762146"/>

conf/modules/airspeed_ms45xx_i2c.xml
```diff
@@ -6,6 +6,8 @@
       MS45XX differential or gauge type of pressure sensor for e.g measuring airspeed.
       This sensor module is using the MS45xxDO digital pressure sensor with readings via I2C.
       Needs to be a versions with 14bit pressure and 11bit temperature data output.
+
+      If the airspeed reading is badly calibrated, use the MS45XX_PRESSURE_SCALE to compensate.
```
dewagter, line 10:
```suggestion
      If the airspeed reading is badly calibrated, use the MS45XX_PRESSURE_SCALE to compensate.
      Use the MS45XX_AIRSPEED_SCALE only to change air density.
```

sw/airborne/modules/sensors/airspeed_ms45xx_i2c.c
gautierhattenberger: I would be in favor of an ABI message to go through air_data module.

dewagter: removed the ```stateSetAirspeed```, checked all airframes using this sensor that it has the ```<air_data>``` module. Removed all ```USE_MS45XX`` and also removed the ```MS45XX_SYNC_SEND``` since it now is sent via ```telemetry.xml``` and fixed some docs.

dewagter:
Cleanup ready: ```ms45xx`` now goes through ```<air_data>``` and the pressure needs to be calibrated and not the airspeed.

PR #3099, opened 2023-09-22
[periph] Invensense2 add timebase correction for PLL

fvantienen:
This corrects for the internal PLL errors in the invensense V2 chips. 
For the Invensense V3 there is no such register anymore. For those and other chips it might be good to average out x amount of measurement timings and calculate a timebase correction factor. Based on measurements with a generated frequency the PLL can deviate easily more than 1%


dewagter (approved):
Great to get the correct integration time. Can you add this to the DOC of the module please?

dewagter:
Is there a way to store this PLL info in the log as well?

PR #3101, opened 2023-09-24
Same message in FULL INDI and INDI Simple

dewagter:
- [x] https://github.com/paparazzi/pprzlink/pull/177


gautierhattenberger:
Looks good. Waiting for corrections, merging and ref update for pprzlink.

EwoudSmeur:
Thanks for these efforts, great improvement!

Actually, now that I'm thinking about it, maybe we can take this one step further, and fully merge the files. INDI simple is a special case of full INDI with an identity control effectiveness matrix, if we send the output to the motor mixing via the stabilization commands. I'm fine with merging this and making the other change later.

EwoudSmeur (approved):
See my previous comment

dewagter:
```make test``` all OK

PR #3105, opened 2023-09-25
[gps] add a timeout on gps fix lost

gautierhattenberger:
We already have a timeout when messages are not received anymore. This adds a possible timeout when GPS fix is lost (but we still have messages). It can prevent to switch to failsafe at the first gps lost message.
Disabled by default to keep the same behavior than before, but tested during real flights.

### DIFF
--- a/sw/airborne/modules/gps/gps.c
+++ b/sw/airborne/modules/gps/gps.c
@@ -277,6 +277,17 @@ void gps_periodic_check(struct GpsState *gps_s)
 #endif
 }
 
+bool gps_fix_valid(void)
+{
+  bool gps_3d_timeout_valid = false;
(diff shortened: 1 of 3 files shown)


dewagter (approved):
A 2-second timeout sounds like a good idea. I guess it can still be seen in the GPS telemetry that the accuracy gets worse.

gautierhattenberger:
I have implemented this because for some unknown reasons, our ublox modules were giving no_fix spikes (with good signal levels). I suspect a side effect of non ideal config (a mix of PVT and other DGPS messages) and maybe a bug in the driver. In any case, it feels safer to have more than one no_fix confirmation before starting the failsafe modes.

PR #3106, opened 2023-09-25
[ins] register new message and add ins_reset for EKF2

gautierhattenberger:
- Register `AHRS_QUAT_INT` message for telemetry
- Implement the `ins_reset_local_origin` function


dewagter (commented):
Sending the quaternion is great.
@fvantienen / @EwoudSmeur: wasn't there an issue with ```ekf.setEkfGlobalOrigin``` or was it only in z?

dewagter (approved):

PR #3108, opened 2023-09-26
[nav] small improvements to rotorcaft nav

gautierhattenberger:
- macro for oval count
- set in_flight to true during takeoff with module
- specify the height of hybrid survey in a parameter


gautierhattenberger:
I'll make a PR for the IMAV flight plans using this later.

PR #3115, opened 2023-09-27
[wls] make WLS lib, add support for guidance_indi_hybrid

gautierhattenberger:
Make a reentrant lib from WLS and a module so it can be reused by several controllers.

Use it as an option for guidance indi hybrid.

This version is only tested in simulation, but WLS for guidance with duplicated code was used in flight during IMAV.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
@@ -134,6 +134,7 @@ float inv_eff[4];
 
 // Max bank angle in radians
 float guidance_indi_max_bank = GUIDANCE_H_MAX_BANK;
+float guidance_indi_min_pitch = GUIDANCE_INDI_MIN_PITCH;
 
 /** state eulers in zxy order */
 struct FloatEulers eulers_zxy;
@@ -147,9 +148,26 @@ Butterworth2LowPass accely_filt;
 
 struct FloatVect2 desired_airspeed;
 
-struct FloatMat33 Ga;
-struct FloatMat33 Ga_inv;
+float Ga[3][3];
 struct FloatVect3 euler_cmd;
+
+#if GUIDANCE_INDI_HYBRID_USE_WLS
+#include "math/wls/wls_alloc.h"
+float du_min_gih[3];
+float du_max_gih[3];
+float du_pref_gih[3];
+float *Bwls_gih[3];
+#ifdef GUIDANCE_INDI_HYBRID_WLS_PRIORITIES
+float Wv_gih[3] = GUIDANCE_INDI_HYBRID_WLS_PRIORITIES;
+#else
+float Wv_gih[3] = { 100.f, 100.f, 1.f }; // X,Y accel, Z accel
+#endif
+#ifdef GUIDANCE_INDI_HYBRID_WLS_WU
+float Wu_gih[3] = GUIDANCE_INDI_HYBRID_WLS_WU;
+#else
+float Wu_gih[3] = { 1.f, 1.f, 1.f };
+#endif
+#endif
 
 float filter_cutoff = GUIDANCE_INDI_FILTER_CUTOFF;
 
@@ -168,7 +186,7 @@ struct FloatVect3 indi_vel_sp = {0.0, 0.0, 0.0};
 float time_of_vel_sp = 0.0;
 
 void guidance_indi_propagate_filters(void);
-static void guidance_indi_calcg_wing(struct FloatMat33 *Gmat);
+static void guidance_indi_calcg_wing(float Gmat[3][3]);
 static float guidance_indi_get_liftd(float pitch, float theta);
 
 #if PERIODIC_TELEMETRY
@@ -208,6 +226,12 @@ void guidance_indi_init(void)
   init_butterworth_2_low_pass(&pitch_filt, tau, sample_time, 0.0);
   init_butterworth_2_low_pass(&thrust_filt, tau, sample_time, 0.0);
   init_butterworth_2_low_pass(&accely_filt, tau, sample_time, 0.0);
+
+#if GUIDANCE_INDI_HYBRID_USE_WLS
+  for (int8_t i = 0; i < 3; i++) {
+    Bwls_gih[i] = Ga[i];
+  }
+#endif
 
 #if PERIODIC_TELEMETRY
   register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_GUIDANCE_INDI_HYBRID, send_guidance_indi_hybrid);
@@ -281,11 +305,6 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
   sp_accel.z = -(radio_control.values[RADIO_THROTTLE]-4500)*8.0/9600.0;
 #endif
 
-  // Calculate matrix of partial derivatives
-  guidance_indi_calcg_wing(&Ga);
-  // Invert this matrix
-  MAT33_INV(Ga_inv, Ga);
-
   struct FloatVect3 accel_filt;
   accel_filt.x = filt_accel_ned[0].o[0];
   accel_filt.y = filt_accel_ned[1].o[0];
@@ -309,8 +328,41 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
 #endif
 #endif
 
-  //Calculate roll,pitch and thrust command
-  MAT33_VECT3_MUL(euler_cmd, Ga_inv, a_diff);
+  // Calculate matrix of partial derivatives
+  guidance_indi_calcg_wing(Ga);
+
+#if GUIDANCE_INDI_HYBRID_USE_WLS
+  // Set lower limits
+  du_min_gih[0] = -guidance_indi_max_bank - roll_filt.o[0]; // roll
+  du_min_gih[1] = RadOfDeg(guidance_indi_min_pitch) - pitch_filt.o[0]; // pitch
+  du_min_gih[2] = (MAX_PPRZ - actuator_state_filt_vect[0]) * g1g2[3][0] + (MAX_PPRZ - actuator_state_filt_vect[1]) * g1g2[3][1] + (MAX_PPRZ - actuator_state_filt_vect[2]) * g1g2[3][2] + (MAX_PPRZ - actuator_state_filt_vect[3]) * g1g2[3][3];
+
+  // Set upper limits limits
+  du_max_gih[0] = guidance_indi_max_bank - roll_filt.o[0]; //roll
+  du_max_gih[1] = RadOfDeg(GUIDANCE_INDI_MAX_PITCH) - pitch_filt.o[0]; // pitch
+  du_max_gih[2] = -(actuator_state_filt_vect[0]*g1g2[3][0] + actuator_state_filt_vect[1]*g1g2[3][1] + actuator_state_filt_vect[2]*g1g2[3][2] + actuator_state_filt_vect[3]*g1g2[3][3]);
+
+  // Set prefered states
+  du_pref_gih[0] = 0.; // prefered delta roll angle
+  du_pref_gih[1] = -pitch_filt.o[0]; // prefered delta pitch angle
+  du_pref_gih[2] = du_max_gih[2];
+
+  float v_gih[3] = { a_diff.x, a_diff.y, a_diff.z };
+  float du_gih[3];
+  int num_iter UNUSED = wls_alloc(
+      du_gih, 3, v_gih, 3, du_min_gih, du_max_gih,
+      Bwls_gih, 0, 0, Wv_gih, Wu_gih, du_pref_gih, 100000, 10);
+  euler_cmd.x = du_gih[0];
+  euler_cmd.y = du_gih[1];
+  euler_cmd.z = du_gih[2];
+
+#else
+  // compute inverse matrix of Ga
+  float Ga_inv[3][3] = {};
+  float_mat_inv_3d(Ga_inv, Ga);
+  // Calculate roll,pitch and thrust command
+  float_mat3_mult(&euler_cmd, Ga_inv, a_diff);
+#endif
 
   AbiSendMsgTHRUST(THRUST_INCREMENT_ID, euler_cmd.z);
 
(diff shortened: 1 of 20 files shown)


dewagter:
Oh no... our next PR is super similar (but incompatible). In other words: "good idea" ...

gautierhattenberger:
Regarding the WLS part, I think my implementation is better since it makes the code fully reentrant. I don't even understand why it was not the case from the start. There is no point in keeping the global define for the vectors' size, these vectors are not global variables anyway.

For the rest, with the quadplane config, shouldn't you set all of these from the airframe file ? It will be a pain in the future if all specific configs are hardcoded in the source code.

sw/airborne/math/wls/wls_alloc.c
```diff
@@ -107,44 +102,44 @@ void qr_solve_wrapper(int m, int n, float** A, float* b, float* x) {
...
   if(!imax) imax = 100;
 
-  int n_c = CA_N_C;
-  int n_u = CA_N_U;
-  int n_v = CA_N_V;
+  int n_c = n_u + n_v;
 
-  float A[CA_N_C][CA_N_U];
-  float A_free[CA_N_C][CA_N_U];
+  float A[n_c][n_u];
```
dewagter, line 118: Does this still work with older compilers?

dewagter (commented):
- WLS as submodule and inverse to math are great.
 - Hiding the printf is important.
 - More parameters to airframes instead of hardcoding is great.
 - Dynamic memory allocation ```float A[n_c][n_u];``` does this work with all used compilers? Is it as fast? We proposed to statically allocate something large enough for all loops.
 - you need to include ```g1g2``` from the stabilization?

sw/airborne/math/wls/wls_alloc.c
EwoudSmeur: Not sure, but the question is if you want this in the first place. This creates a Variable Length Array, which creates a ton of overhead in the code to manage the pointers to the memory. Though possible, It has been called 'actively stupid' by some 😄 
@gautierhattenberger This was changed in #2523

Dennis-Wijngaarden:
I would like to further comment in addition to @dewagter. We indeed prepared a PR implementing the WLS in the guidance_indi_hybrid outer loop code. This was mainly done to implement quadplane types of drones having an over-actuated outer loop (to increase speed, you can pitch or use a pusher propeller). Therefore a WLS is needed to assign weights to this specific inputs and outputs. 

Also when using WLS in the outer loop in the PR @dewagter mentioned,  the control objective v is converted to body axis such that with weights more priority can be given to (air)speed instead of pitch for example.

Another remark I have on how the du_max_gih[2] and du_min_gih[2] are defined is that actuators are hardcoded here. This should depend on the type of drone and configuration you are flying instead of hardcoding it. It uses the actuator_state_filt_vect  and g1g2 variables which makes the outer loop wls dependent on stabilization_indi as well.

sw/airborne/math/wls/wls_alloc.c
gautierhattenberger: At least it compiles with gcc-arm-none-eabi 15:9-2019-q4-0ubuntu1 from Ubuntu 20.04, which is the oldest version officially supported. In the original code, only the size is static, right ? Maybe it helps for compiler optimization, but you still need to allocate on the stack at each call, so I'm not sure it makes a big difference.

gautierhattenberger: @EwoudSmeur hum, I forgot about these changes. Then, it can be restored and be part of the wls module configuration. The user will have to make sure the matrix is big enough for all cases for a given configuration.
Did you make performance tests back then ?

gautierhattenberger:
@Dennis-Wijngaarden For the part in guidance_hybrid, it is just a copy-paste from a branch that Ewoud gave me. I agree this is far from a good way to implement this. First I can add more options for the configuration (e.g. min / max, etc). And for the rest, I open to suggestion. I can also reopen the PR from the paparazzi repo so you can also push changes to it.

sw/airborne/math/wls/wls_alloc.c
EwoudSmeur: I did not do explicit performance checks on this code, but after someone explained to me why VLAs create extra code I had a look at the machine code that is generated from a simple program with a VLA. It explodes quite quickly even for simple operations, The problem is not that the memory needs to be allocated, but that at compile time it is not clear where the memory is going to be. All kinds of clever pointer magic has to be done to keep track, not only of the VLA memory itself, but all the memory that comes after.

gautierhattenberger:
I have restored the fixed size arrays. Concerning the definition of matrices, I think the question is still open. See also #3117 .

sw/airborne/math/wls/wls_alloc.c
```diff
@@ -107,44 +110,46 @@ void qr_solve_wrapper(int m, int n, float** A, float* b, float* x) {
  * control vector (sqare root of gamma)
  * @param imax Max number of iterations
  *
- * @return Number of iterations
+ * @return Number of iterations, -1 upon failure
```
dewagter, line 113:
```suggestion
 * @return Number of iterations, negative number of iterations upon failure
```

dewagter, line 113:
```suggestion
 * @return Number of iterations, negative number of iterations upon failure. If failure is due to max-iterations, the solution is still valid. If it failed for another reason (e.g. min > max), then the solution is invalid.
```

sw/airborne/math/wls/wls_alloc.h
```diff
@@ -56,6 +48,7 @@ void qr_solve_wrapper(int m, int n, float** A, float* b, float* x);
  *
  * @return Number of iterations, -1 upon failure
```
dewagter, line 49:
```suggestion
 * @return Number of iterations, negative number of iterations upon failure. If failure is due to max-iterations, the solution is still valid. If it failed for another reason (e.g. min > max), then the solution is invalid.
```

dewagter (dismissed):

Dennis-Wijngaarden:
I was thinking about making the guidance_indi_hybrid more flexible to enable flying fixed wings, tailsitters and quadplanes within the same guidance indi hybrid code.

First of all, I think moving the wls to modules is a good idea.

Secondly, each type of drone I mentioned before needs a slightly different effectiveness matrix in the guidance with a different rotation order. Can't we maybe define some often used matrices (tailistters, fixed-wing, quadplane) in a file such that the user can select one of those matrices or create its own one if the drone is unconventional. I'm willing to help setting this up.

sw/airborne/math/wls/wls_alloc.c
```diff
@@ -311,11 +315,11 @@ int wls_alloc(float* u, float* v, float* umin, float* umax, float** B,
     }
   }
   // solution failed, return negative one to indicate failure
-  return iter;
+  return -1;
```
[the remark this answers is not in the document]
EwoudSmeur: I think we should just return the number of iterations here. The comment is misleading, the algorithm is just out of iterations.

sw/airborne/math/wls/wls_alloc.c
```diff
@@ -45,22 +44,26 @@
...
+#if WLS_VERBOSE
+#include <stdio.h>
+static void print_final_values(int n_u, int n_v, float* u, float** B, float* v, float* umin, float* umax);
+static void print_in_and_outputs(int n_c, int n_free, float** A_free_ptr, float* d, float* p_free);
+#endif
 
-// Problem size needs to be predefined to avoid having to use VLAs
-#ifndef CA_N_V
-#error CA_N_V needs to be defined!
+#ifndef WLS_N_U
```
EwoudSmeur, line 58: I think we should keep the comment about VLAs, for the next time we forget why we did it this way.

I think removing the error if N_V or N_U is not defined will lead to problems, someone can forget and try to fly an octocopter with this.

Also, I think it is now a little too easy to make a mistake with N_U or N_V, because you have to check yourself (a) weather you use both stab and guidance WLS (b) which one is larger. Can we not force the user to explicitly define both, and then we pick the largest one?

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -309,8 +328,41 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
...
-  //Calculate roll,pitch and thrust command
-  MAT33_VECT3_MUL(euler_cmd, Ga_inv, a_diff);
+  // Calculate matrix of partial derivatives
+  guidance_indi_calcg_wing(Ga);
+
+#if GUIDANCE_INDI_HYBRID_USE_WLS
+  // Set lower limits
+  du_min_gih[0] = -guidance_indi_max_bank - roll_filt.o[0]; // roll
+  du_min_gih[1] = RadOfDeg(guidance_indi_min_pitch) - pitch_filt.o[0]; // pitch
+  du_min_gih[2] = (MAX_PPRZ - actuator_state_filt_vect[0]) * g1g2[3][0] + (MAX_PPRZ - actuator_state_filt_vect[1]) * g1g2[3][1] + (MAX_PPRZ - actuator_state_filt_vect[2]) * g1g2[3][2] + (MAX_PPRZ - actuator_state_filt_vect[3]) * g1g2[3][3];
```
EwoudSmeur, line 338: This will only work for quads, this should loop through all actuators instead.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -309,8 +328,41 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
...
+
+  // Set upper limits limits
```
EwoudSmeur, line 340:
```suggestion
  // Set upper limits
```

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -309,8 +328,41 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
...
+  du_max_gih[0] = guidance_indi_max_bank - roll_filt.o[0]; //roll
+  du_max_gih[1] = RadOfDeg(GUIDANCE_INDI_MAX_PITCH) - pitch_filt.o[0]; // pitch
+  du_max_gih[2] = -(actuator_state_filt_vect[0]*g1g2[3][0] + actuator_state_filt_vect[1]*g1g2[3][1] + actuator_state_filt_vect[2]*g1g2[3][2] + actuator_state_filt_vect[3]*g1g2[3][3]);
+
+  // Set prefered states
+  du_pref_gih[0] = 0.; // prefered delta roll angle
```
EwoudSmeur, line 346: I think it would be better to make this ```- roll_filt.o[0]```. By doing this we put some cost on rolling away from 0, which will be helpful if you have direct lateral thrust control. In that case and with the desired delta set to zero, you could drift to a case where you are rolling and compensating the roll with lateral thrust, with no encouragement to be energy efficient.

EwoudSmeur (changes requested):
Thanks for the PR Gautier! Please have a look at N_V and N_U, I think we should make it as fool-proof as possible

dewagter:
To continue rebasing our quadplanes, we need this PR to be resolved. I solved merge conflicts in https://github.com/paparazzi/paparazzi/tree/rebase_wls

sw/airborne/math/wls/wls_alloc.h
dewagter: Will be addressed in https://github.com/paparazzi/paparazzi/pull/3120

sw/airborne/math/wls/wls_alloc.c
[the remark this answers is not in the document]
dewagter: Addressed in https://github.com/paparazzi/paparazzi/pull/3120

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
dewagter: Will be addressed in the quadplane pull request.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
dewagter: Solved in https://github.com/paparazzi/paparazzi/tree/rebase_wls

gautierhattenberger:
@dewagter Thank you for implementing the corrections. I have merge your branch, fix conflicts with master and pushed to enac branch. It should be fine now.

sw/airborne/math/wls/wls_alloc.c
dewagter: I added size checks in stabilization and in guidance to be sure the matrix is big enough.

dewagter (approved):
Looks good. A few WLS comments are solved in the NaN PR. The sum of actuators is solved in the quadplane PR which will follow.

Ready to merge this one.

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -91,6 +91,15 @@ struct guidance_indi_hybrid_params gih_params = {
 #endif
 float guidance_indi_max_airspeed = GUIDANCE_INDI_MAX_AIRSPEED;
 
+#if 3 > WLS_N_U
```
dewagter, line 94:
```suggestion
// If using WLS, check that the matrix size is sufficient
#if GUIDANCE_INDI_HYBRID_USE_WLS
#if 3 > WLS_N_U
```

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -91,6 +91,15 @@ struct guidance_indi_hybrid_params gih_params = {
...
+#error Matrix-WLS_N_U too small: increase WLS_N_U in airframe file
+#endif
+
+#if 3 > WLS_N_V
+#error Matrix-WLS_N_V too small: increase WLS_N_V in airframe file
+#endif
```
dewagter, line 100:
```suggestion
#endif
#endif
```

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -85,6 +85,14 @@
 #warning SetAutoCommandsFromRC not used: STAB_INDI overwrites actuators
 #endif
 
+#if INDI_NUM_ACT > WLS_N_U
```
dewagter, line 88:
```suggestion
#if !STABILIZATION_INDI_ALLOCATION_PSEUDO_INVERSE
#if INDI_NUM_ACT > WLS_N_U
```

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -85,6 +85,14 @@
...
+#error Matrix-WLS_N_U too small: increase WLS_N_U in airframe file
+#endif
+
+#if INDI_OUTPUTS > WLS_N_V
+#error Matrix-WLS_N_V too small: increase WLS_N_V in airframe file
+#endif
```
dewagter, line 94:
```suggestion
#endif
#endif
```

dewagter (commented):
Only check matrix size if actually using WLS

dewagter (approved):

PR #3116, opened 2023-09-28
[indi] ABI thrust message to 3D vector

dewagter:
Upgrade to 3D thrust vector

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -89,7 +89,7 @@
     </message>
 
     <message name="THRUST" id="16">
-      <field name="thrust_increment" type="float" unit="m/s^2"/>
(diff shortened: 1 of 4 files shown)


gautierhattenberger (changes requested):
I'm all fine to update the thrust increment to vector and the stabilization part looks fine. But for guidance `guidance_indi_du` doesn't seem to be defined, and can't we fine a better way to set this vector than making specific cases with `#if` blocks ?
Also, normal `guidance_indi` should be updated for ABI message.

dewagter:
Done

gautierhattenberger (approved):

dewagter:
All tests completed.

PR #3117, opened 2023-09-28
[guidance_indi] lift effectiveness and set stabilization limits in WEAK

dewagter:
- [x] LIFT_D set by external module if not constant
 - [x] GUIDANCE_INDI_CLIMB_SPEED_FWD settings in airframe (will become a setting in a later pull request)
 - [x] stabilization_indi_set_wls_setting (is overridden for some platforms)

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -248,10 +248,6 @@
       <field name="vel_sp" type="struct FloatVect3 *" unit="m/s"/>
     </message>
 
-    <message name="LIFT_D" id="37">
-      <field name="liftd" type="float"/>
-    </message>
-
   </msg_class>
 
 </protocol>
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -248,6 +248,10 @@
       <field name="vel_sp" type="struct FloatVect3 *" unit="m/s"/>
     </message>
 
+    <message name="LIFT_D" id="37">
+      <field name="liftd" type="float"/>
+    </message>
+
   </msg_class>
 
 </protocol>
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
@@ -130,6 +130,17 @@ static void guidance_indi_filter_thrust(void);
 #endif
 #endif
 
+#ifndef GUIDANCE_INDI_CLIMB_SPEED_FWD
+#define GUIDANCE_INDI_CLIMB_SPEED_FWD 4.0
+#endif
+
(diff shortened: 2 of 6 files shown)


dewagter:
```make test_coverity``` and ```make test``` all OK.

gautierhattenberger:
Something I realized after working a bit with guidance hybrid, is that "hybrid" is really a bad name. It should be more like "something that is close enough from a tail-sitter". So we should either change the name (and make variations for future configurations) or we find a nice way to make it generic. For now, I can see some options:
- use ABI messages to ask some external module to feed the matrices
- make weak functions that must be implemented for each type
- initialized the controller with callbacks

And in addition, we need to deal with pseudo_inverse / wls configuration...

For now, I don't have a strong opinion. ABI message can be flexible way if it is done correctly, but the example of thrust increment between stab and guidance shows that it can make it hard to understand. Weak function seems the fast and easy way. I like callbacks as it is closer to a functional approach, but configuration might be painful. Also, I'm not sure that a solution mixing several options will make things very clear.

dewagter:
- The difficulty with weak functions is that it might get confusing which weak function is being used and which one is overridden.
 - ABI is good for sensor data, but to fill a different matrix in an outerloop, that is not as great.

What is your idea for the callbacks: ```stabilization->get_wls_limits()``` becomes a function pointer that can be implemented by a module?

dewagter:
Here it is with only WEAK functions.

EwoudSmeur:
> Something I realized after working a bit with guidance hybrid, is that "hybrid" is really a bad name. It should be more like "something that is close enough from a tail-sitter". So we should either change the name (and make variations for future configurations) or we find a nice way to make it generic. For now, I can see some options:

I think it is not that bad, as it can quite easily be adapted to fly other drones as well. We've been flying the rotating wing drone with it, which is certainly not a tailsitter. The main thing that is different is the control effectiveness matrix, which we may want to put outside of this file or define weak or something like that.

EwoudSmeur (approved):

gautierhattenberger:
@dewagter My idea with callbacks was something like: `guidance_indi_setup(get_eff_cb)`, where `get_eff_cb` is function pointer provided by an appropriate module that depends on the type of aircraft you are using. Maybe, several functions are required. Even the part where you actually compute the inverse matrix could be a function configured at startup, so it is the module that configures the airframe who choose to use pseudo-inverse, WLS or any other allocation methods. The only requirements is of course to define common prototypes for these functions.

dewagter:
Sound good. Having the functionality in separate (for now WEAK) functions is a required first step. If OK for you, I propose to first push the functions (pseudoinverse/WLS/WLS-quadplane/WLS-reconfigurable-quadplane) within #ifdefs as this code is already flight tested, and then develop a framework to add callbacks.

PR #3118, opened 2023-09-28
Bound function also bounds nan

EwoudSmeur:
We have a lot of ```Bound``` in the code, but this macro doesn't handle NaN values. In fact in many cases NaN values are not considered, even though they can be disastrous if they stick in integrators. It might not technically make sense to assign the max or min to a NaN value, but at least the output of the Bound should be within the given range, and not NaN in my opinion.

With this subtle change NaN will become max in the Bound function.

### DIFF
--- a/sw/include/std.h
+++ b/sw/include/std.h
@@ -135,8 +135,8 @@ typedef uint8_t unit_t;
 
 #define BoundUpper(_x, _max) { if (_x > (_max)) _x = (_max);}
 
-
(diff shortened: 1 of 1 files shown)


dewagter (dismissed):
Great idea.
 - minus NaN will now also be max
 - not 100% sur all compilers behave the same, but most say (NaN > 100) is false.

EwoudSmeur:
Perhaps we would rather prefer NaN became the min value, it seems safer in case of actuator bounds

dewagter (approved):

PR #3120, opened 2023-10-01
WLS update NaN

dewagter:
Upon testing more implementations, we found some conditions that can trigger NaN in the output of the WLS. This is solved with this PR.

(requires https://github.com/paparazzi/paparazzi/pull/3115 to be merged first)


dewagter:
- [x] Giving the matrix size is essential since stabilization and guidance use different sizes.
 - [x] ```n_free``` could sometimes get negative -> segfault
 - [x] only ```check limits``` for free actuators
 - [x] ```alpha``` should only scale down (so it initializes at 1)
 - [x] numerical rounding errors in float can cause scaled-back actuators to still exceed bounds: so bound after scaling back: this also means there is no more need for margins when testing actuator limits

dewagter:
Of course!

PR #3122, opened 2023-10-02
quadplane guidance

dewagter:


gautierhattenberger (approved):
I kind of agree with this implementation as a short term option so you can bring back all your code to master. However, we really need to do something to avoid blocks of optional code. This should be split into specialized modules with a common base if possible, and weak functions or callbacks to parametrize the controller.

dewagter:
Hi @gautierhattenberger,

Would it be much better if I already made a module now for the quadplane control with the 2 WEAK functions that are in ifdef's? Do note that the vast majority of the code is the same. 

To get an idea, the next PR will add a morphing quadplane in a module https://github.com/paparazzi/paparazzi/pull/31240.

gautierhattenberger:
If you already have a third option option in a module, maybe it would have been better to put the quadplane also in a module. The question is if we should keep the tailsitter config as a default, or make it an explicit module. I prefer the second option, but it means to avoid breaking several airframes, it should be at least the default `suggested` module.

PR #3125, opened 2023-10-03
[simulator] morphing quadplane

dewagter:


conf/simulator/jsbsim/aircraft/rotwing25.xml
```diff
@@ -0,0 +1,448 @@
...
+        <z>0</z>
+      </location>
+      <direction>
+        <x>1</x>
+        <y>0</y>
+        <z>0</z>
+      </direction>
+    </force>
+
+    <force name="back_motor_couple1" frame="BODY" unit="LBS">
```
gautierhattenberger, line 339: @dewagter @EwoudSmeur It is now possible to use `moment` directly, it makes the configuration easier. See https://github.com/paparazzi/paparazzi/blob/master/conf/simulator/jsbsim/aircraft/simple_quad_wind.xml#L214

EwoudSmeur: That is nice, though I think it was already possible inside the aerodynamics section like this:
https://github.com/paparazzi/paparazzi/blob/master/conf/simulator/jsbsim/aircraft/Systems/aerodynamics_cyfoam.xml#L4

PR #3126, opened 2023-10-03
Negative airspeeds from negative differential pressures.

dewagter:


gautierhattenberger:
Are we sure that there is no side effects in using a negative airspeed somewhere else in the code ? (e.g. in control)

dewagter:
I most controllers (e.g. fixedwing) you have something where it doesn't matter that can become negative: e.g.:
```float err_airspeed = v_ctl_auto_airspeed_setpoint - stateGetAirspeed_f();```

In ```rotorcraft``` it is sometimes used to compute the yaw rate for a given bank angle, but there are bounds:
```
float airspeed_turn = stateGetAirspeed_f();
Bound(airspeed_turn, 10.0f, 30.0f);
```

In some settings, it is used to change for instance the descending speed or the mode:
```if (airspeed < INDI_HROTTLE_LIMIT_AIRSPEED_FWD)```

In some of the ```ctrl_eff_sched_XXX``` the ```airspeed^2``` is used. I will add some ```Bound``` for now, but we could actually also measure the effectiveness with flow from the back as this could improve the control.

dewagter:
Double-checked all instances to ```stateGetAirspeed```. Most are already protected against dividing by zero with lower bounds or form simple comparisons that are just as valid. In a few instances when unsure, I added some protection.

Since we have quadplanes that can hover with a tailwind, this becomes very useful to measure. You can for instance automatically identify a tailwind before take-off.

dewagter:
I don't feel a lot of enthusiasm for this one. Any particular fears or things I should still check? Or Bounds to add?

 - in control loops with positive setpoints it does not make a difference
 - in scheduling with if ```airspeed < X``` is also does not make a difference
 - in hover and before take-off, (in particular for quadplanes) it is very helpful to know
 
We used to have negative airspeeds before in e.g. the Delftacopter, but it got bound to zero once the ```air_data``` module was made.

gautierhattenberger (approved):
It looks acceptable to try this. I'm not sure how to warn people for future code that it might be negative, in case it can be a problem (when doing a square root for instance).

dewagter:
Isn't protecting against division by zero and negative sqrt standard practice? That is why I did not need to change hardly any code except in some unsigned telemetry messages.

PR #3129, opened 2023-10-04
UAVCAN telemetry

dewagter:


sw/airborne/modules/actuators/actuators_uavcan.h
```diff
@@ -30,4 +30,6 @@ extern void actuators_uavcan_init(struct uavcan_iface_t *iface);
 extern void actuators_uavcan_commit(struct uavcan_iface_t *iface, int16_t *values, uint8_t nb);
 extern void actuators_uavcan_cmd_commit(struct uavcan_iface_t *iface, int16_t *values, uint8_t nb);
 
+extern float getActuatorPosition(int nr);
```
dewagter, line 33: Push via ABI instead!

dewagter:
@fvantienen do you want a different message/field for the servo positions?

fvantienen (approved):

fvantienen:
One note is that the rpm_sensor module is not functioning since we are missing the `get_pwm_input_period_in_usec` function. This was already before this PR and can be solved later when we want to use the sensor.

fvantienen (approved):

gautierhattenberger (approved):

PR #3131, opened 2023-10-05
[boards] orange cube changed some outputs to inputs

dewagter:
@fvantienen : this PR had to be done in a different way?


fvantienen:
Yes this is future work/problems but it has to be done different else we end up with a lot of different board files for the same board.

dewagter:
Let's close this one then.

PR #3134, opened 2023-10-06
[electrical]: extra ADC

dewagter:
@fvantienen is there a more generic way (to e.g. also add even more in the future)?


fvantienen (approved):
I think this is fine for now

PR #3141, opened 2023-10-16
[airframe] Airframe update, missing default telemetry & fix reading wing angle

dewagter:


gautierhattenberger (approved):

fvantienen:
Tested and works!

PR #3142, opened 2023-10-16
[imu] Add fast logging

fvantienen:
This adds the ability to log multiple IMU's at the highest speeds possible (32kHz, 8kHz, 1kHz etc.). It will log the accelerometer and gyroscope at the sampling frequencies onto the flightlogger SD card. The timestamps are not 100% correct, but if you average out the timestamp over time you will get the correct frequencies in the log.


dewagter (approved):
Small risk if enabled by default to flood the logging device.
Can it show a warning when compiling?

dewagter:
@fvantienen : can you try to keep the pull request number in the commit's title when merging? Tnx

PR #3143, opened 2023-10-18
[modules] wing-rotation controller with servo

dewagter:


dewagter:
How do you enable ```ADC_5``` in the ```test``` architecture?

fvantienen:
You should make `ADC_CHANNEL_WING_ROTATION_POSITION` a configure. Then you can do something similar to:
```
    <configure name="ADC_CHANNEL_WING_ROTATION_POSITION" default="ADC_5" case="lower|upper"/>
    <define name="ADC_CHANNEL_WING_ROTATION_POSITION" value="$(ADC_CHANNEL_WING_ROTATION_POSITION_UPPER)"/>
    <define name="USE_$(ADC_CHANNEL_WING_ROTATION_POSITION_UPPER)"/>
```

fvantienen:
That way the ADC is enabled by itself instead of doing in the airframe

fvantienen:
Btw the naming of these defines etc. could be a lot better. Usually it is best to start with the module name, such that they are all grouped together

sw/airborne/modules/rot_wing_drone/wing_rotation_controller_servo.c
```diff
@@ -0,0 +1,254 @@
...
+
+void wing_rotation_init(void)
+{
+  // your init code here
+#if !USE_NPS
+  adc_buf_channel(ADC_CHANNEL_WING_ROTATION_POSITION, &buf_wing_rot_pos, ADC_CHANNEL_WING_ROTATION_POSITION_NB_SAMPLES);
+#endif
+
+  // Init wing_rotation_controller struct
+  wing_rotation.servo_pprz_cmd = 0;
```
fvantienen, line 95: Just do `wing_rotation = {0};` after that just set the variables which are not `0` or `false`

sw/airborne/modules/rot_wing_drone/wing_rotation_controller_servo.c
```diff
@@ -0,0 +1,254 @@
...
+  wing_rotation.adc_wing_rotation = 0;
+  wing_rotation.wing_angle_rad = 0;
```
fvantienen, line 97: Is it really needed to have everything both in rad and deg?

sw/airborne/modules/rot_wing_drone/wing_rotation_controller_servo.h
```diff
@@ -0,0 +1,61 @@
...
+#include "std.h"
+
+extern void wing_rotation_init(void);
+extern void wing_rotation_periodic(void);
+extern void wing_rotation_event(void);
+extern bool set_wing_rotation_scheduler(bool rotation_scheduler_on);
+extern bool set_wing_rotation_scheduler_nav(bool rotation_scheduler_on);
+
+// Paramaters
+struct wing_rotation_controller {
```
fvantienen, line 38: This struct seems to contain way too many variables which are not needed.

fvantienen (changes requested):

sw/airborne/modules/rot_wing_drone/wing_rotation_controller_servo.h
dewagter: removed unneeded: all the rest is used.

dewagter:
@fvantienen update complete

gautierhattenberger:
is is still a draft ?

dewagter:
Working on it. Want to convert it into an actuator_soft_servo if you agree.

dewagter:
Cleanup complete.

gautierhattenberger (approved):

PR #3144, opened 2023-10-18
[modules] Fix error in Rotating Wing message

fvantienen:
This is fixing memory errors and makes it possible to run the simulator.


dewagter (approved):
Absolutely!

PR #3145, opened 2023-10-25
Support for rovers and third order Bézier splines in GVF parametric

alfredoFBW:
1. Support for rovers in GVF parametric added.
2. Support for third order Bézier splines in GVF parametric added.
3. GVF and GVF parametric can now be used in the same vehicle via gvf_common. Created structures for curvature and orientation error in gvf_common.
4. Original GVF parametric telemetry message size has not been changed. For that purpose telemetry of Bézier data is sent piecewise, because some radio modules do not allow for relatively big data frames.

### DIFF
--- a/sw/airborne/modules/guidance/gvf_common.h
+++ b/sw/airborne/modules/guidance/gvf_common.h
@@ -17,37 +17,37 @@
  * along with paparazzi; see the file COPYING.  If not, see
  * <http://www.gnu.org/licenses/>.
  */
- 
- #ifndef GVF_COMMON_H
- #define GVF_COMMON_H
- 
- // uint32_t
+
+#ifndef GVF_COMMON_H
+#define GVF_COMMON_H
+
+// uint32_t
 #include "std.h"
 
- /** @typedef gvf_common_omega
+/** @typedef gvf_common_omega
 * @brief Horizontal control signal for both gvf
 * @param omega is the horizontal control signal
 */
 
- typedef struct{
- 	float omega;
- } gvf_common_omega;
- 
+typedef struct {
+  float omega;
+} gvf_common_omega;
+
 extern gvf_common_omega gvf_c_omega;
 
- /** @typedef gvf_common_params
+/** @typedef gvf_common_params
 * @brief Different parameters obtained from gvfs. dot means d/dt
 * @param kappa is the curve's curvature
 * @param ori_err is the orientation error
 */
-typedef struct{
-	float kappa;
-	float kappa_dot;
-	float ori_err;
-	float ori_err_dot;
+typedef struct {
+  float kappa;
+  float kappa_dot;
+  float ori_err;
+  float ori_err_dot;
 } gvf_common_params;
 
 extern gvf_common_params gvf_c_info;
- 
- #endif // GVF_COMMON_H
+
+#endif // GVF_COMMON_H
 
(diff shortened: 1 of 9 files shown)


sw/airborne/modules/guidance/gvf_common.h
```diff
@@ -0,0 +1,53 @@
...
+ // uint32_t
+#include "std.h"
+
+ /** @typedef gvf_common_omega
+* @brief Horizontal control signal for both gvf
+* @param omega is the horizontal control signal
+*/
+
+ typedef struct{
+ 	float omega;
```
gautierhattenberger, line 33: We use 2 spaces for indentation. You can use the script `fix_code_style.sh` in root folder or refer to https://docs.paparazziuav.org/latest/styleguide.html

sw/airborne/modules/guidance/gvf_parametric/gvf_parametric.cpp
```diff
@@ -345,14 +361,85 @@ bool gvf_parametric_2D_trefoil_wp(uint8_t wp, float w1, float w2, float ratio, f
 { 
   gvf_parametric_trajectory.p_parametric[7] = wp;
   gvf_parametric_plen_wps = 1;
-
+  #if defined(FIXEDWING_FIRMWARE)
   gvf_parametric_2D_trefoil_XY(waypoints[wp].x, waypoints[wp].y, w1, w2, ratio, r, alpha);
+  #elif defined(ROVER_FIRMWARE)
+  gvf_parametric_2D_trefoil_XY(WaypointX(wp), WaypointY(wp), w1, w2, ratio, r, alpha);
+  #endif
```
gautierhattenberger, line 368: The macros `WaypointX` and `WaypointY` exist for all firmwares, so it should work with `gvf_parametric_2D_trefoil_XY(WaypointX(wp), WaypointY(wp), w1, w2, ratio, r, alpha);` in all cases.

sw/airborne/modules/guidance/gvf_parametric/gvf_parametric.cpp
```diff
@@ -345,14 +361,85 @@ bool gvf_parametric_2D_trefoil_wp(uint8_t wp, float w1, float w2, float ratio, f
...
+	gvf_parametric_trajectory.type = BEZIER_2D;
+	float fx, fy, fxd, fyd, fxdd, fydd;
+	gvf_parametric_2d_bezier_splines_info(gvf_bezier_2D, &fx, &fy, &fxd, &fyd, &fxdd, &fydd);
+	gvf_parametric_control_2D(gvf_parametric_2d_bezier_par.kx, gvf_parametric_2d_bezier_par.ky, fx, fy, fxd, fyd, fxdd, fydd);
+	return true;
+}
+
+// TODO: Improve scalability (pass an array of wp) ??
+bool gvf_parametric_2D_bezier_wp(uint8_t wp0, uint8_t wp1, uint8_t wp2, uint8_t wp3, uint8_t wp4, uint8_t wp5, uint8_t wp6, uint8_t wp7, uint8_t wp8, uint8_t wp9,
+				  uint8_t wp10, uint8_t wp11, uint8_t wp12)
```
gautierhattenberger, line 384: This function is not very convenient to use. I suggest to pass only 2 parameters: the id of the first WP and the number of WP. It means that all WPs used for bezier curves must be defined in the correct order in the flight plan, but this is acceptable constraint (already used for other navigation routines like https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/nav/nav_survey_poly_rotorcraft.h#L45).

gautierhattenberger (changes requested):

sw/airborne/modules/guidance/gvf_parametric/gvf_parametric.cpp
alfredoFBW: Since the number of points is fixed by the number of Bézier segments, I've only passed the first point of the spline and then the user shall define in the airframe file the number of segments. Thanks for the help, using this new way improves scalability a lot.

gautierhattenberger (approved):

PR #3148, opened 2023-10-26
[nav] use a low speed for goto with hybrid

gautierhattenberger:
@dewagter @EwoudSmeur @fvantienen This is a proposal to use different speeds for `goto` and `route` to prevent unnecessary transition to forward flight when just going to a waypoint. In most flight plan, the goto is used for small displacement, while using a route is more suitable for longer trajectories where forward flight is useful.
The `force_forward` is still possible.

### DIFF
--- a/sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
+++ b/sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
@@ -35,9 +35,10 @@
 #define NAV_MAX_SPEED (GUIDANCE_INDI_MAX_AIRSPEED + GUIDANCE_INDI_NAV_SPEED_MARGIN)
 float nav_max_speed = NAV_MAX_SPEED;
 
-// Max ground speed in hover mode (goto/stay)
-#ifndef GUIDANCE_INDI_HOVER_SPEED
-#define GUIDANCE_INDI_HOVER_SPEED 5.f
+// Max ground speed in with goto/stay instruction
+// by default, same as route speed
+#ifndef GUIDANCE_INDI_GOTO_SPEED
+#define GUIDANCE_INDI_GOTO_SPEED NAV_MAX_SPEED
 #endif
 float nav_hover_speed = GUIDANCE_INDI_HOVER_SPEED;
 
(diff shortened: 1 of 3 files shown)


EwoudSmeur:
I am not in favor of making this the default behavior. Like you say, if the goto is used only for a small displacement, depending on the position gain the drone will not go to forward flight. If the distance is larger, I see no reason to limit the commanded ground speed. Note that if there is wind, the drone may even be in forward flight regardless.
The other changes seem fine to me!

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -225,7 +232,7 @@ static void nav_hybrid_circle(struct EnuCoor_f *wp_center, float radius)
     } else {
       // close to circle, speed function of radius for a feasible turn
       // MAX_BANK / 2 gives some margins for the turns
```
EwoudSmeur, line 234:
```suggestion
      // MAX_BANK * 0.8 gives some margins for the turns
```

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -35,6 +35,12 @@
 #define NAV_MAX_SPEED (GUIDANCE_INDI_MAX_AIRSPEED + GUIDANCE_INDI_NAV_SPEED_MARGIN)
 float nav_max_speed = NAV_MAX_SPEED;
 
+// Max ground speed in hover mode (goto/stay)
+#ifndef GUIDANCE_INDI_HOVER_SPEED
+#define GUIDANCE_INDI_HOVER_SPEED 5.f
```
EwoudSmeur, line 40:
```suggestion
#define GUIDANCE_INDI_HOVER_SPEED NAV_MAX_SPEED
```
I'm not fully convinced with the name 'hover_speed'.

EwoudSmeur (changes requested):

gautierhattenberger:
@EwoudSmeur I changed the name, made the max speed the default as you suggested, and add some doc

dewagter (approved):
I'm OK with this extra option. The default when ```GUIDANCE_INDI_GOTO_SPEED``` is not defined remains the same so that is great.

EwoudSmeur (approved):
Great, thanks!

PR #3151, opened 2023-10-31
fix bug that gave zero effectiveness matrix

EwoudSmeur:
Since #3122 the calculation of the control effectiveness matrix in the guidance was removed. This adds it back by default. 

https://github.com/paparazzi/paparazzi/blame/4ae40567ed96b71c28f258ce1bb4af6315b74a6c/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c#L843


gautierhattenberger (approved):
I'm fine to have the "regular" hybrid code as default, but I think that having 2 weak functions protected with `#if`/`"else` is a bit strange. Quadplane specific function should come from a dedicated module. It can be for an other PR thought.

EwoudSmeur:
You're right. Let's merge this now so that the code works again, then it should be improved as you suggest.

PR #3152, opened 2023-11-01
[imu] Fix Invensense V2 accelerometer DLPF config

fvantienen:
The table for the Invensense V2 accelerometer was shifter by one, since it has two options to set 265Hz.
Screenshot 2023-11-01 at 10 47 35


gautierhattenberger:
The gyro dlpf config also have a minus one, while 0 seems a valid configuration value. Is is correct ?

dewagter (approved):
This PR improves things -> LGTM
But why only the ACC?

fvantienen:
Because with the gyro this is correct. Since we added the `INVENSENSE2_GYRO_DLPF_OFF` (https://github.com/paparazzi/paparazzi/blob/04c67e373518dd7238caf4b5fec3937dc7d014d3/sw/airborne/peripherals/invensense2.h#L81) at the start it needs to subtract `1`. But because the accelerometer has two 265Hz options at the start this doesn't need to happen there.

Issue #3091, opened 2023-09-21
Guidance in NED frame does not work for long distances


EwoudSmeur:
### The problem
We are used to doing the guidance (in paparazzi also called navigation) in the NED frame. This causes issues when flying long distance due to the fact that the earth is not flat:
- When flying far away from the NED origin, the NED frame is tilted with respect to the ground, which means that down is no longer really down. This leads to horizontal paths in NED becoming inclined in the local tangent plane of the drone's location.
- Due to the above issue, the attitude coming from stateGetNedToBodyQuat is either not what the name says it is, or it does not give the attitude wrt. the local tangent plane.
- The waypoints that are defined in LLA in the flight plan, are correctly converted to the NED frame. However, this means that if a drone is sent to a waypoint far away at the same altitude, it is first going to descend to the right z coordinate in the NED frame, before it covers any significant distance horizontally, and will climb slowly to the right altitude, keeping the same NED z coordinate.
- Guidance in a NED frame far away leads to other problems as well: two points perfectly north of each other, have both a north and east difference in a NED frame at a different longitude. Due to this, I think a drone will not fly the shortest path between two points.

Note that additionally, there is a mismatch with some of the INS systems: EKF2 takes hmsl from the GPS and puts it in the z coordinate, so then the state position is actually not in proper NED.

### The proposed solution
Though the NED coordinates can be intuitive when flying in close proximity to the takeoff location, it doesn't make sense to use these for guidance on a larger scale. I propose to modify the way the guidance works. Instead of strictly adhering to the NED frame, we should calculate North and East errors based on the local tangent plane of the current position of the drone. The vertical error should be based on height above the geoid. This way, you can fly to a waypoint far away with the same geoid altitude, and nicely follow the earth curvature.
Instead of using the NED coordinates of waypoints to compute errors for the guidance, we can do the following procedure:
- Obtain the LLA coordinate of a waypoint.
- Obtain the LTP of the current drone position.
- Convert the waypoint LLA to this LTP frame.
- Use the North and East position directly as error for the guidance.

### Required changes
I quickly hacked this in and it works in simulation. But I found there are a lot of things to change, as the NED coordinates are used everywhere, most notably in how waypoint coordinates are shared, in the guidance reference system and in navigation (which should be called guidance) functions.

Though a big change, I think it is better to solve the issue in a proper way than to put some bandage (e.g. moving the NED origin). We can then also make it a bit more clear what altitude definitions we use e.g. in the flight plan. 

[...]

PR #3153, opened 2023-11-01
Long distance altitude control

EwoudSmeur:
Following up on #3091, I made some code to control the ellipsoid altitude, and obtain position errors in a local NED frame centered at the drone (not at the origin), because if the origin is far away from the drone it will be tilted wrt the earth surface. I only changed the hybrid code so far.

Now though it is somewhat straightforward to calculate the horizontal position errors and the altitude, this has some implications:
- Position reference positive up (following the ellipsoid alt definition)
- The horizontal reference model will become more complicated, as the position reference can no longer be a NED coordinate. 
- I am not sure what that would mean for nav routines such as the circle.

An alternative could be to move the NED origin every kilometer or so and do the vertical control in altitude above ellipsoid, but keep the horizontal control in NED. Every time the origin is moved, the waypoints and reference models need to be reset. This could lead to some difficulty with the GCS, as it needs to be communicated over the datalink that the origin has moved. It will also lead to jumps in all the NED signals.

Another issue is that EKF2 estimates altitude in the NED frame, but puts the altitude above the ellipsoid there. If you get the current LLA altitude from the state, it will be too low if the drone is not close to the origin.

Any input is welcome! @gautierhattenberger @dewagter @noether

### DIFF
--- a/conf/flight_plans/tudelft/nederdrone_valkenburg.xml
+++ b/conf/flight_plans/tudelft/nederdrone_valkenburg.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
 
-<flight_plan alt="60" ground_alt="0" lat0="52.168595" lon0="4.412444" max_dist_from_home="5000" name="Nederdrone Valkenburg" security_height="2">
+<flight_plan alt="60" ground_alt="0" lat0="52.168595" lon0="4.412444" max_dist_from_home="500000" name="Nederdrone Valkenburg" security_height="2">
   <header>
     #include "autopilot.h"
     #include "modules/datalink/datalink.h"
@@ -8,7 +8,8 @@
   <waypoints>
     <waypoint name="HOME" x="12.6" y="-48.7"/>
     <waypoint name="CLIMB" x="62.6" y="-80.4"/>
-    <waypoint name="STDBY" x="106.0" y="-55.1"/>
+    <!-- <waypoint name="STDBY" x="106.0" y="-55.1"/> -->
+    <waypoint name="STDBY" lat="51.168595" lon="4.412444"/>
     <waypoint name="p1" x="322.0" y="-254.3"/>
     <waypoint name="p2" x="696.0" y="-148.1"/>
     <waypoint name="p3" x="600.4" y="73.4"/>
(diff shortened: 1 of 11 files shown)


noether:
Mmm, maybe I am wrong here because I do not remember all the "nav"s in Paparazzi.

When the GPS is on, do not we have MSL altitude? I mean, the problem is only in the XY because the error by the tangent plane to the sphere is noticeable, right?

noether:
Regardless of this hack :P. 

It looks like we should divide (add a tag or dependency?) between code/modules that support operations far from the initial and fixed ground station and local operations.

What do you think @gautierhattenberger ? Do you see it feasible? Or will it break most of the code? :P.

gautierhattenberger:
If you want to make a proper navigation on a non flat ground, you need to have your guidance/navigation in lla directly. Years ago, I tried a hack where the x/y coordinates and differences where computed using the Haversine formula that @EwoudSmeur have also used here. So the x/y where not orthonormal Cartesian coordinates anymore, and it kind of works. It was for fixedwing and before having the state interface. Now, I think it is a bit more complex.
Probably we need to redefine the different coordinates:
- LLA based on ellipsoid
- LTP tangent plane frame at LLA0/ECEF0
- NED something that always have Z pointing down and X/Y:
-

gautierhattenberger:
If you want to make a proper navigation on a non flat ground, you need to have your guidance/navigation in lla directly. Years ago, I tried a hack where the x/y coordinates and differences where computed using the Haversine formula that @EwoudSmeur have also used here. So the x/y where not orthonormal Cartesian coordinates anymore, and it kind of works. It was for fixedwing and before having the state interface. Now, I think it is a bit more complex.
Probably we need to redefine the different coordinates:
- LLA based on ellipsoid
- LTP tangent plane frame at LLA0/ECEF0
- NED something that always have Z pointing down and X/Y following the earth curvature; something else ?
- UTM certainly a bad idea, but for now required for fixedwing

Then the altitude to track: z in LTP, alt above ellipsoid or hmsl ? Fixedwing firmware is flying with hmsl. It seems to be the best choice to me, also compatible with EKF2, and probably easy to adapt to our filters if needed.

gautierhattenberger:
sorry, for close/reopen, wrong button

noether:
I can make it tomorrow at 11.00 too. At least, it is nice to see your faces again.

fvantienen:
Tomorrow at 11:00 is fine with me as well

PR #3154, opened 2023-11-01
[WLS] Fix error check and define orders.

Fabien-B:
- Explicitly include generated airframe
- Move error check


gautierhattenberger:
WLS should be independent of the algorithms using it. So you can't have a check for STABILIZATION_INDI in WLS. What was the reason to move this ?

Fabien-B:
Ok, I changed the PR.
WLS_N_U and WLS_N_V were  defined in the C file, so not accessible from an other file (stabilization_indi.c in my case).

dewagter (approved):
OK

PR #3156, opened 2023-11-03
Ground-detect-fix

dewagter:


EwoudSmeur (approved):
I'm fine with these changes, but I think we need to go towards a ground detect method that can take information from multiple sources in order to reliably detect the ground. 

I made a start with such a module in https://github.com/tudelft/paparazzi/blob/dronisos/sw/airborne/modules/nav/ground_detect.c
That could be extended to e.g. 3 sources of information, and a minimum of 2 need to be true to detect ground. What do you think?

dewagter:
> I'm fine with these changes, but I think we need to go towards a ground detect method that can take information from multiple sources in order to reliably detect the ground.
> 
> I made a start with such a module in https://github.com/tudelft/paparazzi/blob/dronisos/sw/airborne/modules/nav/ground_detect.c That could be extended to e.g. 3 sources of information, and a minimum of 2 need to be true to detect ground. What do you think?

I totally agree that merging more independent sources is required. This PR already solves some problems with existing code.

PR #3161, opened 2023-11-04
hybrid quadplanes and tailsitters explicit

dewagter:
As discussed in #3122, it was desired to reduce the number of ```#ifdef``` blocks. Here I propose to split the unique code for ```tailsitter``` and ```quadplane``` in separate modules that automatically include and configure ```hybrid```

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
@@ -1,0 +1,157 @@
+/*
+ * Copyright (C) 2015 Ewoud Smeur <ewoud.smeur@gmail.com>
+ *
+ * This file is part of paparazzi.
+ *
+ * paparazzi is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * paparazzi is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with paparazzi; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+/**
+ * @file firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
+ *
+ */
+
+#include "firmwares/rotorcraft/guidance/guidance_indi_hybrid.h"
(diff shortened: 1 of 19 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_tailsitter.c
```diff
@@ -0,0 +1,100 @@
...
+  v_gih[2] = a_diff.z;
+}
+
+
+void WEAK guidance_indi_hybrid_set_wls_settings(float body_v[3] UNUSED, float roll_angle, float pitch_angle)
+{
+  // Set lower limits
+  du_min_gih[0] = -guidance_indi_max_bank - roll_angle; // roll
+  du_min_gih[1] = RadOfDeg(guidance_indi_min_pitch) - pitch_angle; // pitch
+  du_min_gih[2] = (MAX_PPRZ - actuator_state_filt_vect[0]) * g1g2[3][0] + (MAX_PPRZ - actuator_state_filt_vect[1]) * g1g2[3][1] + (MAX_PPRZ - actuator_state_filt_vect[2]) * g1g2[3][2] + (MAX_PPRZ - actuator_state_filt_vect[3]) * g1g2[3][3];
```
gautierhattenberger, line 89: It seems that `actuator_state_filt_vect` and `g1g2` from `firmwares/rotorcraft/stabilization/stabilization_indi.h` are not visible here.

dewagter: The ```WLS``` version is incompatible with ```stabilization_indi_simple```

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
```diff
@@ -0,0 +1,157 @@
...
+
+
+/**
+ * Perform WLS
+ *
+ * @param Gmat Dynamics matrix
+ * @param a_diff acceleration errors in earth frame
+ * @param body_v 3D vector to write the control objective v
+ */
+void WEAK guidance_indi_calcg_wing(float Gmat[GUIDANCE_INDI_HYBRID_V][GUIDANCE_INDI_HYBRID_U], struct FloatVect3 a_diff, float body_v[GUIDANCE_INDI_HYBRID_V]) {
```
gautierhattenberger, line 86: A weak function only makes sense if you have a default implementation, but since you have made explicit modules for each vehicle types, it is not needed I think. Only one of them will implement this function.
Maybe, these modules can even be declared in conflicts in the module XML files.

dewagter: I did try to declare conflicts. The tail sitter implementation should indeed not be weak anymore. I will remove it. The settings for the quadplane on the other hand still need to be WEAK since the skewing quadplane overrides this in ctrl_eff modules.

gautierhattenberger (approved):

dewagter:
Suggested follow-up: https://github.com/paparazzi/paparazzi/issues/3168

PR #3162, opened 2023-11-05
CommandsFromRC overwritten in INDI

dewagter:
```stabilization_indi``` overwrites the ```CommandsFromRC``` from the main loop.
For now: this fixes it.
@EwoudSmeur @gautierhattenberger: any better idea?

Symptom: we need a 5th command for quadplanes.

### DIFF
--- a/conf/airframes/tudelft/rot_wing_25kg.xml
+++ b/conf/airframes/tudelft/rot_wing_25kg.xml
@@ -20,10 +20,11 @@
             <module name="flight_recorder"/>
 
             <!-- RC switches -->
-            <define name="RADIO_TH_HOLD"        value="RADIO_AUX1"/>
-            <define name="RADIO_KILL_SWITCH"    value="RADIO_AUX1"/>
-            <define name="RADIO_FMODE"          value="RADIO_AUX2"/>
-            <define name="RADIO_FBW_MODE"       value="RADIO_AUX3"/>
+            <define name="RADIO_TH_HOLD"          value="RADIO_AUX1"/>
+            <define name="RADIO_KILL_SWITCH"      value="RADIO_AUX1"/>
+            <define name="RADIO_FMODE"            value="RADIO_AUX2"/>
+            <define name="RADIO_FBW_MODE"         value="RADIO_AUX3"/>
+            <define name="RADIO_CONTROL_THRUST_X" value="RADIO_AUX4"/>
 
             <!-- EKF2 configure inputs -->
             <define name="INS_EKF2_GYRO_ID"     value="IMU_CUBE1_ID"/>
@@ -156,9 +157,9 @@
     </commands>
     
 
-    <auto_rc_commands>
-        <set VALUE="@AUX4" COMMAND="THRUST_X"/>
-    </auto_rc_commands>
+    <rc_commands>
+        <set COMMAND="THRUST_X" VALUE="@AUX4"/>
+    </rc_commands>
 
 
     <command_laws>
--- a/conf/airframes/tudelft/rot_wing_25kg.xml
+++ b/conf/airframes/tudelft/rot_wing_25kg.xml
@@ -20,10 +20,10 @@
             <module name="flight_recorder"/>
 
             <!-- RC switches -->
-            <define name="RADIO_TH_HOLD"          value="RADIO_AUX1"/>
-            <define name="RADIO_KILL_SWITCH"      value="RADIO_AUX1"/>
-            <define name="RADIO_FMODE"            value="RADIO_AUX2"/>
-            <define name="RADIO_FBW_MODE"         value="RADIO_AUX3"/>
+            <define name="RADIO_TH_HOLD"        value="RADIO_AUX1"/>
+            <define name="RADIO_KILL_SWITCH"    value="RADIO_AUX1"/>
+            <define name="RADIO_FMODE"          value="RADIO_AUX2"/>
+            <define name="RADIO_FBW_MODE"       value="RADIO_AUX3"/>
             <define name="RADIO_CONTROL_THRUST_X" value="RADIO_AUX4"/>
 
             <!-- EKF2 configure inputs -->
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by gautierhattenberger** [comment]
It is a bit hard to follow, but it seems that you want to set by hand the THRUST_X when using `RC_DIRECT` (so not in NAV) . Am I correct ? `SetAutoCommandsFromRC` is applied to `commands` (and not `stabilization_cmd` in regular code) and is supposed to be used for RC commands that you want to use in NAV. So that is a bit the opposite here ?
I'm not sure what is the best way. Is it something that you want to always have for quadplanes, or only if "configured" in airframe file ? In which modes is it active and where is it applied (stab, guidance, both) ?

**Comment by dewagter** [comment]
Totally agree it should have been ```rc_commands``` and not ```rc_auto_commands```.

```stabilization_indi``` uses either the thrust command from ```ABI``` if it is set, or reads the ```stabilization_cmd[THRUST]``` and ```stabilization_cmd[THRUST_X]``` is a manual mode is active. These must be set PRIOR to running stabilization, which is why the current CommandsFromRC macro called at the end of the main loop does not appear to work.

One option is to add the forward thrust in ```guidance_v.rc_delta_t_x;``` just like it is done for the thrust. Or we use the airframe file ```rc_commands``` to be more flexible (the name of the stick is not yet standard and might change a lot).

File: sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c [hunk 1]
```diff
@@ -232,6 +232,10 @@ void guidance_v_run(bool in_flight)
     case GUIDANCE_V_MODE_RC_DIRECT:
       guidance_v.z_sp = stateGetPositionNed_i()->z; // for display only
       stabilization_cmd[COMMAND_THRUST] = guidance_v.rc_delta_t;
+
+#ifdef SetCommandsFromRC
```
EwoudSmeur, line 236: This doesn't make sense to me, first this is set
```stabilization_cmd[COMMAND_THRUST]```
And then it is overwritten. I also don't understand why this would be part of vertical guidance. I think it would be better if there was an explicit check if some thrust_x command is available, and otherwise (explicitly) take a certain RC channel. In my eyes that would be more transparent than a macro that is quite hard to find and understand.

Comment (dewagter) [reply sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c]: Ok, in ```guidance_h``` as requested and a clear ```RADIO_CONTROL_THRUST_X``` instead of a hidden define.

File: sw/airborne/firmwares/rotorcraft/guidance/guidance_v.c [hunk 2]
```diff
@@ -232,6 +232,7 @@ void guidance_v_run(bool in_flight)
     case GUIDANCE_V_MODE_RC_DIRECT:
       guidance_v.z_sp = stateGetPositionNed_i()->z; // for display only
       stabilization_cmd[COMMAND_THRUST] = guidance_v.rc_delta_t;
+
```
dewagter, line 235:
```suggestion
```

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
```diff
@@ -298,6 +298,11 @@ void guidance_h_run(bool  in_flight)
       if ((!(guidance_h.mode == GUIDANCE_H_MODE_FORWARD)) && transition_percentage > 0) {
         transition_run(false);
       }
+
+#ifdef COMMAND_THRUST_X
+      stabilization_cmd[COMMAND_THRUST_X] = radio_control.values[RADIO_CONTROL_THRUST_X];
+#endif
+
```
gautierhattenberger, line 305: I really don't like to have something so specific here. What about adding this line here :https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c#L628, which corresponds to the case where stabilization indi is called without guidance if I'm not wrong ?

conf/airframes/tudelft/rot_wing_25kg.xml
```diff
@@ -156,9 +157,9 @@
     </commands>
     
 
-    <auto_rc_commands>
-        <set VALUE="@AUX4" COMMAND="THRUST_X"/>
-    </auto_rc_commands>
+    <rc_commands>
+        <set COMMAND="THRUST_X" VALUE="@AUX4"/>
+    </rc_commands>
```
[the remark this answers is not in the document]
dewagter: Correct, with the proposed solution this is not needed. I did not remove it yet only because I was not convinced. What I wanted is really ```rc_command```: as in when you are in a manual mode just copy the stick to the actuator. That is exactly what is needed, but with ```stabilization_indi```, that does not work (since stabilization writes all actuators before the main loop copies commands to actuators).

sw/airborne/firmwares/rotorcraft/guidance/guidance_h.c
dewagter: OK, I moved it.

dewagter:
I now read the ```radio_control.value``` directly in ```stabilization_indi``` as proposed.

I don't particularly like that ```THRUST``` is read in ```guidance_v``` while ```THRUST_X``` is in ```stabilization_indi.c```. And that stabilization now basically takes outerloop ```radio``` commands and puts it directly into ```actuators```, bypassing all command settings in main and everything in ```guidance_v``` except ```stabilization_cmd[THRUST]```, which (sometimes: when not receiving ABI thrust commands) serves as input while it also overwrites the variable.

Works for now, but might be worth re-thinking.

gautierhattenberger:
I have a work in progress to get rid of the ABI thrust trick. Maybe it will make things better for this case, but I'm not sure.
I agree it is not the best solution, but at least the impact looks minimal on the code, and doesn't change the behavior for general case.

gautierhattenberger (approved):

PR #3166, opened 2023-11-06
fixed nav_hybrid bug, renamed nav_hover_speed to nav_goto_max_speed.

alessandro-mancinelli:
Fixed nav_hybrid bug, renamed nav_hover_speed to nav_goto_max_speed consistently in both .xml, .c and .h nav_hybrid files. Master compiles now.


fvantienen:
Double

PR #3167, opened 2023-11-06
[sim] moving the old ocaml simulator to NPS

gautierhattenberger:
- the exact same basic model is now a NPS FDM
- sim target still woks, it is just an alias to NPS with the proper FDM
- the old ocaml files are removed
- AHRS and INS are bypassed, since the accelerations are not well calculated by the model

The main reason to drop the ocaml implementation is the dependency to deprecated gtk2. A side effect is that now we don't have the slider to easily change the battery voltage to test safety procedures.


fvantienen (approved):
I like the cleanup!

PR #3164, opened 2023-11-06
[imu] Add accel and gyro filtering options

fvantienen:

PR #3170, opened 2023-11-08
[imu] Fix accelerometer and gyroscope filters

fvantienen:
This fixes a mistake from #3164 because the filtering is done on the raw values it can't be filtered using the integer functions.

### DIFF
--- a/sw/airborne/modules/imu/imu.c
+++ b/sw/airborne/modules/imu/imu.c
@@ -389,7 +389,7 @@ void imu_init(void)
       float tau = 1.0 / (2.0 * M_PI * imu.gyros[i].filter_freq);
       float sample_time = 1 / imu.gyros[i].filter_sample_freq;
       for(uint8_t j = 0; j < 3; j++)
-        init_butterworth_2_low_pass_int(&imu.gyros[i].filter[j], tau, sample_time, 0.0);
+        init_butterworth_2_low_pass(&imu.gyros[i].filter[j], tau, sample_time, 0.0);
     }
 
     /* Copy accel calibration if needed */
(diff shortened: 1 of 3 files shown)


fvantienen:
Not a problem with overflow, but see what happens below:
Screenshot 2023-11-07 at 16 08 33
We need to be able to represent values in between the raw sensor values for the filter. There is another option to first scale the sensor values, then this problem would not be there.

fvantienen:
The other problem with the `_int` version is that the arguments of the functions are different. Instead of `tau` it accepts a `cutoff_freq`, which was also incorrect in the current implementation.

sw/airborne/modules/imu/imu.c
```diff
@@ -579,12 +579,12 @@ static void imu_gyro_raw_cb(uint8_t sender_id, uint32_t stamp, struct Int32Rates
...
     struct Int32Rates data_filtered[samples];
     for(uint8_t i = 0; i < samples; i++) {
-      update_butterworth_2_low_pass_int(&gyro->filter[0], data[i].p);
-      data_filtered[i].p = get_butterworth_2_low_pass_int(&gyro->filter[0]);
-      update_butterworth_2_low_pass_int(&gyro->filter[1], data[i].q);
-      data_filtered[i].q = get_butterworth_2_low_pass_int(&gyro->filter[1]);
-      update_butterworth_2_low_pass_int(&gyro->filter[2], data[i].r);
-      data_filtered[i].r = get_butterworth_2_low_pass_int(&gyro->filter[2]);
+      update_butterworth_2_low_pass(&gyro->filter[0], data[i].p);
+      data_filtered[i].p = get_butterworth_2_low_pass(&gyro->filter[0]);
```
gautierhattenberger, line 583: side note: `update` function already returns the filtered value, so you can write `data_filtered[i].p = update_butterworth_2_low_pass(&gyro->filter[0], data[i].p);`

fvantienen: Thanks updated

gautierhattenberger (approved):
It seems to me that the problem is the implementation of the integer version that is a bit strange and most likely not well tested.
In the meantime, let's go for the float version if you are happy with it.

PR #3171, opened 2023-11-09
Joystick file for the radio master tx12

tmldeponti:
Simple joystick file for the radio master tx12 that we use at the mavlab for simulations and for the bebops in datalink

### DIFF
--- a/conf/joystick/radiomaster_tx12.xml
+++ b/conf/joystick/radiomaster_tx12.xml
@@ -30,8 +30,8 @@ The basis of steering is the standard signs of aerospace convention
     <axis index="0" name="LeftStickVertical"/>
     <axis index="1" name="RightStickHorizontal"/>
     <axis index="2" name="RightStickVertical"/>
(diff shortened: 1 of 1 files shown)


conf/joystick/radiomaster_tx12.xml
```diff
@@ -0,0 +1,51 @@
+<!--
+SM600 6CH RC Transmitter shaped USB HID joystick Mode 2
```
gautierhattenberger, line 2: Can you just update the comments to match this joystick ?

gautierhattenberger (changes requested):

conf/joystick/radiomaster_tx12.xml
dewagter, line 2:
```suggestion
RadioMaster TX12 Transmitter - Used as Mode 2 Joystick
```

tmldeponti:
Thank you for the change! :)

gautierhattenberger (approved):

PR #3174, opened 2023-11-14
[imu] Initialize filters to false by default

dewagter:
For @fvantienen


fvantienen (approved):

fvantienen:
Is a nice to have, but functions without

PR #3178, opened 2023-11-15
Control camera's over MAVLink

dewagter:
Not sure how to solve this: to control camera's MAVLink is used but the AP should then not identify as PPRZ


dewagter:
```make test```
```All tests successful```

Issue #3179, opened 2023-11-21
Dependency on INDI. Should be made indpendent


tmldeponti:
https://github.com/paparazzi/paparazzi/blob/7b54c09acc82b15a583150cf27b2c9c88c73f6fe/sw/airborne/modules/rot_wing_drone/wing_rotation_controller_servo.c#L138C3-L138C49

look at how i fixed it:
rw3c_master
46cd43b31537f2f0b1a085934fb3bde5a2d8a396

tmldeponti:
PR opened

dewagter:
see PR

PR #3180, opened 2023-11-21
Rot mech fix pr

tmldeponti:
This is to fix the depndency of the rot controller servo module (NPS) from INDI related variables. The IDX of the servo of the rotation mechanism can (and has) to be defined in the airframe, not dependent on assumptions regarding the number of INDI actuators.

### DIFF
--- a/conf/airframes/tudelft/rot_wing_v3b.xml
+++ b/conf/airframes/tudelft/rot_wing_v3b.xml
@@ -119,7 +119,9 @@
         <!-- Other -->
         <module name="sys_id_doublet"/>
         <module name="sys_id_auto_doublets"/>
-        <module name="wing_rotation_adc_sensor"/>
+        <module name="wing_rotation_adc_sensor"
+            <define name = "ROT_MECH_IDX" value = "9"/>
+        </module>
         <module name="rot_wing_automation"/>
         <module name="ground_detect_sensor"/>
         <module name="rotwing_state"/>
--- a/sw/airborne/modules/rot_wing_drone/rotwing_state.c
+++ b/sw/airborne/modules/rot_wing_drone/rotwing_state.c
@@ -579,7 +579,7 @@ void rotwing_state_skew_actuator_periodic(void)
 #endif
 
 #if USE_NPS
-  actuators_pprz[INDI_NUM_ACT] = (rotwing_state_skewing.servo_pprz_cmd + MAX_PPRZ) / 2.; // Scale to simulation command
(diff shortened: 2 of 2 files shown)


dewagter:
@tmldeponti Thanks for the proposed update. This is indeed a problem!

I think that the existing code is writing outside the array (fortunately only in simulation, but still)!

Why can the simulator not simply use the same actuator IDX as the autopilot code? This way it would first of all not be outside the array (crucial), be similar to the flight code, but also not require extra defines.

tmldeponti:
@dewagter Indeed manually specifying an index without any checks can result in out of bounds. I also agree that the ap can be coded to use actuator_pprz[whatever_index_the_rot_mech_is] instead of wing_rotation_controller.servo_pprz_cmd as currently coded in the 3b airframe. I could though see some problems with that:
1) How do we make sure that control codes that loop through the number of actuators do not consider the index of the rotation mech? Maybe something like looping through ACT_NUM -1 if rot_mech exists and assuming that it is the last actuator (which might not be the case)?
2) The number of actuators still has to be defined as something independent of INDI (or my oneloop controller). I belive there should be a distinction between the total number of actuators and the actual number of actuator used in control allocation. Maybe, building on what i have done in the oneloop andi controller (see https://github.com/tudelft/paparazzi/blob/395ddc3df232cf3eaa4bc67e0f3dab3d454d38bc/conf/airframes/tudelft/rot_wing_v3c_oneloop_simulation.xml#L113), actuator number should have subcategories like :
ANDI_NUM_REAL_ACT = 4; // four quad motors
ANDI_NUM_VIRTUAL_ACT = 2; // Phi and Theta
ANDI_NUM_PASSIVE_ACT = 1; // Rot Mech
ANDI_NUM_ACT = ANDI_NUM_REAL_ACT + ANDI_NUM_VIRTUAL_ACT; // For control allocation loops
ANDI_NUM_ACT_TOT =  ANDI_NUM_ACT + ANDI_NUM_PASSIVE_ACT; // For definition of important arrays like actuator_pprz

ANDI can of course be removed from the name. 
Let me know what you think and i can code a solution and commit it for further review.

dewagter:
Addresses #3179

dewagter:
I agree there must be a way to have any combination of
 - actuators set by control: e.g. INDI
 - actuators set by modules
 - actuators set by RC directly
 - virtual actuators

To me, in paparazzi, this list is called ```COMMANDS``` which in paparazzi already is super flexible and can do everything mentioned above.

I just think that adding defines about which element of the array to use is so terribly error prone it should not be the way to go. Here the index must match the first place after the actuators set by control, but should still fit in the array. This is all not checked and will lead to bugs.

One possible alternative is suggested in #3203 ... and if that is not working out we will need another way.

PR #3181, opened 2023-11-21
we need force forward also in NPS

tmldeponti:


dewagter:
```make test``` successful.

PR #3182, opened 2023-11-22
[scheduling] harmonize efficiency scheduling modules

gautierhattenberger:
- harmonize names
- make a 'generic' module with basic linear scheduling
- old ctrl_effectiveness_scheduling was very specific to cyfoam, so renamed
- add falcon scheduler, related airframe will come later

@dewagter I have not renamed the rot_wing scheduling module as I don't want to break tudelft conf, but I suggest to change its name as well if you can


dewagter:
Hi @gautierhattenberger : since now all our rotwing code is in master, it is ok to rename it too. If ```make test_tudelft``` keeps working, then anything can be renamed to harmonize the project.

dewagter (dismissed):
Agree with the suggested change. It would be even better if all modules are renamed if that is possible?

conf/airframes/tudelft/cyfoam.xml
dewagter, line 1: a pitty to have 2 airframes with so little difference ... hopefully one day we can have a simple elegant solution for this.

dewagter (approved):
Thank you!

dewagter:
Will check bebop. That used to work. 25kg will be fixed with #3177.

PR #3184, opened 2023-11-23
[ins] fix ekf2 status flag when using external vision positionning

gautierhattenberger:
It fixes the display of IMU status in the GCS when the OPTITRACK mode of EKF2 is used.


fvantienen (approved):
Seems like a good fix for the GCS. Though it would be better in the future to distinguish between a GPS and EV position fix, since the EKF2 can have both at the same time or a single one.

gautierhattenberger:
We probably need to revise the status messages, not only for position...

PR #3186, opened 2023-11-24
[supervision] Add log extractor.

Fabien-B:
Add a IHM for SD log extraction in the paparazzi center.
It allows to specify output path with variable substitutions.
The default path can be changed in the settings.

image


gautierhattenberger:
Some remarks and suggestions:
- The tab name `Tools` is a bit misleading with tools in operation panel. Replace by `Utilities` or something else ?
- Replace `Logs` by something more explicit: "Extract logs from SD card / flight recorder"
- Add some lines in the help popup about what you are suppose to do: "Select a file from flight recorder in pprzlog format and choose the destination folder for extracted data" (or something equivalent)
- The size of the rectangle in the tab is changing with the size of the whole window, which is a bit strange. Is possible to have at least a fixed vertical size ?

Fabien-B:
I made the suggested changes.

gautierhattenberger (approved):

PR #3190, opened 2023-11-27
INDI actuator dynamics in continuous time

EwoudSmeur:
It is much more understandable and convenient if the first order actuator dynamics in INDI are specified in continuous time, through the corner frequency of the first order model. Then it is also possible to change the periodic frequency without having to change the actuator filter parameters.

I tried some find-replace in all the airframe files, but it got complex because of the different periodic frequencies. I decided to just add an error if the old name is used.

### DIFF
--- a/conf/airframes/tudelft/bebop_indi.xml
+++ b/conf/airframes/tudelft/bebop_indi.xml
@@ -156,9 +156,9 @@
     <define name="ESTIMATION_FILT_CUTOFF" value="8.0"/>
 
     <!-- first order actuator dynamics -->
-    <define name="ACT_DYN_P" value="0.1"/>
-    <define name="ACT_DYN_Q" value="0.1"/>
-    <define name="ACT_DYN_R" value="0.1"/>
+    <define name="ACT_FREQ_P" value="52.0"/>
+    <define name="ACT_FREQ_Q" value="52.0"/>
+    <define name="ACT_FREQ_R" value="52.0"/>
 
     <!-- Adaptive Learning Rate -->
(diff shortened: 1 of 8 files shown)


gautierhattenberger:
I'm all in favour of changing this parameter to something independent of sampling time. However, if you have an error when the new name is not defined, this will break many conf including in compilation tests.
If new name is not used, I suggest to keep initialising with the old value and print a deprecated warning to ask replacing the airframe define with the provided formula. At least it will show the failing conf so we can update them by hand if needed.

EwoudSmeur:
Ok, maybe a time should also come to throw some old airframes away, that would make updating the remaining ones less painful haha

fvantienen:
If we deprecate things like this, can we also put an end date on it like a year? Since it just complicated the code and is only applicable to a very small amount of people usually.

For me I don't like the whole deprecation scheme we have right now. We are always trying to keep it working for old airframes and users who don't want to update, but I think the current user base that has these problems is so small that it seems useless to me. I would prefer to update and renew things and just get rid of the old and live with the fact that sometimes you have to solve problems like this as a user. Currently we are not the most user friendly autopilot system, so why do we try to be this with the airframe's and defines?

EwoudSmeur:
haha ok yeah I can also see that.

Now that I made it already, maybe we can merge as is? I'm fine with removing the warning later. Also, some of these airframes that I saw are long gone, like the 'frog' airframe in BR conf XD

gautierhattenberger (approved):
I'm okay to merge this, with the objective to fix most airframes (or remove them) before next release to avoid unnecessary noise during compilation tests

PR #3194, opened 2023-11-30
[pwm] use different names to activate PWM drivers in chibios and pprz

gautierhattenberger:
ChibiOS PWM driver is required for DShot driver, but activating it should not result in an activation of the PWM driver by paparazzi. Otherwise, the last driver to be called at init will set the parameters and eventually overwrite the correct values.

Not tested yet.


Fabien-B (approved):
Tested on a Tawaki.

PR #3195, opened 2023-12-04
[fix] fix some bugs from #3190

gautierhattenberger:


EwoudSmeur:
Thanks for fixing, sorry for the trouble...

PR #3196, opened 2023-12-07
Fix continuous actuator

dewagter:
- ```thrust_dyn``` only exists when not using ```GUIDANCE_INDI_SPECIFIC_FORCE_GAIN```
 - ```act_dyn_discrete``` is set at the beginning from the new ```ACT_FREQ``` define, but should be computed during initialization from ```ACT_DYN``` if that old define is still present.


EwoudSmeur (approved):
Thanks for fixing! I should check my PRs better... 😅

dewagter:
No worries: it's a nice update. Thanks for checking.

PR #3198, opened 2023-12-08
[modules] UAVCAN range sensor and cleanup

fvantienen:
This adds a range sensor over UAVCAN and also cleans up the airspeed sensor.


dewagter (approved):
Ok!

PR #3199, opened 2023-12-12
Support for TawakiV2 (STM32H7)

Fabien-B:


conf/modules/board_tawakiV2.xml
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="board_tawakiV2" dir="boards">
```
gautierhattenberger, line 3: as mentioned, maybe change the name of this module

sw/airborne/arch/chibios/mcu_periph/hal_stm32_dma.h
```diff
@@ -444,7 +448,9 @@ struct DMADriver {
   /**
    * @brief controller associated with stream
    */
+//#if STM32_DMA_SUPPORTS_DMAMUX
```
gautierhattenberger, line 451: remove if not needed ?

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
```diff
@@ -345,8 +345,14 @@ void i2c3_hw_init(void)
 
 #if USE_I2C4
 PRINT_CONFIG_VAR(I2C4_CLOCK_SPEED)
+
+#if defined(STM32H7XX)
+// Local variables (in DMA safe memory)
+static IN_BDMA_SECTION(struct i2c_init i2c4_init_s) = {
```
gautierhattenberger, line 351: is this true for all H7 boards ?

Fabien-B:
I will try looking into your suggestions, but it may not be easy.
First annoying thing:

**common_board.h**:
```
#define LED_1_GPIO_ON     gpio_clear
#define LED_1_GPIO_OFF    gpio_set
```
**tawaki_v2.0.h**:
```
#define LED_1_GPIO_ON     gpio_set
#define LED_1_GPIO_OFF    gpio_clear
```

sw/airborne/arch/chibios/mcu_periph/spi_arch.c
```diff
@@ -487,6 +487,31 @@ void spi4_arch_init(void)
 }
 #endif
 
+#if USE_SPI6
+
+#if defined(STM32H7XX)
+// Local variables (in DMA safe memory)
+static IN_BDMA_SECTION(struct spi_init spi6_init_s) = {
```
gautierhattenberger, line 494: same question. Is it for all H7 boards ? only for Tawaki ? what about the other SPI or I2C peripherals ?

gautierhattenberger (changes requested):

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
Fabien-B: yes

sw/airborne/arch/chibios/mcu_periph/spi_arch.c
Fabien-B: yes, SPI6 is on BDMA.

alex31:
> What about the defines in `mcuconf.h` is this not enough for you to set the nocache regions?:
> 
> ```
>  * Memory attributes settings.
>  */
> #define STM32_NOCACHE_ENABLE                TRUE
> #define STM32_NOCACHE_MPU_REGION            MPU_REGION_6
> #define STM32_NOCACHE_RBAR                  0x24000000U
> #define STM32_NOCACHE_RASR                  MPU_RASR_SIZE_64K
> ```
> 
> Because I prefer to keep it then the same for every board, either we do it ourselves or use the ChibiOS features to set the MPU.
> 
> Next to that your [tawaki_v2.0.h](https://github.com/paparazzi/paparazzi/pull/3199/files#diff-75a9a301594643dbefcdd4ea7ecc73e73135ef3ea9e6d6818a2a17058339a254) is almost the same as the `common_board.h`. Isn't it better to change the names in `board.cfg` such that `common_board.h` can be used and the features that are missing can be added there?
> 
> These are just some suggestions, but I really like the fixes that you have done :)

The H7 provides 3 different domains, and we need "nocache" regions in 2 regions, to support all of our dma bound peripherals. Chibios mechanism permit to define only one region via mcuconf.h.  
We define 2 regions in our custom code, so STM32_NOCACHE_ENABLE should be set to FALSE in mcuconf.h

sw/airborne/arch/chibios/mcu_periph/i2c_arch.c
alex31: yes I2C4 and SPI6 are peripherals linked to BDMA on all STM32H7 whatever the card that use them.
the rational behind the use of I2C4 and SPI6 on the tawakiV2 is to migrate maximum number of peripherals from DMA to BDMA to free DMA channels that we want to later use with BIDIR DSHOT which need N+1 DMA channels for each timer using Bidir DSHOT, so up to 10 DMA channels for octocopter.

sw/airborne/arch/chibios/mcu_periph/spi_arch.c
alex31: cf I2C4 comment

conf/airframes/ENAC/quadrotor/TawakiV2.xml
```diff
@@ -0,0 +1,205 @@
...
+    <target name="ap" board="tawaki_2.0">
+      <define name="RADIO_KILL_SWITCH" value="RADIO_GAIN1"/>
+    </target>
+
+    <target name="nps" board="pc">
+      <module name="fdm" type="jsbsim"/>
+      <module name="radio_control" type="ppm"/>
+    </target>
+
+    <module name="board" type="tawakiV2">
```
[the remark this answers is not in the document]
Fabien-B: There is `boards/tawaki2.0.xml` which is different from `board_tawakiV2.xml`.
I tried reproducing the same files as for Tawaki.
What should I do ?

gautierhattenberger (approved):
Can be merged once tested in flight

Fabien-B:
Tested multiple times, it works.

PR #3203, opened 2023-12-13
Remove ```actuators_pprz```

dewagter:
1) there are now 3 types of numbers of actuators and 2 types of control command arrays:

   - ```SERVO_NAME_IDX``` -> ```actuators[]``` -> must also be used in ABI messages
   - INDI_MATRIX_SEQUENCE -> ```actuators_pprz[]``` is called actuator but contains controller outputs which not always map 1 to 1 to actuators
   - ```SERVO_NAME no=""``` -> CAN: should never be used in code. I would like to remove the ```define``` from the ```airframe.h``` file to avoid errors
   - the sequence of commands defined in the airframe file is often used but not always.

2) there are now problems with adding virtual actuators when the number of commands becomes larger than the number of actuators and there are no checks if ```INDI_NUM_ACT``` is smaller than the actuator array size: see #3180

3) ```actuators_pprz``` index numbers are often the same numbers as the actuators, so many users confuse them, but they are not always the same. One is the sequence in the INDI and the other is the list of servos.

4) ```SERVO_NAME_IDX``` and ```SERVO_NAME``` are sometimes confused in code: #3206 

5) Safety features on commands are OVERRULED (e.g. ```FAILSAFE```) when using INDI: command-laws in xml must re-implement functions like kill

6) Command laws contain raw INDI indexes: not very readable + eff_sheduling relies on a specific order but cannot check this.

7) ```RC_COMMANDS``` and ```RC_AUTO_COMMANDS``` are broken with INDI

8) The simulator needs harmonized actuator commands (```-MAX_PPRZ``` -> ```MAX_PPRZ```), which required the ```actuators_pprz``` but this can be solved differently: #3205

9) some modules write PWM directly to actuators and ignore the servo MIN and MAX
  - [ ] rotrocraft_cam
  - [ ] switch_servo

10) there are unused commands in many airframe files: ##3208 

-> so I want to remove ```actuators_pprz```: and instead do what was always the case before

control -> ```stabilization_cmd``` with the command sequence of the airframe file, and then set the  ```actuators``` in the ```ActuatorsFromCommands``` block, just like the name says.

You can then:
 - make as many (virtual) commands as you want by adding commands
 - use default kill/failsafe functions
 - read the order of commands in the airframe instead of the code
 - forced to define which command goes where so no confusion with different numbering in actuator and actuator_pprz
 - no double numbering in actuators and actuators_pprz
 - ```RC_COMMANDS``` etc repaired...

### DIFF
--- a/conf/airframes/tudelft/bebop_indi_actuators.xml
+++ b/conf/airframes/tudelft/bebop_indi_actuators.xml
@@ -45,7 +45,7 @@
     </module-->
   </firmware>
 
-  <commands>
+  <commands><!-- Order must match the INDI ctrl_eff: INDI_NUM_ACT -->
     <axis name="TOP_LEFT"     failsafe_value="0"/>
     <axis name="TOP_RIGHT"    failsafe_value="0"/>
     <axis name="BOTTOM_RIGHT" failsafe_value="0"/>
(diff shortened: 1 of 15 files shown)


gautierhattenberger:
@dewagter I'm all in favor of removing `actuators_pprz` !
* In the "reference" design, the output of stabilization always goes to the `commands` vector, which is mapped to actuators with the `command_law` section of the airframe (and this is what you did). I also would like to be able to compile without the ROLL, PITCH, YAW command when they are not used. THRUST command is annoying because I thing it is used for the GCS display, but we could probably improve that as well.
* From what I can see, you are now feeding the `stabilization_cmd` array. I will be needed to copy it to the real `commands` array at some point, and `set_rotorcraft_commands`
* I'm not sure about the proper way to deal with the `motors_on` directly implemented in the airframe. It is really annoying and incompatible with dual MCU support (FBW doesn't have access to this variable).
* Last point, I would like to change the `actuators` array of `pprz_t` by an array of structures (with the type of actuators, feedback value if available, current value in pprz and actuators units, etc). This can be done later I think, but could also help to manage the different status (kill, failsafe, motor on/off, running, etc).

dewagter:
> @dewagter I'm all in favor of removing `actuators_pprz` !
> 
> * In the "reference" design, the output of stabilization always goes to the `commands` vector, which is mapped to actuators with the `command_law` section of the airframe (and this is what you did). I also would like to be able to compile without the ROLL, PITCH, YAW command when they are not used. THRUST command is annoying because I thing it is used for the GCS display, but we could probably improve that as well.
> * From what I can see, you are now feeding the `stabilization_cmd` array. I will be needed to copy it to the real `commands` array at some point, and `set_rotorcraft_commands`
> * I'm not sure about the proper way to deal with the `motors_on` directly implemented in the airframe. It is really annoying and incompatible with dual MCU support (FBW doesn't have access to this variable).
> * Last point, I would like to change the `actuators` array of `pprz_t` by an array of structures (with the type of actuators, feedback value if available, current value in pprz and actuators units, etc). This can be done later I think, but could also help to manage the different status (kill, failsafe, motor on/off, running, etc).

PR #3205, opened 2023-12-14
actuators[ ] array of structs with pprz units and driver units

dewagter:
In simulation / NPS, we would like to simulate all actuators. The ```actuators[ ]``` array can not be used because it has different units: e.g. PWM 1100 - 1900.

With the new system of sending actuators to different actuator drivers, the conversion to actuator_driver units is best done when sending to the actuator_driver.

 - this means the separate copy of ```actuators_pprz``` is not needed.
 - this allows to simulate actuators directly and remove dependencies on control allocation.

### DIFF
--- a/conf/settings/setup_actuators.xml
+++ b/conf/settings/setup_actuators.xml
@@ -5,14 +5,14 @@
 <settings target="setup_actuators">
   <dl_settings>
     <dl_settings NAME="Actuators" >
-      <dl_setting VAR="actuators[0]" MIN="900" STEP="1" MAX="2100" module="modules/actuators/actuators" shortname="chan0"/>
-      <dl_setting VAR="actuators[1]" MIN="900" STEP="1" MAX="2100" shortname="chan1"/>
-      <dl_setting VAR="actuators[2]" MIN="900" STEP="1" MAX="2100" shortname="chan2"/>
-      <dl_setting VAR="actuators[3]" MIN="900" STEP="1" MAX="2100" shortname="chan3"/>
-      <dl_setting VAR="actuators[4]" MIN="900" STEP="1" MAX="2100" shortname="chan4"/>
-      <dl_setting VAR="actuators[5]" MIN="900" STEP="1" MAX="2100" shortname="chan5"/>
-      <dl_setting VAR="actuators[6]" MIN="900" STEP="1" MAX="2100" shortname="chan6"/>
-      <dl_setting VAR="actuators[7]" MIN="900" STEP="1" MAX="2100" shortname="chan7"/>
+      <dl_setting VAR="actuators[0].driver_val" MIN="900" STEP="1" MAX="2100" module="modules/actuators/actuators" shortname="chan0"/>
+      <dl_setting VAR="actuators[1].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan1"/>
+      <dl_setting VAR="actuators[2].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan2"/>
+      <dl_setting VAR="actuators[3].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan3"/>
+      <dl_setting VAR="actuators[4].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan4"/>
+      <dl_setting VAR="actuators[5].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan5"/>
+      <dl_setting VAR="actuators[6].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan6"/>
+      <dl_setting VAR="actuators[7].driver_val" MIN="900" STEP="1" MAX="2100" shortname="chan7"/>
     </dl_settings>
   </dl_settings>
 </settings>
--- a/sw/airborne/modules/actuators/actuators.h
+++ b/sw/airborne/modules/actuators/actuators.h
@@ -57,10 +57,17 @@ struct act_feedback_t {
 extern uint32_t actuators_delay_time;
 extern bool   actuators_delay_done;
 
+// Actuator feedback structure for ABI Message
+struct actuator_t {
+  pprz_t pprz_val;                  ///< Actuator value in PPRZ units
+  int16_t driver_val;               ///< Actuator value in driver units (scaling from servo in airframe.h)
+};
+
+
 /** Actuators array.
  * Temporary storage (for debugging purpose, downlinked via telemetry)
  * */
-extern int16_t actuators[ACTUATORS_NB];
+extern struct actuator_t actuators[ACTUATORS_NB];
 
 /** PPRZ command to each actuator
  * Can be used to directly control actuators from the control algorithm
@@ -72,7 +79,7 @@ extern int16_t actuators_pprz[ACTUATORS_NB];
  * @param _n actuators name as given in airframe file, servos section
  * @param _v new actuator's value
  */
-#define _ActuatorSet(_n, _v) Set_ ## _n ## _Servo(_v)
+#define _ActuatorSet(_n, _v) Set_ ## _n ## _Servo(_v,_v)
 #define ActuatorSet(_n, _v) _ActuatorSet(_n, _v)
 
 #endif /* ACTUATORS_NB */
--- a/sw/airborne/modules/actuators/actuators.h
+++ b/sw/airborne/modules/actuators/actuators.h
@@ -79,7 +79,7 @@ extern int16_t actuators_pprz[ACTUATORS_NB];
  * @param _n actuators name as given in airframe file, servos section
  * @param _v new actuator's value
  */
-#define _ActuatorSet(_n, _v) Set_ ## _n ## _Servo(_v,_v)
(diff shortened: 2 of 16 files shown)


gautierhattenberger:
The issue with this (actuators in pprz_t) is that the actuators message is now also in pprz_t. So if you need to read the actuators position to adjust the trim (which should be in "driver" unit), you will have a good chance to make errors doing the conversion or just forgetting to convert it to the correct scale.
An actuator structure could hold the two values, that could be used with proper format when needed...

dewagter:
Agreed

dewagter:
Changed to actuator struct. If you have naming recommendations let me know.

sw/tools/generators/gen_airframe.ml
```diff
@@ -271,9 +271,10 @@ let preprocess_value = fun s v prefix ->
 let print_actuators_idx = fun out ->
   Hashtbl.iter (fun s (d, i) ->
     (* Set servo macro *)
-    fprintf out "#define Set_%s_Servo(_v) { \\\n" s;
-    fprintf out "  actuators[SERVO_%s_IDX] = Clip(_v, SERVO_%s_MIN, SERVO_%s_MAX); \\\n" s s s;
-    fprintf out "  Actuator%sSet(SERVO_%s_DRIVER_NO, actuators[SERVO_%s_IDX]); \\\n" d s s;
+    fprintf out "#define Set_%s_Servo(_pprzv,_s) { \\\n" s;
+    fprintf out "  actuators[SERVO_%s_IDX].pprz_val = ClipAbs( _pprzv, MAX_PPRZ); \\\n" s;
```
EwoudSmeur, line 275: Is there a need to do this here? Normally it should already be within this range, and clipping it can no longer influence the servo driver value.

EwoudSmeur (dismissed):
Great development!

sw/airborne/modules/actuators/actuators.h
```diff
@@ -72,7 +79,7 @@ extern int16_t actuators_pprz[ACTUATORS_NB];
  * @param _n actuators name as given in airframe file, servos section
  * @param _v new actuator's value
  */
-#define _ActuatorSet(_n, _v) Set_ ## _n ## _Servo(_v)
+#define _ActuatorSet(_n, _v) Set_ ## _n ## _Servo(_v,_v)
```
gautierhattenberger, line 82: Here you pass the value `_v` twice, but they should be pprz and driver values. It seems that something is wrong. We either do the mapping pprz to driver should be done in `Set_XXX_Servo` functions (and they take pprz inputs), or we store the `pprz_val` before calling this function (and they take driver inputs as it is currently the case).

sw/airborne/modules/actuators/actuators.h
```diff
@@ -57,10 +57,17 @@ struct act_feedback_t {
 extern uint32_t actuators_delay_time;
 extern bool   actuators_delay_done;
 
+// Actuator feedback structure for ABI Message
+struct actuator_t {
+  pprz_t pprz_val;                  ///< Actuator value in PPRZ units
+  int16_t driver_val;               ///< Actuator value in driver units (scaling from servo in airframe.h)
+};
```
gautierhattenberger, line 64: This is nice to have an actuator structure, and it could / should hold even more information, like the type of actuators, the status, its position or speed if measured, etc. Should we start already with this (and maybe change it later again), or already try to add future elements, even if not used yet ?

conf/settings/setup_actuators.xml
```diff
@@ -5,14 +5,14 @@
...
     <dl_settings NAME="Actuators" >
-      <dl_setting VAR="actuators[0]" MIN="900" STEP="1" MAX="2100" module="modules/actuators/actuators" shortname="chan0"/>
-      <dl_setting VAR="actuators[1]" MIN="900" STEP="1" MAX="2100" shortname="chan1"/>
-      <dl_setting VAR="actuators[2]" MIN="900" STEP="1" MAX="2100" shortname="chan2"/>
-      <dl_setting VAR="actuators[3]" MIN="900" STEP="1" MAX="2100" shortname="chan3"/>
-      <dl_setting VAR="actuators[4]" MIN="900" STEP="1" MAX="2100" shortname="chan4"/>
-      <dl_setting VAR="actuators[5]" MIN="900" STEP="1" MAX="2100" shortname="chan5"/>
-      <dl_setting VAR="actuators[6]" MIN="900" STEP="1" MAX="2100" shortname="chan6"/>
-      <dl_setting VAR="actuators[7]" MIN="900" STEP="1" MAX="2100" shortname="chan7"/>
+      <dl_setting VAR="actuators[0].driver_val" MIN="900" STEP="1" MAX="2100" module="modules/actuators/actuators" shortname="chan0"/>
```
gautierhattenberger, line 8: It would be nice to have a getter function or macro for that, in case the structure is modified in the future

sw/airborne/modules/actuators/actuators.h
dewagter: This macro is weird indeed and points to an inconsistency in the project: it is used in code that sets actuator values directly: e.g. switch servo sets the actuator to 1200 us. The correct way would be to put it to MAX_PPRZ. I can harmonize that.

sw/airborne/modules/actuators/actuators.h
dewagter: It is very easy to add things later. I'd rather only add variables while that are used.
In this PR, as a first step, my goal is to be able to run the simulator directly from actuators everywhere (both INDI and rotorcraft)

dewagter: But I agree the rpm/position feedback could move to this struct a.s.a.p.  
Anybody able to help?

sw/tools/generators/gen_airframe.ml
dewagter: I was not sufficiently sure that all code everywhere did this properly, so I left this double clipping, hoping that the compiler would optimize it away if not needed.

sw/airborne/modules/actuators/actuators.h
gautierhattenberger: I'm in favor of harmonizing the servo setter with PPRZ values. It would make the code actuator independent.

sw/airborne/modules/actuators/actuators.h
gautierhattenberger: I was planing to do this after finishing the reorganization of stabilization, if I can finish that someday...

dewagter: stabilization reorganization merged.

sw/airborne/modules/actuators/actuators.h
dewagter: It was now moved to the ```Set_XXX_Servo``` function. Now all pan-tilt and switch servo code becomes much simpler.

dewagter:
Sample generatle ```airframe.h```

```
#define Set_AIL_LEFT_Servo(actuator_value_pprz) { \
  actuators[SERVO_AIL_LEFT_IDX].pprz_val = ClipAbs( actuator_value_pprz, MAX_PPRZ); \
  command_value = actuator_value_pprz * (actuator_value_pprz>0 ? SERVO_AIL_LEFT_TRAVEL_UP_NUM : SERVO_AIL_LEFT_TRAVEL_DOWN_NUM); \
  command_value /= actuator_value_pprz>0 ? SERVO_AIL_LEFT_TRAVEL_UP_DEN : SERVO_AIL_LEFT_TRAVEL_DOWN_DEN; \
  servo_value = SERVO_AIL_LEFT_NEUTRAL + command_value; \
  actuators[SERVO_AIL_LEFT_IDX].driver_val = Clip(servo_value, SERVO_AIL_LEFT_MIN, SERVO_AIL_LEFT_MAX); \
  ActuatorUavcan2CmdSet(SERVO_AIL_LEFT_DRIVER_NO, actuators[SERVO_AIL_LEFT_IDX].driver_val); \
}
```

dewagter:
> The issue with this (actuators in pprz_t) is that the actuators message is now also in pprz_t. So if you need to read the actuators position to adjust the trim (which should be in "driver" unit), you will have a good chance to make errors doing the conversion or just forgetting to convert it to the correct scale. An actuator structure could hold the two values, that could be used with proper format when needed...

This is now sent in ```driver_val```.

gautierhattenberger (approved):

Fabien-B:
Hi @dewagter, following this PR, I have this error when building the CobraV2 aircraft from ENAC/conf_enac.xml.
Can you look into it please ?
```
In file included from modules/switching/switch_servo.h:27,
from modules/switching/switch_servo.c:22:
modules/switching/switch_servo.c: In function 'switch_servo_periodic':
/home/fabien/paparazzi/var/aircrafts/CobraV2/ap/generated/airframe.h:162:3: error: 'command_value' undeclared (first use in this function)
162 | command_value = actuator_value_pprz * (actuator_value_pprz>0 ? SERVO_SWITCH_TRAVEL_UP_NUM : SERVO_SWITCH_TRAVEL_DOWN_NUM); \
| ^~~~~~~~~~~~~
```

dewagter:
Should be solved in #3280 ...
if someone accepts it, you can merge

PR #3216, opened 2023-12-18
Fix compile warnings in ```make test```

dewagter:


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -128,13 +128,14 @@ static void guidance_indi_filter_thrust(void);
 #warning "The thrust dynamics are now specified in continuous time with the corner frequency of the first order model!"
 #warning "define GUIDANCE_INDI_THRUST_DYNAMICS_FREQ in rad/s"
 #warning "Use -log(1 - old_number) * PERIODIC_FREQUENCY to compute it from the old value."
+#define GUIDANCE_INDI_THRUST_DYNAMICS_FREQ (-ln(1-GUIDANCE_INDI_THRUST_DYNAMICS)*PERIODIC_FREQUENCY)
```
gautierhattenberger, line 131: Shouldn't it be `ln` in the comment` and `log` in the define rather than the opposite ?

dewagter: Yes: this was found and solved in #3217

PR #3218, opened 2023-12-18
[xml] Airframe updates after testflights

dewagter:


conf/flight_plans/tudelft/rotating_wing_EHVB.xml
```diff
@@ -27,6 +33,32 @@
     <waypoint lat="52.164861" lon="4.427268" name="C7"/>
     <waypoint lat="52.170422" lon="4.427511" name="C8"/>
     <waypoint lat="52.172276" lon="4.424011" name="C9"/>
+    <!-- t'Harde
+    <waypoint lat="52.4034734" lon="5.9067552" name="C1"/>
```
gautierhattenberger, line 37: It seems that you have to deal with multiple flight zones. Instead of commenting the waypoints in every flight plans, I recommend you to use includes like this:
https://github.com/paparazzi/paparazzi/blob/master/conf/flight_plans/competitions/IMAV2023_data.xml
https://github.com/paparazzi/paparazzi/blob/master/conf/flight_plans/competitions/IMAV2023_carto.xml#L105

dewagter: **Option 1:** local FP includes common list of blocks:
 - problem: procedure does not allow all normal flightplan options: it can not accept spaces in names, fails with Lexing error in ```'(GetPosAlt(),``` errors in Lexing error in ```'(agl_dist_valid&&(agl_dist_value```
 
 **Option 2:** common main flightplan including a list of local WP
 - problem: can not change the ```alt="60" ground_alt="0" lat0="52.965523" lon0="4.474033"```

gautierhattenberger: Procedures are indeed limited in terms of parsing, but you can at least easily use definition of (global) waypoints and sectors.

dewagter: This only seems to work when the WP are close to the FP HOME.

Any fundamental reason why included Flight plans are parsed differently?

dewagter:
Flight plan cleaned. Ready to merge.

gautierhattenberger (approved):

conf/flight_plans/tudelft/rotating_wing_EHVB.xml
gautierhattenberger: Not sure what you mean by close to HOME. Is it for local waypoints or global ones (or both ?)

dewagter: I mean the flightplan location: ```<flight_plan alt="70" ground_alt="0" lat0="52.1681551" lon0="4.4126468"``` is the most important difference.

PR #3220, opened 2023-12-19
[flightplans] tudelft flightplans

dewagter:
@Fabien-B new GCS not working well with includes : tudelft-conf - rotwing3D -> xmlParseEntityRef on exception


Fabien-B:
@dewagter Oh yeah, I remember Gautier told me that.
I will try looking into it, but I'm quite annoyed by how we handle the configuration files at the moment : in my opinion, all this handling of the configuration files should be done by the server. We should being able to query it to have the full airframe file, with the include resolved.
Same thing for the flight plans with the procedures, that are not working neither in the new GCS at the moment.
I wanted to do that but I was discouraged by how much work is needed to replace the current server.
Maybe I can start small and just do the merging of the configuration files ?

dewagter:
Unable to split.

Option 1: separate wp files with common blocks fails in GCS
Option 2: common flightplan including separate wp files has wrong origin (WP to far from HOME) and needs to be modified by hand every time.
Option 3: separate full flightplans which are often synced with difftools -> selected option.

dewagter:
@Dennis-Wijngaarden changed ```GetPosAlt``` to ```GetPosHeight```

PR #3225, opened 2023-12-21
Time of week in EXTERNAL_POSE message

FlorianSan:
Timestamps are remapped to the time of week to obtain the correct date when using sd2log.


dewagter (commented):
The idea is very nice. However, you now seem to use the current python client time instead of the more exact measurement time ```stamp``` of when the measurement was made.

Could you not compute the ```tow`-stamp`` at the start and then add the offset?

gautierhattenberger:
I agree that the time stamp should be based on optitrack value with a possible offset based on the real date. This should be an option to have it or not, but if you need to synchronise data recorded by optitrack software with your log, it is easier to have the same time reference. As we probably don't do that so often, I suggest to have the tow offset enabled by default so that SD log time is correct.

PR #3228, opened 2024-01-03
[generator] add support for matrix and struct in airframe

gautierhattenberger:
This is a proposal for support of matrix and structures in airframe XML files. See #3189.

It updates:
- code for indi G1 matrix and IMU calib
- ENAC airframes


dewagter (approved):
Great addition, thanks.
For the calibration, this only had advantages.
For the G1 matrix, it is now less clear in the airframe xml which row is which axis. But this can be solved with #3214

gautierhattenberger:
When you make a matrix, you can keep the attribute `name` if it helps, it is just ignored (see https://github.com/paparazzi/paparazzi/pull/3228/files#diff-d9977922b66e09abe58816357c358e7364493e0fc93492541e68966aea432ca5L165)

PR #3231, opened 2024-01-26
Fix natnet2ivy rotation order

tblaha:
@tmldeponti recently found that the pitch and roll components of the `natnet2ivy.py` are sometimes wrong. Further investigation showed that this happens due to a wrong rotation order that shows up when _not_ using options `--x_side right` and `--ac_nose far`, and when the craft is yawed away from pointing to the "north". Also, the yaw angle component seemed unaffected, which is why this bug hasn't shown up before (optitrack pitch/roll components are not typically used by estimators AFAIK?).

The diff looks huge, but actually the only functional change is post-multiplying `q_nose_correction` instead of pre-multiplying here:
https://github.com/paparazzi/paparazzi/blob/41998a12b3dd9de7dc061b532eece9addb1bef6b/sw/ground_segment/python/natnet3.x/natnet2ivy.py#L310-L312

To really make sure this time that it works for all options, I generated test cases, and encapsulated some more script code in functions so they can be tested if `--test` is given as argument (in which case all other arguments are ignored and only the tests are performed)

Todo:
- [ ] test flight with --x_angle set to some numerical value

### DIFF
--- a/sw/airborne/modules/gps/gps_datalink.c
+++ b/sw/airborne/modules/gps/gps_datalink.c
@@ -134,15 +134,15 @@ void gps_datalink_parse_EXTERNAL_POSE(uint8_t *buf)
   enu_speed.y = DL_EXTERNAL_POSE_enu_yd(buf);
   enu_speed.z = DL_EXTERNAL_POSE_enu_zd(buf);
 
-  struct FloatQuat body_q;
+  struct FloatQuat body_q; // converted to NED
   body_q.qi = DL_EXTERNAL_POSE_body_qi(buf);
   body_q.qx = DL_EXTERNAL_POSE_body_qy(buf);
   body_q.qy = DL_EXTERNAL_POSE_body_qx(buf);
-  body_q.qz = DL_EXTERNAL_POSE_body_qz(buf);
+  body_q.qz = -DL_EXTERNAL_POSE_body_qz(buf);
 
   struct FloatEulers body_e;
   float_eulers_of_quat(&body_e, &body_q);
-  float heading = -body_e.psi; // convert heading in ENU to NED
+  float heading = body_e.psi;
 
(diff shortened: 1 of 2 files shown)


tblaha:
Thanks! Why is this different from `ins_ekf2.cpp`? And what is the purpose of `gps_datalink.c` if the INS also read that message directly?
https://github.com/paparazzi/paparazzi/blob/fe35af73825a32b8a6e594f9053c5ce975358257/sw/airborne/modules/ins/ins_ekf2.cpp#L711-L714

The way I see it, `natnet2ivy.py` has always been sending attitude in ENU, meaning the rotation of a Right-Forward-Up local frame with respect to the East-North-Up frame. This can be converted to Forward-Right-Down-wrt-North-East-Down by simple exchanging x and y, and negating z. (the same way as you would convert a position).
This choice is not configurable in `natnet2ivy.py`, but what _is_ configurable is how that ENU frame is oriented with respect to the observer. But position and attitude should always be consistent in that frame.

tblaha:
Ah, I didnt see the theta=-theta line a few lines down:
https://github.com/paparazzi/paparazzi/blob/fe35af73825a32b8a6e594f9053c5ce975358257/sw/airborne/modules/ins/ins_ext_pose.c#L181-L187

Then it all makes sense, I think and is also the same as @tmldeponti 's change in #3233

tblaha:
Hey Gautier! Perfect, I'm available for testing new stuff as well. On a related note, we have been developing a C++ Natnet client with to goal of keeping the CLI as similar as possible between different target robot/drone operating systems in operation at TU Delft. It's also not completely finished, and I'm not quite sure if it's possible/desirable to include it with paparazzi as well, but I'd like to hear your thoughts.

https://github.com/tudelft/UnifiedOptitrackClients

Fabien-B (approved):

PR #3208, opened 2023-12-14
Remove command_pitch and command_roll and command_yaw

dewagter:
Appears feasible to remove ```COMMAND_ROLL``` ```PITCH``` and ```YAW``` when not used.

PR #3234, opened 2024-02-02
[nav] remove nav_manual mode

gautierhattenberger:
This mode is not really used and not even recommended. It is also not implemented in all firmwares. If a direct call to stabilization command is needed, it should be done with a specific function but not from a built-in instruction from the flight plan.

It was discussed in #3208, but not applied because of `nav_heli_spinup`. A new h_mode 'NONE' allows to disable guidance and stabilization while in NAV mode, which allows to set directly the commands.


dewagter (dismissed):
Great, thanks.

dewagter (approved):
Thanks for fixing the heli-spinup too

PR #3235, opened 2024-02-05
[conf] update Enac rotorcraft so that they work in simulation as well

gautierhattenberger:
- update Enac rotorcraft airframes
- remove dead aircraft
- add jsbsim model files
- update bebop2 example


conf/airframes/ENAC/quadrotor/anton_indi_aruco.xml
```diff
@@ -11,6 +11,8 @@
   </description>
 
   <firmware name="rotorcraft">
+    <!--autopilot name="rotorcraft_autopilot"/-->
```
dewagter, line 14: why?

conf/airframes/ENAC/quadrotor/anton_indi_aruco.xml
```diff
@@ -267,9 +245,9 @@
   </section>
 
   <section name="SIMULATOR" prefix="NPS_">
-    <define name="ACTUATOR_NAMES" value="nw_motor, ne_motor, se_motor, sw_motor" type="string[]"/>
-    <define name="JSBSIM_MODEL" value="simple_x_quad_ccw" type="string"/>
-    <define name="SENSORS_PARAMS" value="nps_sensors_params_default.h" type="string"/>
+    <define name="ACTUATOR_NAMES" value="ne_motor, se_motor, sw_motor, nw_motor" type="string[]"/>
```
dewagter, line 248: why is the order different than most other airframes?

conf/airframes/examples/bebop2_indi.xml
```diff
@@ -22,9 +22,7 @@
     <module name="imu" type="bebop"/>
     <module name="gps" type="ublox"/>
     <module name="stabilization" type="indi_simple"/>
-    <module name="ins" type="ekf2">
-      <define name="INS_EKF2_OPTITRACK" value="true"/>
```
dewagter, line 26: In other airframes you moved the define to the target, but here is it removed?

dewagter (commented):
Thanks for cleaning up. Nice to be able to simulate all airframes.
 - PS: what is the status of ```dshot```? Is it set to 300 by default?

conf/airframes/ENAC/quadrotor/anton_indi_aruco.xml
gautierhattenberger: It was for testing purpose in simulation, but not validated in real flight yet

conf/airframes/ENAC/quadrotor/anton_indi_aruco.xml
gautierhattenberger: Because the motors are not declared in the same order in most of our airframes compared to quadrotors in examples. I don't know why it is like this, but I preferred to change the order of actuators in sim (that was broken) instead of the order in `servos` section (that was working on real platforms).

conf/airframes/examples/bebop2_indi.xml
gautierhattenberger: Yes, since in this airframe, the gps that is declared in `ublox` and not `optitrack`. So the EKF2 flag shouldn't be defined at all. In other airframes, the gps `opitrack` or `datalink` is used for real flights, and is replaced by `nps` for simulation. The EKF2 flag is not compatible with that, so I moved it in `ap` target.

gautierhattenberger:
Normal DShot is working and the default is set to 300. With 600 some ESC can have troubles (and 1200 is officially removed from the specs).
We are still working on the bidirectional dshot with test code that is working and that should be ported to Paparazzi.

dewagter:
> Normal DShot isNormal DShot is working and the default is set to 300. With 600 some ESC can have troubles (and 1200 is officially removed from the specs). We are still working on the bidirectional dshot with test code that is working and that should be ported to Paparazzi.

Is that #2717 ?

dewagter (approved):
LGTM

gautierhattenberger:
> > Normal DShot isNormal DShot is working and the default is set to 300. With 600 some ESC can have troubles (and 1200 is officially removed from the specs). We are still working on the bidirectional dshot with test code that is working and that should be ported to Paparazzi.
> 
> Is that #2717 ?

No. I need to check, but maybe this PR is not relevant anymore.

PR #3236, opened 2024-02-06
[fix] guidance sp was not correct in speed mode

gautierhattenberger:
After #3197, the functions guidance_h_set_xxx are reseting the state so it creates inconsistent behavior when abusing the 'set_pos' to update the gcs display.


dewagter (approved):
I agree not to overwrite ```pos``` in ```speed``` mode. Thanks.

PR #3240, opened 2024-02-16
Add vl53l5cx driver. Based on St ULD.

Fabien-B:
Driver for 8x8 range sensor VL53L5CX.
Based on St ULD.

Also fix some errors in tawakiV2 board.


sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+
+}
+
+
+
+static void thd_lidar_vl53l5cx(void* arg) {
+  chRegSetThreadName("vl53l5cx");
+  VL53L5CX_Configuration* dev = (VL53L5CX_Configuration*) arg;
```
gautierhattenberger, line 128: local `dev` and global `Dev` are a bit confusing. Maybe choose a more explicit name for the global one ?

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
```diff
@@ -0,0 +1,168 @@
...
+
+char* VL53L5CX_ERROR_MSGS[] = {
+  "VL53L5CX_NO_ERROR",
+  "VL53L5CX_NOT_DETECTED",
+  "VL53L5CX_ULD_LOADING_FAILED",
+  "VL53L5CX_SET_RESOLUTION_FAILED",
+  "VL53L5CX_RUNTIME_ERROR",
+};
+
+static void lidar_cb(uint8_t sender_id __attribute__((unused)),
```
gautierhattenberger, line 59: if this function is a datalink report function, why make it an ABI callback ?

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
```diff
@@ -0,0 +1,168 @@
...
+
+  // Create thread
+  Dev.platform.thread_handle = chThdCreateStatic(wa_thd_lidar_vl53l5cx, sizeof(wa_thd_lidar_vl53l5cx),
+                    NORMALPRIO, thd_lidar_vl53l5cx, (void *)&Dev);
+}
+
+void lidar_vl53l5cx_periodic(void)
+{
+
+  if(Dev.platform.thread_handle != NULL) {
```
gautierhattenberger, line 95: run indent script on files (at least the one that are not from ST)

sw/airborne/arch/chibios/modules/lidar/vl53l5cx_api.c
```diff
@@ -0,0 +1,1311 @@
+/**
+  *
+  * Copyright (c) 2021 STMicroelectronics.
```
gautierhattenberger, line 3: These files are supposed to be platform independent. Maybe they should be in `peripherals` directory, on in the modules in non-arch part ?

gautierhattenberger (changes requested):

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
Fabien-B: Ok, done.

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
Fabien-B: Ok, done.

sw/airborne/arch/chibios/modules/lidar/vl53l5cx_api.c
Fabien-B: Ok, done.

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
Fabien-B: It was meant to be an example. Should I remove it ?
How can I document the fact that this module's output is an ABI message ?

gautierhattenberger: The description node in module's xml is probably the best place.
For debug message, I'd rather not leave it commented. Either remove it or make it optional with a flag to activate.

gautierhattenberger:
I think you also need to specify the correct `arch` attribute for the test node

sw/airborne/arch/chibios/modules/lidar/lidar_vl53l5cx.c
Fabien-B: Ok

Fabien-B:
> I think you also need to specify the correct `arch` attribute for the test node

Adding `arch` to the test do not help. Its because the compiled file is not arch dependent, but the included file is.
I don't known exactly how the build system is working, but I think the problem come from the test.

gautierhattenberger:
You can try to include `modules/lidar/vl53l5cx_platform.h`, since the include path in test script it `arch/<arch_name>` (https://github.com/paparazzi/paparazzi/blob/master/tests/modules/test_modules.py#L148)

Fabien-B:
Not working, because it's a `file`, and not a `file_arch`.

gautierhattenberger:
The i2c modification is not really a bug fix, but more like a hack to use the i2c from a thread with active way by skipping the normal transaction queue. So for the long term I would even prefer an other option, but we'll see. The reasons for the test to file are identified, we just need to move some code and files. It should be done soon.

gautierhattenberger (approved):

PR #3242, opened 2024-02-20
Use altitude waypoints for EKF2

EwoudSmeur:
As https://github.com/paparazzi/paparazzi/pull/3153 has gone a bit stale, here a quick fix for EKF2 altitude.

EKF2 does not provide real NED coordinates, but vertically the real altitude is put in the PPRZ NED coordinates. That means that currently if you fly long distance, the waypoints become lower. If you use an INS that outputs real NED coordinates, this doesn't happen.

This can be solved by instead of providing the NED coordinates of waypoints, give the altitude, which is already available from the LLA representation.

### DIFF
--- a/conf/modules/ins_ekf2.xml
+++ b/conf/modules/ins_ekf2.xml
@@ -25,7 +25,7 @@
     <define name="INS_EKF2_FLOW_POS_X" value="0" description="Flow sensor X offset from CoG position [m]"/>
     <define name="INS_EKF2_FLOW_POS_Y" value="0" description="Flow sensor Y offset from CoG position [m]"/>
     <define name="INS_EKF2_FLOW_POS_Z" value="0" description="Flow sensor Z offset from CoG position [m]"/>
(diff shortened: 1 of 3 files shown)


EwoudSmeur:
@Dennis-Wijngaarden tested and it works, so good to merge?

Dennis-Wijngaarden:
> @Dennis-Wijngaarden tested and it works, so good to merge?

Yes, tested and it works

EwoudSmeur:
Regarding #3091, I think it depends on the distances we want to be compatible with. This solution with EKF2 is actually quite ok for long distance, as long as you don't cover _many_ degrees latitude or longitude (so flying at the poles would not be ideal).

What I think is worse if that now we're not consistent with the definitions. There is the 'strict' NED (used by most INS) and the curved NED (used by EKF2). Also the implications for attitude functions like stateGetNedToBodyEulers are not entirely clear.

PR #3249, opened 2024-03-01
Update natnet SDK.

Fabien-B:


gautierhattenberger (approved):

alessandro-mancinelli:
Hi Fabien,

Unfortunately, we are experiencing problems using the OptiTrack pose estimation with Paparazzi and Motive 2. Reverting this commit solved the problem. Did you test the changes with Motive 3, or is this perhaps an issue only TUDelft is experiencing?

gautierhattenberger:
Our current version of Motive is 3.1.0 Beta2. Even if this new natnet2ivy is not the most stable, without these changes it is not working at all for us. What kind of problems do you have ?

alessandro-mancinelli:
image
That's the output of the natnet2ivy.py script with the -v flag up. It is not able to properly identify the stream coming from motive and shuts down the interface. Perhaps the pose packets are different in the old SDK. Could we somehow make a flag (like the current "-o" flag) to allow backward compatibility? 
Probably we could also solve by purchasing a motive 3.x license but it's nice to have the capability of rolling back and use motive 2.x for any reason.

PR #3250, opened 2024-03-01
fix omega discontinuous

EwoudSmeur:
There were jumps in omega (reference heading rate), because in one line the current roll angle is used, and in this other piece of code the commanded angles are used. This lead to jumps in the feed forward rate command.

With everything in current Euler angles this should be solved.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
@@ -473,8 +473,8 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
   // Use the current roll angle to determine the corresponding heading rate of change.
   float coordinated_turn_roll = eulers_zxy.phi;
 
-  if( (guidance_euler_cmd.theta > 0.0f) && ( fabs(guidance_euler_cmd.phi) < guidance_euler_cmd.theta)) {
-    coordinated_turn_roll = ((guidance_euler_cmd.phi > 0.0f) - (guidance_euler_cmd.phi < 0.0f)) * guidance_euler_cmd.theta;
+  if( (eulers_zxy.theta > 0.0f) && ( fabs(eulers_zxy.phi) < eulers_zxy.theta)) {
+    coordinated_turn_roll = ((eulers_zxy.phi > 0.0f) - (eulers_zxy.phi < 0.0f)) * eulers_zxy.theta;
   }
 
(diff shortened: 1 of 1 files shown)


gautierhattenberger (dismissed):
I fine with merging this, but could you also add a comment explaining what these lines are suppose to do and in which situation ? It is not so obvious to me. Thanks

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -473,8 +473,8 @@ struct StabilizationSetpoint guidance_indi_run(struct FloatVect3 *accel_sp, floa
   // Use the current roll angle to determine the corresponding heading rate of change.
   float coordinated_turn_roll = eulers_zxy.phi;
 
-  if( (guidance_euler_cmd.theta > 0.0f) && ( fabs(guidance_euler_cmd.phi) < guidance_euler_cmd.theta)) {
-    coordinated_turn_roll = ((guidance_euler_cmd.phi > 0.0f) - (guidance_euler_cmd.phi < 0.0f)) * guidance_euler_cmd.theta;
+  if( (eulers_zxy.theta > 0.0f) && ( fabs(eulers_zxy.phi) < eulers_zxy.theta)) {
+    coordinated_turn_roll = ((eulers_zxy.phi > 0.0f) - (eulers_zxy.phi < 0.0f)) * eulers_zxy.theta;
```
Fabien-B, line 477: Just a personal opinion:
If I understood correctly, the line 477 is equivalent to this snippet of code:
```C
if(eulers_zxy.phi > 0.0f) {
    coordinated_turn_roll = eulers_zxy.theta;
} else {
    coordinated_turn_roll = -eulers_zxy.theta;
}
```
If so, can we stop doing this kind of code style in favor of more "readable" code ?
The performance gain is most probably negligible, specially if compiled with optimizations.

EwoudSmeur: Thanks for the comment, I agree it is more readable like that!

gautierhattenberger (approved):

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
gautierhattenberger: Even if the roll = 0 case is not explicit, this is a case that should never happen in practice due to noise. My only concern is maybe, that due to noise, if roll is close enough to 0, the final command might oscillate between +/- theta. But again, in practice, it shouldn't be a real issue.

PR #3253, opened 2024-03-19
Bidirectionnal DShot for H7

Fabien-B:
Add support for bidirectionnal DShot telemetry on H7 based autopilots.
- [x] Test on the TawakiV2 (SRVB connector).
- [x] Test that it doesn't break normal DShot on TawakiV1 (F7 based)
- [x] Test with SRVA connector on TawakiV2
- [x] Better automagic GPT driver activation


conf/modules/actuators_dshot.xml
```diff
@@ -19,8 +20,15 @@
     <file name="actuators_dshot.h"/>
   </header>
   <makefile>
+    <configure name="DSHOT_BIDIR" default="FALSE"/>
+    <define name="DSHOT_BIDIR" value="$(DSHOT_BIDIR)"/>
+    <define name="USE_GPT7" value="TRUE" cond="ifeq ($(DSHOT_BIDIR),TRUE)"/>
```
Fabien-B, line 25: @gautierhattenberger How could it be automagic?

Fabien-B:
- done
- regular dshot was already complicated, it's just marginally more complicated :smile:.
I have still some things to address to make it simpler to use
- Added a comment in the module xml. Bidirectionnal DSHOT transmit rpm telemetry at high frequency, but do not transfers voltage and current. UART telemetry is still useful if you need this data.
EDT (extended DSHOT telemetry) allows to transmit voltage and current, but almost no ESC firmware supports it...
- thanks !

Fabien-B:
@gautierhattenberger Is it okay to do something like that ? https://github.com/paparazzi/paparazzi/pull/3253/commits/7fc6c6b95122be00bbc384be5f71b8d1e9f25788
There is a default configuration for the first DSHOT, and for the other(s) DSHOT, the user must define, e.g :
```XML
<define name="USE_GPT8" value="TRUE"/>
<define name="DSHOT2_BIDIR_GPT" value="GPTD8"/>
```

Fabien-B:
Ready to merge!

gautierhattenberger (approved):

PR #3254, opened 2024-03-20
[chibios] Update and add Pixhawk 6X autopilot

fvantienen:
- Updated chibios for support for newer processors
- Restructured `mcuconf.h` and made a common h7 conf which can be used
- Added support for inverted LEDs in the common board
- Added the Pixhawk V6X autopilot

Still needs some more testing and cleanup


Pixhawk6X_bl.bin
gautierhattenberger, line 1: root folder is probably not the best place for these binary files

sw/airborne/arch/chibios/STM32H753xI.ld
Fabien-B, line 1: Could you use this file ? https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/arch/chibios/STM32H743xI.ld
If this is not possible, can you mirror the memory section `ram0nc` ?

gautierhattenberger:
@fvantienen see comment from @Fabien-B, then can be merged

PR #3255, opened 2024-03-25
[board] fix crazyfly board files

### LINKED ISSUE #3252: Crazyflie ChibiOS firmware crashing
## The problem
Crazyflie 2.1 does not work. All LEDs except one are lit without heartbeat (there are 2 red LEDs, 2 green LEDs and 2 blue LEDs. If I'm not mistaken, the blue LED corresponding to the system timing is the one turned off)

My supervisor @noether suggested me to open an issue here mentioning @Fabien-B. I would also thank anyone trying to flash the firmware to a Crazyflie they have lying around to see if the problem is on my end and not in Paparazzi or ChibiOS.

## What I tried
Flashing the firmware on one Crazyflie with Debian 12 and Ubuntu 22, two different laptops. In both laptops I tried ap, test_gpio, test_sys_time and test_led. I also tried the profile from ```conf/airframes/test_boards``` and ```conf/airframes/OPENUAS```.  Changing these XML (for example, removing unnecessary modules) result in same behaviour.   

## What I found out
I recently debugged with a STLink V2 and OpenOCD the Crazyflie with ```test_led``` and ```ap``` firmwares flashed.

- ~~In the case of ```test_led``` the Crazyflie ends up stuck in an infinite loop in the function ```idle_thread``` [(link)](https://chibiforge.org/doc/21.11/full_rm/group__instances.html#ga0ea7aefc26b9e8ff69c343baa2acb0ef) from ChibiOS (```chinstances.c``` file)~~ 
- ~~I also recently tried ```test_gpio``` and ```test_sys_time_timer```. Same behaviour as ```test_led```~~
- ~~For ```ap```, it ends up in a ```unhandled_exception``` function~~
- There is heartbeat on the ```test_led``` after some changes to the ```chibios_test_led.c``` file. See comment below. 
- There seems to be no context switches to the threads using the UART6, which are responsible of sending and receiving data from the nRF chip in the Crazyflie. Every other thread seems to work normally (idle, sys_time and I2C). No LED heartbeat yet.

~~In both cases, OpenOCD ```reset halt``` command does not work, so I can only either reset the board or halt the board manually after a reset, so I can't trace how both firmwares end up in those functions.~~ 
I'm now using Cortex-Debug plugin for VSCode. It does an incredible job debugging the Crazyflie and ChibiOS.

Recently, I also tried using Paparazzi branch v6.2, since has minor modifications since this commit https://github.com/paparazzi/paparazzi/commit/45b020d678b918a74c00d407100faa83930da40d, but same behaviour as latest master

## Will try soon
- ~~Try to fix OpenOCD/GDB for actual, proper debugging~~ I've been using Cortex-Debug plugin for VSCode, works way better
- ~~Debug test_sys_time~~
- ~~Debug test_gpio~~
- ~~Try to compile the firmware with Paparazzi around this commit (seems to be the most recent Crazyflie commit that seems to still work): https://github.com/paparazzi/paparazzi/commit/45b020d678b918a74c00d407100faa83930da40d~~ 

### Side note 1
I've managed to install Paparazzi on Debian 12 with some minor adaptations. Should I open a new issue or PR in order to help porting Paparazzi to Debian 12 (could also work for Debian 11)?

### Side note 2
The Python script for the Crazyradio needs ```cflib``` . There is little documentation about this, I believe there should be a little more information elsewhere about how to use the Crazyflie. Just a minor suggestion.

gautierhattenberger:
- use correct main file with chibios
- fix pin definitions

it should fix #3252


Pelochus:
Fixed LEDs, but same behaviour with the radio (receives empty packets, so I'm assuming correct communication within the radios but not the Crazyflie radio with the STM32 chip)

I've explained it better, along with some possible causes, in the issue https://github.com/paparazzi/paparazzi/issues/3252

gautierhattenberger:
I'm merging this as it already solves important issues. The other problem is not yet clearly identified so we can keep the related issue opened.

PR #3256, opened 2024-03-26
Restructure INDI to limit estimated disturbances

EwoudSmeur:
I restructured the INDI controller, keeping the workings exactly the same. The benefit of the new structure is that we replace the increments by a disturbance estimate, such that in the control allocation we allocate the complete control objective (specific angular accelerations and thrust), not increments. This allows us to limit the specific angular accelerations (read: moments) that we allow the drone to allocate.

As an example, the rotating wing drone has a tail, which will automatically reduce the sideslip. We don't want the drone to stick to a specific heading, and in doing that 'fight' with the tail. With this update, we can limit the estimated disturbance such that we only control yaw to a limited extent.

What is also included is a feed forward on the yaw rate from RC if the yaw attitude gain is 0, otherwise it would not be possible to control yaw anymore with the sticks if you want to fly without a fixed yaw attitude setpoint.

### DIFF
--- a/conf/airframes/tudelft/rot_wing_v3b.xml
+++ b/conf/airframes/tudelft/rot_wing_v3b.xml
@@ -203,21 +203,21 @@
 
     <!-- Can bus 1 actuators -->
     <servos driver="Uavcan1">
-        <servo no="0"  name="MOTOR_FRONT"     min="0"     neutral="1000" max="8191"/>
-        <servo no="1"  name="MOTOR_RIGHT"     min="0"     neutral="1000" max="8191"/>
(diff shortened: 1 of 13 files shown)


EwoudSmeur:
Hi @gautierhattenberger, the impact of this particular PR should be minimal, at least that is the idea, but I'm happy to discuss how to best approach merging the PRs. I can do today in the afternoon, Friday afternoon, and Monday, just let me know when you prefer!

gautierhattenberger:
Thanks, replied by email

EwoudSmeur:
Now without the yaw stuff!

gautierhattenberger (approved):

PR #3259, opened 2024-04-07
[rot_wing nps] add more aerodynamic drag in rotwing simulation

Dennis-Wijngaarden:
Added aerodynamic drag in rotating wing simulation

### DIFF
--- a/conf/simulator/jsbsim/aircraft/rotwing3.xml
+++ b/conf/simulator/jsbsim/aircraft/rotwing3.xml
@@ -827,6 +827,7 @@
           <value>47.9</value> <!-- Conversion to pascals -->
           <value>0.0151</value> <!-- CD x Area (m^2) -->
           <value>0.224808943</value> <!-- N to LBS -->
+          <value>3</value>
         </product>
       </function>
     </axis>


conf/simulator/jsbsim/aircraft/rotwing3.xml
```diff
@@ -827,6 +827,7 @@
           <value>47.9</value> <!-- Conversion to pascals -->
           <value>0.0151</value> <!-- CD x Area (m^2) -->
           <value>0.224808943</value> <!-- N to LBS -->
+          <value>3</value>
```
gautierhattenberger, line 830: It is a bit strange to add an extra parameter rather than multiplying the CD coef by 3. In addition, without comments it might be a problem in the future to understand where this value is coming from. I recommend to change the CDxArea by the proper value, unless there is a good reason to do otherwise.

gautierhattenberger (changes requested):

conf/simulator/jsbsim/aircraft/rotwing3.xml
dewagter: Changed!

dewagter (approved):
OK!

PR #3260, opened 2024-04-07
[guidance_indi_hybrid] Added secondary airspeed filtering in guidance

Dennis-Wijngaarden:
Added a secondary airspeed filter i guidance_indi_hybrid to for example filter airspeed more in forward flight to reduce vibration in actuator command

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
@@ -118,6 +118,9 @@ bool take_heading_control = false;
 
 bool force_forward = false;
 
+bool guidance_indi_secondary_airspeed_filtering = false;
+
+
 struct FloatVect3 sp_accel = {0.0,0.0,0.0};
 #ifdef GUIDANCE_INDI_SPECIFIC_FORCE_GAIN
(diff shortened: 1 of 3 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
```diff
@@ -175,6 +182,7 @@ Butterworth2LowPass roll_filt;
 Butterworth2LowPass pitch_filt;
 Butterworth2LowPass thrust_filt;
 Butterworth2LowPass accely_filt;
+Butterworth2LowPass secondary_airspeed_filt;
```
gautierhattenberger, line 185: I think this "secondary_airspeed" name is misleading. It a regular filter on the airspeed data (that can be filter or not depending on the driver), not a second source of measures. So I recommend to remove the `secondary` from all names.

gautierhattenberger (changes requested):

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid.c
dewagter: Agreed: this was changed into ```guidance_indi_airspeed_filt```

dewagter (approved):
LGTM

PR #3261, opened 2024-04-08
Oneloop Controller: some updates

tmldeponti:
This pull request contains some updates and improvements to the oneloop controller. 

1. Airframes for the rotating wing drone C
2. Simulation jsbsim for the RW3C
3. Generated autopilots to run the oneloop with either INDI or ANDI
4. Some slight improvements in nav hybrid rotorcraft: cleaner defines and specific settings for indoor flight

In the future I plan to open another pull request to better integrate the oneloop controller with the existing rotating wing drone modules (rotwing state and effectiveness scheduling). Please refer to the issue: https://github.com/paparazzi/paparazzi/issues/3237#issue-2130261570

PR #3265, opened 2024-04-11
Use Commands instead of actuator_pprz 

tmldeponti:
This Pull request is the first of a series of pull requests to make sure that modules such as Rotwing State and Rotwing Effectveess scheduling can be used by both the onleoop controller as well as a more conventional INDI cascaded controller. 
This specific pull requests aims to solve:

1. Have shared modules which do not use actuator_pprz hardcoded references ( e.g. actuator_pprz[6] ) . This can be problematic as different actuator orders will not work. 
2. Have a way to clearly specify how many actuators there are and what type of actuator they are (e.g. Real, Virtual, Passive and Other) so that variables such as "INDI_NUM_ACT" are not used outside of INDI files and cannot mistakely be assigned a wrong value. 

With @dewagter we thought to revert back the controllers to assign commands to the commands[ ] array rather than actuator_pprz[ ]. Therefore, one would define a command per actuator to be controlled and use that in the command laws. 
It would then be possible to make sure that the shared modules are always using the correct actuator index (e.g. commands[ROT_MECH])

I have also changed the gen_airframe file to add the attribute "actuator_type" to the commands in the airframe, and have counters for each sub-type. Then one can replace the hardcoded definitions such as "ANDI_NUM_ACT" or "INDI_NUM_ACT" with the gnerated definitions from the airframe (See how i have changed my oneloop controller). This is my first time writing in OCALM, the changes i made work but i am open to improvements.
This is what is generated from the airframe:
```c
#define COMMAND_MOTOR_FRONT 0
#define COMMAND_MOTOR_RIGHT 1
#define COMMAND_MOTOR_BACK 2
#define COMMAND_MOTOR_LEFT 3
#define COMMAND_MOTOR_PUSHER 4
#define COMMAND_ROLL 5
#define COMMAND_PITCH 6
#define COMMAND_ROT_MECH 7
#define COMMAND_YAW 8
#define COMMAND_THRUST 9
#define COMMANDS_NB_OTHER 2
#define COMMANDS_NB_PASSIVE 1
#define COMMANDS_NB_VIRTUAL 2
#define COMMANDS_NB_REAL 5
#define COMMANDS_NB 10
#define COMMANDS_FAILSAFE {0,0,0,0,0,0,0,0,0,0}
``` 
An example commands section of an airframe would then look something like: 

```c
   <commands>
        <!-- Real actuators commands -->
        <axis name="MOTOR_FRONT"  actuator_type ="REAL" failsafe_value="0"/> <!-- IDX 0--> 
        <axis name="MOTOR_RIGHT"  actuator_type ="REAL" failsafe_value="0"/> <!-- IDX 1-->
        <axis name="MOTOR_BACK"   actuator_type ="REAL" failsafe_value="0"/> <!-- IDX 2-->
        <axis name="MOTOR_LEFT"   actuator_type ="REAL" failsafe_value="0"/> <!-- IDX 3-->
        <axis name="MOTOR_PUSHER" actuator_type ="REAL" failsafe_value="0"/> <!-- IDX 4-->
        <!-- Virtual actuators commands -->
        <axis name="ROLL"         actuator_type ="VIRTUAL" failsafe_value="0"/> <!-- ID X 5-->
        <axis name="PITCH"        actuator_type ="VIRTUAL" failsafe_value="0"/> <!-- IDX 6-->
        <!-- Passive actuators commands -->
        <axis name="ROT_MECH"     actuator_type ="PASSIVE" failsafe_value="0"/> <!-- IDX 7-->
        <!-- Legacy commands-->
        <axis name="YAW"          actuator_type ="OTHER" failsafe_value="0"/>
        <axis name="THRUST"       actuator_type ="OTHER" failsafe_value="0"/>
    </commands>
``` 

I have an airframe modified like this but I need my other PR (#3261) to be approved first.

### DIFF
--- a/sw/simulator/nps/nps_fdm_jsbsim.cpp
+++ b/sw/simulator/nps/nps_fdm_jsbsim.cpp
@@ -316,7 +316,18 @@ static void feed_jsbsim(double *commands, int commands_nb __attribute__((unused)
     FDMExec->GetPropertyManager()->GetNode(property)->SetDouble("", commands[i]);
   }
 #else /* use COMMAND names */
+#ifdef NPS_USE_COMMANDS /*Use generated command names*/
+  char buf[64];
+  const char *names[] = COMMAND_NAMES;
+  string property;
 
+  int i;
+  for (i = 0; i < commands_nb; i++) {
+    sprintf(buf, "fcs/%s", names[i]);
+    property = string(buf);
+    FDMExec->GetPropertyManager()->GetNode(property)->SetDouble("", commands[i]);
+  }
+#else /*Use manually defined commands*/
   // get FGFCS instance
   FGFCS *FCS = FDMExec->GetFCS();
 
@@ -355,6 +366,7 @@ static void feed_jsbsim(double *commands, int commands_nb __attribute__((unused)
   FCS->SetDfCmd(commands[COMMAND_FLAP]);
 #endif /* COMMAND_FLAP */
 
+#endif /* NPS_USE_COMMANDS */
 #endif /* NPS_ACTUATOR_NAMES */
 }
 
--- a/sw/simulator/nps/nps_fdm_jsbsim.cpp
+++ b/sw/simulator/nps/nps_fdm_jsbsim.cpp
@@ -80,6 +80,19 @@
 #define JSBSIM_PATH(_x) _x
 #endif
 
+/* Actuator naming*/
+#ifdef NPS_USE_COMMANDS
+#define NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES TRUE
+#endif
+
+#ifdef NPS_ACTUATOR_NAMES
+#define NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES TRUE
+#endif
+
+#ifndef NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES
+#define NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES FALSE
+#endif
+
 /** Name of the JSBSim model.
  *  Defaults to the AIRFRAME_NAME
  */
@@ -304,23 +317,14 @@ void nps_fdm_set_temperature(double temp, double h)
  */
 static void feed_jsbsim(double *commands, int commands_nb __attribute__((unused)))
 {
-#ifdef NPS_ACTUATOR_NAMES
+#if NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES
   char buf[64];
+#if defined(NPS_USE_COMMANDS)
+  const char *names[] = COMMAND_NAMES; 
+#elif defined(NPS_ACTUATOR_NAMES)
   const char *names[] = NPS_ACTUATOR_NAMES;
+#endif 
   string property;
-
-  int i;
-  for (i = 0; i < commands_nb; i++) {
-    sprintf(buf, "fcs/%s", names[i]);
-    property = string(buf);
-    FDMExec->GetPropertyManager()->GetNode(property)->SetDouble("", commands[i]);
-  }
-#else /* use COMMAND names */
-#ifdef NPS_USE_COMMANDS /*Use generated command names*/
-  char buf[64];
-  const char *names[] = COMMAND_NAMES;
-  string property;
-
   int i;
   for (i = 0; i < commands_nb; i++) {
     sprintf(buf, "fcs/%s", names[i]);
@@ -366,8 +370,7 @@ static void feed_jsbsim(double *commands, int commands_nb __attribute__((unused)
   FCS->SetDfCmd(commands[COMMAND_FLAP]);
 #endif /* COMMAND_FLAP */
 
-#endif /* NPS_USE_COMMANDS */
-#endif /* NPS_ACTUATOR_NAMES */
+#endif /* NPS_AUTOMATIC_JSBSIM_ACTUATOR_NAMES */
 }
 
 /**
--- a/sw/tools/generators/gen_airframe.ml
+++ b/sw/tools/generators/gen_airframe.ml
@@ -31,7 +31,7 @@ open Printf
 open Xml2h
 
 type channel = { min : float; max : float; neutral : float }
-type control = { failsafe_value : int; foo : int; actuator_type : string}
+type control = { failsafe_value : int; foo : int; group : string}
 
 let fos = float_of_string
 let sof = fun x -> if mod_float x 1. = 0. then Printf.sprintf "%.0f" x else string_of_float x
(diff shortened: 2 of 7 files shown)


sw/tools/generators/gen_airframe.ml
```diff
@@ -31,7 +31,7 @@ open Printf
 open Xml2h
 
 type channel = { min : float; max : float; neutral : float }
-type control = { failsafe_value : int; foo : int }
+type control = { failsafe_value : int; foo : int; actuator_type : string}
```
gautierhattenberger, line 34: I think `type` or `group` would be enough for the attribute name. You also need to add it to the DTD file `conf/airframes/airframe.dtd`.

sw/tools/generators/gen_airframe.ml
```diff
@@ -337,9 +337,17 @@ let parse_ap_only_commands = fun out ap_only ->
...
+  let actuator_type = ExtXml.attrib_or_default command "actuator_type" "REAL" in 
+  define_out out command_name (string_of_int no);
+  { failsafe_value = failsafe_value; foo = 0 ; actuator_type = actuator_type }  
+
+let count_commands_by_type commands_params =
+  List.fold_left (fun acc cmd ->
+    let subtype = cmd.actuator_type in
+    let count = try List.assoc subtype acc with Not_found -> 0 in
+    (subtype, count + 1) :: (List.remove_assoc subtype acc)
+  ) [("REAL", 0); ("VIRTUAL", 0); ("PASSIVE", 0); ("OTHER", 0)] commands_params
```
gautierhattenberger, line 350: Is it really necessary to have default categories. The list is not even exhaustive since it should work with any name. I suggest that the initial list is just empty. If the actuator is not defining a type, the name `REAL` will be used by default and only this type will be generated. If a controller needs more, it is up to the user to define more in the airframe file.

sw/tools/generators/gen_airframe.ml
```diff
@@ -337,9 +337,17 @@ let parse_ap_only_commands = fun out ap_only ->
...
   let command_name = "COMMAND_"^ExtXml.attrib command "name" in
-  define_out out command_name (string_of_int no);
   let failsafe_value = int_of_string (ExtXml.attrib command "failsafe_value") in
-  { failsafe_value = failsafe_value; foo = 0}
+  let actuator_type = ExtXml.attrib_or_default command "actuator_type" "REAL" in 
+  define_out out command_name (string_of_int no);
+  { failsafe_value = failsafe_value; foo = 0 ; actuator_type = actuator_type }  
+
+let count_commands_by_type commands_params =
+  List.fold_left (fun acc cmd ->
```
gautierhattenberger, line 346: You could use `Array.fold_left` to avoid an extra conversion between list and array type

gautierhattenberger (commented):
no so bad for a first try with Ocaml

sw/tools/generators/gen_airframe.ml
tmldeponti: Changed to group and added it to the dtd

sw/tools/generators/gen_airframe.ml
tmldeponti: Thanks, done that!

tmldeponti:
After a talk with @dewagter we realized that this is also a good opportunity to streamline the use of the defined commands in the airframe directly in NPS. Currently we provide in the airframe redundant info like the name and number of actuators: 
```c
        <define name="ACTUATOR_NAMES"  value="front_motor, right_motor, back_motor, left_motor, pusher" type="string[]"/> 
        <define name="JSBSIM_MODEL"    value="rotwingv3c_SI" type="string"/>
        <define name="SENSORS_PARAMS"  value="nps_sensors_params_default.h" type="string"/>
        <define name="COMMANDS_NB"     value="5"/> 
        <define name="USE_COMMANDS"    value="TRUE"/>
        <define name="JS_AXIS_MODE"    value="4"/>
    </section>
``` 
This can be a source of errors, as for example one can provide the names of the actuators in the wrong order. The proposed changes first allow for the automatic generation of an array for strings from the commands names. The result is something like:
```c
#define COMMAND_NAMES { "MOTOR_FRONT", "MOTOR_RIGHT", "MOTOR_BACK", "MOTOR_LEFT", "MOTOR_PUSHER", "ROLL", "PITCH", "ROT_MECH", "YAW", "THRUST" }
``` 
This array is then used in NPS to set always in the right order the actuator names when "NPS_USE_COMMANDS" is defined, resulting in a more streamlined nps section in the airframe:
```c
        <define name="JSBSIM_MODEL"    value="rotwingv3c_SI" type="string"/>
        <define name="SENSORS_PARAMS"  value="nps_sensors_params_default.h" type="string"/>
        <define name="USE_COMMANDS"    value="TRUE"/>
        <define name="JS_AXIS_MODE"    value="4"/>
    </section>
``` 
Finally the JSBSIM model should be modified so that the properties match the commands names, e.g.:
```c
    <property>fcs/MOTOR_FRONT</property>
    <property>fcs/MOTOR_FRONT_lag</property>
    <property>fcs/MOTOR_FRONT_d</property>
    <property>fcs/MOTOR_RIGHT</property>
    <property>fcs/MOTOR_RIGHT_lag</property>
    <property>fcs/MOTOR_RIGHT_d</property>
    <property>fcs/MOTOR_BACK</property>
    <property>fcs/MOTOR_BACK_lag</property>
    <property>fcs/MOTOR_BACK_d</property>
    <property>fcs/MOTOR_LEFT</property>
    <property>fcs/MOTOR_LEFT_lag</property>
    <property>fcs/MOTOR_LEFT_d</property>
``` 
I will be able to push a sample airframe and JSBSIM model for the rotating wing drone C once PR (https://github.com/paparazzi/paparazzi/pull/3261) is approved.

sw/tools/generators/gen_airframe.ml
```diff
@@ -338,16 +338,25 @@ let parse_ap_only_commands = fun out ap_only ->
...
+  ) [] commands_params
+
+let generate_command_names = fun out commands ->
+  let command_names = ref [] in
+  Array.iter (fun axis ->
+    let name = ExtXml.attrib axis "name" in
+    command_names := ("\"" ^ name ^ "\"") :: !command_names
+  ) commands;
+  let command_array = String.concat ", " (List.rev !command_names) in
+  fprintf out "#define COMMAND_NAMES { %s }\n\n" command_array
```
gautierhattenberger, line 359: You can use `map` function for this
```suggestion
  let command_names = Array.map (fun axis -> "\"" ^ (ExtXml.attrib axis "name") ^ "\"") commands in
  let command_names = String.concat ", " (Array.to_list command_names) in
  fprintf out "#define COMMAND_NAMES { %s }\n\n" command_names
```

sw/simulator/nps/nps_fdm_jsbsim.cpp
```diff
@@ -316,7 +316,18 @@ static void feed_jsbsim(double *commands, int commands_nb __attribute__((unused)
...
+  char buf[64];
+  const char *names[] = COMMAND_NAMES;
+  string property;
 
+  int i;
+  for (i = 0; i < commands_nb; i++) {
+    sprintf(buf, "fcs/%s", names[i]);
+    property = string(buf);
+    FDMExec->GetPropertyManager()->GetNode(property)->SetDouble("", commands[i]);
+  }
```
gautierhattenberger, line 329: This is duplicated, maybe there is a way to avoid this ?

gautierhattenberger:
I don't mind having command names also used for NPS. If it works for specific frames, maybe you can harmonise other airframes later for the generic jsbsim models.

sw/simulator/nps/nps_fdm_jsbsim.cpp
tmldeponti: Done!

dewagter (approved):
Good step towards using COMMANDS to have virtual, real or passive commands driven independently.
I particularly like not needing to specify the simulator order of commands.
This should not damage anything, so it can be merged.

gautierhattenberger (approved):

PR #3271, opened 2024-04-30
Rotwing guidance bank fix

Dennis-Wijngaarden:
- Limit lateral acceleration in guidance_indi_hybrid
- Fix bank angle in WLS outer loop(s)

@EwoudSmeur, please verify

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
@@ -114,7 +114,7 @@ struct FloatMat33 Ga_inv;
 struct FloatVect3 control_increment; // [dtheta, dphi, dthrust]
 
 float filter_cutoff = GUIDANCE_INDI_FILTER_CUTOFF;
-float guidance_indi_max_bank = GUIDANCE_H_MAX_BANK;
+float guidance_indi_max_bank = DegOfRad(GUIDANCE_H_MAX_BANK);
 
 float time_of_accel_sp_2d = 0.0;
 float time_of_accel_sp_3d = 0.0;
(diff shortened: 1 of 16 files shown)


sw/airborne/modules/checks/preflight_checks.c
```diff
@@ -43,8 +43,13 @@
 #define PREFLIGHT_CHECK_INFO_TIMEOUT 5
 #endif
 
-static struct preflight_check_t *preflight_head = NULL;
+#ifdef PREFLIGHT_CHECK_BYPASS
+bool preflight_bypass = PREFLIGHT_CHECK_BYPASS;
+#else
 bool preflight_bypass = FALSE;
+#endif
```
gautierhattenberger, line 50: For coherence, can it be written like this ?
```suggestion
#ifndef PREFLIGHT_CHECK_BYPASS
#define PREFLIGHT_CHECK_BYPASS FALSE
#endif

bool preflight_bypass = PREFLIGHT_CHECK_BYPASS;
```

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
```diff
@@ -118,7 +118,7 @@ struct FloatMat33 Ga_inv;
 struct FloatVect3 control_increment; // [dtheta, dphi, dthrust]
 
 float filter_cutoff = GUIDANCE_INDI_FILTER_CUTOFF;
-float guidance_indi_max_bank = GUIDANCE_H_MAX_BANK;
+float guidance_indi_max_bank = DegOfRad(GUIDANCE_H_MAX_BANK);
```
gautierhattenberger, line 121: I don't see why this should be converted to degrees, to convert it again to radians later

gautierhattenberger (dismissed):

sw/airborne/firmwares/rotorcraft/guidance/guidance_indi.c
dewagter: The setting in the GCS is otherwise in radians.

gautierhattenberger:
You have `unit` and `alt_uni` attributes for settings to tell how it should be displayed and sent from the GCS

dewagter:
PR updated!

EwoudSmeur:
Thanks @dewagter and @gautierhattenberger for looking into this, didn't know the setting could also be configured with an alt_unit. Like this it works!

gautierhattenberger:
Some stuff seem to be merged with #3274 already. Can you rebase on master to check ? Or we just believe in git...

dewagter:
Rebased on master works fine.

dewagter (approved):
Bypass of preflight in NPS is OK.
Setting MAX_BANK is now done the clean way.
Guidance bank bug is resolved.

PR #3275, opened 2024-05-23
add nav_hybrid module to rotorcraft disco

EwoudSmeur:


dewagter (approved):
Great that the simulation is fixed. Tnx.

PR #3276, opened 2024-05-24
[pfc] Add ground checklists

fvantienen:
This adds the option for checklists which are shown in the Paparazzi GCS, see https://github.com/paparazzi/PprzGCS/pull/25


dewagter (approved):
Perfect

PR #3286, opened 2024-06-03
[supervision] remove VT100 escape codes.

Fabien-B:
Remove VT00 escape codes from console.
Still not ideal, but better.

Before:
image

After:
image


Fabien-B:
This problem occurs if the program that runs in the console output color escape sequences.
For example, the STM32 Cube Programmer.

PR #3290, opened 2024-06-05
[ocaml] use glib from gtk3 if gtk2 is not found

gautierhattenberger:
gtk3 doesn't support all features of gtk2 (or have a different interface), so if only gtk3 is found, GUI tools are not compiled, only CLI tools based only on glib.
If no gtk found, tools based on glib/gtk are not compiled.

This should help to use old CLI tools on ubuntu 24.04 until they are replaced.


Fabien-B (approved):
Tested on Ubuntu 24.04.

PR #3291, opened 2024-06-06
[supervision] Use PyQtWebEngine for the doc webview.

Fabien-B:
Use PyQtWebEngine for the doc webview, QtWebKit beeing deprecated.


Fabien-B:
Sorry, I don't know why it changed. Fixed in #3293.

Issue #3294, opened 2024-06-08
Compile all in master


dewagter:
Still 3 airframes that do not compile:
 - [x] bebop_flip
 - [x]  quadshot
 - [x] rot_wing_25kg

PR #3296, opened 2024-06-09
Fix collection errors from ci

gautierhattenberger:
A couple of fixes for
- missing changes in guidance_hybrid #3294 
- bebop flip (removed) #3294 
- natnet python (after #3249)
- warning while compiling with chibios http://forum.chibios.org/viewtopic.php?t=5830

The issues with rotwing airframe (see #3294) is a problem with the definition of servo index in simulation. This should be fixed by someone from TUDelft.


dewagter:
Double of #3298 and fixing #3294

dewagter (approved):

PR #3248, opened 2024-02-27
[control] rework the stabilization for rotorcraft

gautierhattenberger:
Introducing several changes to the control stask for the stabilization:
- new structures: stabilization, stab (att and/or rates) and thrust (vector)
- more functional approach: stabilization receives the stab and thrust setpoints from guidance and fill command vector at the end
- modes and sub-modes are redefined (split stab and guidance modes), some AP modes are removed (FLIP and MODULE)
- the AP_MODULE mode is replaced by generated autopilot
- the ABI hack for guidance indi is not needed anymore
- AP generator supports a new 'store' attribute to use return values of functions
- cleaning of the stabilization API (for setter functions)

Most of the changes should have been tested in simulation, but not much in real flight. Even if it is a quite big changes, I think it is an overall simplification of the control architecture.

PR #3297, opened 2024-06-10
[indi] fix thrust filtering after #3248

gautierhattenberger:
guidance indi thrust is broken when using specific force gain


Fabien-B (approved):

dewagter:
@EwoudSmeur can you also review this?

PR #3303, opened 2024-06-13
Herelink command

dewagter:
@gautierhattenberger : any idea on how not to have the paparazzi sw path automatically added in front of commands? How to store ```vlc``` and ```adb shell``` in controlpanel?


gautierhattenberger:
It should work if you add `$` before the command, see https://github.com/paparazzi/paparazzi/blob/master/conf/tools/http_server.xml#L1
Also, you should remove programs from the custom control panel files if they use default parameters, the tools in `conf/tools` are then used.

dewagter:
@gautierhattenberger Thanks a lot. I now also cleaned all TUDelft control_panels

gautierhattenberger (approved):

PR #3305, opened 2024-06-14
Python virtual environment

Fabien-B:
I propose to ease the use of a python virtual environment.
It will allow a much more clear python dependencies management.

I did not integrate it in the Makefile to allow testing it without risking breaking anything.

To use it, create en venv: `python3 setup.py`
Then source `pprzEnv/bin/activate`.
It can be done in your .bashrc: `source $PAPARAZZI_HOME/pprzEnv/bin/activate`

What are you though on that PR ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@
 .cache.xml
 .directory
 
+# python venv
+pprzEnv/
+
 # Eclipse IDE project files
 *.cproject
 *.project
--- a/conf/system/term_init.sh
+++ b/conf/system/term_init.sh
@@ -1,0 +1,19 @@
+
+source ~/.profile
+
+if [ -d "$PAPARAZZI_HOME/pprzEnv" ]; then
+    source $PAPARAZZI_HOME/pprzEnv/bin/activate
+fi
+
+shell=$(echo $SHELL | cut -d / -f 3)
+
+if [ "$shell" = "bash" ]; then
+    # enter bash specific commands here
+    echo $shell > /dev/null
+elif [ "$shell" = "zsh" ]; then
+    # enter zsh specific commands here
+    echo $shell > /dev/null
+else
+    # fallback commands
+    echo $shell > /dev/null
+fi;
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,52 @@
 #!/bin/bash
 
-sudo apt-get install -f -y python3 python3-pyqt5
+USE_VENV=true
+BASHRC_SOURCE_VENV=false
+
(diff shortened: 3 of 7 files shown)


dewagter:
@Fabien-B looks like a good idea because more and more people have python conflicts. However, in the current version it looks like you need quite some manual work to get things working. Can that be improved? It would also (temporary) make switching branches quite hard doesn't it?

Fabien-B:
@dewagter I don't think a 2 step process that you only have to do **once** is "*quite some manual work*" :smiley: .
I can be added to the makefile, but it's maybe best to have some return on experience before doing that.
The venv files are not tracked by git, so there is absolutely no problem switching branches.
The only annoying thing is that on a branch where the `.gitignore` have not been updated, git will show all venv files as "untracked". But it's the only issue I can see.

dewagter (dismissed):
Ideally, it would be added to the ```install.sh``` script. And it would need some communication.

Fabien-B:
Something like this: a04674bd8bff370a6d90b395068988b18f7e04f1 ?

dewagter:
> Something like this: [a04674b](https://github.com/paparazzi/paparazzi/commit/a04674bd8bff370a6d90b395068988b18f7e04f1) ?

Great thank! Can this be run several times, or will keep adding bash variables?

dewagter (dismissed):
Ideally the install-window can be opened several times.

dewagter (dismissed):

Fabien-B:
I added a guard to prevent creating the venv from itself (which can't be done), and adding the source to the .bashrc.
However, if the venv is deactivated, it is still possible to add the source multiple times in .bashrc.
As this is optional, defaulting to `false`, and harmless anyway, I would argue this is a reasonable compromise.

Fabien-B:
Rebased on master

Fabien-B:
@dewagter If it's ok for you, I replaced `lsb_release` by `distro` in the install script because `lsb_release` is not installable from pypi.
It works pretty much the same.

dewagter:
> @dewagter If it's ok for you, I replaced `lsb_release` by `distro` in the install script because `lsb_release` is not installable from pypi. It works pretty much the same.

OK for me. I would just like the script to adapt to the distribution. For me, a project should be installed with ```git clone CODE``` followed by and by ```install``` without needing a full WIKI to survive.

Fabien-B:
Can we merge this ?

dewagter:
Yes

dewagter (approved):

PR #3309, opened 2024-06-14
[stabilization] fix reporting of attitude setpoint in hybrid mode

gautierhattenberger:
also add stab message to flight recorder

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
@@ -212,7 +212,6 @@ struct FloatRates angular_rate_ref = {0., 0., 0.};
 float angular_acceleration[3] = {0., 0., 0.};
 float actuator_state[INDI_NUM_ACT];
 float indi_u[INDI_NUM_ACT];
-float rate_vect_prev[3] = {0., 0., 0.};
 
 float q_filt = 0.0;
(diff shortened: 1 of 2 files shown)


sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -432,11 +433,6 @@ void stabilization_indi_init(void)
  */
 void stabilization_indi_enter(void)
 {
-  /* reset psi setpoint to current psi angle */
-  stab_att_sp_euler.psi = stabilization_attitude_get_heading_i();
```
dewagter, line 436: Why is this removed?

gautierhattenberger: This is meaningless here. `stab_att_sp_euler` is not the global variable used to set command, it is only needed to store the last command value (passed as a parameter of the run function) to send it later over telemetry.
If this reset is still needed (and maybe it is, @EwoudSmeur ?), it as to be done from upper level (guidance or rc_input).

dewagter: If when switching back to INDI (which we do e.g. when testing neural controllers) the heading does not jump then it is fine.

dewagter (approved):

PR #3310, opened 2024-06-14
PFC fix for Rotwing3

NoahWe:
Change the actuator feedback values such that all drones pass the PFC
change v3d rudder min command such that is has the same range of motion as the other rotwings


dewagter (approved):
OK

gautierhattenberger:
As a side note, since https://github.com/paparazzi/paparazzi/pull/3228 you can use a more readable definition of structure in airframe files

dewagter:
> As a side note, since #3228 you can use a more readable definition of structure in airframe files

Solved here: https://github.com/paparazzi/paparazzi/pull/3315

PR #3311, opened 2024-06-14
Solve wing vibration of the 25kg rotating wing drone

NoahWe:


dewagter (approved):
Crucial update for 25kg

PR #3316, opened 2024-06-16
Actuators in PPRZ values

dewagter:
We would like to have PPRZ units in telemetry and logs. @gautierhattenberger @fvantienen @Fabien-B 
 - a) can this be the default?  
 - b) should we make a define to select this
 - c) should I make 2 messages: driver-value and PPRZ value

### DIFF
--- a/conf/telemetry/AGGIEAIR/aggieair_fixedwing.xml
+++ b/conf/telemetry/AGGIEAIR/aggieair_fixedwing.xml
@@ -3,7 +3,7 @@
 <telemetry>
   <process name="Ap">
     <mode name="default">
-      <message name="ACTUATORS"           period=".5"/> <!-- For trimming -->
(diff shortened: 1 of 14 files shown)


gautierhattenberger:
Since it is a downlink message, you can send several variable length arrays (only parsing from C code is not working), an other option is:
- d) send driver and pprz values in ACTUATORS message

dewagter:
Implemented d)

dewagter:
Using https://github.com/paparazzi/pprzlink/pull/187

dewagter:
@fvantienen @gautierhattenberger Now implemented option C.
 - uses less telemetry / log space. I added slow MESSAGES_RAW by default.
 - you can still trim: look at the RAW message just like when checking RAW sensor data.
 - highspeed logs are in unified PPRZ units.

gautierhattenberger (approved):

PR #3327, opened 2024-07-04
Implement classic GVF for rotorcrafts

Pelochus:
I've also added/modified some files (`bebop2_gvf.xml`, mainly) for trying it out on simulation. Additionally I added a warning for the `crazyradio2ivy.py` script: Crazyflies need a certain original firmware version before flashing the Paparazzi firmware for the nRF radio to work with Paparazzi)


conf/airframes/examples/bebop2_gvf.xml
```diff
@@ -0,0 +1,187 @@
...
+	  <!-- Subsystem section -->
+    <module name="telemetry" type="transparent_udp"/>
+    <module name="radio_control" type="datalink"/>
+    <module name="motor_mixing"/>
+    <module name="actuators" type="bebop"/>
+    <module name="imu" type="bebop"/>
+    <module name="gps" type="ublox"/>
+    <module name="stabilization" type="indi_simple"/>
+    <module name="ins" type="ekf2"/>
+    <module name="gvf_module"/>
```
gautierhattenberger, line 26: Is this module the only difference with other a bebop 2 examples. If yes, the module can be loaded directly from the flight plan. It makes the FP more portable and avoid to add an extra airframe example that is very close to other ones.

conf/flight_plans/demo_gvf_rotorcraft.xml
```diff
@@ -0,0 +1,104 @@
...
+  <waypoints>
+    <waypoint name="HOME" x="0.0" y="0.0"/>
+    <waypoint name="STDBY" x="0.1" y="0.8"/>
+    <waypoint name="S1_START" x="0.0" y="1.0"/>
+    <waypoint name="S1_END" x="0.0" y="5.0"/>
+    <waypoint name="S2_START" x="2.0" y="1.0"/>
+    <waypoint name="S2_END" x="2.0" y="5.0"/>
+    <waypoint name="TD" x="-1.0" y="1.5"/>
+  </waypoints>
+
```
gautierhattenberger, line 15: As mentioned before, you can load the required GVF module here.

conf/flight_plans/demo_gvf_rotorcraft.xml
```diff
@@ -0,0 +1,104 @@
...
+      <exception cond="GetPosHeight() @LT 0.3" deroute="Ramp down"/>
+      <exception cond="!nav_is_in_flight()" deroute="Kill Engine"/>
+      <stay climb="nav.descend_vspeed" vmode="climb" wp="TD"/>
+    </block>
+
+    <block name="Ramp down" key="r">
+      <exception cond="guidance_v.rc_delta_t @LT 0.1*9600." deroute="Kill Engine"/>
+      <set var="fp_throttle" value="guidance_v.rc_delta_t/9600."/>
+      <stay throttle="fp_throttle-0.1*stage_time" vmode="throttle" wp="TD"/>
+    </block>
```
gautierhattenberger, line 96: You should remove this block (and the related exception in previous block). I doubt that it is really working since the last changes in guidance. I know that it comes from one of our example, but we really have too fix that as well.

conf/telemetry/GVF/gvf_default_rotorcraft.xml
gautierhattenberger, line 1: What is the difference with the default rotorcraft telemetry ? Is it only `GVF` and `GVF_PARAMETRIC` ?

sw/airborne/modules/guidance/gvf/gvf.c
```diff
@@ -178,37 +180,19 @@ void gvf_control_2D(float ke, float kn, float e,
...
 
-  // Use parameter kn as the speed command
-  nav.speed.x = md_x * kn;
-  nav.speed.y = md_y * kn;
+  // Speed-based control, acceleration based control not implemented yet
+  nav.speed.x = gvf_control.speed * md_x;
+  nav.speed.y = gvf_control.speed * md_y;
 
-
-  // Acceleration induced by the field with speed set to kn (!WIP!)
```
gautierhattenberger, line 190: The control in acceleration is possible with some controllers (INDI) but not all (PID). Still I think the two options should be kept, either selected with a compilation flag (something like `GVF_ROTORCRAFT_USE_ACCEL`, or any better name), or a dynamic selection at runtime. Unless there is a good reason not to keep the accel based command ?

conf/flight_plans/demo_gvf_rotorcraft.xml
```diff
@@ -0,0 +1,104 @@
...
+
+    <block name="GVF Segment 1">
+      <call fun="gvf_segment_loop_wp1_wp2(WP_S1_START, WP_S1_END, gvf_segment_par.d1, gvf_segment_par.d2)"/>
+    </block>
+
+    <block name="GVF Segment 2">
+      <call fun="gvf_segment_loop_wp1_wp2(WP_S2_START, WP_S2_END, gvf_segment_par.d1, gvf_segment_par.d2)"/>
+    </block>
+
+    <!-- Can't go below 1 because int var -->
```
gautierhattenberger, line 64: I don't get this comment, it seems that radius (or ellipse axis) are given in float ?

gautierhattenberger (changes requested):

conf/airframes/examples/bebop2_gvf.xml
Pelochus: Yes, thanks for pointing it out, didn't know that possibility.

conf/flight_plans/demo_gvf_rotorcraft.xml
Pelochus: Ok!

conf/telemetry/GVF/gvf_default_rotorcraft.xml
Pelochus: Yes, is there any other cleaner way of doing it?

conf/flight_plans/demo_gvf_rotorcraft.xml
Pelochus: Sorry, my fault for not removing some leftover comments.

sw/airborne/modules/guidance/gvf/gvf.c
Pelochus: We decided to use the speed control due to simplicity, however, I can include the older implementation with the flag as you said. Apart from that and AFAIK, the older implementation doesn't work, does it?

gautierhattenberger (approved):

sw/airborne/modules/guidance/gvf/gvf.c
gautierhattenberger: It was supposed to work, but not as good as expected. We will test and fix it later.

PR #3331, opened 2024-07-12
Correct Heading of Rotating Wing Drone with Optitrack

tmldeponti:
The wingbox of the rotating wing drone rotates according to the desired skew angle. At the same time, the wing is the most visible place where to put optitrack markers. This pull requests allows the INS EKF2 and EXT POSE to correct the streamed quaternion of the optitrack with the readings of the skew sensor


gautierhattenberger:
I agree with @fvantienen

tmldeponti:
> This seems not the correct place to implement this. It is very specific for the rotating wing and the way this is currently implemented inside the INS seems very restricted.
> 
> What I suggest is to have some sort of rotation matrix which is applied to the external pose, which can be set by another module (For example the rotating_wing_state). This way it is more generic and code is not duplicated.

@fvantienen Thanks for the advice! I understand your point. I have implemented something along that line using weak functions. The default does nothing but if you define the function in your scheduling module, then you can apply whatever rotation. I think this is a good compromise between usability and complexity.

tmldeponti:
> This is still not really a nice implementation with the WEAK functions, which we try to avoid as much as possible. Next to that generating (optional) dependencies from INS to some other module is not nice. You should try to write it the other way around. My suggestion is to do something along these lines.
> 
> `eff_scheduling_rotwing.c`:
> 
> ```
> // INS external pose rotation based on the wing angle
> #ifdef INS_EXT_POSE_ROT
> struct FloatQuat ins_ext_pose_rot;
> #endif
> 
> static void wing_position_cb(uint8_t sender_id UNUSED, struct act_feedback_t *pos_msg, uint8_t num_act)
> {
>   for (int i=0; i<num_act; i++){
>     if (pos_msg[i].set.position && (pos_msg[i].idx == SERVO_ROTATION_MECH_IDX))
>     {
>       // Get wing rotation angle from sensor
>       eff_sched_var.wing_rotation_rad = 0.5 * M_PI - pos_msg[i].position;
> 
>       // Bound wing rotation angle
>       Bound(eff_sched_var.wing_rotation_rad, 0, 0.5 * M_PI);
> 
> #ifdef INS_EXT_POSE_ROT
>       // Define an INS external pose quaternion rotation from the wing rotation angle
>       struct FloatEulers rot_e = {0, 0, eff_sched_var.wing_rotation_rad};
>       float_quat_of_eulers(&ins_ext_pose_rot, &rot_e);
> #endif
>     }
>   }
> }
> ```
> 
> `....ins.c`
> 
> ```
> #ifdef INS_EXT_POSE_ROT
> extern struct FloatQuat ins_ext_pose_rot;
> #endif
> 
> void ins_ekf2_parse_EXTERNAL_POSE(uint8_t *buf) {
>   if (DL_EXTERNAL_POSE_ac_id(buf) != AC_ID) { return; } // not for this aircraft
> 
>   sample_ev.time_us = get_sys_time_usec(); //FIXME
>   sample_ev.pos(0) = DL_EXTERNAL_POSE_enu_y(buf);
>   sample_ev.pos(1) = DL_EXTERNAL_POSE_enu_x(buf);
>   sample_ev.pos(2) = -DL_EXTERNAL_POSE_enu_z(buf);
>   sample_ev.vel(0) = DL_EXTERNAL_POSE_enu_yd(buf);
>   sample_ev.vel(1) = DL_EXTERNAL_POSE_enu_xd(buf);      
>   sample_ev.vel(2) = -DL_EXTERNAL_POSE_enu_zd(buf);
>   sample_ev.quat(0) = DL_EXTERNAL_POSE_body_qi(buf);
>   sample_ev.quat(1) = DL_EXTERNAL_POSE_body_qy(buf);
>   sample_ev.quat(2) = DL_EXTERNAL_POSE_body_qx(buf);
>   sample_ev.quat(3) = -DL_EXTERNAL_POSE_body_qz(buf);
>   sample_ev.posVar.setAll(INS_EKF2_EVP_NOISE);
>   sample_ev.velCov = matrix::eye<float, 3>() * INS_EKF2_EVV_NOISE;
>   sample_ev.angVar = INS_EKF2_EVA_NOISE;
>   sample_ev.vel_frame = velocity_frame_t::LOCAL_FRAME_FRD;
> 
> #ifdef INS_EXT_POSE_ROT
>   // Rotate the quaternion
>   FloatQuat body_q = {sample_ev.quat(0), sample_ev.quat(1), sample_ev.quat(2), sample_ev.quat(3)};
>   FloatQuat rot_q;
>   float_quat_comp(&rot_q, &ins_ext_pose_rot, &body_q);
>   sample_ev.quat(0) = rot_q.qi;
>   sample_ev.quat(1) = rot_q.qx;
>   sample_ev.quat(2) = rot_q.qy;
>   sample_ev.quat(3) = rot_q.qz;
> #endif
> 
>   ekf.setExtVisionData(sample_ev);
> }
> ```

@fvantienen Thank you very much for the suggestion, I have implemented it and doubled check the rotation direction, let me know what you think!

PR #3332, opened 2024-07-12
6 channel rc 

tmldeponti:
In simulation I work with the radiomaster_TX12 which is nice because it has many switches and buttons. This pull request adds the possibility to have a 6ch rc. I for example use the added channel to switch between autopilots modes (e.g. ANDI and INDI) but it can be used for other purposes as well.


gautierhattenberger:
Instead of adding a `RC_6CH` with 7 fields, maybe we could just send an array of values. That would fit to all case that are not handled by the 3 and 4 channels messages. I can't tell if everything will work straight with the `input2ivy` tool, but would be better to have a more generic solution.

tmldeponti:
@fvantienen @gautierhattenberger I think I have a working solution to have a general RC message. The hardest part (at least for me because I am not familiar at all with ocaml) was to adapt the input2ivy.ml to take arrays as inputs in the field of the joystick xml.  The code is not yet clean (it has a lot of unecessary comments that I used for debugging) but the content is all there. What do you think? 
I tested it with RW3F and joystick TX12 in simulation.

sw/airborne/modules/radio_control/rc_datalink.c
```diff
@@ -62,6 +62,29 @@ void rc_datalink_parse_RC_4CH(uint8_t *buf)
...
+  LED_TOGGLE(RADIO_CONTROL_DATALINK_LED);
+#endif
+  parse_rc_6ch_datalink(DL_RC_6CH_mode(buf),
+      DL_RC_6CH_throttle(buf),
+      DL_RC_6CH_roll(buf),
+      DL_RC_6CH_pitch(buf),
+      DL_RC_6CH_yaw(buf),
+      DL_RC_6CH_kill(buf),
+      DL_RC_6CH_aux7(buf));
+}
```
gautierhattenberger, line 77: can we remove this one if the generic version is working ?

sw/airborne/modules/radio_control/rc_datalink.h
```diff
@@ -29,48 +29,76 @@
...
+#define RADIO_PITCH     1  // Default: Pitch
+#define RADIO_YAW       2  // Default: Yaw
+#define RADIO_THROTTLE  3  // Default: Thrust
+#define RADIO_MODE      4  // Default: MODE
+#define RADIO_AUX1      5  // Default: KILL
+#define RADIO_AUX2      7  // Default: NA
+#define RADIO_AUX4      8  // Default: NA
+#define RADIO_AUX5      9  // Default: NA
+#define RADIO_AUX6      10 // Default: NA
+#define RADIO_AUX7      6  // Default: AP_MODE
```
gautierhattenberger, line 48: I don't yet this ordering ? Why AUX7 is also AP_MODE if there is already a MODE channel ?

sw/ground_segment/joystick/input2ivy.ml
gautierhattenberger, line 1: It looks fine in general, but can you remove the commented lines from debug ?

sw/ground_segment/joystick/input2ivy.ml
```diff
@@ -219,19 +219,30 @@ let parse_input = fun input ->
...
     Printf.printf "parse_msg_field: field %s not found\n" name;
     raise (Failure "field not found") in
-
-  let value = eval_settings_and_blocks field_descr (parse_value (Xml.attrib field "value")) in
-  (name, field_descr.PprzLink._type, value)
+  (* Printf.printf "check 2\n"; *)
+  let value = List.map (eval_settings_and_blocks field_descr) (parse_value (Xml.attrib field "value")) in
+  let field_type = field_descr.PprzLink._type in
+  let field_type_str = PprzLink.string_of_type field_type in
+  (* Printf.printf "Field name: %s, Type: %s\n" name field_type_str; *)
```
gautierhattenberger, line 244: `field_type_str` can be removed (and commented line below as well), same thing below in `execute_action` function.

gautierhattenberger (changes requested):

tmldeponti:
> Really cool that you got it to work. The only thing I saw is that you mapped the switches still from 0-2, I would change that from -127 to 127 such that all channels can be parsed similar on the drone side.

@fvantienen Thanks! I have changed how the switches are sent so that the pasing can be done in a cleaner forloop. Is that what you meant?

sw/airborne/modules/radio_control/rc_datalink.c
tmldeponti: Yes!

sw/airborne/modules/radio_control/rc_datalink.h
tmldeponti: I have added some extra description to the definitions.

sw/ground_segment/joystick/input2ivy.ml
tmldeponti: Done

sw/ground_segment/joystick/input2ivy.ml
tmldeponti: Done

sw/airborne/modules/radio_control/rc_datalink.h
tmldeponti: Making RADIO_AUX7 channel 6 means that in the joystick file you do not have to add any extra zeros to the array. This also results in a smaller array.

tmldeponti: see : https://github.com/paparazzi/paparazzi/blob/master/conf/autopilot/rotorcraft_oneloop_switch.xml

tmldeponti: If the naming is confusing it can be changed!

tmldeponti:
> We could also get rid of the other RC messages going up now, and clean it up fully if we want to. As they can all use the same message.

Indeed, are you ok with it @gautierhattenberger ?

sw/logalizer/ctrlstick.c
gautierhattenberger, line 1: I think it is better to just remove this, it is the historical tool before `input2ivy`. Also remove rule from the makefile.

sw/misc/rctx/main_rctx.c
gautierhattenberger, line 1: This one can be removed as well, unless someone wants to use an autopilot to make a bridge between a PPM RC receiver and a XBee modem...

gautierhattenberger (changes requested):
If you can make some extra cleaning, the rest looks fine. Thanks.

sw/logalizer/ctrlstick.c
tmldeponti: removed

sw/misc/rctx/main_rctx.c
tmldeponti: removed

gautierhattenberger:
I have merged https://github.com/paparazzi/pprzlink/pull/191, so you have to update pprzlink again

tmldeponti:
> I have merged [paparazzi/pprzlink#191](https://github.com/paparazzi/pprzlink/pull/191), so you have to update pprzlink again

Done.

gautierhattenberger (approved):

PR #3335, opened 2024-07-12
Some changes to NAV Hybrid

tmldeponti:
I have worked on the nav hybrid functions such that we can:

1. Specify different max acceleration limits depending on the which nav function we are performing. For example, I have found to be beneficial to limit the acceleration fo the rotating wing drone when moving in a line. However, once the drone is doing a circle, these limits have to be relaxed to not fly a big radius. 
2. Reworked the circling function such that the commanded radius, in combination with the desired airspeed and expected maximum constant wind, will not result in a desired side acceleration that exceed the imposed limit. This also means that the circle in the GCS will better represent the circle that the drone will achieve.


sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -54,11 +54,17 @@ float nav_max_speed = NAV_MAX_SPEED;
...
+#endif
+
+#ifndef NAV_HYBRID_MAX_ACCELERATION
+#define NAV_HYBRID_MAX_ACCELERATION 4.0
 #endif
 
 float nav_max_deceleration_sp = NAV_HYBRID_MAX_DECELERATION;
+float nav_max_acceleration_sp = NAV_HYBRID_MAX_ACCELERATION;
 
+float nav_hybrid_max_expected_wind = 5.0f;
```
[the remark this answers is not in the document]
tmldeponti: Added the definition.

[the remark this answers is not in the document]
tmldeponti: It is true we usually fly at higher constant wind (~10 m/s), but I found this value to work best. The value is the outcome of the trade-off between perfoming an achievable circle (e.g. do not exceed maximum acceleration limits and actually flying the circle displayed on the ground station) and doing a circle of small radius. 

1. Setting it too low will result in the drone not flying the planned circle but continuosly overshooting it. Not nice from a ground operator perspective especially when you have geofences. 
2.  Setting it too high (or just perfectly on the expected 10 m/s) will make sure that the drone flies the expected circle but the radius migh be too big for some test fields (e.g. Valkenburgh)
3. Setting it to something inbetween makes sure that the circle has an ok radius while also making sure that the drone maintains a good tracking of the trajectory for most of the circle.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -87,11 +93,23 @@ bool force_forward = 0.0f;
 #define NAV_HYBRID_GOTO_MODE NAV_SETPOINT_MODE_SPEED
 #endif
 
+#ifndef NAV_OPTITRACK
+#define NAV_OPTITRACK FALSE
```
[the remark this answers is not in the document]
tmldeponti: 1. Removed Optitrack relation
2. Added description

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -115,16 +132,21 @@ static void nav_hybrid_goto(struct EnuCoor_f *wp)
...
+    max_h_speed = Min(nav_max_speed, max_speed_decel); // use hover max speed
   }
   float_vect2_bound_in_2d(&speed_sp, max_h_speed);
-
   VECT2_COPY(nav.speed, speed_sp);
   nav.horizontal_mode = NAV_HORIZONTAL_MODE_WAYPOINT;
-  nav.setpoint_mode = NAV_HYBRID_GOTO_MODE;
+
+  // In optitrack tests use position mode for more precise hovering
+  if (NAV_OPTITRACK){
```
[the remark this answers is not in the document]
tmldeponti: Done and done :)

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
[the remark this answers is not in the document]
tmldeponti: done

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -87,11 +93,23 @@ bool force_forward = 0.0f;
 #define NAV_HYBRID_GOTO_MODE NAV_SETPOINT_MODE_SPEED
 #endif
 
+#ifndef NAV_OPTITRACK
+#define NAV_OPTITRACK FALSE 
+#endif
+
+static void update_max_acc(bool full_acc){
```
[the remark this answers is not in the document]
tmldeponti: done

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -115,16 +139,21 @@ static void nav_hybrid_goto(struct EnuCoor_f *wp)
     // The time in which it does this is: T = V / a_max
     // The maximum speed at which to fly to still allow arriving with zero
     // speed at the waypoint given maximum deceleration is: V = sqrt(2 * a_max * d)
-    float max_speed_decel2 = fabsf(2.f * dist_to_wp * nav_max_deceleration_sp); // dist_to_wp can only be positive, but just in case
+    float max_speed_decel2 = fabs(2.f * dist_to_wp * nav_max_deceleration_sp); // dist_to_wp can only be positive, but just in case
```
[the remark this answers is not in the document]
tmldeponti: not needed anymore

conf/modules/nav_hybrid.xml
```diff
@@ -14,12 +14,16 @@
       <define name="GOTO_SPEED" value="nav_max_speed" description="maximum speed when flying goto/stay routines (default is max speed)"/>
       <define name="NAV_LINE_DIST" value="50" description="distance coefficient for line routine"/>
       <define name="NAV_CIRCLE_DIST" value="40" description="distance coefficient for circle routine"/>
+      <define name="NAV_HYBRID_MAX_EXPECTED_WIND" value="5" description="Maximum expected constant wind. Used to calculate minimum circle radius to not exceed acceleration bound."/>
```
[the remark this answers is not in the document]
tmldeponti: Yes sorry! Corrected.

tmldeponti:
> Is this tested also in simulation with other controllers/airframes that this works as well? Since this impacts multiple airframes I think

the changes are tested with INDI guidance and stabilization + RW3F in simulation. Drone performs the circles as expected. 
The only thing that does not work on this controller is the limiting of the circle radius (which is by default off). This is due to the old rotwing state module setting the nav_max_speed (see https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/rot_wing_drone/rotwing_state.c#L482).

dewagter (dismissed):

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -83,15 +94,21 @@ float nav_hybrid_pos_gain = 1.0f;
...
+/* Use max target groundspeed and max acceleration to limit circle radius*/
+#ifndef NAV_HYBRID_LIMIT_CIRCLE_RADIUS
+#define NAV_HYBRID_LIMIT_CIRCLE_RADIUS FALSE 
+#endif
 /** Implement basic nav function for the hybrid case
  */
 
 static void nav_hybrid_goto(struct EnuCoor_f *wp)
 {
+  nav_max_acceleration_sp = nav_hybrid_max_acceleration / 2.0;
```
EwoudSmeur, line 111: Why this magic division by two?

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -118,13 +135,19 @@ static void nav_hybrid_goto(struct EnuCoor_f *wp)
     float max_speed_decel2 = fabsf(2.f * dist_to_wp * nav_max_deceleration_sp); // dist_to_wp can only be positive, but just in case
     float max_speed_decel = sqrtf(max_speed_decel2);
     // Bound the setpoint velocity vector
-    max_h_speed = Min(nav_goto_max_speed, max_speed_decel); // use hover max speed
```
EwoudSmeur, line 121: I think this will break the functionality of nav_goto_max_speed, for those users that expect it to work.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -248,28 +274,30 @@ static void nav_hybrid_circle(struct EnuCoor_f *wp_center, float radius)
...
+  Bound(desired_speed, 0.0f, nav_max_speed);
   // compute speed vector from target position
   struct FloatVect2 speed_unit;
   VECT2_DIFF(speed_unit, nav.target, *stateGetPositionEnu_f());
   float_vect2_normalize(&speed_unit);
   VECT2_SMUL(nav.speed, speed_unit, desired_speed);
 
   nav_rotorcraft_base.circle.center = *wp_center;
-  nav_rotorcraft_base.circle.radius = radius;
+  nav_rotorcraft_base.circle.radius = sign_radius * abs_radius;
```
EwoudSmeur, line 300: Why is this better than just radius?

EwoudSmeur (commented):
Cool development to change the circle radius in case of wind!

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
tmldeponti: I have made it more clear now. I have experienced that some platforms (like my RW3C) benefit from a lower acceleration bound in certain phases but still need the full acceleration limit to perform tight turns. I made a new variable that if not defined defaults to the maximum acceleration.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
tmldeponti: I have checked in master who uses this nav_goto_max_speed and actually nobody uses it. Only us (Rotatingwing) "mis"-use it the state machine assigning it the sane value as nav_max_speed. I would remove it completly?

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
tmldeponti: Because abs_radius is the bounded variable.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
gautierhattenberger: I have added this because our hybrid had issues when doing small moves. So it assumes that a goto is short move that should be done close to the hover attitude, while a segment is for fast forward flights.

dewagter: I  would prefer if this was automatic. If a hybrid drone type has a distance that is too short to transition (partially) and should stay in hover mode while moving the small distance, then define
```NAV_HYBIRD_MIN_DISTANCE_TO_TRANSISITON```

```
#if NAV_HYBRID_MIN_DISTANCE_TO_TRANSISTION
  static bool may_transisition = false;
  if (wp_dist > NAV_HYBRID_MIN_DISTANCE_TO_TRANSISTION) {
    may_transition = true;
  }
  if ((wp_dist < NAV_HYBRID_MIN_DISTANCE_TO_TRANSISTION) && (may_transisition == false)) {
    max_h_speed = Min(nav_goto_max_speed, max_speed_decel);
  }
  if (max_speed_decel < nav_goto_max_speed) {
    may_transition = false;
  }
#endif
```

File: conf/modules/nav_hybrid.xml [hunk 10]
```diff
@@ -7,19 +7,24 @@
...
+      <define name="EXT_VISION_SETPOINT_MODE" value="FALSE" description="Use position setpoints when using external vision systems."/>
     </section>
     <section name="GUIDANCE_INDI_HYBRID" prefix="GUIDANCE_INDI_">
       <define name="MAX_AIRSPEED" value="15." description="maximum airspeed (required)"/>
       <define name="NAV_SPEED_MARGIN" value="10." description="maximum commanded ground speed is MAX_AIRSPEED + NAV_SPEED_MARGIN"/>
       <define name="GOTO_SPEED" value="nav_max_speed" description="maximum speed when flying goto/stay routines (default is max speed)"/>
       <define name="NAV_LINE_DIST" value="50" description="distance coefficient for line routine"/>
       <define name="NAV_CIRCLE_DIST" value="40" description="distance coefficient for circle routine"/>
-    </section>
+      </section>
```
dewagter, line 20:
```suggestion
    </section>
```

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -57,7 +57,23 @@ float nav_goto_max_speed = NAV_HYBRID_GOTO_MAX_SPEED;
...
+#define NAV_HYBRID_MAX_ACCELERATION 4.0
+#endif
+
+#ifndef NAV_HYBRID_SOFT_ACCELERATION
+#define NAV_HYBRID_SOFT_ACCELERATION NAV_HYBRID_MAX_ACCELERATION
+#endif
+
+float nav_max_deceleration_sp      = NAV_HYBRID_MAX_DECELERATION;  //Maximum deceleration allowed for the set-point
+float nav_max_acceleration_sp      = NAV_HYBRID_MAX_ACCELERATION;  //Maximum acceleration allowed for the set-point
+float nav_hybrid_soft_acceleration = NAV_HYBRID_SOFT_ACCELERATION; //Soft acceleration limit allowed for the set-point (Equal to the maximum acceleration by default)
```
dewagter, line 70: If ```nav_hybrid_max_acceleration``` is for circles and ```nav_hybrid_soft_acceleration``` is for lines, it would be better if the names reflect this.

dewagter (commented):
At least the ```nav_goto_max_speed``` should be addressed before we can merge.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
tmldeponti: I would not say that they are meant to be used either for lines or for circles. For example nav_hybrid_soft_acceleration is used in circles when we are far away from it. With these defines we can design a part of a navigation function to have aggressive characteristics, while another part can have more gentle ones.

conf/modules/nav_hybrid.xml
tmldeponti: Done

tmldeponti:
I have implemented @dewagter 's idea for the minimum trnasition idstance. What do you think? @dewagter @gautierhattenberger

tmldeponti:
I have also removed the nav_goto_max_speed from the state machines for the rotwing as we do not rteally use it and is not needed anymore. @fvantienen

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
gautierhattenberger: I agree that going for a transition should be automatic for "normal" use, and setting a distance threshold is not a bad option to start with.
Still, I can add to my previous comment that I was using this goto=slow speed, route = high speed as a "debug" feature. If the forward mode doesn't really work, I can still navigate back to home or standby like a regular rotorcraft. It is not an ideal situation, and probably not a good "default" behavior, but I think it is worth keeping this as an option for now (e.g. a config flag should be available to force this behavior if needed).

tmldeponti: @gautierhattenberger I have added a definition that allows you to force max goto speed so that you can use the goto as a "debug". Let me know if it is ok with you.

EwoudSmeur: From my perspective, this is not a general difference between goto and route: for instance flying back home due to an exception I would do with a goto, but this can be from far away.
With NAV_HYBIRD_MIN_DISTANCE_TO_TRANSISITON I see the problem that the transition should be a function of airspeed, and a transition to forward flight may already be needed in order to hover. On the other hand, I can imagine that for vehicles that can fly very fast in hover and have an airspeed range in which they cannot fly (read: they need some dynamic transition maneuver) this can make sense. Not sure if we have any of those vehicles in Paparazzi though.

Personally I would vote for remove this difference between route and goto for these reasons, but if it is useful for some I'm ok with keeping it.

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
```diff
@@ -118,13 +144,27 @@ static void nav_hybrid_goto(struct EnuCoor_f *wp)
     float max_speed_decel2 = fabsf(2.f * dist_to_wp * nav_max_deceleration_sp); // dist_to_wp can only be positive, but just in case
     float max_speed_decel = sqrtf(max_speed_decel2);
     // Bound the setpoint velocity vector
-    max_h_speed = Min(nav_goto_max_speed, max_speed_decel); // use hover max speed
+    max_h_speed = Min(nav_max_speed, max_speed_decel);
+#ifdef NAV_HYBRID_MIN_DISTANCE_TO_TRANSISTION
+    if (dist_to_wp < NAV_HYBRID_MIN_DISTANCE_TO_TRANSISTION){
+      max_h_speed = Min(nav_goto_max_speed, max_h_speed);
```
EwoudSmeur, line 150: the #if NAV_HYBRID_FORCE_GOTO_MAX_SPEED is already needed here

EwoudSmeur: I think it is worth checking in which case this will actually be useful: if this distance*position_gain is smaller than the nav_goto_max_speed it will never take effect.

EwoudSmeur (commented):
I think the max_h_speed part can be made a bit nicer, then I'm good for merging!

sw/airborne/modules/nav/nav_rotorcraft_hybrid.c
gautierhattenberger: I agree that the question is not about goto or route, nor preventing a "transition" if it is required to hover due to wind condition. But sometimes it is required to set the desired and/or maximum speed, independently from the type . This should be done from the flight plan in the first place I think. At the moment, the different parameters must be changed each time before calling a flight instruction, which is a pain. I guess enhancing the flight plans is required at longer (short ?) term.
In the meantime, I just would like to preserve a "debug" mode where go and route have different maximum ground speed, even if it is very basic. Why the current state where `goto_max_speed` is equal to `max_speed` unless specified otherwise is a problem ?
Sorry if I'm a bit lost, but I'm not sure what was the original issue here...

EwoudSmeur: I agree it is a pain to do it after each flight plan instruction, but if it is just for debugging it is maybe not a big problem? Just thinking that the more simple the code is the easier it is to understand and maintain. If you still disagree we can stick to the goto_max_speed implementation.

gautierhattenberger: Ok, let's keep the code simple.

gautierhattenberger (approved):

PR #3345, opened 2024-07-29
Add pprz-py-plotter a new logplotter based on Python and NumPy

Pelochus:
I've developed a new logplotter in Python, based on NumPy and Qt5. It still is a bit immature, but it works. Usage is as follows (there's also a README with more information):

- Open pprz-py-plotter
- Select one folder which contains a `.log` file and a `.data` file. If there is more than one it will take the very first 2
- The program will parse for telemetry and datalink messages
- It will also save some files (depending if using CLI or GUI version) to the `tmp` and/or `output` directories, with information of the telemetry/datalink messages and the data in a plain `.npy` file that can be read with NumPy directly
- Inside the program select which ID do you want to see data for
- Select which variables do you want displayed with checkboxes. This is a little bit cumbersome to do, but works.
- Refresh the plot and see the data

Some extra info here: 
https://github.com/Swarm-Systems-Lab/pprz-py-plotter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,6 +195,9 @@ Cargo.lock
 
 /sw/airborne/test/ahrs/run_ahrs_on_synth
 
+# tmp
+/tmp
+
 # Mac OS X
 .DS_Store
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -53,3 +53,6 @@
 [submodule "sw/ext/libcanard"]
 	path = sw/ext/libcanard
 	url = https://github.com/UAVCAN/libcanard.git
+[submodule "sw/ext/pprz-py-plotter"]
+	path = sw/ext/pprz-py-plotter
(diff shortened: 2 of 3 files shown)


Pelochus:
I thought of adding it as a submodule since I'm still developing and improving it. Making it a submodule would reduce the number of PRs I'd need to make, just suggesting an update to the submodule every now and then.

gautierhattenberger:
Replacing the old plotter is definitely a way to go. However, I have some remarks on the current status of this tool:
- I also would like to have it as part of the main code when ready
- It seems to be very preliminary and not easy to use at the moment compared to the previous one
So, if you are willing to continue working and improving this tool, I suggest that you have a look at the repository mentioned by @fvantienen and make a PR for integration inside main code once it is more mature.

Pelochus:
Will take a look into the other repo. For now, I'll try to improve my version to make it at least decent. Once I've achieved enough features/bugfixes I'll consider either integrating into the other repo or just PR here

PR #3346, opened 2024-07-29
[jevois] tag ids can contains '0'.

Fabien-B:


sw/airborne/modules/computer_vision/tag_tracking.c
```diff
@@ -179,8 +179,8 @@ struct tag_info {
 };
 
 struct wp_tracking {
-  uint8_t wp_id;
-  uint8_t tag_id;
+  uint16_t wp_id;
+  int16_t tag_id;
```
gautierhattenberger, line 183: `int16` is fine for tags, but why `uint16` for wp id ?

gautierhattenberger (approved):

PR #3347, opened 2024-07-29
[actuators] Update controller faulhaber

fvantienen:
- Fixes initialization and homing which is done at startup
- Changed to velocity control on the MC5004 and added a position gain


gautierhattenberger (approved):
You can merge if it works for you @fvantienen

PR #3356, opened 2024-09-12
Gtk to pyqt : Add alternatives to deprecated GTK GUI elements

spacewolfXfr:
New tools to replace deprecated GUI based on GTK, namely:

- `sw/simulator/gaia.py`: alternative to `sw/simulator/gaia[.ml]`, same functionalities, but based on PyQT instead of GTK
- `sw/ground_segment/messages-gui` : folder containing alternative to both  `sw/ground_segment/tmtc/messages[.ml]` and `sw/logalizer/plotter[.ml]` (and `sw/logalizer/rt_plotter.py`). Also based on PyQT (PyQTGraph for plotting), it adds new functionalities for message display (and also some to the plotter, but is still mostly a work in progress).

The messages replacement is located at `sw/ground_segment/messages-gui/messages.py` and the one for the plotter is at `sw/ground_segment/messages-gui/rtplotter.py`. For messages, it is already configured as the new default for the Paparazzi Center (and the previous one is considered to be "Messages (Legacy)" in the Paparazzi Center).

New functionalities added to the messages tool are : 
- RegExp Filtering (messages name and fields)
- Expandable messages, allowing to look at fields of several messages at once
- Message grouping, allowing to use the field of one message as an additional key when logging (and so log the same message several times)
- Message/Field pinning, allowing to show only selected messages/fields, eventually across several aircraft

The new messages tool still support drag'n'drop actions, compatible with the previous plotters.


spacewolfXfr:
Removed unrelated commit  fd19eab08a939268e76548f3d50076cf85cf3392

EwoudSmeur:
Great stuff! I have the following issues:
- If I start the new gaia tool in paparazzi center, it doesn't include the .py part. Probably because in some control panel confs this seems to be overwritten, e.g. in conf/userconf/tudelft/RW_control_panel.xml:
```38:    <program name="Environment Simulator" command="sw/simulator/gaia"/>```
- I need to first run ```export PAPARAZZI_HOME=~/paparazzi &&
export PAPARAZZI_SRC=~/paparazzi```
for gaia.py to find pprzlink. Is this just the way it is or can this be automatic in a way?

gautierhattenberger:
Indeed, if you have an environment simulator in your local control panel, it will overwrite the new one by default. If you run the program from the paparazzi center, it should set the HOME and SRC paths. If you run it from a terminal, you have to do it yourself.
One question I have with this PR is that the performances of the new message tool is quite poor due to the Python implementation. The concepts for filtering are really nice, so I really would like to keep that. So the possibilities are:
- implement everything in C++/Qt, efficient but is annoying in terms of compilation as it requires a dependency on qt dev libraries
- implement the backend with Ivy in C++ and mix with Python/Qt for the graphical part (then Qt is only used as a binary binding (that we are already using)
- we try to improve the python implementation or accept lower performances

spacewolfXfr:
Update:  The added plotter has been removed, in favor of using `sw/logalizer/rt_plotter.py` instead.

PR #3359, opened 2024-09-23
Fuel-Cell Monitoring over CAN

dewagter:
Requires https://github.com/paparazzi/pprzlink/pull/195


gautierhattenberger (approved):
can be merged if ref to pprzlink is correct

PR #3360, opened 2024-09-23
Airspeed Filtering

dewagter:


sw/airborne/modules/sensors/airspeed_ms45xx_i2c.c
```diff
@@ -221,13 +223,22 @@ void ms45xx_i2c_init(void)
...
 {
+#ifdef USE_AIRSPEED_LOWPASS_FILTER
+  // Update the filter if needed
+  if(ms45xx.lowpass_tau != ms45xx.lowpass_tau_set) {
+    ms45xx.lowpass_tau_set = ms45xx.lowpass_tau;
+    init_butterworth_2_low_pass(&ms45xx_filter, ms45xx.lowpass_tau,
+                              MS45XX_I2C_PERIODIC_PERIOD, get_butterworth_2_low_pass(&ms45xx_filter));
+  }
+#endif
+
```
gautierhattenberger, line 234: maybe using a setting handler is a better way to update the time constant and call the init with the newly received value ?

gautierhattenberger (changes requested):

sw/airborne/modules/sensors/airspeed_ms45xx_i2c.c
dewagter: Do we have the guarantee that the telemetry callback and i2c_periodic are not executed in parallel in chibios?

gautierhattenberger: They are both part of the mainloop, only the low level transactions are using dedicated threads (accessed through protected buffers and transaction queues)

gautierhattenberger (approved):

sw/airborne/modules/sensors/airspeed_ms45xx_i2c.c
dewagter: Updated to use a handler

PR #3361, opened 2024-09-23
Power UAV-CAN

dewagter:


sw/airborne/modules/sensors/power_uavcan.c
```diff
@@ -98,6 +99,28 @@ struct uavcan_circuit_battery_t {
...
+{
+  static uint8_t idx = 0;
+  // Send the circuit status
+  if(circuits[idx].set) {
+    uint8_t cid = circuits[idx].circuit_id;
+    pprz_msg_send_POWER_DEVICE(trans, dev, AC_ID, &circuits[idx].node_id, &cid, &circuits[idx].current, &circuits[idx].voltage);
+  }
+
+  // Go to the next
+  if (rand_uniform() > 0.02) {
```
gautierhattenberger, line 115: why do you need to randomly go to next id ?

dewagter: The random sending overall gives a good overview of the power used both in telemetry and in the log. Is this OK for this PR? I'll link to the ```index per telemetry``` issue for future reference #2501

dewagter: norm > 0.2 means 42% to increment: when read by the telemetry and the logger interchangingly, then most of the time there is an update, and it seldomly skips an index on each, but indeed sometimes sends them twice. As freek said: an index per link would be better but this works in the meantime.

gautierhattenberger (approved):

PR #3367, opened 2024-09-26
UAVCAN actuator update

dewagter:


gautierhattenberger:
@dewagter @fvantienen I can see that you have a few PR related to UAVCAN. There is #3344 that is pending and that should update uavcan to use the correct dronecan repository and start the support for FDCAN. This new code is almost ready but still need some work to pass the CI tests.
I guess it is more important your PR in their current states since they have probably been tested in flight. We'll update them together with #3344 if needed, but will you have the time to test them, you're the one really using CAN devices at the moment ?

dewagter:
Updated with ```RunOnceEvery``` instead of the counter.

Yes we will help test the new dronecan.

gautierhattenberger (approved):

PR #3370, opened 2024-09-27
collection of changes after IMAV2024

gautierhattenberger:
- update airframe and add competition flight plans
- fix flight plan DTD
- mag params for Bristol
- add a set_expo function (might be weak) for digital cam driver
- fix tag tracking for simulation, add getter functions
- add option to skip initial circle in nav survey
- fix jevois driver parsing

### DIFF
--- a/sw/airborne/modules/nav/common_flight_plan.h
+++ b/sw/airborne/modules/nav/common_flight_plan.h
@@ -60,4 +60,7 @@ void nav_goto_block(uint8_t block_id);
 /** Time in s since the entrance in the current block */
 #define NavBlockTime() (block_time)
 
+/** macro to use pointers in flight plans XML */
+#define Ptr(x) (&(x))
+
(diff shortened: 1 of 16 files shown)


sw/airborne/modules/nav/common_flight_plan.h
```diff
@@ -60,4 +60,7 @@ void nav_goto_block(uint8_t block_id);
 /** Time in s since the entrance in the current block */
 #define NavBlockTime() (block_time)
 
+/** macro to use pointers in flight plans XML */
+#define Ptr(x) (&(x))
```
dewagter, line 64: where is this used?

dewagter (approved):

sw/airborne/modules/nav/common_flight_plan.h
gautierhattenberger: In the rotorcraft flight plan, the XML parser don't like the `&` symbol

PR #3371, opened 2024-09-28
[tests] Fix printing summary after testing confs

dewagter:
Improved tools to make an overview of failing airframes.
 - always print the conf name
 - parse ```Error``` and ```failed``` as errors
 - do not show an empty overview if the ```CONF``` name is unknown

### DIFF
--- a/sw/tools/parse_compile_logs.py
+++ b/sw/tools/parse_compile_logs.py
@@ -33,7 +33,7 @@ def parse_log(log_file):
         if conf and airframe:
             if ('error:' in lowerline) or ('failed:' in lowerline):
                 errors.append((conf, airframe, line.strip()))
-            if 'warning:' in lowerline.lower():
+            if 'warning:' in lowerline:
                 errors.append((conf, airframe, line.strip()))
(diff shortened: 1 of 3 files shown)


sw/tools/parse_compile_logs.py
```diff
@@ -30,9 +31,9 @@ def parse_log(log_file):
                 airframe = line.strip()
 
         if conf and airframe:
-            if 'error:' in line:
+            if ('error:' in lowerline) or ('failed:' in lowerline):
                 errors.append((conf, airframe, line.strip()))
-            if 'warning:' in line:
+            if 'warning:' in lowerline.lower():
```
gautierhattenberger, line 36: isn't it already in lower case ?

dewagter, line 36:
```suggestion
            if 'warning:' in lowerline:
```

dewagter: yes. fixed

gautierhattenberger (approved):

PR #3375, opened 2024-09-29
[rotwing] update simulator models

dewagter:
- Improved lateral stability
 - Fixed 25kg model


tmldeponti:
If you use the commands instead of the actuator_pprz you can simplify the simulator section of the common airframes to something like :
``` c
        <define name="JSBSIM_MODEL"    value="rotwingv3c_SI" type="string"/>
        <define name="SENSORS_PARAMS"  value="nps_sensors_params_default.h" type="string"/>
        <define name="USE_COMMANDS"    value="TRUE"/>
        <define name="JS_AXIS_MODE"    value="4"/>
    </section>

tmldeponti:
As for the simulation model itself, I think some things can be implemented from "rotwingv3c_SI.xml". 

1. Use COMMAND names (e.g. MOTOR_FRONT instead of front_motor). Like that you do not have to provide the names of the actuators in simulator section of the airframe. 
2. Use SI units where possible. Makes it a little bit more understandable I think. 
3. Useful property definitions like airspeed and trig functions of skew.
4. Modelling of Elevator stall. 
5. Modelling of changing Inertia with skew. 
I do not think it should be difficult to implement these changes also for the 25kg. The 7kg file can be used as a model and the actual values can be changed to match the weight and force/moment curves of the big drone.

dewagter:
Completely agree. I added it to the list. But please first get the best-tested flying code in, and only change airframes after. So I put it in an issue: #3376

EwoudSmeur (approved):
I agree with all comments. I tried with the SI units, but I cannot get those to work in JSBSIM, and I got so frustrated that I gave up altogether.

PR #3377, opened 2024-10-01
[uavcan] actuators - fix

dewagter:


dewagter:
You prefer 

```#define ActuatorsUavcan2Commit()  RunOnceEvery(XXX,actuators_uavcan_commit``` than having it in the driver like now or slowing down the actuator freq?

dewagter:
@fvantienen Like this?

PR #3381, opened 2024-10-01
Updates of Oneloop Controller

tmldeponti:
This pull request is to update some airframes to correctly use the oneloop controller. 
The oneloop controller is updated to use the new WLS structure. 
General Clean Up on the oneloop controller.


dewagter (approved):
Nice.

PR #3374, opened 2024-09-29
Rotwing control update made compatible with master

dewagter:
- hard to split up even further
 - requires: https://github.com/paparazzi/pprzlink/pull/194

PR #3382, opened 2024-10-01
RotWing Demo + RotMech index correction

tmldeponti:
This pull request contains code to start a cosine signal to the cmd skew angle for demo purposes. 
Moreover, this pull requests fixes the rotation mech IDX in the Abi messages so that RotWings with either the adc sensor or the supercan board can use the same code.


dewagter (approved):

dewagter:
Good for now. Would be great if after #3374 you could investigate if you really need a ```state_V2.h``` or can re-use the default one.

PR #3385, opened 2024-10-02
[flight plan] add on_enter and on_exit functions for nav blocks

gautierhattenberger:
The functions passed to the attributes 'on_enter' and 'on_exit' will be called once every time the flight plan state machine enters or leaves a block, even if the block change is requested by the operator or if it is caused by an exception.
It allows to properly activate and then disable services related to a specific block.
Another good side effect is that the block and stage indexes are now private and can't be modified by external modules. Old variables nav_block and nav_stage can still be used as read-only.

Based on the work from:
Baptiste Pollien, Christophe Garion, Gautier Hattenberger, Pierre Roux, Xavier Thirioux. A Verified UAV Flight Plan Generator. 2023 IEEE/ACM 11th International Conference on Formal Methods in Software Engineering (FormaliSE), May 2023, Melbourne, Australia. pp.130-140, [10.1109/FormaliSE58978.2023.00021](https://www.doi.org/10.1109/FormaliSE58978.2023.00021)


dewagter (approved):
Nice to protect the ```nav_block``` and have guaranteed ```on_exit``` functions.
Can this already be used to not allow e.g. to start motors if the geo_init is not finished like described in the paper?

gautierhattenberger:
The protection of block jump is another feature (`forbidden deroute` in the paper). It is not integrated in this PR. It may not be so difficult, but this functionality still has some conceptual issues: do you want to prevent deroute from the user only, or also from the "normal" behavior of the FP logic ? what to do if a deroute is refused ?
The easy option is to limit to user deroute to prevent wrong actions from the operator, but will not cover a wrong FP logic (if it is even possible...).

Issue #3386, opened 2024-10-03
Re-think ```MAX_AIRSPEED``` and ```SPEED_MARGIN```


dewagter:
**rotwing_state:**
 - ```ROTWING_QUAD_MAX_AIRSPEED```  is the max allowed airspeed in quad mode
 - ```ROTWING_FW_MAX_AIRSPEED``` is the max allowed airspeed in FW mode
**nav_rotorcraft_hybrid:**
 - ```NAV_MAX_AIRSPEED```
 - ```NAV_HYBRID_MAX_AIRSPEED``` should  ```== GUIDANCE_INDI_MAX_AIRSPEED```, can be unified by making a define that is not specific for the module.
 - ```NAV_SPEED_MARGIN```
 - ```NAV_HYBRID_SPEED_MARGIN``` is the amount the maximum commanded ground speed can be larger than the airspeed. If there is 10 m/s wind, this should at least be larger than 10
 **guidance_indi_hybrid:**
 - ```GUIDANCE_INDI_MAX_AIRSPEED```
 - ```GUIDANCE_INDI_NAV_SPEED_MARGIN```
 - ```GUIDANCE_INDI_COORDINATED_TURN_MAX_AIRSPEED```  is used to limit the minimum turn rate due to rolling. 
 - ```GUIDANCE_INDI_COORDINATED_TURN_AIRSPEED_MARGIN``` if the airspeed sensor would be broken and outputs 1000 m/s, because then the turn rate would go to zero otherwise. Not sure if we want to handle that in this case, error checking on the airspeed sensor would make more sense IMO. Then GUIDANCE_INDI_COORDINATED_TURN_AIRSPEED_MARGIN can be removed too.

OpenUAS:
Done AFAIK

PR #3387, opened 2024-10-03
[flight plan] add a max_speed attribute to fp instruction

gautierhattenberger:
This is only for the rotorcraft firmware at the moment, and it is replacing the explicit call to `guidance_h_SetMaxSpeed` to simplify the flight plans.

Maybe the issue  in #3386 should also be considered here, but I'm not sure it is directly related nor incompatible.

### DIFF
--- a/conf/flight_plans/flight_plan.dtd
+++ b/conf/flight_plans/flight_plan.dtd
@@ -171,7 +171,8 @@ pre_call CDATA #IMPLIED
 post_call CDATA #IMPLIED
 nav_type CDATA #IMPLIED
 nav_params CDATA #IMPLIED
-until CDATA #REQUIRED>
+until CDATA #REQUIRED
+max_speed CDATA #IMPLIED>
 
 <!ATTLIST attitude
 roll CDATA #REQUIRED
@@ -207,7 +208,8 @@ pre_call CDATA #IMPLIED
 post_call CDATA #IMPLIED
 nav_type CDATA #IMPLIED
 nav_params CDATA #IMPLIED
-until CDATA #IMPLIED>
+until CDATA #IMPLIED
+max_speed CDATA #IMPLIED>
 
 <!ATTLIST path
 wpts CDATA #REQUIRED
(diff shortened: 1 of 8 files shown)


dewagter:
I like the development a lot, but I just have a few questions:
 - As soon as some blocks have some speed definition, basically, all blocks should have it, otherwise, a block might act differently depending on what you did before. e.g. after a ```slow_search``` block where the nav speed was reduced, a ```return_home``` will also be slow unless it also defines the high speed again. This can be easily solved by a ```push``` and ```pop``` mechanism.  Whenever entering a block with a ```max_speed```, the default value is stored, and when entering any block without ```max_speed``` the default is reset.
 - Why are we setting the ```max_speed``` and not a speed setpoint. Would that not make more sense in a flightplan?
 - Could you replace all the ```guidance_h_SetMaxSpeed``` from the (mainly) IMAV flightplans?

gautierhattenberger:
>     * As soon as some blocks have some speed definition, basically, all blocks should have it, otherwise, a block might act differently depending on what you did before. e.g. after a `slow_search` block where the nav speed was reduced, a `return_home` will also be slow unless it also defines the high speed again. This can be easily solved by a `push` and `pop` mechanism.  Whenever entering a block with a `max_speed`, the default value is stored, and when entering any block without `max_speed` the default is reset.

Yes, as soon as you start to play with speed limits, you should do for most of the flight plan. An alternative is to use the `on_enter` and `on_exit` functions from #3385 to guarantee that you restore a default parameter upon leaving a block.
As you suggested, I wondered if I could define/use a default speed or restore the previous value if nothing is specified. The implementation is probably a bit more complex, so for a first concept, it is merely a macro for the `SetMaxSpeed` function.

> 
>     * Why are we setting the `max_speed` and not a speed setpoint. Would that not make more sense in a flightplan?

I wondered, if I should introduce several speeds (e.g. max_speed, speed, airspeed). At the moment, the rotorcraft control is really doing a "bounded" control where it flies as fast as possible, limited by the max speed. If you set this maximum too high, maybe the drone will not reached it, depending on the navigation and guidance gains. Again, I'm not introducing new functionalities in the control, only an ease of writing for the flight plans. It could be extended later if the underlying controller supports it.

> 
>     * Could you replace all the `guidance_h_SetMaxSpeed` from the (mainly) IMAV flightplans?

I prefer not, if possible I would like keep these flight plans in their original states, as historical markers... I would be good to improve the other examples to better reflect the new features, thought.

PR #3388, opened 2024-10-03
Fix Python tools - file name consistency - compile master without warnings - airframe fix

dewagter:
- [x] Moving defines to *.h files requires compiler defines. airframe.h in driver is not working.
 - [x] More flightplan renames to ```rotwing_```
 - [x] update IMU locations
 - [x] Fix python tools
 - [x] redundant ```gih_param```
 - [x] removed unfixable bebop
 - [x] typo ```13 .0```
 - [x] fix rotwing_state_V2 
 - [x] ```ins_ext_vision``` only works in AP
 - [x] ```rotwing_viewer.py``` update got lost

PR #3389, opened 2024-10-03
DroneCAN overhaul

Fabien-B:
Replace #3344, merged with master and all commit squashed together.
Not tested yet, probably broken.


Fabien-B:
@dewagter could you test this branch ?
Probably a lot of things are broken.
The RotatingWingV3D is building, but I think you broke other airframes like Nederdrone8 in your lasts PR :
`guidance_indi_hybrid.c: undefined reference to 'rotwing_state_pusher_motor_running'`

dewagter:
@Fabien-B Thanks for rebasing. We will test as soon as possible. PS: sorry about the merge errors that were not present in the separate branches. I fixed all I could find in #3388

Fabien-B:
@dewagter I rebased again on master, can you test this branch when you have the time?
You need to install python packages before running make:
- empy==3.3.4
- pexpect

Fabien-B:
@dewagter are you using just the standard CAN (8 data bytes) or CAN-FD (64 data bytes) ?
It works on my desk but just with the CAN, not with CAN-FD (but I don't have any CAN-FD sensor to investigate)

fvantienen:
We only use the standard CAN for now and also don't have devices with CAN-FD for now.

Fabien-B:
Ok, could you test this PR sometimes ?
It mainly update uavcan with a recent version. (this is why all was renames as "dronecan")
The serialisation code is now generated from DSDL definitions, so you will need the python packages `empy==3.3.4` and `pexpect`.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Cortex Debug",
+            "cwd": "${workspaceFolder}",
+            "executable": "${workspaceFolder}/var/aircrafts/CobraV2/ap/obj/ap.elf",
```
fvantienen, line 7: Nice, never used this. But the executable is especially for your aircraft, so maybe we need to leave this out or fix it another way.
Maybe some information on how to set this up correctly would be really nice and help a lot of people.

.vscode/c_cpp_properties.json
```diff
@@ -6,16 +6,19 @@
                 "path": [
                     "${workspaceFolder}"
                 ],
-                "limitSymbolsToIncludedHeaders": true
+                "limitSymbolsToIncludedHeaders": false
             },
             "includePath": [
-                "${workspaceFolder}"
+                "~/paparazzi/sw/**",
```
fvantienen, line 12: Maybe this path `~/paparazzi` is not guaranteed to be the same at every computer. But I like the fixes for VSCode!

fvantienen (commented):
Will try to test this code today/tomorrow

.vscode/launch.json
Fabien-B: This PR is still a bit messy, I will remove this file before merging.
I started a VSCode plugin for paparazzi, but its far from being useful yet...

.vscode/c_cpp_properties.json
Fabien-B: Same comment, just some junk to be removed :smile:

fvantienen:
Quick question by looking at the code. The `iface->mutex` is not locked when calling `canardBroadcastObj` and `canardRequestOrRespondObj`. Then next to that shouldn't we release the mutex during `canTransmitTimeout` to avoid it being locked for 100ms if the transmit timed out.

fvantienen:
Couldn't reproduce the hanging of the autopilot without any dronecan device hanging on the bus, so will test in the drone hopefully today/tomorrow. But as a response to my previous comment I think we should do this: https://github.com/tudelft/paparazzi/commit/a106b362a866e26d002fb177c1b5a13b199ac621

fvantienen:
I tried to debug some more, but there are several problems which I currently can't identify directly.
My current status is:
- Trying to fix the mutex correctly and also release it in the callbacks such that it is handled correctly
- `actuators_dronecan_cmd_commit` only needs to transmit actuators which are set, as this increases the can bus usage significantly
- After fixing some mutexes I only see the `NodeStatus` message once. I think it has to do something with the threads and priorities and keeping things busy in combination with the dronecan actuators. Not sure yet how to properly debug this.
- If I have one of the `dronecan_bind`'s enabled in the `dronecanInitIface` it keeps the device busy

What I currently see on the bus after disabling the `dronecan_cmd` and all of the `dronecan_bind`'s in the `dronecanInitIface` is the raw dronecan actuators at 500hz.

Sorry for the vague debug information, but I need to take a better look at how to properly see what is going on and what is going wrong.

fvantienen:
I think I found a problem with the FDCan driver in chibios:
- It configures the RX fifo buffers to contain multiple messages (10-32 messages, depending on how many CAN busses are enabled)
- The `canReceiveTimeout` is called with `TIME_INFINITE` and waits for the `canp->rxqueue`
- The CAN interrupt handler checks the `FDCAN_IR_RF0F` which is triggered when the RX fifo is **full** (>=10-32 messages)
- This triggers the `canp->rxqueue` and the `canp->rxfull_event` and restarts the `canReceiveTimeout` thread
(This results in the blocking of the RC thread until it receives 10-32 messages and then parsing them all at once)

@Fabien-B What do you think?

fvantienen:
Next to that your driver has some other problems:
- The usage of static variables in functions that are called with an `iface` as argument. This is not possible as the different interfaces should not share variables.
- The `dronecan_bind` should not be called per `iface` but once in just the `dronecan_init`

Fabien-B:
I realize there is a lot of thing I don't know, so  will dive more deeply in it...
It's a code from a summer internship that I rebased on master.

Fabien-B:
Closing in favor of #3424

PR #3392, opened 2024-10-07
Add extra PPM input to Cube

dewagter:


OpenUAS:
Super, thx, now the issue will now be handled by PR #3403

PR #3394, opened 2024-10-14
Changes to Rotwing 7kg simulator and Actuator dynamics

tmldeponti:
This PR is to:

1. Update the dynamics of the motors to the newly identified values
2. Do not use Inertia scheduling for the 7kg oneloop
3. Make jsbsim model much easier to modify for another RW platform
4. Fix small indexing bug for the rot mech when using oneloop and the adc module


EwoudSmeur (approved):
LGTM!

PR #3395, opened 2024-10-14
Start of documentation for adding boards

FlorianSan:
A few lines to describe the process of integrating a new board into paparazzi.

However, no mention is made of the generic files used for the stm32h7.

### DIFF
--- a/doc/sphinx/source/developer_guide/index_developer.rst
+++ b/doc/sphinx/source/developer_guide/index_developer.rst
@@ -15,4 +15,5 @@ TBD
     abi
     logging
     debug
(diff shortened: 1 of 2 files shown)


fvantienen:
Really nice! I might add some things for the general board file, as it can save a lot of time

FlorianSan:
I've just added a bit of text describing the use of common_board.h. 
@fvantienen @Fabien-B  Is this OK for you?

gautierhattenberger (approved):
It looks correct

fvantienen (approved):

Issue #3321, opened 2024-06-19
PPM RC in no longer working on OrangeCube(+)


OpenUAS:
As of PPRZ master of 20240619 (unknown SHA sorry) the PPM input no longer works on an OrangeCube(+). PPM was as always connected to the  AUX OUTPUT  4 with RX a R-XSR  receiver set to PPM output. This way no UART was wasted :). RCIN officially at the time of writing still was not functional AFAIK.

OpenUAS:
@fvantienen , the by you proposed solution is the best, I agree, but we could live with an interim solution since it is a blocking issue for quite some users.

PR #3403, opened 2024-10-30
[chibios] PPM/PWM restructure

fvantienen:
This is changing the setup for the PPM and PWM in chibios to be more configurable.
- All the pins are now named `SERVO*`
- Every `SERVO*` pin can be used as PWM output or PPM input
- Timers for PWM or ICU are automatically enabled
- By default all pins are still set as PWM if not used by PPM (and timer is also not used by any ICU)

Question is do we want to fix the `stm32` arch? (For me I would like to drop/remove the `stm32` architecture preferably now)


OpenUAS:
* Really appreciated, time to hook-up a compute board to the now free UART again.
* As for  baremetal stm32 architecture drop, yeah fine, gives more motivation to convert remaining boards like lisa-s. 
 Q: But what about FWB firmware like on Pixhawk1?

fvantienen:
@gautierhattenberger and @Fabien-B what do you think about this?

OpenUAS:
@fvantienen still in Draft?else a review welcomed. Now we still need to resort to @alessandro-mancinelli hacked in solution(Which works) to pull in a working branch. @dewagter @others , Can setup the hardware for a desktest if needed.

alessandro-mancinelli:
@OpenUAS I think the PPM solution is already implemented in this branch, which is very close to master: https://github.com/tudelft/paparazzi/tree/rotwing_landing

fvantienen:
This requires a lot of work to fix all boards/airframes and IO think also drops support for the stm32 architecture.

gautierhattenberger:
Breaking the stm32 arch means that we don't support the px4io boards anymore. I don't use these boards, but maybe it is not wise to get rid of them without alternate solution. I don't know the status of ChibiOS for the support of the old STM32F1.
Also, what is the exact objective of this PR, beside a possible simplification of the configuration ? Is it solving a problem or increasing the possibilities ?

fvantienen:
It is doing both a simplification in configuration and also increases the possibilities. This makes it possible to configure every PWM pin as input/output without generating different board files. For the configuration it automatically enables the correct timers and gets everything from the board configuration file.

The ultimate goal is to make the generating of new boards extremely easy and flexible, such that you only need to generate the board.cfg file. Also make it more intuitive for configuring airframe files, such that you can just set your PWM output/input to SERVOx which matches with the name on the PCB. Ideally we also make similar alternative names for all UART's, I2C and other pins possible in the generation.

On the STM32F1 I'm ok with dropping the support, as we are currently not using it anymore. For the projects that we still use I prefer to upgrade the hardware than to keep support for outdated software.

OpenUAS:
For the FBW part of Pixhawk1 STM32F103, it should be able to run Chibios latest. Can test on that and debug on a Discovery board and Lisa/M 2 (not MX) or a Pixhawk1

OpenUAS:
See also https://github.com/paparazzi/paparazzi/issues/3321

PR #3404, opened 2024-10-30
Small fixes rotwing

NoahWe:
- Increased effectiveness scheduling frequency from 10Hz to 500Hz
- Adjusted V3F rotation mech limits
- Fix typo in V3B airframe file
- Fix compilation warnings for 7kg nps airframes

### DIFF
--- a/conf/modules/eff_scheduling_rotwing.xml
+++ b/conf/modules/eff_scheduling_rotwing.xml
@@ -35,7 +35,7 @@
     <file name="eff_scheduling_rotwing.h"/>
   </header>
   <init fun="eff_scheduling_rotwing_init()"/>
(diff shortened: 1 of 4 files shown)


conf/modules/eff_scheduling_rotwing.xml
```diff
@@ -35,7 +35,6 @@
     <file name="eff_scheduling_rotwing.h"/>
   </header>
   <init fun="eff_scheduling_rotwing_init()"/>
-  <periodic fun="eff_scheduling_rotwing_periodic()" freq="500.0"/>
```
fvantienen: Not remove the whole line, only the `freq` part, as it should just run with the main loop frequency.

fvantienen (approved):

PR #3409, opened 2024-11-12
State interface input filter

gautierhattenberger:
This PR is a proposal to add a filtering to the state interface setter functions. It allows to have a fine control of the modules allowed to push data to the state. Key elements are:
- Each module have a unique ID generated at compile time, which can be used to identify the source. In the future, we should consider these values as ID for ABI messages (and remove the hard-coded list).
- The AHRS primary/secondary selection is not centralised anymore. Each AHRS can be selected by settings or internal function calls. The main benefit is that number of AHRS is not limited. They are either primary (allowed to push data, one at a time) or secondary (not allowed, all the others).
- The INS don't have settings yet, but only minimal changes are still required to have multiple INS support.
- The state origin now have getter functions for a cleaner and safer access.

Some remaining work, for this PR or another one:
- [x] propagate ins_reset with ABI messages (for decoupling purpose, e.g. prepare next point)
- [x] support multiple INS

@fvantienen @dewagter @Fabien-B What is your opinion on this ?


fvantienen:
I think this a very nice improvement especially with multiple INS support! :)
I like the module ID generator, which can be really useful. But I see a problem when replacing the ABI id, since some modules need multiple id's. For example a multi-imu setup like the orange cube has 3 ABI id's for the different IMU's.

gautierhattenberger:
I'm working on the modifications required for multi-INS, it shouldn't be long.

Your comment on ABI ids is correct. The options I found are:
- the module requests in the xml file to have more than one ID
- the ABI id are changed to uint16 (will do that someday anyway) and a part of it is reserved for "components"

@fvantienen do you have a preferred solution ?

fvantienen:
Will take a look at it later.
The calibration could include the define of the ID and not the number itself maybe? This would solve the problem

gautierhattenberger:
> The calibration could include the define of the ID and not the number itself maybe? This would solve the problem

yes indeed, it is even better to detect some errors at compile time, if the driver is not loaded

fvantienen (approved):
If this is tested I'm ok with merging it :)

gautierhattenberger:
tested and now working with C++ (EKF)

fvantienen (approved):

PR #3413, opened 2024-11-27
Add betafpv AIO and tmotor AIO

FlorianSan:
- Add betafpv AIO board based on F722 (aiof7_betafpv test board)
- Add tmotor AIO board based on F722 (aiof7_tmotor test board)
- Fix useless define on tawaki_v2 configuration file for "#define STM32_HSE_BYPASS" linked to boardgen documentation
- Add two aiframes (Goose and Crow) on Enac airframe


fvantienen:
I think you could almost use the `common_board.h`(https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/arch/chibios/common_board.h) file instead of your own if you give the pin's the correct namings. Only thing that might be missing is the DSHOT, but you could either put that in the `board.cfg` header section or write a section inside the `common_board.h` to handle this.

This would remove a lot of 'double' code that we have all over our board files and also makes the naming of the pins more consistent. The ideal case is that we only need to create the `board.cfg` file for new boards, which I have been working towards. This currently works for stm32 H7, but because of the DMA's and the F7 you still need a separate mcuconf for that.

fvantienen:
One of the main things that I think is missing in the `board.cfg` generator is the feature to give pin's multiple names with different functions, or multiple names for the same function. For example:
```
TELEM1 -> UART4 # To enable the use of TELEM1 as a define/configure in boards for the correct UART
PA00    UART4_TX            UART    AF:UART4_TX # As default
*PA00   I2C1_SDA            I2C     AF:I2C1_SDA # As secondary option
*PA00 .... # Or even a third option for the same pin
```
* Note that it is currently possible to list multiple AF, but the pin still has the same name. That makes it difficult to identify which one should be enabled.

This would open the opportunity to use pin's for different functions very easily and would already work with the `common_board.h` and `mcuconf.h` as features/pins are only enabled when in use. Also makes it easier for people to find the correct port/name for things like UART's which usually have names on the board as well.

FlorianSan:
> I think you could almost use the `common_board.h`(https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/arch/chibios/common_board.h) file instead of your own if you give the pin's the correct namings. Only thing that might be missing is the DSHOT, but you could either put that in the `board.cfg` header section or write a section inside the `common_board.h` to handle this.
> 
> This would remove a lot of 'double' code that we have all over our board files and also makes the naming of the pins more consistent. The ideal case is that we only need to create the `board.cfg` file for new boards, which I have been working towards. This currently works for stm32 H7, but because of the DMA's and the F7 you still need a separate mcuconf for that.

I've modified the files to use common_board.h!

I've also talked to Alex about modifying the boardGen to handle aliases.

conf/flash_modes.xml
```diff
@@ -91,6 +91,8 @@
       <board name="matek_.*"/>
       <board name="holybro_.*"/>
       <board name="cube_.*"/>
+      <board name="betafpv_aiof7"/>
+      <board name="tmotor_aiof7"/>
```
gautierhattenberger, line 95: Is `SWD` the only flashing mode available ?

conf/flight_plans/ENAC/ENAC_stadium.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0"?>
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+<flight_plan alt="152" ground_alt="147" lat0="43 33 50.83" lon0="1 28 52.61" max_dist_from_home="100" name="Rotorcraft Basic (Enac)" security_height="2" geofence_sector="Flight_Area" geofence_max_alt="190">
+  <header>
+#include "autopilot.h"
```
gautierhattenberger, line 5: no needed

conf/flight_plans/ENAC/ENAC_stadium.xml
```diff
@@ -0,0 +1,102 @@
...
+    <block name="Holding point">
+      <call_once fun="NavKillThrottle()"/>
+      <attitude pitch="0" roll="0" throttle="0" vmode="throttle" until="FALSE"/>
+    </block>
+    <block name="Start Engine">
+      <call_once fun="NavResurrect()"/>
+      <attitude pitch="0" roll="0" throttle="0" vmode="throttle" until="FALSE"/>
+    </block>
+    <block name="Takeoff" strip_button="Takeoff" strip_icon="takeoff.png">
+      <exception cond="stateGetPositionEnu_f() @DEREF z @GT 2.0" deroute="Standby"/>
```
gautierhattenberger, line 58:
```suggestion
      <exception cond="GetPosHeight() @GT 2.0" deroute="Standby"/>
```

sw/airborne/boards/tawaki/chibios/v2.0/tawaki_2.0.cfg
```diff
@@ -19,7 +19,7 @@ HEADER
 
 #if !defined(STM32_HSECLK)
 #define STM32_HSECLK                8000000U
-#define STM32_HSE_BYPASS            TRUE
+#define STM32_HSE_BYPASS
```
gautierhattenberger, line 22: What is the problem with this ?

conf/airframes/test_boards/aiof7_betafpv.xml
```diff
@@ -0,0 +1,230 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="AIO F7 Test File">
+
+  <firmware name="fixedwing">
```
gautierhattenberger, line 5: Do we really want to test this board with a fixedewing firmware ? (same question for the tmotor board)

gautierhattenberger (changes requested):

conf/flash_modes.xml
FlorianSan: No, in usb dfu too! 
By default, the card is added to the paparazzi center in default mode.

sw/airborne/boards/tawaki/chibios/v2.0/tawaki_2.0.cfg
FlorianSan: In chibios, only the definition is tested 
#if defined(STM32_HSE_BYPASS)

So it makes no sense to set it to True.

conf/airframes/test_boards/aiof7_betafpv.xml
FlorianSan: No, these boards will always be used in a quad.

gautierhattenberger (approved):

PR #3415, opened 2024-12-04
[sys_time] Add GPS tow sync

fvantienen:
Add time of week synchronisation with the GPS for the system timer.

### DIFF
--- a/sw/airborne/arch/chibios/mcu_periph/sys_time_arch.h
+++ b/sw/airborne/arch/chibios/mcu_periph/sys_time_arch.h
@@ -48,4 +48,14 @@ extern void sys_time_usleep(uint32_t us);
 extern void sys_time_msleep(uint16_t ms);
 extern void sys_time_ssleep(uint8_t s);
 
+/**
+ * Get the time in milliseconds since the start of the week.
+ * If no GPS is connected this will return the time since startup
+ * @return Time of the week in milliseconds
+ */
+static inline uint32_t get_sys_time_tow(void)
+{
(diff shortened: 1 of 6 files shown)


sw/airborne/arch/chibios/mcu_periph/sys_time_arch.h
```diff
@@ -48,4 +48,14 @@ extern void sys_time_usleep(uint32_t us);
 extern void sys_time_msleep(uint16_t ms);
 extern void sys_time_ssleep(uint8_t s);
 
+/**
+ * Get the time in milliseconds since the start of the week.
+ * If no GPS is connected this will return the time since startup
+ * @return Time of the week in milliseconds
+ */
+static inline uint32_t get_sys_time_tow(void)
```
gautierhattenberger, line 56: why not adding this function to `sys_time.h` to avoid duplication ?

fvantienen: That was my initial idea, but there is some issues with includes because of me using the `get_sys_time_msec`. But maybe you have a proper way of how that could be fixed

gautierhattenberger: The problem is that a couple of functions are declared in the arch part because the early implementations where using `static inline` rather than normal functions. I think that with the compilers we have nowadays, it is a waste of time and a source of problems to optimize like that. I would move all the definitions (see `sys_time_arch.h` in chibios arch) to `sys_time.h` and implement them in the relevant `sys_time_arch.c` as normal functions (already the case for chibios).

PR #3423, opened 2025-01-22
Threads abstraction

Fabien-B:
This adds a threads abstraction layer to paparazzi.
Since we decided to drop bare metal support (e.g in #3403) , it is time to add a paparazzi abstraction for threads!

This PR add for both ChibiOS and linux with the same API:
- thread spawning and joining
- mutexes
- binary semaphores

It also adds a test module to experiment with it. Set the initial value of a Syracuse suit from the settings, and watch it count in the chat widget, cadenced by the main AP thread.

TODO:
- [x] Support linux arch -> Is it possible to use the same *_arch* files for both the `linux` and `sim` arch ?

Future uses:
One of the goals is to integrate the support of thread in the mcu_periphs, (with dedicated "blocking" functions) to allow writing peripheral drivers in a much simple way.


gautierhattenberger:
There are already a few files in `sim` that are symbolic links to the `linux` arch (uart, udp and pipe).

gautierhattenberger (approved):

fvantienen (approved):

PR #3424, opened 2025-01-28
Uavcan update

Fabien-B:
Updating UAVCAN to a recent version.
UAVCAN (a.k.a DroneCAN now) now generate serialization code from DSDL descriptions.

@fvantienen, @dewagter  I updated airspeed, power and range sensor, but I have trouble updating actuators.
The code is complex and I feel I will mess thing up.
Could you try to take a look at what I did in the last 3 commits and see if it could possible to update this module in the same way ?

Also, could you test this branch if you are using the modules I updated ?

Make sure you have these python modules installed, and run `make` to generate the (de)serialization code from the DSDL definitions.
```
empy==3.3.4
pexpect
```

The goal is to first update uavcan in this PR, then enable using FDCAN to benefit of the increased bandwitdth (in a future PR).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,6 +50,15 @@
 [submodule "sw/ext/matrix"]
 	path = sw/ext/matrix
 	url = https://github.com/PX4/Matrix.git
-[submodule "sw/ext/libcanard"]
-	path = sw/ext/libcanard
-	url = https://github.com/UAVCAN/libcanard.git
+[submodule "sw/ext/dronecan/DSDL"]
+	path = sw/ext/dronecan/DSDL
+	url = https://github.com/dronecan/DSDL.git
+[submodule "sw/ext/dronecan/dronecan_dsdlc"]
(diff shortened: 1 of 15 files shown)


fvantienen:
I will take a look later when I'm back from my next holiday

fvantienen:
I still need to test this, but the actuators are now also updated.

Would be nice to also change:
- The names to `dronecan` everywhere
- Use the `can` module
- Enable support for FDCAN

Fabien-B:
Cool !
This is the plan but I would like to do it step by step, so the PR are easier to understand.
I propose to merge this PR once you confirmed that it works, then I will look to use the `can` module (I already started). And then I can refactor all this to "dronecan" instead of "uavcan".

fvantienen (approved):

fvantienen:
@Fabien-B How far are you with the *can* module update? Because I might want to change some of the actuator/feedback code

PR #3430, opened 2025-02-10
CAN abstraction

Fabien-B:
Paparazzi abstraction for the CAN bus.
The structures are inspired by SocketCAN in linux.
Only the ChibiOS arch is implemented at the moment, linux will follow.
The plan is to make UAVCAN arch agnostic.
An other PR will follow for an SLCAN bridge, allowing to use DroneCAN GUI.

### DIFF
--- a/sw/airborne/mcu_periph/can.c
+++ b/sw/airborne/mcu_periph/can.c
@@ -56,7 +56,7 @@ uint8_t can_len_to_dlc(uint8_t len) {
 }
 
 
-void pprz_can_init()
+void can_init()
 {
   can_hw_init();
 }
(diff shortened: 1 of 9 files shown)


sw/airborne/mcu_periph/can.c
```diff
@@ -25,24 +25,68 @@
...
+  }
+  // no available slot
+  return -1;
 }
 
-void _can_run_rx_callback(uint32_t id, uint8_t *buf, uint8_t len)
-{
-  if (can_rx_callback) {
-    can_rx_callback(id, buf, len);
+int can_add_callback(can_rx_frame_callback_t callback, struct pprzaddr_can* src_addr, void* user_data) {
```
gautierhattenberger, line 78: name should be changed to avoid confusion with function above

gautierhattenberger (changes requested):
the can_arch files in stm32 are not compatible anymore, better remove them or fail at compilation with an error message

fvantienen (dismissed):

Fabien-B:
> the can_arch files in stm32 are not compatible anymore, better remove them or fail at compilation with an error message

It should be working now, although this is not tested.

gautierhattenberger (approved):

PR #3438, opened 2025-03-11
Airframe fixes TUD

dewagter:
- joystick mode sends commands of 0, 75 and 150 PPRZ units instead of -9600, 0, 9600
 - ```rotorcraft_control_loop``` has 2 options for ```MODE1```
 - ```autopilot``` modes now set in autopilot xml
 - Insufficient range for opticalflow_landing

### DIFF
--- a/conf/autopilot/rotorcraft_control_loop.xml
+++ b/conf/autopilot/rotorcraft_control_loop.xml
@@ -60,6 +60,7 @@
       <exception cond="RCLost()" deroute="FAILSAFE"/>
     </mode>
 
+<!--
     <mode name="ATTITUDE_Z_HOLD" shortname="A_ZH">
       <select cond="RCMode1()"/>
       <on_enter>
@@ -76,7 +77,7 @@
       </control>
       <exception cond="RCLost()" deroute="FAILSAFE"/>
     </mode>
-
+   -->
 
     <mode name="MODULE">
       <select cond="RCMode1()"/>
--- a/conf/autopilot/rotorcraft_control_loop.xml
+++ b/conf/autopilot/rotorcraft_control_loop.xml
@@ -60,25 +60,6 @@
       <exception cond="RCLost()" deroute="FAILSAFE"/>
     </mode>
 
-<!--
-    <mode name="ATTITUDE_Z_HOLD" shortname="A_ZH">
-      <select cond="RCMode1()"/>
-      <on_enter>
-        <call fun="guidance_h_mode_changed(GUIDANCE_H_MODE_NONE)"/>
-        <call fun="guidance_v_mode_changed(GUIDANCE_V_MODE_HOVER)"/>
-        <call fun="stabilization_mode_changed(STABILIZATION_MODE_ATTITUDE, STABILIZATION_ATT_SUBMODE_HEADING)"/>
-      </on_enter>
-      <control freq="NAVIGATION_FREQUENCY">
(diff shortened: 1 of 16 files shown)


conf/autopilot/rotorcraft_control_loop.xml
```diff
@@ -60,6 +60,7 @@
       <exception cond="RCLost()" deroute="FAILSAFE"/>
     </mode>
 
+<!--
     <mode name="ATTITUDE_Z_HOLD" shortname="A_ZH">
```
fvantienen, line 64: Not sure how this works and does this affect other drones?

gautierhattenberger: The purpose of this autopilot file is the MODULE mode, so their is no problem to remove this, I should have done that in the first place.
If someone think it is still useful, you can do something like this: https://github.com/paparazzi/paparazzi/commit/27860ffa4970d1e3685e78d4dbb87b82221ae6cd#diff-48f82a8991c7345f6c46ad66d35c26c069d40a5a3665f536ddf7e2b652b53057

conf/joystick/extreme_3d_pro.xml
```diff
@@ -20,9 +20,9 @@
 
 	<variables>
 		<var name="mode" default="0" />
-		<set var="mode" value="0" on_event="b3" />
-		<set var="mode" value="1" on_event="b4" />
-		<set var="mode" value="2" on_event="b5" />
+		<set var="mode" value="-127" on_event="b3" />
+		<set var="mode" value="0" on_event="b4" />
+		<set var="mode" value="127" on_event="b5" />
```
gautierhattenberger, line 25: It seems that all joystick files should be updated after #3332 ...

conf/autopilot/rotorcraft_control_loop.xml
dewagter: > Not sure how this works and does this affect other drones?

I did a grep and only bebop opticflow airframes use this.

dewagter: > The purpose of this autopilot file is the MODULE mode, so their is no problem to remove this, I should have done that in the first place. If someone think it is still useful, you can do something like this: [27860ff#diff-48f82a8991c7345f6c46ad66d35c26c069d40a5a3665f536ddf7e2b652b53057](https://github.com/paparazzi/paparazzi/commit/27860ffa4970d1e3685e78d4dbb87b82221ae6cd#diff-48f82a8991c7345f6c46ad66d35c26c069d40a5a3665f536ddf7e2b652b53057)

There are now two options that both have the same conditions of mode==1

conf/joystick/extreme_3d_pro.xml
dewagter: Wow, that sounds important. I'll have more joysticks updated then.

conf/autopilot/rotorcraft_control_loop.xml
fvantienen: Then instead of commenting just delete it?

conf/joystick/extreme_3d_pro.xml
dewagter: I updated all joystick files with variables

fvantienen (approved):

gautierhattenberger (approved):

PR #3440, opened 2025-03-14
[uavcan] Use CAN abstraction

Fabien-B:


Fabien-B:
I'm not sure  what to do about error handling.
It seems that you are not really using the `txempty_event`, apart maybe for waking up transmission when a can mailbox is available ?
https://github.com/paparazzi/paparazzi/blob/5facb3c93bdc3a85c7a9c9f55bc693def41cd321/sw/airborne/arch/chibios/modules/uavcan/uavcan.c#L152
I suppose in that case that the retransmission will try again a little bit later. Is this enough ?
https://github.com/paparazzi/paparazzi/blob/9c97520777e8fcb7f490bf8ace6833ec2b808dfb/sw/airborne/arch/chibios/modules/uavcan/uavcan.c#L118-L134

Fabien-B:
What should be done about the `error_event` ?
https://github.com/paparazzi/paparazzi/blob/5facb3c93bdc3a85c7a9c9f55bc693def41cd321/sw/airborne/arch/chibios/modules/uavcan/uavcan.c#L164-L167
I suppose it should be handled in the CAN module.
Should I check it in the Rx thread of the can module, and clear it like do did there, and maybe keep track of the number of errors ?

Fabien-B:
I tested it with an airspeed sensor.
I do not have any CAN ESC to test it with, so can you test it?
What do you think about the error handling? (See the last commit)

fvantienen:
I will try this soon and will look at the error handling. Thanks a lot for all the updates!

Fabien-B:
Hey @fvantienen, did you had the time to test this PR ?
I would like to add support for FDCAN at some point, but after this PR is merged.

Fabien-B:
@fvantienen do you think it could be merged ? Did you had the time to test it ?

fvantienen:
I will test now quickly and will merge if working :) Sorry for the delays, was very busy

fvantienen:
Tested and works, can be merged :) Once merged we will test it more

PR #3441, opened 2025-03-14
[doc] Update install instructions.

Fabien-B:
Removed instructions for Ubuntu18.04, which is EOL.


gautierhattenberger (approved):

dewagter:
I would love that you can install paparazzi with just:

``` git clone
install.sh```

This adapts to the OS, and allows to install only the desired parts.

PR #3462, opened 2025-06-03
[nav] add an option for takeoff height

gautierhattenberger:
Add a parameter to the takeoff function to specify the height at which the takeoff is considered finished. A negative value tells that the default value (configurable from airframe file) should be used.


Fabien-B (commented):
Maybe we should choose a different threshold for the default value ?
We will probably not be flying in the dead sea, but it may be an issue for flying in the Netherlands ?

gautierhattenberger:
It is the height above takeoff point, not the absolute altitude, so it should be fine even below sea level as long as you climb :)

Fabien-B (approved):

PR #3464, opened 2025-06-03
[hybrid] add basic support for Heewing T1 Ranger VTOL

gautierhattenberger:
The Heewing T1 Ranger VTOL is a tilt-rotor hybrid plane. ( https://www.heewing.com/products/t-1-ranger-vtol-pnp )As a first basic approach, two separated controls are used for hovering (INDI) and forward flight (PID). The mixing is done by a dedicated module and the autopilot is customized to support the specific modes.
A simple PID controller is added to the rotorcraft firmware as it is not possible to directly reuse the one from fixedwing firmware. In a future work, a full INDI version might be used, but this approach can still serves as an example of customized autopilot. A simulation model for JSBSim is also provided.

To avoid confusion,  there also exist a regular Heewing T1 Ranger, just fixedwing code can fly this plane.

### DIFF
--- a/conf/airframes/ENAC/hybrid/ranger_1.xml
+++ b/conf/airframes/ENAC/hybrid/ranger_1.xml
@@ -1,0 +1,356 @@
+<!DOCTYPE airframe SYSTEM "../../airframe.dtd">
+
+<airframe name="RANGER_T1">
+
+  <description>
+    HEEWING Ranger T1
+    Hybrid Quad/Plane with tilt motors
+    Tawaki v1.0 Chibios
+    Xbee API
+    Ublox
+    SBUS Futaba
+  </description>
+
+  <firmware name="rotorcraft">
+    <autopilot name="heewing_vtol"/>
+
+    <configure name="PERIODIC_FREQUENCY" value="500"/>
+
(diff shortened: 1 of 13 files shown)


OpenUAS:
Nice! The sooner PR honored the better since we have a few X450 ( https://www.wl-toys.com/XK-X450-RC-Plane-41488.html ) already paparazidized FC and also https://oscarliang.com/fimi-manta-vtol/ and now also acquired a T1 Ranger VTOL (https://www.heewing.com/products/t-1-ranger-vtol-pnp ). The Autopilot naming can be fixed later IMHO, love to testfly these aforementioned airframes ASAP. After that INDI would be welcomed.

fvantienen:
The basic PID fixedwing controller is really nice. It would be really awesome if we can make an example fixedwing plane just using that. This means that most of the old 'fixedwing' firmware becomes obsolete and in theory could be removed in the future.

gautierhattenberger:
@fvantienen I think the main issue with the fixedwing is not really the flight control laws, but the representation of waypoints (and altitude), in particular for long distance flight. I recall we have some tricks for that, but it could be better.
In the meantime, you agree if I merge this one ?

fvantienen (approved):

PR #3465, opened 2025-06-05
Mag viewer

Fabien-B:
Add a 3d visualization of the raw mag direction, to help during magnetometer calibration dance.
All mag data are normalised in order to be scale invariant.

image

### DIFF
--- a/sw/supervision/python/mag_viewer.py
+++ b/sw/supervision/python/mag_viewer.py
@@ -90,8 +90,14 @@ try:
             clear_button.clicked.connect(self.reset_data)
 
             self.distance_auto_chk = QtWidgets.QCheckBox("size auto", self)
+            self.distance_auto_chk.setSizePolicy(QSP.Policy.Fixed, QSP.Policy.Fixed)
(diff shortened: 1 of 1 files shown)


fvantienen (approved):
Ideally this can also do the calculation of the values directly and give the result

gautierhattenberger:
Before merging, I suggest some additions:
- protect the paparazzicenter from crashing if the required libraries are not installed
- add an option to filter on the sensor ID
- add a start/stop button to bind/unbind the messages and decrease the load on Ivy bus
- maybe placed the "Utilities" in different tabs ?

Of course, it can be improve later to perform the full calibration from the GUI.

Fabien-B:
- [x] protect the paparazzicenter from crashing if the required libraries are not installed
- [x] add an option to filter on the sensor ID
- [x] add a start/stop button to bind/unbind the messages and decrease the load on Ivy bus
- [x] maybe placed the "Utilities" in different tabs ?

image

dewagter (approved):
Works great. Used many times already.

PR #3468, opened 2025-06-11
[build] fix compilation warnings and remove unused code

gautierhattenberger:
Some code from base64 module in ocaml is deprecated but is only used by the outdated tool broadcaster. Better remove them.


Fabien-B (approved):
Always in favour of removing some old code :wink:

fvantienen (approved):

PR #3471, opened 2025-06-16
Threads I2C and SPI

Fabien-B:
Update the blocking I2C and SPI functions to use the new OS functionalities.
Using blocking functions in a thread (not the AP thread!) should simplify writing drivers, avoid the boilerplate of switch/cases.
Driver for SHT25 (temperature and humidity sensor) rewritten to demonstrate the use of the blocking functions.
Driver for AMT22 encoders added to test the SPI.

Tested on ChibiOS only.

### DIFF
--- a/sw/airborne/mcu_periph/i2c.c
+++ b/sw/airborne/mcu_periph/i2c.c
@@ -229,11 +229,6 @@ bool i2c_transceive(struct i2c_periph *p, struct i2c_transaction *t,
   return i2c_submit(p, t);
 }
 
-/** Default timeout for blocking I2C transactions */
-#ifndef I2C_BLOCKING_TIMEOUT
-#define I2C_BLOCKING_TIMEOUT 1.f
-#endif
-
 static enum I2CTransactionStatus i2c_blocking_submit(struct i2c_periph *p, struct i2c_transaction *t, float timeout) {
   if (!i2c_submit(p, t)) {
     return I2CTransFailed;
--- a/sw/airborne/mcu_periph/i2c.c
+++ b/sw/airborne/mcu_periph/i2c.c
@@ -234,68 +234,45 @@ bool i2c_transceive(struct i2c_periph *p, struct i2c_transaction *t,
 #define I2C_BLOCKING_TIMEOUT 1.f
 #endif
 
-bool i2c_blocking_transmit(struct i2c_periph *p, struct i2c_transaction *t,
-                           uint8_t s_addr, uint8_t len)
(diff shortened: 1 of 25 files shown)


sw/airborne/mcu_periph/i2c.h
```diff
@@ -330,8 +335,8 @@ extern bool i2c_transceive(struct i2c_periph *p, struct i2c_transaction *t,
  * @param len number of bytes to transmit
  * @return TRUE if insertion to the transaction queue succeeded
```
gautierhattenberger: Update documentation

sw/airborne/mcu_periph/i2c.c
```diff
@@ -234,68 +234,45 @@ bool i2c_transceive(struct i2c_periph *p, struct i2c_transaction *t,
...
   if (!i2c_submit(p, t)) {
-    return false;
+    return I2CTransFailed;
   }
 
   // Wait for transaction to complete
-  float start_t = get_sys_time_float();
-  while (t->status == I2CTransPending || t->status == I2CTransRunning) {
-    if (p->spin) p->spin(p);
-    if (get_sys_time_float() - start_t > I2C_BLOCKING_TIMEOUT) {
```
gautierhattenberger, line 252: remove `I2C_BLOCKING_TIMEOUT` define

sw/airborne/mcu_periph/spi.h
```diff
@@ -296,19 +298,7 @@ extern bool spi_submit(struct spi_periph *p, struct spi_transaction *t);
  * @param t spi transaction
  * @return TRUE if transaction completed (success or failure)
```
gautierhattenberger: update doc, remove `SPI_BLOCKING_TIMEOUT` define

Fabien-B:
Can it be merged ?

fvantienen:
Nice I really like this approach!

fvantienen:
We should be careful that the ABI calls, message calls, logging calls etc. in the drivers are then run in a different thread. So only thread safe functions should be called, as this could lead to problems in the future. Not sure if we can make a list of commonly used thread safe functions?

Fabien-B:
I agree that we should do a list of thread safe functions, and also a list of thread unsafe functions ! (at least the current ones).
I think it should be done in the sphinx documentation. If I have some time, I will start a list in a PR and wait for for inputs before merging.
It could be also a good starting point to make more functions thread-safe.

The idea for now is that you can use a semaphore to try "waiting" on it in a non blocking way in a periodic or event function.
So in the thread, you just have to signal the semaphore.

PR #3475, opened 2025-06-18
[rotwing] make setup for demo's simpler

NoahWe:
1. More simple set up procedure for the rotating wing drone when performing demonstrations.
2. Control the skew angle from an RC switch.

### DIFF
--- a/sw/airborne/modules/rotwing_drone/rotwing_demo.c
+++ b/sw/airborne/modules/rotwing_drone/rotwing_demo.c
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2025 Noah Wechtler <noahwechtler@tudelft.nl>
+ *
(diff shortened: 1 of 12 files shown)


sw/airborne/modules/rotwing_drone/rotwing_demo.c
fvantienen, line 1: We don't need a whole file for this. Just add some stuff to the flightplan instead. You can check the RC there also and also change the skew. Also the startup of the actuator check can be done like that.
There is no need for a full module, as the flightplan can do most.

Also I will not fully bypass the preflight checks! Just show the state of the checks with for example the elevator if they are successful and 'disable' the checks that are not needed specifically for the cyberzoo in the airframe file and or flightplan for the demo.

PR #3449, opened 2025-04-07
[rotwing] V3B Delivery

NoahWe:
Changes for Rotwing V3B Delivery:

- Adjusted low/critical battery levels
- Failsafe Throttle setpoint in static AP
- Relative flight plan
- More strict bank, climb and descend limits during transition
- Generated AP for rotwings

PR #3478, opened 2025-06-20
Airspeed consistency checks

NoahWe:
Create module for airspeed consistency checks as suggested in #3449.
~~Requires Pprzlink #205~~

### DIFF
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -265,6 +265,10 @@
       <field name="stamp" type="uint32_t" unit="us"/>
       <field name="relpos" type="struct RelPosNED *"/>
     </message>
+
+    <message name="AIRSPEED_CONSISTENCY" id="40">
+      <field name="gs_to_as_ratio" type="float"/>
+    </message>
   </msg_class>
 
 </protocol>
--- a/conf/abi.xml
+++ b/conf/abi.xml
@@ -265,10 +265,6 @@
       <field name="stamp" type="uint32_t" unit="us"/>
       <field name="relpos" type="struct RelPosNED *"/>
     </message>
-
-    <message name="AIRSPEED_CONSISTENCY" id="40">
-      <field name="gs_to_as_ratio" type="float"/>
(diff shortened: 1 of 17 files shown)


fvantienen:
Can't we add this ratio to the airspeed message? Seems a bit overkill to create a message just for that.

gautierhattenberger:
This is a bit confusing. It looks like an airspeed and wind estimator based on the same principal than the ground station estimator (https://journals.sagepub.com/doi/10.1260/1756-8293.4.1.15), but only used to check that True Airspeed to shifting from measured airspeed (which is normal if you don't compensate for air density variations).
I agree with @fvantienen that creating a telemetry message for that is overkill, especially considering that me are short on messages ID. But also, can you clarify what is the end goal of this code ? Why do you need an ABI as well ?

NoahWe:
The goal is to have a tool which lets us identify if our airspeed sensor is calibrated correctly, i.e. for us that is to check if MS45XX_PRESSURE_SCALE has the correct value, so that we can change it if necessary.

gautierhattenberger:
My concern is that it seems wrong to compare the airspeed from `stateGetAirspeed` which is en equivalent airspeed (EAS) and the radius of your ground speed adjusted circle that is the true airspeed (TAS) (https://docs.paparazziuav.org/latest/air__data_8h.html#ae9803888a95a6268efcb76c8adc5cff4). I might be wrong, but I think you should use `air_data.tas` instead.
About message part, since it seems to be a kind of warning, what about using directly the `INFO_MSG`. It will appear by default in the message widget of the GCS, and you can still bind to it in your custom interface and display the message.

NoahWe:
I think you're right about using `air_data.tas`.

gautierhattenberger:
It looks good to me, thank you for the changes.
Since I have just updated pprzlink in master, it seems you are out of sync again on this one. If you can rebase the branch on master without touching pprzlink ref that would be perfect.

PR #3479, opened 2025-06-23
Disable GPS_Fix for a given number of seconds

dewagter:
- Only in ABI messages
 - For testing purposes
 - Restore after X seconds (even if loss of telemetry)
 - Ground station still shows the real position (for safety)

### DIFF
--- a/sw/airborne/modules/gps/gps.c
+++ b/sw/airborne/modules/gps/gps.c
@@ -83,7 +83,7 @@ uint8_t multi_gps_mode;
 int gps_disable_fix = 0;              ///< Disable fix
 
 void gps_periodic_fix_counter(void) {
-#ifdef GPS_LOSS_TEST_TIMER
+#if GPS_LOSS_TEST_TIMER
   if (gps_disable_fix > 0) {
     gps_disable_fix--;
   }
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
It looks a bit dangerous. What is the expected outcome of this test ? Trigger the gps lost conditions ?

dewagter:
Several types of projects need to be tested to fly without GPS.

 - computer-vision-only projects
 - some larger drones flying for instance over sea, prefer to try to continue flying in alternate navigation modes than trigger ```FAILSAFE```. 

Both need to be tested. To make things as safe as possible, the GPS is disabled only for a short time and always restores automatically, even if datalink is lost. In the mean time, datalink still has the real GPS data to track status.

sw/airborne/modules/gps/gps.c
```diff
@@ -80,6 +80,13 @@ static uint8_t current_gps_id = GpsId(PRIMARY_GPS);
 #endif
 
 uint8_t multi_gps_mode;
+int gps_disable_fix = 0;              ///< Disable fix
```
gautierhattenberger, line 83: Can you add some more detailed comments about this feature, as it is dangerous if wrongly used.
What do you think about adding a flag to disable the feature by default ?

dewagter: OK

gautierhattenberger (approved):

PR #3480, opened 2025-06-23
Activate barometric altitude in fixedwing Disco

dewagter:


gautierhattenberger:
I would like to get rid of these baro_board modules as much as possible. Would it be possible to just load the `baro_ms5611_i2c` module in your airframe if you need it ?

dewagter:
Replaced board with sensor module. Not sure it still works. If this is the way to go I'll test it.

gautierhattenberger (approved):
can be merged if it works for you @dewagter

dewagter:
Was tested on the real drone. It now works, but needs another review.

OpenUAS:
Screenshot from 2025-06-27 14-45-06

Works for me on Disco now.

OpenUAS (approved):

PR #3482, opened 2025-06-23
16 bit telemetry

dewagter:


fvantienen:
The mavlink messages have ID's in 16bit. This adds support for that

sw/airborne/modules/datalink/telemetry.c
```diff
@@ -48,9 +48,9 @@ struct periodic_telemetry pprz_telemetry = { TELEMETRY_PPRZ_NB_MSG, telemetry_cb
  * @param _cb callback function, called according to telemetry mode and specified period
  * @return -1 on failure to register, index of callback otherwise
  */
-int8_t register_periodic_telemetry(struct periodic_telemetry *_pt, uint8_t _id, telemetry_cb _cb)
```
gautierhattenberger, line 51: The input id can be extended to uint16, but the return type corresponds to the index in the registration table. So unless you plan to register more than 127 different messages in your telemetry file, it is not necessary to change its type.

sw/airborne/modules/datalink/telemetry.c
```diff
@@ -112,7 +112,7 @@ void telemetry_reporting_task(void)
  * @param _mode telemetry mode
  * @param _id id of the message in telemetry system (see var/<AC>/generated/periodic_telemetry.h)
  */
-void periodic_telemetry_err_report(uint8_t _process, uint8_t _mode, uint8_t _id)
+void periodic_telemetry_err_report(uint8_t _process, uint8_t _mode, uint16_t _id)
```
gautierhattenberger, line 115: note that this parameter is cast to uint8 to be sent with the `TELEMETRY_ERR` message

sw/airborne/modules/datalink/telemetry_common.h
```diff
@@ -60,7 +60,7 @@ struct telemetry_cb_slots {
  *  and the list of registered callbacks
  */
 struct periodic_telemetry {
-  uint8_t nb;                     ///< number of messages
+  uint16_t nb;                    ///< number of messages
```
gautierhattenberger, line 63: this is the maximum number of messages to be registered, not related to the actual message ID.

sw/airborne/modules/datalink/telemetry.c
fvantienen: I know, didn't want to change the message. But it might be better to do that

dewagter:
See : https://github.com/paparazzi/paparazzi/pull/3486/files#diff-aac293192fbb00fab60e037225c39ae3ef8620a0b2e9279c94ab1d2b1bc5f783R151

PR #3483, opened 2025-06-23
Fast Reboot When in-flight

dewagter:


sw/airborne/arch/chibios/mcu_arch.c
```diff
@@ -54,6 +54,31 @@ static void init_ram_areas(const ram_init_area_t *rap);
...
+CH_IRQ_HANDLER(BusFault_Handler)
+{
+  mcu_reboot(MCU_REBOOT_FAST);
+}
+
+CH_IRQ_HANDLER(UsageFault_Handler)
+{
+  mcu_reboot(MCU_REBOOT_FAST);
+}
+
```
gautierhattenberger, line 81: It seems to be colliding with the hard fault recovery mode. What is the purpose of this ?

sw/airborne/arch/linux/mcu_periph/sys_time_arch.c
```diff
@@ -46,6 +46,12 @@ void *sys_time_thread_main(void *data);
 
 #define NSEC_OF_SEC(sec) ((sec) * 1e9)
 
+
+
+void mcu_set_fast_reboot(void) {
```
gautierhattenberger, line 51: why this in `sys_time_arch` ?

sw/airborne/autopilot.c
```diff
@@ -160,6 +161,13 @@ void autopilot_periodic(void)
     // after takeoff for fixedwing
     // when in flight for other firmwares
     RunOnceEvery(PERIODIC_FREQUENCY, autopilot.flight_time++);
+
+    if(!fast_reboot) {
```
gautierhattenberger, line 165: Why not in the `autopilot_init` function ? It seems to be called only once.

sw/airborne/arch/chibios/mcu_arch.c
fvantienen: This needs some rewrite, but we don't want the hard fault recovery. But still reboot and skip the bootloader

sw/airborne/arch/linux/mcu_periph/sys_time_arch.c
fvantienen: Nees to move, was a hack

sw/airborne/autopilot.c
fvantienen: Needs to be called only when `in_flight` else you will always skip the bootloader at a reboot. That makes it very hard to flash the autopilot. But will find a correct place

sw/airborne/arch/chibios/mcu_arch.c
gautierhattenberger: If you activate the `USE_HARD_FAULT_RECOVERY` flag, but set the `hard_fault` to `false` (with a configurable value, default to `true`), do you have the expected result ? I would avoid duplicating the reset handlers.

dewagter:
The goal is to regain ```RC Manual``` control in less than a second upon airborne reboot. This is now not the case.

gautierhattenberger:
Do you mean RC manual for rotorcraft or fixedwing ? The recovery mode should exactly do that (and only that) for fixedwing (as long as you set your RC switch to manual). For rotorcraft, the alignment of the AHRS/INS makes it complicated to have control at all.

dewagter:
I mean ```fixedwing```. In our tests, it takes several seconds to pass the bootloader by default. On top of that, when the RC is in manual, it does not drive the servos until there is GPS/AHRS (not sure which) or it gets stuck in HOME mode. I can show you a video if you want where we deliberately trigger a segfault and measure how long it takes for servo's to start moving again while in fixedwing RC-direct. With this PR (and some deeper modifications in the HOME logic) there was just a tiny 0.1s jitter on servos upon reboot. The reboot HOME logic will be in a separate PR. We can try a different cleaner way to get the same results, but basically, I'd like that if RC is OK and in manual in fixedwing, there is a possibility to have it work instantly.

OpenUAS:
Very welcomed if this could be integrated, albeit in the best way, since in the age of hybrid airframes, where we could have saved and skipped the rebuild process :(( since even a second is sometime too long... Then we could spend time on where the Hardfault came from instead of re-building :)

gautierhattenberger:
We don't use a bootloader and in this case the hard fault recovery works as expected, and probably just like you want as well.
The problem I see here is that it is not compatible with the recovery mode since it is redefining some handler. From what I see, it is not even exclusive (one configuration or the other).
Ideally, There should be only one recovery mode that works with and without bootloader, with some options if needed like the `fast_reboot`  function (and move it outside of `sys_time` module).

PR #3485, opened 2025-06-23
Air Tag On/Off

dewagter:


fvantienen (changes requested):
I would close this PR as it only works with the PX4 bootloader and is an extremely specific hack.

PR #3487, opened 2025-06-23
Wait for shaking to take-off

dewagter:


fvantienen:
Wouldn't it be better to not listen to the accel but request the `getBodyAccel` from the state? Because you are only using the last value anyways.

sw/airborne/modules/nav/nav_shakestart.c
```diff
@@ -0,0 +1,112 @@
...
+  // If shake_state is even, we are waiting for a positive acceleration
+  if (shake_state % 2 == 0) {
+    // Check if the acceleration is above a threshold
+    if (nav_shakestart_accel > 12.0f) { // Threshold of 2g
+      timer = 15;
+      shake_state++;
+    }
+  } else {
+    // Check is the acceleration is below a threshold
+    if (nav_shakestart_accel <= 12.0f) { // Threshold of -2g
```
gautierhattenberger: I don't really get the threshold of 12 for +/- 2g ?

sw/airborne/modules/nav/nav_shakestart.c
```diff
@@ -0,0 +1,112 @@
...
+    nav_shakestart_detected = true;
+  }
+
+}
+
+bool nav_shakestart_run(void)
+{
+  NavAttitude(RadOfDeg(0.f));
+  NavVerticalAutoThrottleMode(0.5f);
+  NavVerticalThrottleMode(0.f);
```
gautierhattenberger, line 108: Shouldn't you do that in your flight plan ? Here the code is only working for fixed-wing because of the nav header include.

sw/airborne/modules/nav/nav_shakestart.c
dewagter:
```suggestion
    if (nav_shakestart_accel <= 12.0f) { // Threshold of 1.2g
```

sw/airborne/modules/nav/nav_shakestart.c
dewagter: Yes, this module was meant for fixedwings. I'm not sure it is also applicable to rotorcraft. But maybe it can be indeed used to trigger a flightplan exception rather than as a navigation block. Would you prefer that?

gautierhattenberger: It's true that it looks dangerous to start a rotorcraft like this, so it is up to you...
By the way, there is a takeoff module based on pitch angle (https://docs.paparazziuav.org/latest/module__takeoff_detect.html). Not sure if it is useful for you, depends on the plane...

dewagter: I can fuse both (add the shaking functionality to the take-off-detect).

gautierhattenberger (approved):
If someone ever needs this for rotorcraft, the necessary changes can be done later.

OpenUAS:
Tried it on fixed wing and works well, IMHO can be merged. Since the module also does not auto-magical lands up in one's own airframe, the risk of incorrect use is somewhat low.

PR #3492, opened 2025-07-01
[natnet] use scipy rotation module instead of pyquaternion

gautierhattenberger:
scipy is already part of the dependencies and can be installed from system packages which is not the case for pyquaternion only available with pip

@fvantienen or @dewagter, could you test it for your setup at TUDelft ?


Fabien-B (approved):
tested in ENAC

gautierhattenberger:
@dewagter or @fvantienen , did you get a chance to check that it is still working for you in your flight arena ?

PR #3477, opened 2025-06-19
Imu scale float

Fabien-B:
Imu scales in float.

Still some stuff to fix.

PR #3495, opened 2025-07-02
[imu] fix some bugs introduced with imu float refactor

NoahWe:
- Fix some bugs introduced during rewrite
- Use FloatRates for gyros


fvantienen:
Errors introduced in #3477

dewagter (approved):
Urgent and important.

fvantienen (approved):

Fabien-B:
I'm not sure I agree with using FloatRates for the gyro scaling (nor strongly disagree).
A rate should be in rad/s or something similar, but the scales values are not exactly in that unit.

fvantienen:
It was more for the use of `.p, .q, .r` and since the int32 version was in rates we chose to change this. Having `x, y, z` with rates feels weird.

PR #2890, opened 2022-07-23
[imu] Multi IMU support

gautierhattenberger:
Reopening #2883 in the paparazzi repository

Multi IMU support
+ fixing compilation in default "single" IMU case

PR #3496, opened 2025-07-03
Current compensation code is removed 

dewagter:
- The following defines don't do anything anymore.
 - Was the current compensation for magnetometers removed on purpose?


gautierhattenberger:
The functionality was dropped with the support of multi IMU #2890. If we add this back, it needs to be done with the new way define sensor calibration anyway, so we can remove the old definitions.

PR #3497, opened 2025-07-03
[ins] cleaning of external pose INS

gautierhattenberger:
- refactor some code
- more configurable
- reduce the number of lines with values set to 0 (replaced by memset)
- fix state update (pointer cast this way is undefined behavior)


dewagter:
Are you also flying this code indoors?

gautierhattenberger:
We are just starting to test it for indoor flights and found a bug (also related to #3498). Now it seems to work all good.

PR #3499, opened 2025-07-06
Fix warning of nested extern declaration of fn get_sys_time_float

OpenUAS:
Would fix this:

```
arch/stm32/modules/core/threads_arch.c: In function 'pprz_bsem_wait_timeout':
arch/stm32/modules/core/threads_arch.c:55:20: warning: implicit declaration of function 'get_sys_time_float' [-Wimplicit-function-declaration]
55 | float time_end = get_sys_time_float() + timeout;
| ^~~~~~~~~~~~~~~~~~
arch/stm32/modules/core/threads_arch.c:55:20: warning: nested extern declaration of 'get_sys_time_float' [-Wnested-externs]
```

### DIFF
--- a/sw/airborne/arch/stm32/modules/core/threads_arch.c
+++ b/sw/airborne/arch/stm32/modules/core/threads_arch.c
@@ -13,7 +13,7 @@
 #include "modules/core/threads.h"
 #include "modules/core/threads_arch.h"
 #include "stdbool.h"
-
+#include "mcu_periph/sys_time.h"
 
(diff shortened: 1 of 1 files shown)


OpenUAS:
Appriciate if we could set  -Werror-implicit-function-declaration , -Wall is not enough for that, [take a look at this great article](https://mcuoneclipse.com/2021/02/02/no-excuse-for-implicit-function-declaration-warning-in-c/)

PR #3505, opened 2025-07-15
[can fuelcell] use DSDL definition

Fabien-B:
Add custom definition of dronecan message to paparazzi.
Changed message id to comply with the [ID distribution convention](https://dronecan.github.io/Specification/5._Application_level_conventions/#id-distribution).


fvantienen:
Wow really nice, thanks!

fvantienen (approved):

Fabien-B:
Don't forget to test it, I don't have a fuelcell to test it, and I changed the message ID, so the fuelcell must be updated as well !

PR #3508, opened 2025-07-24
[uavcan] memory offset can lines

NoahWe:


Fabien-B:
Hey @NoahWe can you explain what the issue was ? I'm having trouble with the CAN, so I wonder if it's linked or not to that ?

Fabien-B:
I see there was an error in this function, there was still some hardcoded `can1_arch_s.` that should be replaced by `cas->`

fvantienen:
There were still errors, but the memory for the two can interfaces should not overlap/use the same memory. That is why the memory offset needs to be there.
We had issues with consistency of transmitting/receiving on the CAN1 bus when also CAN2 is enabled, but I guess the mistake you mentioned might fix the problems we encountered.

Fabien-B:
My issue is that I can't send a proper CAN frame. With the logic analyser I see this:
image
All the data are missing.

Fabien-B:
I also have an issue on reception.
I have a sensor that send continuously the same message, and I subcribed a callback for that message.
Although `shouldAcceptTransfer` returns `true` at each message, `onTransferReceived` is not called one out of two times.

In `can_frame_cb` I replaced
```C
canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
```
by
```C
  int16_t ret = canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
  if(ret != CANARD_OK) {
    chSysHalt("angry duck");
  }
```
And now  `shouldAcceptTransfer` is called for every messages, and it never halts (so I guess that `ret==CANARD_OK`)

NoahWe:
Hi, @Fabien-B it should be fixed now can you verify?

fvantienen (approved):
There was both the problem of the misconfiguration of CAN2 and not setting the `CANTxFrame` to `0`. Which both caused problems when transmitting frames.

Fabien-B:
Yeah it works !
Thanks for fixing it, I was loosing my mind over it :smile: 
I still have the issue that the callback is not called if I don't use the return value of `canardHandleRxFrame`, but I will fix it in a future PR.
I am currently implementing the dynamic ID allocation, I will do a PR _soon™_

Fabien-B (approved):

PR #3510, opened 2025-07-28
Add Bayesian Obstacle Grid Mapping with TFmini Lidar Support and Simulation

alrochas1:
This PR introduces a Bayesian obstacle grid mapping module for rovers, integrated with a TFmini lidar capable of servo-based scanning. It includes simulation support, new configuration files for testing, and all required interfaces with the GCS.

###  Obstacle Grid Mapping
- New module: `rover_obstacles`
  - Generates a Bayesian grid map of obstacles.
  - Provides utilities to fill, update, and send grid cells to the GCS.
  - Communicates via a new ABI message with the Lidar module.

### TFmini Lidar Enhancements + Simulation
- Extended `tfmini` driver:
  - Optional servo control to enable 2D scanning.
  - Added a callback interface to decouple data handling (used by the grid mapper).
  - Lidar + servo simulation added to NPS.

### Configurations & Demo Files
- New demo configuration for testing the new functionality:
  - Use `conf/airframes/UCM/conf.xml` to see the map.
  - The map is drawn below the tile source layer by default, so you may need to disable the tile source layer to view it clearly.

### PR Related

- [paparazzi/pprzlink#207](https://github.com/paparazzi/pprzlink/pull/207) – Adds new messages for obstacle cell transmission.
- [paparazzi/PprzGCS#29](https://github.com/paparazzi/PprzGCS/pull/29) – Adds visualization of the obstacle grid map in the GCS.

### DIFF
--- a/sw/airborne/modules/lidar/tfmini.c
+++ b/sw/airborne/modules/lidar/tfmini.c
@@ -36,9 +36,6 @@
 #include "pprzlink/messages.h"
 #include "modules/datalink/downlink.h"
 
-// Test
-#include "firmwares/rotorcraft/autopilot_static.h"
-
 
 // Horizontal distance from the IMU to the LiDAR (in meters)
 #ifndef LIDAR_OFFSET
--- a/sw/simulator/nps/nps_autopilot.h
+++ b/sw/simulator/nps/nps_autopilot.h
@@ -54,13 +54,8 @@ extern struct NpsAutopilot nps_autopilot;
 
 extern bool nps_bypass_ahrs;
 extern bool nps_bypass_ins;
-extern bool nps_bypass_lidar;
 extern void sim_overwrite_ahrs(void);
 extern void sim_overwrite_ins(void);
-
-#ifdef NPS_BYPASS_LIDAR
-extern void sim_overwrite_lidar(void);
-#endif
 
 extern void nps_autopilot_init(enum NpsRadioControlType type, int num_script, char *js_dev);
 extern void nps_autopilot_run_step(double time);
--- a/sw/simulator/nps/nps_autopilot_rover.c
+++ b/sw/simulator/nps/nps_autopilot_rover.c
@@ -179,7 +179,7 @@ void nps_autopilot_run_step(double time)
 
 
 void sim_overwrite_ahrs(void)
-{
+{ 
   stateSetInputFilter(STATE_INPUT_ATTITUDE, MODULE_NPS_ID);
 
   struct FloatQuat quat_f;


conf/abi.xml
```diff
@@ -265,6 +265,13 @@
       <field name="stamp" type="uint32_t" unit="us"/>
       <field name="relpos" type="struct RelPosNED *"/>
     </message>
+
+    <message name="LIDAR_SERVO" id="40">
```
gautierhattenberger: There is already a `OBSTACLE_DETECTION` message that is very close to this one, except for the timestamp. Would it be possible to use it here, eventually by adding the `stamp` field if needed ?

conf/airframes/UCM/rover_lidar.xml
```diff
@@ -0,0 +1,180 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+
+<airframe name="boat">
+
+  <firmware name="rover"> 
+    <autopilot name="rover_steering.xml"/>
+    <define name="USE_INS_NAV_INIT"/>
```
gautierhattenberger: note that the `.xml` extension is not needed, nor the `USE_INS_NAV_INIT` which is set to TRUE by default in all INS filters.

conf/modules/obstacle_rover.xml
```diff
@@ -0,0 +1,40 @@
...
+  
+  
+  <header>
+     <file name="rover_obstacles.h"/>
+  </header>
+  <periodic fun="decay_map()" freq="100" autorun="TRUE"/>
+   
+  <makefile firmware="fixedwing">
+     <file name="rover_obstacles.c"/>
+  </makefile>
```
gautierhattenberger: Are you sure about using this module for fixedwing firmware ?

sw/airborne/firmwares/rover/obstacles/rover_obstacles.h
```diff
@@ -0,0 +1,89 @@
...
+
+#ifndef N_COL_GRID
+#define N_COL_GRID 100
+#endif
+
+#ifndef N_ROW_GRID
+#define N_ROW_GRID 100
+#endif
+
+#include "firmwares/rover/navigation.h"
```
gautierhattenberger: move to C file if not needed here

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,417 @@
...
+/** @file rover_obstacles.c
+ *  @brief functions to create a grid map
+ *
+ */
+
+
+// Depends of tfmini lidar(could be replaced by other lidar)
+
+#include "rover_obstacles.h"
+#include "modules/lidar/tfmini.h"
```
gautierhattenberger: This include should not be needed, that is the point of using an ABI internal message.

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,417 @@
...
+
+  memset(obstacle_grid.world, 0, sizeof(obstacle_grid.world));
+
+	#if PERIODIC_TELEMETRY
+		register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_GRID_INIT, send_grid_init);
+	#endif
+
+	#ifndef USE_EKF_SLAM
+		init_walls(); // Initialize the walls for NPS
+		AbiBindMsgLIDAR_SERVO(AGL_LIDAR_TFMINI_ID, &lidar_ev, lidar_cb);
```
gautierhattenberger: Don't bind explicitly to TFMINI sensor but rather to the broadcast address and make it configurable.
See: https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/modules/range_forcefield/range_forcefield.c#L91

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,417 @@
...
+    convert_walls_to_ltp();
+  }
+
+	// In case there is real measured for the Lidar
+	if (distance == 0.0){
+		return;
+	}
+
+  // Everything is in ENU
+  float x_rover = POS_FLOAT_OF_BFP(stateGetPositionEnu_i()->x);
```
gautierhattenberger: use `stateGetPositionEnu_f()` to have position in float directly

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,417 @@
...
+
+// Global variables (to avoid recalculating log all the time)
+float POCC = 0;
+float PFREE = 0;
+float PT = 0;
+
+int8_t LT, LOCC, LFREE;
+
+world_grid obstacle_grid;
+uint8_t grid_block_size = GRID_BLOCK_SIZE;
```
gautierhattenberger: These variables should be declared `static`

sw/airborne/modules/lidar/tfmini.c
gautierhattenberger, line 1: I guess what you are doing here is put a tfmini lidar on a servo to make a planar scan ?
It is fine but you can't integrate this servo control in the lidar driver, these are two separated things. You should make a new module dedicated to moving a lidar or range sensor and producing the required output (an ABI message with direction and distance if I'm not wrong).

sw/simulator/nps/nps_autopilot.h
```diff
@@ -54,6 +54,7 @@ extern struct NpsAutopilot nps_autopilot;
 
 extern bool nps_bypass_ahrs;
 extern bool nps_bypass_ins;
+extern bool nps_bypass_lidar;
```
gautierhattenberger: This variable don't seem to be used, at least in NPS (it doesn't have support for this kind of sensor anyway). If you want to produce a simulation mode for the sensor, it should be done in your driver module.

sw/simulator/nps/nps_autopilot_rover.c
```diff
@@ -179,8 +185,7 @@ void nps_autopilot_run_step(double time)
 
 
 void sim_overwrite_ahrs(void)
-{ 
-  stateSetInputFilter(STATE_INPUT_ATTITUDE, MODULE_NPS_ID);
```
gautierhattenberger: Why removing this ?

sw/airborne/modules/lidar/slam/lidar_correction.c
```diff
@@ -0,0 +1,206 @@
...
+    psi_counter++;
+  }
+  #endif
+  return min_distance;
+}
+
+
+
+// Example of obstacles (this need to be structure in a better way)
+void init_walls(void) {
```
gautierhattenberger: I agree, these "walls" shouldn't be hard-coded !

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
gautierhattenberger, line 1: Please use the script `./fix_code_style.sh` to adjust the formatting to the coding rules (see https://docs.paparazziuav.org/latest/stylec.html)

conf/modules/lidar_tfmini.xml
gautierhattenberger, line 1: Please don't change the default values for this module (especially if you can redefine them for your needs in the airframe file).
Also, the `lidat_correction` part should be a separated module (it can depend explicitly to this module if it is tightly related, though).

gautierhattenberger (changes requested):

gautierhattenberger:
As a side note, we are currently working with the rover firmware to add support to some features like "mission" and "guided" modes. Because of that, I found an error in the simulator where the heading/steering direction are wrong (due to a strange mix of NED/ENU frames) but "compensated" by a minus sign in the macro `GetCmdFromDelta` (and maybe other errors in the servo direction in the airframe file).
I have fixed this our dev branch (https://github.com/enacuavlab/paparazzi/tree/rover_steering_pid). I'll make a pull request for that around the end of the month, when other features are also integrated.

conf/abi.xml
alrochas1: That's right, now it used the `OBSTACLE_DETECTION` message. `LIDAR_SERVO` deleted (stamp wasn't needed).

conf/airframes/UCM/rover_lidar.xml
alrochas1: Corrected

conf/modules/obstacle_rover.xml
alrochas1: That was a typo. Corrected.

sw/airborne/firmwares/rover/obstacles/rover_obstacles.h
alrochas1: Moved

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: That's right. Deleted.

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: Corrected.

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: The variables `POCC`, `PFREE`, `PT`, `LT`, `LOCC`, and `LFREE` should be static. Both `obstacle_grid` and `grid_block_size` are needed in the `obstacle_rover.xml` settings. Updated

sw/simulator/nps/nps_autopilot_rover.c
alrochas1: Another typo. Added back

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: Done.

sw/airborne/modules/lidar/tfmini.c
alrochas1: I've refactored this into three separate modules.
- The basic `lidar_tfmini` driver. Removed all servo-related code from tfmini.c
- A new `servo_lidar` module
- The `lidar_correction` module for simulation support

Changes are in [69778fb1fa370c5dce4db19798ee434a6c9d3048]

sw/simulator/nps/nps_autopilot.h
alrochas1: It is now used since I have separated the lidar simulation from the `tfmini` module.

conf/modules/lidar_tfmini.xml
alrochas1: About the default values - I'm not entirely sure which specific default values you're referring to. If its related with the `lidar_correction` part, its already removed from here (see [69778fb1fa370c5dce4db19798ee434a6c9d3048]).

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: Done [c114d0221c4d811627f0b9348f8a6bb29765e58f]

sw/airborne/modules/lidar/slam/lidar_correction.c
alrochas1: This a little harder. I've moved the hard-coded obstacles to a .h file, and ntroduced a modular configuration system, allowing multiple map configurations, see [ea775c471ebbbff65e7c7c51fe263d0dd12a2413].
- The obstacles are now at `sw/airborne/modules/lidar/maps/ucm_football.h` (all maps should be in this folder)
- There is a new file `sw/airborne/modules/lidar/maps/maps_common.h` with the structures defined
- In the conf file `lidar_correction.xml`, you can change the variable `OBSTACLES_CONFIG_FILE` to defined which map to use.

Ideally, the maps would be moved to a conf file, and the .h file would be automatically generated, but for now, I think this should work.

alrochas1:
Regarding the heading/steering direction issue mentioned in your note - I had observed this same behavior in our testing. Interestingly, after implementing these changes, the issue appears to have resolved itself unexpectedly (not sure why).

sw/airborne/modules/lidar/tfmini.h
```diff
@@ -60,9 +62,22 @@ struct TFMini {
 
 extern struct TFMini tfmini;
 
+struct TFMiniServo {
+  int32_t pos;
+  float ang;
+  bool dir;
+};
+
+
```
gautierhattenberger: It should be removed or moved to another module

conf/modules/servo_lidar.xml
```diff
@@ -0,0 +1,34 @@
...
+
+  <header>
+    <file name="servo_lidar.h"/>
+  </header>
+  
+  <init fun="servoLidar_init()"/>
+  <periodic fun="servoLidar_periodic()" freq="100" autorun="TRUE"/>
+
+  <makefile target="ap|nps">
+    <configure name="SERVO_LIDAR_SPEED" default="5"/>
```
gautierhattenberger: This line has no effect at the moment. It creates a Makefile variable with a default value but don't use it to make a compile flag (e.g. with `define` node)

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,426 @@
...
+#endif
+
+
+#define DECAY_INTERVAL 5000 // ms for obstacle probability to decay
+#define DECAY 5 // Decay per second
+static uint32_t last_s = 0;
+
+
+#ifdef USE_EKF_SLAM
+#include "modules/ins/ins_slam_ekf.h"
```
gautierhattenberger: I don't see this INS in your pull request, I guess it is a local development for you, but if you don't plan to make a pull request for that, I'd prefer that you remove it for now.

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
```diff
@@ -0,0 +1,426 @@
...
+
+#ifndef USE_EKF_SLAM
+
+
+void ins_update_lidar(float distance, float angle)
+{
+
+  if (!ins_int.ltp_initialized) {
+    return;
+  }
```
gautierhattenberger: Why is it explicitly referring to `ins_int` ? If it is necessary to run the conversion function below, better use `stateIsLocalCoordinateValid()` function which is independent of the INS filter.

sw/airborne/modules/lidar/maps/ucm_football.h
```diff
@@ -0,0 +1,55 @@
+// modules/lidar/maps/ucm_football.h
+
+// TODO: Maybe try to use a code to automatically generate this file
+// Its not perfect, but at least its a separated file
```
gautierhattenberger: The airframe file allows you to define initialisation values for structures and arrays.
See some examples:
- https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/ENAC/quadrotor/cobraV2.xml#L94
- https://github.com/paparazzi/paparazzi/blob/master/conf/airframes/ENAC/quadrotor/cobraV2.xml#L171

You can also check my comment in this discussion: https://github.com/paparazzi/paparazzi/discussions/3189#discussioncomment-7685872
With this, you should be able to declare your walls directly in your airframe file.

sw/airborne/modules/lidar/slam/lidar_correction.c
```diff
@@ -0,0 +1,202 @@
...
+  wall_system.converted_to_ltp = false;
+}
+
+#ifdef USE_GRID
+// TODO: Find a easy way to this
+// void fill_known_grid(){
+
+
+// }
+#endif // USE_GRID
```
gautierhattenberger: probably not needed

sw/airborne/modules/lidar/slam/lidar_correction.c
```diff
@@ -0,0 +1,202 @@
...
+ *                                                                             *
+ *  Wall functions                                                             *
+ *                                                                             *
+ ******************************************************************************/
+
+
+// Parse of obstacles (defined in the map file)
+void init_walls(void)
+{
+#include INCLUDE_FILE(OBSTACLES_CONFIG_FILE)  // Include the obstacles configuration file
```
gautierhattenberger: not needed if using structure initialisation from generated code

sw/airborne/modules/lidar/slam/lidar_correction.c
```diff
@@ -0,0 +1,202 @@
...
+// void fill_known_grid(){
+
+
+// }
+#endif // USE_GRID
+
+
+void convert_walls_to_ltp(void)
+{
+  if (wall_system.converted_to_ltp || !ins_int.ltp_initialized) { return; }
```
gautierhattenberger: same comment on `ins_int`

sw/airborne/modules/lidar/tfmini.c
```diff
@@ -35,12 +36,20 @@
...
 
+// Test
+#include "firmwares/rotorcraft/autopilot_static.h"
+
+
+#define LIDAR_MIN_RANGE 0.1
+#define LIDAR_MAX_RANGE 12.0
+
+#define LIDAR_OFFSET 0.20   // Horizontal distance from the IMU to the LiDAR (in meters)
+#define LIDAR_HEIGHT 0.24   // Height of the LiDAR above the ground (in meters)
```
gautierhattenberger: I'm fine with adding offset values, but they should be configurable and set to 0 by default. Then use the airframe file to adjust to your own setup.

sw/airborne/modules/lidar/tfmini.c
```diff
@@ -155,16 +168,32 @@ static void tfmini_parse(uint8_t byte)
...
           }
 
-          // send message (if requested)
+          #ifndef USE_SERVO_LIDAR
+          //send message (if requested, and there is not servo module)
           if (tfmini.update_agl) {
-            AbiSendMsgAGL(AGL_LIDAR_TFMINI_ID, now_ts, tfmini.distance);
+            AbiSendMsgOBSTACLE_DETECTION(AGL_LIDAR_TFMINI_ID, tfmini.distance, 0, 0);
           }
+          #endif
```
gautierhattenberger: I'm not sure what you want to do here, but you can't change this part. The `OBSTACLE_DETECTION` message is supposed to be sent by you new module `servo_lidar`, so why here ?
If you need the raw distance, just disable the compensation and either read the value or bind to `AGL` message (even if it is not really AGL value for you).

sw/simulator/nps/nps_autopilot_rover.c
```diff
@@ -210,3 +228,43 @@ void sim_overwrite_ins(void)
   // Here we make sure that ENU states are recalculated
   stateCalcPositionEnu_i();
 }
+
+
+#ifdef NPS_BYPASS_LIDAR
+// If you want to use the lidar simulation, you need to use the lidar correction module
+void sim_overwrite_lidar(void)
```
gautierhattenberger: I don't think NPS is the right place for this, especially this file.
It seems to me that what you do is either use the real lidar or make a simulation. So in the end, what you need is a simulation mode directly from the lidar driver (or your obstacle module, I'm not sure).
You have to options for that:
- make a dedicated simulation module that you load only in the `nps` target (and the tfmini in the `ap` section)
- use the `target` attribute of the existing driver to compile different files for `ap` and `nps` targets.

gautierhattenberger (changes requested):

sw/airborne/modules/lidar/tfmini.h
alrochas1: I think I push at the same time that your comment. Its already been deleted [7a88a3a96ae652a3bc0780e4890c079ce98ca8da]

sw/airborne/modules/lidar/tfmini.c
alrochas1: The idea is to avoid depending on the servo module (so the map works even without it). It's true that the `AGL` message is necessary, so I added it back (I encapsulated this part in the function `tfmini_send_abi`). Now:
- It always sends the `AGL` message (if `tfmini.update_agl` is true, as before)
- It sends `OBSTACLE_DETECTION` only when the servo module is disabled (to avoid dependency on this module).

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: It's in fact a local development. I've deleted the references to it

sw/airborne/firmwares/rover/obstacles/rover_obstacles.c
alrochas1: It`s makes more sense in our local development. Replaced

sw/airborne/modules/lidar/tfmini.c
```diff
@@ -155,16 +179,28 @@ static void tfmini_parse(uint8_t byte)
...
+            if(fabs(theta) < 0.01){
+              ground_distance = 100;  // If it is 0 it is straight
+            }
+            else{
+              ground_distance = lidar_height/sinf(-theta) - lidar_offset;
+            }
+            
+            if ((tfmini.distance >= ground_distance) && (ground_distance > 0)) {
+              tfmini.distance = 0;
+            }
```
gautierhattenberger: I still don't get what kind of compensation you are doing here. Can you explain it ? Is it still compatible with AGL measurement from a flying drone ?

sw/airborne/modules/lidar/tfmini.c
```diff
@@ -178,3 +214,34 @@ static void tfmini_parse(uint8_t byte)
...
+void tfmini_send_abi(void)
+{
+  uint32_t now_ts = get_sys_time_usec();
+  if (tfmini.update_agl) {
+    AbiSendMsgAGL(AGL_LIDAR_TFMINI_ID, now_ts, tfmini.distance);
+  }
+  #ifndef USE_SERVO_LIDAR
+  //send message (if there is not servo module)
+  AbiSendMsgOBSTACLE_DETECTION(AGL_LIDAR_TFMINI_ID, tfmini.distance, 0, 0);
+  #endif
```
gautierhattenberger: Are you really using this sensor without the "servo_lidar" module ?

conf/modules/servo_lidar.xml
gautierhattenberger: I saw that this parameter is now configurable with a `define`, but this `configure` still has no effect. I suggest to just remove this line.

gautierhattenberger (changes requested):
As I mentioned before, the simulation code for this lidar module should not be in NPS. It is currently too specific to the rover and tfmini sensor. You should make a dedicated module for that purpose.

sw/airborne/modules/lidar/tfmini.c
alrochas1: When the rover pitches (due to uneven terrain), the LiDAR might point toward the ground. This compensation prevents that. It’s true that I should preserve the original compensation, so I’ve added an option to choose between the rover and the drone.

sw/airborne/modules/lidar/tfmini.c
alrochas1: We might need it—mounting the LiDAR alone on our rovers is easier than with the servo, and the map will still work without it.

sw/simulator/nps/nps_autopilot_rover.c
alrochas1: I've a clean separation now:
- Created dedicated `tfmini_nps.c` for simulation
  - The NPS module doesn’t have compensation (because there is no model to simulate).
  - It doesn’t send the AGL message for the same reason (with a drone, it doesn’t make sense to simulate the lidar, as we don’t have a model for it).
- Kept original `tfmini.c` for real hardware
- Used target attributes in the XML to select the appropriate implementation

sw/airborne/modules/lidar/tfmini.c
```diff
@@ -35,12 +36,27 @@
 #include "pprzlink/messages.h"
 #include "modules/datalink/downlink.h"
 
+// Test
+#include "firmwares/rotorcraft/autopilot_static.h"
+
```
gautierhattenberger: Why do you need this ? Is it some test that you forgot to remove ?

sw/simulator/nps/nps_autopilot.h
gautierhattenberger: I think this file could be restored to its original version now since you don't change NPS any more

gautierhattenberger (changes requested):
Thank you for the updates. I only have 2 minor comments left.

gautierhattenberger (approved):

PR #3513, opened 2025-07-29
[uavcan] buffer tx messages in FIFO

Fabien-B:
Add a FIFO for Tx uavcan messages.

The issue was a deadlock when calling `uavcan_broadcast` from the reception callback of an other uavcan message (`iface->mutex` was locked twice).

My solution is to add a FIFO for the messages to be broadcasted.
`uavcan_broadcast` add the messages to this FIFO, and the `uavcan_tx` thread read the messages from the FIFO.
The FIFO have its own mutex, independent from the canard mutex.

I implemented the dynamic node ID allocation (future PR), it broadcast messages from a reception callback.

### DIFF
--- a/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
+++ b/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
@@ -90,7 +90,7 @@ static void can_frame_cb(struct pprzcan_frame* rx_msg, UNUSED struct pprzaddr_ca
 
   pprz_mtx_lock(&iface->mutex);
 
-  CanardCANFrame rx_frame;
+  CanardCANFrame rx_frame = {0};
   memcpy(rx_frame.data, rx_msg->data, 8);
   rx_frame.data_len = rx_msg->len;
   if (rx_msg->can_id & CAN_FRAME_EFF) {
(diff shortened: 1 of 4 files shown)


sw/airborne/arch/chibios/modules/uavcan/uavcan.c
```diff
@@ -92,11 +100,17 @@ static void can_frame_cb(struct pprzcan_frame* rx_msg, UNUSED struct pprzaddr_ca
   }
 
   // Let canard handle the frame
-  canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
+  int ret = canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
+  if(ret != CANARD_OK) {
+    asm("NOP");
```
fvantienen: Is this needed?

Fabien-B: It shouldn't be, and this is not the solution, as this doesn't solve the problem when building in release mode.
For some reason, the message callbacks are not called for each received message, but only for 1/3 or 1/2 of the messages.
When I add that and I build in debug, the callback are called for every message.
I don't know what is causing this problem.

Fabien-B:
I found the issue !
It was exactly the same thing than for the broadcast : the `CanardCANFrame` must be initialized to 0.
Now it's working, without some weird checks.

gautierhattenberger (approved):
It looks fine

OpenUAS:
This will help testflights a lot, thx @Fabien-B @gautierhattenberger and sure @fvantienen

PR #3515, opened 2025-07-31
Uavcan node id allocator

Fabien-B:
Dynamic node id allocation, from this specification : https://dronecan.github.io/Specification/6._Application_level_functions/#dynamic-node-id-allocation

Multiple allocators are not supported.

Depends on  #3513

### DIFF
--- a/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
+++ b/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
@@ -90,7 +90,7 @@ static void can_frame_cb(struct pprzcan_frame* rx_msg, UNUSED struct pprzaddr_ca
 
   pprz_mtx_lock(&iface->mutex);
 
-  CanardCANFrame rx_frame;
+  CanardCANFrame rx_frame = {0};
   memcpy(rx_frame.data, rx_msg->data, 8);
   rx_frame.data_len = rx_msg->len;
   if (rx_msg->can_id & CAN_FRAME_EFF) {
--- a/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
+++ b/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
@@ -100,10 +100,7 @@ static void can_frame_cb(struct pprzcan_frame* rx_msg, UNUSED struct pprzaddr_ca
   }
 
   // Let canard handle the frame
-  int ret = canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
-  if(ret != CANARD_OK) {
-    asm("NOP");
-  }
+  canardHandleRxFrame(&iface->canard, &rx_frame, rx_msg->timestamp);
 
   pprz_mtx_unlock(&iface->mutex);
 }
--- a/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
+++ b/sw/airborne/arch/chibios/modules/uavcan/uavcan.c
@@ -27,6 +27,7 @@
 #include "uavcan.h"
 #include "mcu_periph/can.h"
 #include "modules/core/threads.h"
+#include "modules/uavcan/uavcan_allocator.h"
 
 #ifndef UAVCAN_NODE_ID
 #define UAVCAN_NODE_ID    100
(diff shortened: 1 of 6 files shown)


sw/airborne/modules/uavcan/uavcan_allocator.c
```diff
@@ -0,0 +1,269 @@
...
+  uavcan_bind(UAVCAN_PROTOCOL_DYNAMIC_NODE_ID_ALLOCATION_ID, UAVCAN_PROTOCOL_DYNAMIC_NODE_ID_ALLOCATION_SIGNATURE,
+    &id_alloc_ev, &id_alloc_uavcan_cb);
+  
+  uavcan_bind(UAVCAN_PROTOCOL_GETNODEINFO_RESPONSE_ID, UAVCAN_PROTOCOL_GETNODEINFO_RESPONSE_SIGNATURE,
+    &node_info_ev, &node_info_resp_cb);
+}
+
+
+
+void uavcan_allocator_periodic(void) {
```
fvantienen: Why this empty periodic?

sw/airborne/modules/uavcan/uavcan_allocator.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
fvantienen: Maybe add a correct header?

sw/airborne/arch/chibios/modules/uavcan/uavcan.c
```diff
@@ -109,6 +121,26 @@ static void uavcan_tx(void* p)
...
+    // read the Tx FIFO to canard
+    pprz_mtx_lock(&iface->tx_fifo_mutex);
+    while(true) {
+      struct uavcan_msg_header_t header;
+      int ret = circular_buffer_get(&iface->_tx_fifo, (uint8_t*)&header, sizeof(header));
+      if(ret < 0) {break;}
+      if(header.payload_len >= UAVCAN_MSG_MAX_SIZE) {
+        chSysHalt("UAVCAN_MSG_MAX_SIZE too small");
+      }
+      ret = circular_buffer_get(&iface->_tx_fifo, msg_payload, UAVCAN_MSG_MAX_SIZE);
```
fvantienen, line 132: Do you always need to get the MAX_SIZE?

sw/airborne/arch/chibios/modules/uavcan/uavcan.h
```diff
@@ -30,6 +30,17 @@
...
 #include "modules/core/threads.h"
+#include "utils/circular_buffer.h"
+
+
+#ifndef UAVCAN_TX_FIFO_SIZE
+#define UAVCAN_TX_FIFO_SIZE 1024
+#endif
+
+#ifndef UAVCAN_MSG_MAX_SIZE
+#define UAVCAN_MSG_MAX_SIZE 256
```
fvantienen: Isn't there a define in UAVCAN/DRONECAN itself that contains it from the specification?

sw/airborne/modules/uavcan/uavcan_allocator.c
```diff
@@ -0,0 +1,269 @@
...
+
+
+/**
+ *  Dynamic node ID allocation.
+ *  See https://dronecan.github.io/Specification/6._Application_level_functions/#dynamic-node-id-allocation
+ */
+
+
+#ifndef UAVCAN_DYN_NODE_NB
+#define UAVCAN_DYN_NODE_NB 10
```
fvantienen: Maybe name it UAVCAN_MAX_NODES, since it also contains the nodes of the fixed ids? And by default make it a bit higher, since we are usually way above this number. So for us 50 would make more sense.

fvantienen (changes requested):
Really nice that you implemented this! Did you also test this in combination with the slcan?

sw/airborne/arch/chibios/modules/uavcan/uavcan.h
Fabien-B: I don't remember any max size being specified, but maybe I'm wrong ?

sw/airborne/modules/uavcan/uavcan_allocator.c
Fabien-B: some remains of old code... Cleaned.

sw/airborne/arch/chibios/modules/uavcan/uavcan.c
Fabien-B: I changed that a bit, so if the buffer is too small to contains the message, the message is simply discarded. It won't work as expected, but at least it won't crash the autopilot, and it won't affect the other messages.

Fabien-B:
> Really nice that you implemented this! Did you also test this in combination with the slcan?

I just did, it works !

image

Fabien-B:
Close in favour of #3579

PR #3517, opened 2025-08-01
[test] improve module tests

Fabien-B:


OpenUAS:
Will benefit @fvantienen and the new hardware to be tested a lot, thank Fabien, awesome!

Issue #3437, opened 2025-03-11
Support for new Invensense ICM45686 - e.g. in Pixhawk 6X


dewagter:

dewagter:
See: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_InertialSensor/AP_InertialSensor_Invensensev3.cpp

OpenUAS:
This will satisfy the request :
 https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/peripherals/invensense3_456.c 
and all related files of e.g. imu_cube.c

PR #3519, opened 2025-08-07
[imu/board] Add support for ICM45686 and LTS release of the Cube Orange+

lthoma5:
The CubePilot Cube Orange+ has two versions: the original version (matte surface) and the LTS version (metallic/shiny finish). The main difference between these two versions is a change in the IMU boad, with the LTS release moving to a triple ICM45686 ([CubePilot ECN-03](https://discuss.cubepilot.org/t/ecn-03-balanced-gyro-for-cube-orange-plus/11625)).

This PR introduces (a first version of) support for the **ICM45686** IMU, also addressing #3437. This initial version of the driver is largely based on the original invensense3 driver by @fvantienen and the ArduPilot/PX4 drivers for modifications. Some things to note:
- At present, only the high-resolution (20-bit) mode is supported with 32G/4000dps setting by the driver.
- In light of the large differences between the two drivers (e.g. a different `WHO_AM_I` register address), I decided to split it off into its own driver. I am open to merging it with `invensense3`, but not a huge fan of this.
- It supports SPI & identical rates for both the accelerometer and gyro only.
- The original driver had to configure the anti-aliasing filter itself. The ICM45686 does not expose configuration parameters, but only an enable option. This is used by the driver, in line with the ArduPilot implementation.

It also adds a new `cube_orange_plus_lts` board for the _shiny_ Cube Orange+. Some key differences between the boards have been highlighted below:
- All three IMUs are now ICM45686s.
- Different orientations for the IMUs - see the `imu_cube.c` source code and cross-reference with slave configuration in the LTS Makefile.
- New chip-select pin for the non-isolated IMU: `PG01`, set up as `SPI_SLAVE9`.

So far, these changes have been tested on a PR and seem to be consistent with each other/reasonable values. Further testing (e.g. validating the data rate) might not be a bad idea.


gautierhattenberger (approved):
It looks fine, but since I don't have the hardware to test the driver, I leave it to you to make more tests if needed.

One small remark: you should consider splitting the module `imu_cube`  into several ones for each type of Cube board since it becoming a huge pile of `#if / #else`. Each one can even be a dependency of its respective board module, so you don't risk to load the wrong one.

OpenUAS:
@lthoma5 look at @gautierhattenberger remark and please improve  splitting the module imu_cube into several ones for each type of Cube somewhere this summer, that would be great.

PR #3520, opened 2025-08-13
Add support for common sonar sensors on I2C

OpenUAS:
Sonar on I2C port support. Tested with all famous GY-US42v2, should work on Maxbotix I2C based sensors like 1240 by changing I2C address in Airframe file.

----
Anyhow the behavior of connected sonar on I2C whatever brand connected to flightcontroller before PR
ScreenshotBefore

After PR yields a 100% improvement ;0)
ScreenshotAfter

### DIFF
--- a/conf/modules/sonar_i2c.xml
+++ b/conf/modules/sonar_i2c.xml
@@ -1,0 +1,95 @@
+<!DOCTYPE module SYSTEM "module.dtd">
+
+<module name="sonar_i2c" dir="sonar">
+  <doc>
+    <description>
+    A Sonar with I2C connection can be used as unidirectional rangefinder. Examples of devices are e.g. Maxbotix MB1240 or a GY-US42V2.
+    The sensor can be used as a distance-measuring tool to detect the range (distance) to a surfaces. Usecase for example: AGL hold,terrain following or landing assist.
+
+    To be able to use the I2C version of a Sonar, enable this module in your airframe and set needed parameters.
+
+    Notes:
+    + The maximum useful detection distance is 4m for common sonar based small sized sensors, even if datasheet e.g. states 7m.
+    For most I2C sonar range devices, a sampling rate of at least 15Hz is supported. However, if measurements are taken at longer distances, 
+    the time required for the ultrasonic pulse to return increases, which may limit the maximum achievable sampling rate. 
+    It is recommended to adjust the SONAR_I2C_PERIODIC_FREQUENCY parameter according to the sensor's capabilities and the expected measurement range to ensure reliable readings.
+
+    + If the distance is read but the scale is incorrect, adjust the SONAR_I2C_SCALE parameter to match the sensor's specifications.
+    The scale factor can also be determined experimentally by measuring a known distance and adjusting the scale until the output matches the expected value.
+
+    + If the sensor is used in an airframe with landing gear, it is recommended to set the SONAR_I2C_OFFSET to the height of the landing gear, so that the AGL value is correct. 
+    Keep in mind that most sonar sensors have a deadzone of 0.15m to 0.25m, so the AGL cannot be measured if landinggear shorter than the deadzone range.
+
+    + There is an option to compensate the AGL value for body rotation, which is useful if e.g. a roll maneuver, the sensor is not pointing straight down it compensates the AGL value for the body rotation.
+    
+    + Also an option to filter the sensor output is available, which is useful if the sensor is noisy or has a lot of outliers, in cas of a sonar, almost always the case.
+    
+    + An option to set a different I2C address is available in some hardware but not implemented to set dynamically in this driver. 
+    Use external tools to change the I2C address if needed and set the SONAR_I2C_ADDR parameter accordingly.
+
+    + Distance compensation for pressure and temperature are not implemented, keep in mind that sonar sensors are susceptible to temperature and pressure changes.
+ 
+  </description>
+    <configure name="SONAR_I2C_DEV" value="i2cX" description="I2C device port on flightcontroller to use for this device, e.g i2c2"/>
+    <define name="SONAR_I2C_ADDR" value="0xE0" description="The I2C Slave address in 8bit of this device"/>
+    <configure name="USE_SONAR_I2C_AGL" value="TRUE|FALSE" description="Updates the AGL value in state,if not defined, defaults to FALSE"/>
+    <define name="SONAR_I2C_COMPENSATE_ROTATION" value="TRUE|FALSE" description="Compensate AGL measurements for body rotation. Disabled by default"/>
+    <define name="SONAR_I2C_USE_FILTER" value="TRUE|FALSE" description="If this is enabled, a median filter on the distance output, defaults to TRUE"/>
+    <define name="SONAR_I2C_MEDIAN_SIZE" value="11" description="If median filter is enabled then set this option to filter the output more(or less), default is 7"/>
+    <define name="AGL_SONAR_I2C_ID" value="254" description="Set your own different ID if using both PWM, Analog and I2C version of a sensor at the same time"/>
+    <define name="SONAR_I2C_SCALE" value="0.00006" description="Sensor scale for raw sensor output to a meter unit conversion, "/>
+    <define name="SONAR_I2C_OFFSET" value="0.56"  description="Sensor offset in meters, as in where one wants zero to be, default is 0.0"/>
+    <define name="SONAR_I2C_MIN_RANGE" value="0.44" unit="m" description="If defined, set limit to minimum value reported, default 0.3m"/>
+    <define name="SONAR_I2C_MAX_RANGE" value="3.3" unit="m" description="If defined, set limit to maximum value reported, default 4.0m"/>
+    <define name="MODULE_SONAR_I2C_SYNC_SEND" value="TRUE|FALSE" description="Send RAW and scaled message with each new measurement,useful for debugging sensor issues (default: FALSE)"/>
+  </doc>
+
+  <settings>
+    <dl_settings NAME="Rangefinder">
+      <dl_settings NAME="Sonar I2C">
+        <dl_setting MAX="1" MIN="0" STEP="1" VAR="sonar_i2c.update_agl" shortname="Update AGL"/>
+      </dl_settings>
+    </dl_settings>
+  </settings>
+  <dep>
+    <depends>i2c</depends>
+    <provides>sonar</provides>
+  </dep>
+
+  <header>
+    <file name="sonar_i2c.h"/>
+  </header>
+  <init fun="sonar_i2c_init()"/>
+  <periodic fun="sonar_i2c_periodic()" freq="SONAR_I2C_PERIODIC_FREQUENCY"/>
+  <periodic fun="sonar_i2c_report()" freq="50" autorun="FALSE"/>
+  <event fun="sonar_i2c_event()"/>
+  <makefile>
+    <configure name="SONAR_I2C_PERIODIC_FREQUENCY" default="30"/>
+    <configure name="SONAR_I2C_DEV" default="i2c1" case="lower|upper"/>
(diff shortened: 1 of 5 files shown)


sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+#include "pprzlink/messages.h"
+#endif
+
+#ifdef MODULE_SONAR_I2C_SYNC_SEND
+#include "modules/datalink/downlink.h"
+#endif
+
+#ifndef SONAR_I2C_AGL_ID
+#define SONAR_I2C_AGL_ID AGL_SONAR_I2C_ID
+#endif
```
gautierhattenberger, line 45: When you are the data publisher, you should not allow to redefine the sender ID, it would be like faking your identity.
Also this `AGL_SONAR_I2C_ID` seems to be new, so you have to add it to `modules/core/abi_sender_ids.h`

sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+#define SONAR_I2C_USE_FILTER 0 // Enable filtering for sonar per default is best, sonars tend to give noisy spiking readings
+#endif
+
+#ifdef SONAR_I2C_USE_FILTER
+#include "filters/median_filter.h"
+///The amount of sensor samples to keep in the median filter buffer
+#ifndef SONAR_I2C_MEDIAN_SIZE
+#define SONAR_I2C_MEDIAN_SIZE 7 // Default median filter length of 7 is a good fit for most sonar sensors
+#endif
+//struct MedianFilterInt sonar_i2c_filter;
```
gautierhattenberger, line 93: Better remove if not used

sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+
+  sonar_i2c.status = SONAR_I2C_REQ_DATA;
+
+#ifdef SONAR_I2C_USE_FILTER
+  init_median_filter_f(&sonar_i2c_filter, SONAR_I2C_MEDIAN_SIZE);
+#endif
+
+//#if PERIODIC_TELEMETRY
+  register_periodic_telemetry(DefaultPeriodic, PPRZ_MSG_ID_SONAR, sonar_i2c_send_sonar);
+//#endif
```
gautierhattenberger, line 137: Why commenting the `PERIODIC_TELEMETRY` ? Can you add it back ?

sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+      break;
+    case I2CTransFailed:
+      // set to done
+      sonar_i2c.trans.status = I2CTransDone;
+      break;
+    case I2CTransDone:
+      // do nothing
+      break;
+    default:
+      break;
```
gautierhattenberger, line 168:
```suggestion
  switch (sonar_i2c.trans.status) {
    case I2CTransSuccess:
    case I2CTransFailed:
      // set to done
      sonar_i2c.trans.status = I2CTransDone;
      break;
    default:
      // do nothing
      break;
```

sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+    case SONAR_I2C_READ_DATA:
+      if (sonar_i2c.trans.status == I2CTransDone) {
+        sonar_i2c.trans.buf[1] = 0;
+        sonar_i2c.trans.buf[2] = 0;
+        if (i2c_receive(&SONAR_I2C_DEV, &sonar_i2c.trans, sonar_i2c.addr, 2)) {
+          sonar_i2c.status = SONAR_I2C_PARSE_DATA;
+        }
+        sonar_i2c.raw = (uint16_t)sonar_i2c.trans.buf[2]; // Debugging value, to see if we got a valid reading
+        DOWNLINK_SEND_SONAR(DefaultChannel, DefaultDevice, &sonar_i2c.raw, &sonar_i2c.distance);
```
gautierhattenberger, line 201: This should be removed to avoid overloading telemetry. The message is already registered for periodic telemetry so its frequency can be controlled from the telemetry XML file.

sw/airborne/modules/sonar/sonar_i2c.c
```diff
@@ -0,0 +1,254 @@
...
+      sonar_i2c.distance = sonar_i2c.distance - SONAR_I2C_OFFSET;
+      //sonar_i2c.raw = 33; // Debugging value, to see if we got a valid reading
+      //DOWNLINK_SEND_SONAR(DefaultChannel, DefaultDevice, &sonar_i2c.raw, &sonar_i2c.distance);
+      Bound(sonar_i2c.distance, (float)SONAR_I2C_MIN_RANGE, (float)SONAR_I2C_MAX_RANGE);
+
+      // Compensate range measurement for body rotation
+#if SONAR_I2C_COMPENSATE_ROTATION
+      float phi = stateGetNedToBodyEulers_f()->phi;
+      float theta = stateGetNedToBodyEulers_f()->theta;
+      float gain = (float)fabs((double)(cosf(phi) * cosf(theta)));
```
gautierhattenberger, line 223: I disagree about the `fabs` here, it just gives a wrong value that looks like a good one if you go upside down.
It should be just cos(phi)*cos(theta). Then, if you want some sort of protection, you can set the distance to nan if bank angle is out of range.

conf/modules/sonar_i2c.xml
```diff
@@ -0,0 +1,95 @@
...
+    <define name="SONAR_I2C_MEDIAN_SIZE" value="$(SONAR_I2C_MEDIAN_SIZE)"/>
+    <define name="SONAR_I2C_SCALE" value="$(SONAR_I2C_SCALE)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="SONAR_I2C_COMPENSATE_ROTATION" value="0"/>
+    <define name="AGL_SONAR_I2C_ID" value="254"/>
+    <raw>
+      ifeq ($(SONAR_I2C_DEV),)
+        $(error sonar_i2c module error: please configure SONAR_I2C_DEV in your airframe)
+      endif
+    </raw>
```
gautierhattenberger, line 89: You have set a default value to `SONAR_I2C_DEV` so these lines have no effect

conf/modules/sonar_i2c.xml
```diff
@@ -0,0 +1,95 @@
...
+    <define name="SONAR_I2C_ADDR" value="$(SONAR_I2C_ADDR)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="USE_$(SONAR_I2C_DEV_UPPER)"/>
+    <define name="SONAR_I2C_DEV" value="$(SONAR_I2C_DEV_LOWER)"/>
+    <define name="SONAR_I2C_USE_FILTER" value="$(SONAR_I2C_USE_FILTER)"/>
+    <define name="SONAR_I2C_MEDIAN_SIZE" value="$(SONAR_I2C_MEDIAN_SIZE)"/>
+    <define name="SONAR_I2C_SCALE" value="$(SONAR_I2C_SCALE)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="SONAR_I2C_COMPENSATE_ROTATION" value="0"/>
+    <define name="AGL_SONAR_I2C_ID" value="254"/>
```
gautierhattenberger, line 84: It should not be defined here but in `modules/core/abi_sender_ids.h`

conf/modules/sonar_i2c.xml
```diff
@@ -0,0 +1,95 @@
...
+    <define name="SONAR_I2C_PERIODIC_FREQUENCY" value="$(SONAR_I2C_PERIODIC_FREQUENCY)"/>
+    <define name="SONAR_I2C_ADDR" value="$(SONAR_I2C_ADDR)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="USE_$(SONAR_I2C_DEV_UPPER)"/>
+    <define name="SONAR_I2C_DEV" value="$(SONAR_I2C_DEV_LOWER)"/>
+    <define name="SONAR_I2C_USE_FILTER" value="$(SONAR_I2C_USE_FILTER)"/>
+    <define name="SONAR_I2C_MEDIAN_SIZE" value="$(SONAR_I2C_MEDIAN_SIZE)"/>
+    <define name="SONAR_I2C_SCALE" value="$(SONAR_I2C_SCALE)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="SONAR_I2C_COMPENSATE_ROTATION" value="0"/>
```
gautierhattenberger, line 83: remove this line otherwise it will always be false.

conf/modules/sonar_i2c.xml
```diff
@@ -0,0 +1,95 @@
...
+    <configure name="SONAR_I2C_SCALE" default="0.000044"/>
+    <define name="SONAR_I2C_PERIODIC_FREQUENCY" value="$(SONAR_I2C_PERIODIC_FREQUENCY)"/>
+    <define name="SONAR_I2C_ADDR" value="$(SONAR_I2C_ADDR)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
+    <define name="USE_$(SONAR_I2C_DEV_UPPER)"/>
+    <define name="SONAR_I2C_DEV" value="$(SONAR_I2C_DEV_LOWER)"/>
+    <define name="SONAR_I2C_USE_FILTER" value="$(SONAR_I2C_USE_FILTER)"/>
+    <define name="SONAR_I2C_MEDIAN_SIZE" value="$(SONAR_I2C_MEDIAN_SIZE)"/>
+    <define name="SONAR_I2C_SCALE" value="$(SONAR_I2C_SCALE)"/>
+    <define name="USE_SONAR_I2C_AGL" value="$(USE_SONAR_I2C_AGL)"/>
```
gautierhattenberger, line 82: These configure/define are not matching the `doc` of the module. I suggest to stick to the documentation and remove the `configure` and `define` where the parameter is a `define` in the doc.

sw/airborne/modules/sonar/sonar_i2c.c
OpenUAS: Yes that was the intention. only wanted to start an option if on would have multiple sensors. Will change for now to abi_sender_ids.h and look at multi same sensor (on different I2C address ofcourse)

sw/airborne/modules/sonar/sonar_i2c.c
OpenUAS: Oeps, wil do

sw/airborne/modules/sonar/sonar_i2c.c
OpenUAS: Another hasty error, sorry.

sw/airborne/modules/sonar/sonar_i2c.c
OpenUAS: Note that all current implementations of rangefinders use this method, That does not mean that is great, For now could improve just this I2C driver. A common correct helper function for all rangefinders to use would be best.

conf/modules/sonar_i2c.xml
OpenUAS: Yes, it was left if someone would remove or change module. Will remove it.

conf/modules/sonar_i2c.xml
OpenUAS: indeed, see before remark

conf/modules/sonar_i2c.xml
OpenUAS: OK

PR #3525, opened 2025-08-17
fix errors during pprz installation

EwoudSmeur:
I reinstalled my laptop with Ubuntu 24.04.3 and installed Paparazzi. I ran into some issues:
- [x] pprzlink submodule not checked out before used (fixed)
- [x] lsb_release deprecated, move to distro
- [x] Running ./paparazzi from Inside pprzEnv the window scaling is messed up, fixed by forcing Wayland use
- [ ] xcb QT plugin not found -> fixed with `sudo apt install libxcb-xinerama0` but I'm not sure where to put this
- [x] The installation process creates folder `sw/ext/pprzlink/lib/v2.0/python/src/pprzlink.egg-info/` untracked in git

### DIFF
--- a/sw/tools/dfu/stm32_mem.py
+++ b/sw/tools/dfu/stm32_mem.py
@@ -169,8 +169,16 @@ if __name__ == "__main__":
             continue
         try:
             man = dfudev.handle.getString(dfudev.dev.iManufacturer, 30)
+            if isinstance(man, bytes):
+                man = man.decode("utf-8")
+
             product = dfudev.handle.getString(dfudev.dev.iProduct, 30)
+            if isinstance(product, bytes):
+                product = product.decode("utf-8")
+
             serial = dfudev.handle.getString(dfudev.dev.iSerialNumber, 40)
+            if isinstance(serial, bytes):
+                serial = serial.decode("utf-8")
         except Exception as e:
             print("Whoops... could not get device description.")
             print("Exception:", e)


start.py
gautierhattenberger, line 1: Which can of action can you do with `start.py` that are not supported by the new paparazzi center in python ? Would it be possible to remove the script completely ?

EwoudSmeur: Good point, I think it can be removed. Do you want me to do that here in this PR?

gautierhattenberger: yes, why not

EwoudSmeur: Done!

gautierhattenberger:
I don't have the issue with `libxcb-xinerama0` as it is somehow already installed for me, maybe because I'm using Kubuntu ? Is it really necessary to add it as a dependency of `paparazzi-dev` ? Or just as an extra instruction for people facing the issue ?

The `egg-info` folder should be added to the `.gitignore` file of pprzlink, don't think it can be fixed from main paparazzi.

EwoudSmeur:
Is there anything against adding it? It is a dependency in the end. (honest question)

I made a pull request for pprzlink https://github.com/paparazzi/pprzlink/pull/209

EwoudSmeur:
I also tried to flash an old Lisa/m with DFU (don't ask me why) and fixed an issue with a string comparison

gautierhattenberger:
> Is there anything against adding it? It is a dependency in the end. (honest question)
> 

The honest answer is that it is a pain for me each time I have to update the package `paparazzi-dev` ;p
This is why I'm asking if it is really a blocking point (it is not working without that) or if it is just an annoying warning. Case 1, I will of course update the package. Case 2, I'll add this to the todo list for the next update.

EwoudSmeur:
Ok, I understand! For me it was blocking. But if updating the package is complex it could be a step in the installation script.

gautierhattenberger:
If it is blocking, I will update the dependencies of `paparazzi-dev` soon.

PR #3530, opened 2025-08-29
[rover] improve support for ROVER firmware

gautierhattenberger:
- possibility to use basic PID with steering rover
- fix navigation
- add support for mission mode
- fix simulator (orientation was wrong, leading to opposite steering between sim and reality)
- add support of 2 wheels rover in simulation
- add and update config files

### DIFF
--- a/conf/airframes/ENAC/rover/rover_2_wheels.xml
+++ b/conf/airframes/ENAC/rover/rover_2_wheels.xml
@@ -1,4 +1,4 @@
-<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+<!DOCTYPE airframe SYSTEM "../../airframe.dtd">
 
 <airframe name="Rover">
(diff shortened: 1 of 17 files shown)


gautierhattenberger:
@noether I have fixed an incorrect direction in the simulator and removed the minus sign in the macro `GetCmdFromDelta`. You should check that your rovers are still turning in the correct direction. If not, the servo min/max values are probably swapped.

noether:
> @noether I have fixed an incorrect direction in the simulator and removed the minus sign in the macro `GetCmdFromDelta`. You should check that your rovers are still turning in the correct direction. If not, the servo min/max values are probably swapped.

Cool, sure, we will check it and let you know :P, thanks!

gautierhattenberger:
@noether Did you have a chance to test this ?

PR #3534, opened 2025-09-01
Added and option to get flyble values out of this old ETS sensor

OpenUAS:
Using the ETS (EagleTree Systems) airspeed gave kind of noisy values when compared to modern airspeed sensors. Added and option to get flyable values out of the old sensor by being able to switch on some decent filtering. ETS no longer in business, old hardware but working and quite some sensors used was the reason to improve.


OpenUAS:
Bump for a review...

PR #3540, opened 2025-09-17
Add new support for common sonar sensors on I2C

OpenUAS:
Sonar on I2C port support. Tested with all famous GY-US42v2, should work on Maxbotix I2C based sensors like 1240 by changing I2C address in Airframe file. A new PR with added changes as requested in old PR #3520. Redone since old branch removed. 

Below graph of the the behavior of connected sonar on I2C whatever brand connected to flight-controller before PR
before
and after the PR
after

### DIFF
--- a/conf/modules/sonar_i2c.xml
+++ b/conf/modules/sonar_i2c.xml
@@ -67,7 +67,7 @@
   </header>
   <init fun="sonar_i2c_init()"/>
   <periodic fun="sonar_i2c_periodic()" freq="SONAR_I2C_PERIODIC_FREQUENCY"/>
-  <periodic fun="sonar_i2c_report()" freq="50" autorun="FALSE"/>
+  <periodic fun="sonar_i2c_report()" freq="10" autorun="FALSE"/>
   <event fun="sonar_i2c_event()"/>
   <makefile>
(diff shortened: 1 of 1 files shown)


conf/modules/sonar_i2c.xml
```diff
@@ -0,0 +1,89 @@
...
+    <depends>i2c</depends>
+    <provides>sonar</provides>
+  </dep>
+
+  <header>
+    <file name="sonar_i2c.h"/>
+  </header>
+  <init fun="sonar_i2c_init()"/>
+  <periodic fun="sonar_i2c_periodic()" freq="SONAR_I2C_PERIODIC_FREQUENCY"/>
+  <periodic fun="sonar_i2c_report()" freq="50" autorun="FALSE"/>
```
gautierhattenberger: 50 hertz is even faster than the default sensor update rate. I would suggest 5 or 10 here for low freq report, if enable sync send if needed.

OpenUAS: Set at 50 since for debugging the highest rate would be most useful else a normal sonar message could have been used. The highest sampling rate for sonars available is 50Hz. Anyhow will set to 10.

PR #3541, opened 2025-09-18
Sync guidance indi hybrid and guidance indi hybrid quadplane filters

NoahWe:
- The filter cutoff frequency for the body z acceleration in guidance indi quadplane was hardcoded, this changes it to track the guidance indi hybrid filters by default.
- Use the filtered attitude angles as inputs for the guidance effectiveness matrix.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
+++ b/sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
@@ -44,18 +44,27 @@
 float guidance_indi_thrust_z_eff = GUIDANCE_INDI_THRUST_Z_EFF;
 #endif
 
-float bodyz_filter_cutoff = 0.2;
+#ifndef GUIDANCE_INDI_BODYZ_FILTER_CUTOFF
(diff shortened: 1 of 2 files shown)


sw/airborne/firmwares/rotorcraft/guidance/guidance_indi_hybrid_quadplane.c
```diff
@@ -44,18 +44,27 @@
...
+#ifndef GUIDANCE_INDI_BODYZ_FILTER_CUTOFF
+#ifdef GUIDANCE_INDI_FILTER_CUTOFF
+#define GUIDANCE_INDI_BODYZ_FILTER_CUTOFF GUIDANCE_INDI_FILTER_CUTOFF
+#else
+#define GUIDANCE_INDI_BODYZ_FILTER_CUTOFF 3.0
+#endif
+#endif
 
 Butterworth2LowPass accel_bodyz_filt;
+Butterworth2LowPass yaw_filt;
```
gautierhattenberger: What about adding that to the main `guidance_indi_hybrid` code (with roll and pitch) ? And maybe use the filtered value of attitude for both quadplane and tailsitter efficiency matrices.

PR #3542, opened 2025-09-25
Added noisy ADC filter option and other improvments

OpenUAS:
Old but still in use ADC sonar rangefinder sensors often have very noisy output with one in a while huge spikes. To have more reliable use, a filter option is added. In the same run the rotation compensation is now the same as the Sonar I2C which have already the requested improvements.

Output example of range-find measurements with a Maxbotix EZ1 sonar sensor connected to a FC analog input port. Also displays the Offset option to Ground.

screenshot_new_output_sonar_adc_01

### DIFF
--- a/conf/modules/sonar_adc.xml
+++ b/conf/modules/sonar_adc.xml
@@ -3,15 +3,57 @@
 <module name="sonar_adc" dir="sonar" task="sensors">
   <doc>
     <description>
-     Sonar ADC driver.
-     Reads an anlog sonar sensor and outputs sonar distance in [m]
-    </description>
-    <configure name="ADC_SONAR" value="ADC_X" description="select ADC to use with the sonar"/>
-    <define name="SONAR_OFFSET" value="0"  description="sensor offset in [adc] - default is 0"/>
-    <define name="SONAR_SCALE" value="0.0166" description="sensor scale factor [m/adc] - default is 0.0166"/>
-    <define name="USE_SONAR" value="" description="activate use of sonar in INS extended filter (only rotorcraft)"/>
-  </doc>
+    A Sonar with an ADC output signal can be used as unidirectional rangefinder. Examples of devices are e.g. Maxbotix MB1210 or MB1010
+    The sensor can be used as a control or measuring tool to detect the range (distance) to a surfaces, used for example in AGL hold, terrain following or landing assist.
 
+    To be able to use the Sonar connected to an ADC input, just add this module in your airframe and set needed parameters.
+
+    Notes:
+    + The maximum useful detection distance is 4m for common simple sonar based small sized sensors, even if datasheet e.g. states 7m.
+    For most ADC sonar range devices, a sampling rate of at least 15Hz is supported. However, if measurements are taken at longer distances, 
+    the time required for the ultrasonic pulse to return increases, which may limit the maximum achievable sampling rate. 
+    It is recommended to adjust the SONAR_ADC_PERIODIC_FREQUENCY parameter according to the sensor's capabilities and the expected measurement range to ensure reliable readings.
+
+    + If the distance is read but the scale is incorrect, adjust the SONAR_ADC_SCALE parameter to match the sensor's specifications.
+    The scale factor can also be determined experimentally by measuring a known distance and adjusting the scale until the output matches the expected value. 
+    This Scale value adjustment can be performed real-time via the setting panel which make it a little more user friendly.
+    After the correct value is determined, save setting or manually add the value to the airframe file.
+
+    SONAR_ADC_SCALE for common sensors:
+    Maxbotix XL-MaxSonar part MB1210 where sensor VCC is 5v: 0.001855
+    Maxbotix LV-MaxSonar part MB1010 where sensor VCC is 5v: 0.002315
+
+    + If the sensor is used in an airframe with landing gear, it is recommended to set the SONAR_ADC_OFFSET to the height of the landing gear, so that the AGL zero value is correct to the ground. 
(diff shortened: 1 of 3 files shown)


conf/modules/sonar_adc.xml
```diff
@@ -3,32 +3,84 @@
...
+    <configure name="SONAR_ADC_DEV" value="ADC_X" description="Set ADC device port the sonar is connected to, defaults to ADC_3"/>
+    <configure name="USE_SONAR_ADC_AGL" value="TRUE|FALSE" description="Updates the AGL value in state,if not defined, defaults to FALSE"/>
+    <configure name="SONAR_ADC_COMPENSATE_ROTATION" value="TRUE|FALSE" description="Compensate AGL measurements for body rotation. Disabled by default"/>
+    <configure name="SONAR_ADC_USE_FILTER" value="TRUE|FALSE" description="Enable or disable a filter on the sonar output values, defaults to TRUE"/>
+    <configure name="SONAR_ADC_PERIODIC_FREQUENCY" value="30" description="Sensor readings in HZ"/>
+    <configure name="SONAR_ADC_LOWPASS_TAU" value="0.11" description="If filtering is enabled, then one optionally can set this value to filter the output more(or less). Defaults to 0.16"/>
+    <configure name="SONAR_ADC_SCALE" value="0.001789" description="Sensor scale factor [m/adc], defaults to 0.001855"/>
+    <define name="USE_SONAR_ADC" value="TRUE|FALSE" description="Activate use of sonar in INS extended filter (only rotorcraft), defaults to FALSE"/>
```
gautierhattenberger: The is a `USE_SONAR`, especially in the INS_INT filter, but no `USE_SONAR_ADC`. You should not change the name here.

conf/modules/sonar_adc.xml
```diff
@@ -3,32 +3,84 @@
...
+    <define name="SONAR_ADC_PERIODIC_FREQUENCY" value="$(SONAR_ADC_PERIODIC_FREQUENCY)"/>
+    <define name="SONAR_ADC_USE_FILTER" value="$(SONAR_ADC_USE_FILTER)"/>
     <file name="sonar_adc.c"/>
   </makefile>
-  <makefile target="ap">
-    <define name="ADC_CHANNEL_SONAR" value="$(ADC_SONAR)"/>
-    <define name="USE_$(ADC_SONAR)"/>
+  <makefile target="sim|nps">
+    <define name="USE_SONAR_ADC" value="TRUE"/><!-- in NPS use a virtual sonar to simulate rangeing measurements -->
   </makefile>
```
gautierhattenberger: Here, the `sonar_adc.c` is not compiled anymore, so it won't provide the simulated measures anymore. The `USE_SONAR_ADC` has no effect.

sw/airborne/modules/sonar/sonar_adc.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2010  Gautier Hattenberger, 2013 Tobias Münch
+ * Copyright (C) 2010  Gautier Hattenberger, 2013 Tobias Münch, 2020 OpenUAS
```
gautierhattenberger: 2020, really ? ;)

conf/modules/sonar_adc.xml
```diff
@@ -3,32 +3,84 @@
...
   </header>
-
   <init fun="sonar_adc_init()"/>
-  <periodic fun="sonar_adc_read()" freq="20"/>
-
-  <makefile target="ap|sim">
+  <periodic fun="sonar_adc_periodic()" freq="SONAR_ADC_PERIODIC_FREQUENCY"/>
+  <periodic fun="sonar_adc_report()" freq="SONAR_ADC_PERIODIC_FREQUENCY" autorun="FALSE"/>
+  <makefile target="ap">
+    <configure name="SONAR_ADC_DEV" default="ADC_3" case="lower|upper"/>
```
gautierhattenberger: The previous name of your `SONAR_ADC_DEV` is `ADC_SONAR` and it is used in a few airframe files. You should either keep the same name or update the config files.

sw/airborne/modules/sonar/sonar_adc.c
```diff
@@ -20,68 +20,190 @@
...
+    }
+  } else {
+    // Out of range, set to NAN
+    sonar_adc.distance = NAN;
+  }
+
 #else // SITL
   sonar_adc.distance = stateGetPositionEnu_f()->z;
-  Bound(sonar_adc.distance, 0.1f, 7.0f);
+  Bound(sonar_adc.distance, (float)SONAR_ADC_MIN_RANGE,
```
gautierhattenberger: The real sensor can display zero on ground with proper offset and mounting, while in simulation it is bounded by min range. I suggest to bound the lower value to zero (and keep the max range bound).

sw/airborne/modules/sonar/sonar_adc.c
OpenUAS: Yes ;( never any time to finish stuff started. Doing what we can..

conf/modules/sonar_adc.xml
OpenUAS: For now indeed will make it: USE_SONAR again. Issue for all things sensors still is using a sonar I2C and a sonar ADC then unclear which.  And if I've finished making PR for all commonly used RangeFinder  sensors, we can discuss cleaning up the naming if those... but that thus will not be anytime soon ;)

conf/modules/sonar_adc.xml
OpenUAS: Will make it SONAR_ADC_PORT wherever in all airframes needed. Reason using PORT as toward a bit more unification of naming, e.g. MODEM_PORT, GPS_PORT. A port is after all the name of connection point where to connect the device to the flightcontroller, that is what a common user can determine reading the documentation of the FC.

PR #3544, opened 2025-10-01
[utils] Fix circular buffer and uavcan

fvantienen:
This fixes a mistake from converting the circular buffer size from `uint8_t` to `uint16_t` and not wrapping correctly for unaligned buffers. https://github.com/paparazzi/paparazzi/pull/3512
Also a mistake in the sizing of the buffers in the uavcan module, which was not converted from bits to bytes. https://github.com/paparazzi/paparazzi/pull/3513

### DIFF
--- a/sw/airborne/utils/circular_buffer.c
+++ b/sw/airborne/utils/circular_buffer.c
@@ -51,18 +51,9 @@ int circular_buffer_get(struct circular_buffer *cb, uint8_t *buf, size_t len)
 
 int circular_buffer_put(struct circular_buffer *cb, const uint8_t *buf, size_t len)
 {
-  int available = 0;
-  if (cb->read_offset > cb->write_offset) {
-    available = cb->read_offset - cb->write_offset - 2;
-  } else {
-    available = cb->_buf_len - (cb->write_offset - cb->read_offset) - 2;
-  }
+  size_t available = circular_buffer_available(cb);
 
-  /**
-   * len == available is invalid because it will cause
-   * write_offset to be equal to read_offset, which is considered an empty buffer.
-  */
-  if ((int)len >= available) {
+  if (len > available) {
     return CIR_ERROR_NO_SPACE_AVAILABLE;
   }
 
@@ -104,9 +95,9 @@ int circular_buffer_drop(struct circular_buffer *cb) {
   size_t end_offset = cb->read_offset;
 
   while(end_offset != cb->write_offset) {
-    size_t record_head_offset = end_offset;
+    record_head_offset = end_offset;
     uint16_t msg_len_p = cb->_buf[record_head_offset] | (cb->_buf[(record_head_offset + 1) % cb->_buf_len] << 8);
-    size_t end_offset = record_head_offset + msg_len_p + 2;
+    end_offset = record_head_offset + msg_len_p + 2;
     if (end_offset >= cb->_buf_len) {
       end_offset -= cb->_buf_len;
     }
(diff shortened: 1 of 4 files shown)


fvantienen:
I tried to fix the circular buffer correctly, but it is pretty hard to read and verify by hand. Ideally we add some test cases to verify the correctness of the circular buffer.
Can you maybe take a look and verify @Fabien-B ?

sw/airborne/utils/circular_buffer.c
```diff
@@ -70,17 +71,26 @@ int circular_buffer_put(struct circular_buffer *cb, const uint8_t *buf, size_t l
...
+  cb->_buf[cb->write_offset] = len & 0xFF;
+  cb->_buf[(cb->write_offset + 1) % cb->_buf_len] = (len >> 8) & 0xFF;
+
+  // Write the date no wrapping needed
   if (end_offset > cb->write_offset) {
     memcpy(cb->_buf + cb->write_offset + 2, buf, len);
-  } else {
+  } 
+  // Wrapping data partially
+  else if ((cb->write_offset + 2) < cb->_buf_len) {
```
fvantienen, line 75: Especially also verify this case as I think the `<` is correct and should not be `<=`

fvantienen:
At least these test cases are needed:
- read_offset=0, write_offset=0, buf_len=1024: try put/get of data and dropping (No wrapping at all)
- read_offset=1020, write_offset=1020, buf_len=1024, try put/get of data and dropping (Wrapping data)
- read_offset=1022, write_offset=1022, buf_len=1024: try put/get of data and dropping (Writing length on the end, data at idx=0)
- read_offset=1023, write_offset=1023, buf_len=1024: try put/get of data and dropping (Wrapping length, data at idx=1)

Fabien-B:
I did not think of this case at all...
It starting to be hard to read, wouldn't it be best is we decide that the length must not be wrapped around?
In the case there is only one byte left, it would just be unused.
I can do it, and also make a test program.

fvantienen:
Yes I also thought about that solution, except that makes it also complicated as you need to keep track of when you want to 'skip' that byte. So my guess was that this was the 'easier' solution, but feel free to implement the skip solution.

fvantienen:
An even easier solution is to maybe make the buffer 1 byte smaller than it actually is. That way you solve the problem directly. As the length can be written further than the buffer 'size' and still be valid.
But for me everything is fine with the solutions. As long as we can guarantee that they work.

Fabien-B:
So, I tried my idea but it was a nightmare so I stopped trying.
I wrote some test cases, that allowed me to fix the drop function.
I also added convenience function `clear` and `available` (to get the available space for the next buffer).

sw/airborne/arch/chibios/modules/uavcan/uavcan.c
```diff
@@ -273,16 +273,16 @@ void uavcan_broadcast(struct uavcan_iface_t *iface, uint64_t data_type_signature
     .data_type_signature = data_type_signature,
     .data_type_id = data_type_id,
     .priority = priority,
-    .payload_len = payload_len
+    .payload_len = payload_len // in bits
```
Fabien-B, line 276: Wow, I never realized it was in bits, assuming it was in bytes!

fvantienen: Yes it is a bit weird but they need it because of the weird data types, that's why I added the comment. It also took me a bit to notice this mistake.

PR #3548, opened 2025-10-16
[actuator] change configuration of actuators

gautierhattenberger:
- add a configuration structure to the actuators array
- limit the use of macros
- add a shell command to get and set actuator value
- update config for all boards

### DIFF
--- a/conf/modules/actuators_nps.xml
+++ b/conf/modules/actuators_nps.xml
@@ -11,6 +11,7 @@
     <provides>actuators</provides>
   </dep>
   <makefile target="sim|nps">
+    <file name="actuators_pwm.c"/>
     <file_arch name="actuators_pwm_arch.c"/>
   </makefile>
 </module>


sw/airborne/modules/actuators/actuators_faulhaber.c
```diff
@@ -470,3 +470,11 @@ void actuators_faulhaber_SetMode(uint8_t mode)
   faulhaber.mode = mode;
   faulhaber.state = 0;
 }
+
+void actuators_faulhaber_set(uint8_t idx UNUSED, int16_t value UNUSED)
+{
+#if USE_NPS
```
fvantienen: Should be opposite

sw/tools/generators/gen_airframe.ml
```diff
@@ -266,19 +270,27 @@ let preprocess_value = fun s v prefix ->
...
+      .config = { \\
+        .driver_no = SERVO_%s_DRIVER_NO, \\
+        .servo_idx = SERVO_%s_DRIVER_IDX, \\
+        .min = SERVO_%s_MIN, \\
+        .max = SERVO_%s_MAX, \\
+        .neutral = SERVO_%s_NEUTRAL, \\
+        .travel_up = SERVO_%s_TRAVEL_UP, \\
+        .travel_down = SERVO_%s_TRAVEL_DOWN \\
+      }, \\
+      .set = Actuator%sSet \\
```
fvantienen, line 288: Shouldn't we just use the functions instead of the define.

sw/airborne/modules/actuators/actuators_faulhaber.c
gautierhattenberger: fixed

sw/tools/generators/gen_airframe.ml
gautierhattenberger: This is only for consistency with the other generated macros. That might be changed some day, but better do it all at once in my opinion.

gautierhattenberger:
Tested and working. @fvantienen Are you okay with this ?

fvantienen:
For me, since it is tested I'm ok with merging this

fvantienen:
Bebop still has some problem I see with the tests:
```
usr/bin/ld: /home/pprz/paparazzi/var/aircrafts/bebop2_opticflow/nps/modules/actuators/actuators.o:(.data.rel+0x20): undefined reference to `actuators_pwm_set'
```

gautierhattenberger:
@fvantienen Thanks, it should be fine now

PR #3552, opened 2025-10-22
Add support for Mateksys H743Wing v3 FC

OpenUAS:
A board config for the Mateksys H743Wing v3 FC. Almost the same a Mateksys H743 Slim, but still for additional enhancements a specific board. Board has flown in airframe from another testbranch.

### DIFF
--- a/conf/airframes/OPENUAS/openuas_zohd_talon_nano.xml
+++ b/conf/airframes/OPENUAS/openuas_zohd_talon_nano.xml
@@ -1,0 +1,463 @@
+<!DOCTYPE airframe SYSTEM "../airframe.dtd">
+<airframe name="Ziggy">
+  <description>
+   + Fun airframe to test some OSD functionality and Matek H734 Wing v3 FC
+
+    + MODEL:            ZoHD Talon Nano
+    + FLIGHTCONTROLLER: Mateksys H734 Wing v3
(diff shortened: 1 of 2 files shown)


OpenUAS:
> Maybe you should add an example airframe file which uses the board?
Thx. Yeah good point, my favorite option is to add an example airframe, but in the past I did just that, but then got the comment, please make the PR as concise as possible... Cannot please everyone, my guess :) Anyhow the hopes we can testfly this board on master anytime soon I hope I'll add an airframe asap so I can continue with the rest off all PR we still have in stock in the "Special" branch.

gautierhattenberger:
I agree that we should have a general guideline when adding a new board.
My proposal would be:
- if you have a basic support of a new board working at the lab but not much used for flights, it is acceptable to not push an airframe example
- if the board is really used with a functional plane, better add it
- if the board is widely used, the airframe example can be added to the `test_boards` folder and conf to have it tested by compilation test.

@OpenUAS would it be okay for you ?

OpenUAS:
Thx, yes all proposals fine by me. Just FYI before a PR, I require self to have flown an aircraft in Auto2 at least once without mishap and a desktest of 30min without any issue aka dogfooding.  OK now the time to dig up the baro driver for this board and all sorts of other tasks...

PR #3557, opened 2025-11-04
[ekf2] Changes to mag fusion

NoahWe:
Automatically change the mag fusion mode to Indoor when Optitrack is defined
Set the default mag fusion mode to auto
initialize some mag values to 0


fvantienen:
The ` ekf2.mag_fusion_type` variable can also be removed

gautierhattenberger (approved):

PR #3559, opened 2025-11-06
[supervision] improve start time

Fabien-B:
Shorten a little bit the start time of the paparazzi center.


gautierhattenberger:
With this, the combo box for 'set' selection is reduced to the default side instead of the size required to display correctly all the names. As a result, only a part of the name is displayed and it is not really usable. Would it be possible to resize the box once you know the size of the text inside ?

Fabien-B:
Good catch, that's done!

PR #3561, opened 2025-11-08
Corax - Baby Wing AR Pro - Minimal size maximal endurance - IMAV2025

dewagter:
- 2 hour flight time (sea-level)
 
ARBabyWing

### DIFF
--- a/conf/airframes/OPENUAS/openuas_sonicmodell_baby_ar_wing_pro.xml
+++ b/conf/airframes/OPENUAS/openuas_sonicmodell_baby_ar_wing_pro.xml
@@ -147,7 +147,6 @@ Safety
       <!-- We are using aspirin 2.2, so we set the following barometer configuration: -->
       <configure name="LISA_M_BARO" value="BARO_MS5611_SPI"/>
       <configure name="USE_BARO_BOARD" value="TRUE"/><!-- already true per default -->
-      <!-- <configure name="USE_BARO_MEDIAN_FILTER" value="TRUE"/> -->
       <define name="USE_BARO_MEDIAN_FILTER" value="TRUE"/>
 
       <!-- <configure name="USE_MAGNETOMETER" value="TRUE"/> -->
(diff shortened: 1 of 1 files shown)


conf/flight_plans/tudelft/imav2025.xml
```diff
@@ -0,0 +1,92 @@
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+
+<flight_plan alt="2150" ground_alt="2080" lat0="18.9426688" lon0="-98.1893464" max_dist_from_home="1500" name="Basic" security_height="2130">
```
gautierhattenberger: Except for the ground alt, what is the difference with the basic flight plan ? Is it really necessary to keep it?

conf/airframes/OPENUAS/openuas_sonicmodell_baby_ar_wing_pro.xml
gautierhattenberger, line 1: It seems there is a lot of commented or unused parts in this file. Would it be possible to clean it a bit ?

gautierhattenberger:
@dewagter Side question: do you really see a benefit of using a toroidal propeller ?

conf/flight_plans/tudelft/imav2025.xml
dewagter: @copilot remove this imav2025.xml file and replace references in conf/tudelft/conf.xml with the basic.xml flight plan instead.

dewagter:
> @dewagter Side question: do you really see a benefit of using a toroidal propeller ?

So far, we have shown that it is stronger during landings.

conf/airframes/OPENUAS/openuas_sonicmodell_baby_ar_wing_pro.xml
dewagter: Removed some, but most are actually desired. The airframe does have all the MAG's and rangers. Does it break anything? I thought it was just well documented.

gautierhattenberger: I don't think it breaks anything. If you think all options should be kept, I fine with that.

gautierhattenberger (approved):

PR #3563, opened 2025-11-12
Updated uart default buffer and corrected uart potential issue

alessandro-mancinelli:
According to Freek, I have pushed this UART changes: 

1) Increased default UART buffer. 
2) Corrected a potential issue casting a negative value into uint16_t.

### DIFF
--- a/sw/airborne/arch/chibios/mcu_periph/uart_arch.c
+++ b/sw/airborne/arch/chibios/mcu_periph/uart_arch.c
@@ -1143,7 +1143,7 @@ void uart_put_buffer(struct uart_periph *p, long fd, const uint8_t *data, uint16
     if (space <= 0) {
       space += UART_TX_BUFFER_SIZE;
     }
-    if ((uint16_t)(space - 1) < len) {
+    if ((uint16_t)(space) < len+1) {
       chMtxUnlock(init_struct->tx_mtx);
       return;  // no room
     }
--- a/sw/airborne/mcu_periph/uart.h
+++ b/sw/airborne/mcu_periph/uart.h
@@ -37,7 +37,7 @@
 #if defined(STM32F1)
 #define UART_RX_BUFFER_SIZE 128
 #else
-#define UART_RX_BUFFER_SIZE 256
+#define UART_RX_BUFFER_SIZE 2048
 #endif
 #endif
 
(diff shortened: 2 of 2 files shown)


sw/airborne/arch/chibios/mcu_periph/uart_arch.c
```diff
@@ -1143,7 +1143,7 @@ void uart_put_buffer(struct uart_periph *p, long fd, const uint8_t *data, uint16
     if (space <= 0) {
       space += UART_TX_BUFFER_SIZE;
     }
-    if ((uint16_t)(space - 1) < len) {
```
gautierhattenberger, line 1146: In which case `space-1` is negative ? To me, the worst case is when `space` is equal to `UART_TX_BUFFER_SIZE-1` (with `extract_idx=0` and `insert_idx=UART_TX_BUFFER_SIZE-1 (e.g. the maximum value)`) , and with the line above, it becomes equal to `1`.

OpenUAS:
@fvantienen is it possible to re-review to make sure the case is as @gautierhattenberger states? If so maybewe could close (or fix and close) this one?

sw/airborne/arch/chibios/mcu_periph/uart_arch.c
Fabien-B: I agree that this case should never happens.

Let's say that `insert_idx = 4` and `extract_idx = 5`. 
In `uart_put_byte` we have :
```C
uint16_t temp = (p->tx_insert_idx + 1) % UART_TX_BUFFER_SIZE;
if (temp == p->tx_extract_idx) {
  chMtxUnlock(init_struct->tx_mtx);
  return;  // no room
}
```
So the buffer is unchanged, even in the particular case described by @gautierhattenberger.

`extract_idx` is never equal to `insert_idx`, so at worst, `p->tx_extract_idx - p->tx_insert_idx - 1 == 0`.

On a side note, there are many ring buffer implementations in paparazzi, so I made one more in [here](https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/utils/ring_buffer.h).
At some point we should use it everywhere it make sense, so we are not fixing the same bugs again and again.

Fabien-B:
Seems to not be necessary.
Re-open if the buffer size change is needed.

PR #3565, opened 2025-11-19
[indi] rewrite indi stabilization and guidance for modularity

gautierhattenberger:
This PR allows to redefine the outer-loop for stabilization and guidance, and to specify the guidance model for something else than a quadrotor (but still not an hybrid).
Two use case are proposed:
- replace the PD control of the outer-loop (still used by default) by and Hinfinity controller
- parametrize the guidance for an hexa-copter with tilted motors, allowing to produce side force for lateral motion control.

@EwoudSmeur @fvantienen @dewagter I tried to reorganise the code without breaking things, especially the case of `INDI_OUTPUTS == 5` corresponding to a pusher quadplane or equivalent. I introduced some new elements, like a matrix `ACT_THRUST_MAT` to specify which motors are producing thrust on which axis (`ACT_IS_THRUSTER_X|Y|Z` is still working for now). What do you think about updating your airframes to only have two cases: the default quadrotor style or the full configuration (`INDI_OUTPUTS == 6`) ?

And in general, what do you think about these changes ?


gautierhattenberger:
@fvantienen et al., do you have an opinion about this PR ?

sw/airborne/modules/actuators/actuators.h
```diff
@@ -87,12 +87,6 @@ struct actuator_t {
  * */
 extern struct actuator_t actuators[ACTUATORS_NB];
 
-/** PPRZ command to each actuator
- * Can be used to directly control actuators from the control algorithm
- * if the command_laws are set up appropriately in the airframe file
- */
-extern int16_t actuators_pprz[ACTUATORS_NB];
```
EwoudSmeur, line 94: The goal of this was to have other software (e.g. the simulator) know about the actuator commands without including the specific controller file. Is there a replacement for this functionality?

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -155,36 +195,29 @@ struct Indi_gains indi_gains = {
...
-bool indi_use_adaptive = false;
-#endif
+bool indi_use_adaptive = STABILIZATION_INDI_USE_ADAPTIVE;
 
 #ifdef STABILIZATION_INDI_ACT_RATE_LIMIT
 float act_rate_limit[INDI_NUM_ACT] = STABILIZATION_INDI_ACT_RATE_LIMIT;
 #endif
 
-#ifdef STABILIZATION_INDI_ACT_IS_SERVO
+// Vector of boolean telling is an actuator is a servo
```
EwoudSmeur: is=if

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
```diff
@@ -155,36 +195,29 @@ struct Indi_gains indi_gains = {
...
-#endif
 
-#ifdef STABILIZATION_INDI_ACT_IS_THRUSTER_X
-bool act_is_thruster_x[INDI_NUM_ACT] = STABILIZATION_INDI_ACT_IS_THRUSTER_X;
+// [3 x INDI_NUM_ACT] matrix representing the contribution (True/False) of an
+// actuator to the total 3D thrust vector
+#ifdef STABILIZATION_INDI_ACT_THRUST_MAT
+static bool act_thrust_mat[3][INDI_NUM_ACT] = STABILIZATION_INDI_ACT_THRUST_MAT;
+#else // backward compatibility
+static bool act_thrust_mat[3][INDI_NUM_ACT] = {
```
EwoudSmeur, line 212: The goal of the thruster booleans was to efficiently add those actuator states that contributed to the thrust, but I'm not sure the booleans are really needed. The control effectiveness matrix has all the required information about the actuators' contribution to thrust in any axis, and checking if it is zero should still be fast.

EwoudSmeur: An exception may be control surfaces like flaps though...

EwoudSmeur (commented):
I'm ok with the restructuring. The main thing I'm wondering is if (not just with this PR) things are growing too complex. There are so many options and functions calling other functions that it is easy to get lost. The generated autopilot helps in that regard, allowing for one's own 'lean' controller, but may limit re-usability.

sw/airborne/modules/actuators/actuators.h
gautierhattenberger: It was still working, so the correct header was still included somehow. To be sure, I have included `generated/modules.h` where `actuators_pprz` is used in NPS, so whatever actuator is used, it will be fine.

By the way, my opinion on that is that NPS should only rely on the command vector, even in the case of INDI. It means that `actuators_pprz` should be used to fill the command array, that is eventually directly mapped to that actuators without mixing.

sw/airborne/firmwares/rotorcraft/stabilization/stabilization_indi.c
gautierhattenberger: I tried to get rid of it, or to make it easier to declare, but it was not as easy... I'm afraid we have to keep that.
Also, I tried to make the "thrust" channels coherent for the G matrix, but it is not possible without breaking all your rotwing airframes in particular. So we have with 4 outputs: `[roll, pitch, yaw, Z]`, 5 outputs `[roll, pitch, yaw, Z, X]` and 6 outputs `[roll, pitch, yaw, X, Y, Z]`. I would be happy to only keep the full (6 outputs) or reduced (4 outputs, e.g. "classic" frame) versions, but it is up to you guys.

gautierhattenberger:
> I'm ok with the restructuring. The main thing I'm wondering is if (not just with this PR) things are growing too complex. There are so many options and functions calling other functions that it is easy to get lost. The generated autopilot helps in that regard, allowing for one's own 'lean' controller, but may limit re-usability.

@EwoudSmeur I understand your concern about things getting too complex. It will always be a trade-off between duplicating code or making it more modular. In this case, I though that it was worth allowing to keep the general structure of INDI and adding a way to change the default PD loop.
I flew with it for the Hinf part and it is working correctly (both default PD and Hinf) without breaking compatibility. I can't fully test yet the force allocation part, but it is well separated from the rest, so I'll debug it later if needed.
If possible, I'd like to merge it to avoid divergence of code and preparing future work as well.

gautierhattenberger:
@EwoudSmeur it would help me to have this soon in master. Do you agree with the changes ?

EwoudSmeur:
Yes I'm ok with merging!

PR #3570, opened 2025-12-06
rename circular_buffer to framed_ring_buffer

Fabien-B:
The "circular_buffer" name was misleading, so it was renamed "framed_ring_buffer".
An actual ring buffer is added, named "ring_buffer".


tests/utils/test_framed_ring_buffer.c
gautierhattenberger, line 1: Don't you need to change the name of this file to `test_framed_ring_buffer.c` ?

PR #3572, opened 2025-12-16
[parrot] add support for wpa_supplicant for bebop

gautierhattenberger:
- update script and binary
- add sound and cdc_acm options

### DIFF
--- a/sw/tools/parrot/bebop/pprzstarter
+++ b/sw/tools/parrot/bebop/pprzstarter
@@ -61,6 +61,19 @@ try_join()
   done
 }
 
+# kill wpa_supplicant if needed and reset wifi chip
+# $1 error message
+reset_wifi()
+{
+  echo $1, reverting back to default
+  killall wpa_supplicant
+  /sbin/broadcom_reset.sh
+  BLDC_Test_Bench -n -M -2
+  echo "Reset broadcom: $1" > /data/ftp/internal_000/scripts/MSG_latest_Failed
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
The binary version of `wpa_supplicant` works with some versions of busybox (e.g. `BusyBox v1.25.0 (2017-06-29 09:25:56 CEST)`) but not with others (e.g. `BusyBox v1.25.0 (2018-06-06 11:57:26 UTC)`).
I'll try to find a solution that works for both or try to recompile from source. The version that have been tested is `wpa_supplicant v0.6.10`.

```
ldd /bin/wpa_supplicant 
        /usr/lib/libcrashdump.so (0xb6fc3000)
        libdl.so.2 => /lib/libdl.so.2 (0xb6faf000)
        libgcc_s.so.1 => /lib/libgcc_s.so.1 (0xb6f86000)
        libc.so.6 => /lib/libc.so.6 (0xb6e94000)
        libulog.so => /usr/lib/libulog.so (0xb6e80000)
        libpthread.so.0 => /lib/libpthread.so.0 (0xb6e5c000)
        /lib/ld-linux.so.3 (0xb6fd5000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0xb6d81000)
        libm.so.6 => /lib/libm.so.6 (0xb6d0b000)
```
An other option is to downgrade the parrot firmware: `https://github.com/prgumd/Parrot-Bebop2-Setup/blob/master/Firmware-Instructions.md`

PR #3573, opened 2025-12-16
[bebop] fix timeout for blocking i2c in bebop actuators

gautierhattenberger:
The wrong timeout value for introduced with #3471.

Also use correct settings for imu with 500Hz periodic frequency.

### DIFF
--- a/sw/airborne/boards/disco/actuators.c
+++ b/sw/airborne/boards/disco/actuators.c
@@ -45,6 +45,8 @@
 #warning "Disco actuators require a <servo name=MOTOR>"
 #endif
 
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
I looked at the rest of the IMU code, and only the disco was not updated.
Checking for only two values (500 and 512) is not the best. As you said @fvantienen , their are no real warnings, unless you activate the print config option when compiling.
Maybe changing the test for `PERIODIC_FREQUENCY >= 500` is already a better idea ? and the "low" freq settings in a else case, so all values are covered with 2 different settings, unless the user redefine the values by hand ?

Fabien-B (dismissed):

Fabien-B (approved):

PR #3578, opened 2026-01-21
feat: add ANDI stabilization controller

jpgdubois:
Add Actuator Nonlinear Dynamic Inversion Stabilization (ANDI) controller.

### DIFF
--- a/conf/autopilot/jpg_cyclone_andi.xml
+++ b/conf/autopilot/jpg_cyclone_andi.xml
@@ -1,0 +1,194 @@
+<!DOCTYPE autopilot SYSTEM "autopilot.dtd">
+
+<autopilot name="ANDI Autopilot Cyclone">
+  <state_machine name="ap" freq="PERIODIC_FREQUENCY" gcs_mode="true" settings_mode="true" settings_handler="autopilot_generated|SetModeHandler">
+
+    <modules>
+      <module name="actuators" type="t4_uart"/>
+      <module name="stabilization" type="andi"/>
+    </modules>
+
+    <includes>
+      <include name="autopilot_rc_helpers.h"/>
+      <define name="MODE_MANUAL" value="AP_MODE_ATTITUDE_DIRECT" cond="ifndef MODE_MANUAL"/>
+      <define name="MODE_AUTO1" value="AP_MODE_RATE_DIRECT" cond="ifndef MODE_AUTO1"/>
+      <define name="MODE_AUTO2" value="AP_MODE_RATE_DIRECT" cond="ifndef MODE_AUTO2"/>
+      <define name="RCLost()" value="(radio_control.status == RC_REALLY_LOST)"/>
+    </includes>
+
+    <settings>
+      <dl_setting var="autopilot.kill_throttle" min="0" step="1" max="1" values="Resurrect|Kill"/> <!-- handler="KillThrottle" -->
+    </settings>
+
+    <!-- Commit actuator commands. ANDI/INDI perform allocation themselves; retained so COMMAND_THRUST remains set -->
+    <control_block name="set_commands">
+      <call fun="SetRotorcraftCommands(stabilization.cmd, autopilot_in_flight(), autopilot_get_motors_on())"/>
+    </control_block>
+
+    <control_block name="run_rate_control">
+      <call fun="guidance_v_run(autopilot_in_flight())" store="struct ThrustSetpoint thrust_sp"/>
+      <call fun="stabilization_andi_run(true, autopilot_in_flight(), &stabilization.rc_sp, &thrust_sp, stabilization.cmd)"/>
+    </control_block>
+
+    <control_block name="run_attitude_control">
+      <call fun="guidance_v_run(autopilot_in_flight())" store="struct ThrustSetpoint thrust_sp"/>
+      <call fun="stabilization_andi_run(false, autopilot_in_flight(), &stabilization.rc_sp, &thrust_sp, stabilization.cmd)"/>
+    </control_block>
+
+    <mode_selection>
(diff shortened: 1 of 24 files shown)


conf/autopilot/jpg_cyclone_andi.xml
```diff
@@ -0,0 +1,200 @@
+<!DOCTYPE autopilot SYSTEM "autopilot.dtd">
+
+<autopilot name="ANDI Autopilot Cyclone">
+  <state_machine name="ap" freq="PERIODIC_FREQUENCY" gcs_mode="true" settings_mode="true" settings_handler="autopilot_generated|SetModeHandler">
+
+    <includes>
```
gautierhattenberger, line 11: Maybe you could include some of the required modules directly here (see https://github.com/paparazzi/paparazzi/blob/master/conf/autopilot/heewing_vtol.xml#L7).
The header `modules.h` is already included in the generated autopilot file, so it is not necessary to have manual includes of the headers coming from modules.

conf/flight_plans/JPG/jpg_valkenburg.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0"?>
+<!DOCTYPE flight_plan SYSTEM "../flight_plan.dtd">
+<flight_plan alt="60" ground_alt="0" lat0="52.168212" lon0="4.412785" max_dist_from_home="5000" name="Cyclone2 Valkenburg Basic" security_height="2">
+  <header>
+#include "autopilot.h"
```
gautierhattenberger: it shouldn't be needed

conf/jpg_messages.xml
gautierhattenberger, line 1: This file should not be here. If new messages are required, it should be with a pull request in pprzlink

sw/airborne/filters/complementary_filter.h
```diff
@@ -0,0 +1,307 @@
...
+
+
+/**
+ * @brief First order complementary filter structure.
+ *
+ * This structure contains two first order low-pass filters,
+ * one for the high-pass path and one for the low-pass path.
+ */
+struct FirstOrderComplementary {
+  struct FirstOrderLowPass x_lp_filter; // Low pass filter instance for high pass path
```
gautierhattenberger, line 50: This complementary filter is a good addition. It would have be nice maybe to use/finish/fix the high pass filters in `filters/high_pass_filter.h`. What do you think ?

sw/airborne/filters/low_pass_filter_types.h
```diff
@@ -0,0 +1,544 @@
...
+ * @param[in,out] filter Struct containing first order low-pass filters for x, y, z components.
+ * @param[in] input Pointer to FloatVect3 struct containing new input data.
+ */
+static inline void update_first_order_low_pass_vect3(struct FirstOrderLowPassVect3 *filter,
+    const struct FloatVect3 *input)
+{
+  update_first_order_low_pass(&filter->x, input->x);
+  update_first_order_low_pass(&filter->y, input->y);
+  update_first_order_low_pass(&filter->z, input->z);
+}
```
gautierhattenberger: The `update_xxx_low_pass` functions are returning the last filtered value. What about returning a the values here with the proper type (vect3, rates). It can avoid to call the `get_xxx` function later.

sw/airborne/filters/low_pass_zoh_filter_types.h
```diff
@@ -0,0 +1,66 @@
+#ifndef LOW_PASS_ZOH_FILTER_TYPES_H
```
gautierhattenberger, line 32: please add header

conf/modules/stabilization_andi.xml
```diff
@@ -0,0 +1,170 @@
...
+    <depends>stabilization_rotorcraft,@attitude_command,wls</depends>
+    <provides>commands</provides>
+  </dep>
+  <header>
+    <file name="stabilization_andi.h"/>
+  </header>
+  <init fun="stabilization_andi_init()"/>
+  <makefile target="ap|nps" firmware="rotorcraft">
+    <file name="stabilization_andi.c" dir="$(SRC_FIRMWARE)/stabilization"/>
+    <file name="obm_cyclone.c" dir="$(SRC_FIRMWARE)/stabilization"/>
```
gautierhattenberger, line 88: It would be better to make a module to select the `obm_xxx` file according to the aircraft type.

gautierhattenberger (changes requested):

sw/airborne/filters/low_pass_filter_types.h
jpgdubois: The update functions now directly return the value, with the exception of the array functions.

sw/airborne/filters/complementary_filter.h
jpgdubois: I think this makes sense, I have not found the time to copy the functionality of low_pass_filter.h to the high_pass_filter.h and validate it.

conf/modules/stabilization_andi.xml
jpgdubois: A dedicated module would be the proper implementation.

The obm_xxx modules could then replace the current control effectiveness scheduling functionality used in all other INDI/ANDI controllers. Each controller implementation would remain independent of the specific vehicle or scheduling method, while separate obm_xxx implementations could handle different levels of scheduling (from full scheduling to simple constant matrices).

I would prefer to merge the current implementation and keep this suggestion in mind when making the modules.

jpgdubois:
Thanks! I do no have much time to continue working on this or validating the changes I make at the moment, please merge as-is if okay.

PR #3579, opened 2026-01-27
Uavcan update

Fabien-B:
This is a big update for UAVCAN, I hope it makes sense.
I divided it in a few commits to be easier to understand.

## [[utils] Add const specifier to ring_buffer write.](https://github.com/paparazzi/paparazzi/pull/3579/changes/c01ff46c0d3e579aa69feb143a823bfce7eca465)
Nothing important, just more semantically correct.

## [[utils] Add key-value store.](https://github.com/paparazzi/paparazzi/pull/3579/changes/d1776d81c035c1a8f68482a558cf33fe0154ab69)
Adds a key-value store structure. I use it in the following commits, and I know that something similar is used elsewhere in paparazzi. In the vision of unifying all those little tools rewritten a thousand times, I made that a reusable library (hopefully free of bugs...)

## [[uavcan] radio control over uavcan.](https://github.com/paparazzi/paparazzi/pull/3579/changes/8688f2f3d6fec2d3b54257f7ad5310d0c10aa62c)
Adds a radio control module, that listen the [`dronecan.sensors.rc.RCInput`](https://github.com/dronecan/DSDL/blob/master/dronecan/sensors/rc/1140.RCInput.uavcan) message.
It expects value in PPM range (1000-2000).

## [[uavcan] uavcan tunnel device](https://github.com/paparazzi/paparazzi/pull/3579/changes/90f4708c3f43ccef6c948e85f24448e32e928f55)
Implement a `pprz_device` with [`uavcan.tunnel.Broadcast`](https://github.com/dronecan/DSDL/blob/master/uavcan/tunnel/2010.Broadcast.uavcan) message as backend.
It can be used e.g. to tunnel a serial port.

## [[uavcan] Add responses, fix inout transfer id, use semaphore](https://github.com/paparazzi/paparazzi/pull/3579/changes/b34c88e1e47ffc7fc9ad64894679bd43d22bac50)
The big changes.
- The `transfer_id` was not used correctly: It was a uniq pointer, but there needs to be one `transfer_id` per `{msg type, destination, source}`. I used the kv_store created previously to store them.
- Removed useless a mutex for reception (the reception in single threaded)
- However, RX and TX still share the underlying memory pool. We found the `CANARD_ALLOCATE_SEM` option, that make the memory pool thread-safe. It use a semaphore that we define functions for.
- The ability to send responses to requests was added.
- Finally, the autopilot can respond to `GetNodeInfo` requests, and periodically sends a `NodeStatus` message to comply with the DroneCAN specs. Very usefull to monitor it in DroneCANGuiTool.

## [[uavcan] move out of arch/chibios](https://github.com/paparazzi/paparazzi/pull/3579/changes/0953e9e8219b030aaf52018e72e73aba37179122)
Move uavcan files out of `arch/chibios` as there should be no more dependencies to chibios. The goal is to make it work on the linux arch at least. (not tried yet).

### DIFF
--- a/tests/modules/test_arch/mcu_periph/sys_time_arch.h
+++ b/tests/modules/test_arch/mcu_periph/sys_time_arch.h
@@ -52,7 +52,4 @@ static inline uint32_t get_sys_time_msec(void)
          msec_of_cpu_ticks(sys_time.nb_sec_rem);
 }
 
-
-static inline void sys_time_usleep(uint32_t us __attribute__((unused))) {}
-
 #endif /* SYS_TIME_ARCH_H */


OpenUAS:
Thx, nice! BTW, build error

 ``` /home/pprz/paparazzi/sw/airborne/modules/uavcan/uavcan.h:29:10: fatal error: canard.h: No such file or directory ```

I have 'em in:
 ```sw/ext/libcanard/canard.h ```
and
 ```sw/ext/dronecan/libcanard/canard.h ```

PR #3582, opened 2026-02-02
[GPIO] Put common GPIO function in gpio.h.

Fabien-B:
Move common definitions from `gpio_arch.h` to `gpio.h`.
libopencm3 already define `gpio_set` and `gpio_clear`, so the functions name are prefixed with `pprz_` to avoid conflicts.

### DIFF
--- a/sw/airborne/arch/chibios/mcu_periph/gpio_arch.c
+++ b/sw/airborne/arch/chibios/mcu_periph/gpio_arch.c
@@ -94,3 +94,12 @@ void gpio_setup_pin_analog(ioportid_t port, uint16_t pin)
   chSysUnlock();
 }
 
+void gpio_set(gpio_port_t port, uint16_t pin)
+{
+  palSetPad(port, pin);
+}
+
+void gpio_clear(gpio_port_t port, uint16_t pin)
+{
+  palClearPad(port, pin);
+}
--- a/sw/airborne/arch/chibios/mcu_periph/gpio_arch.h
+++ b/sw/airborne/arch/chibios/mcu_periph/gpio_arch.h
@@ -38,52 +38,39 @@
  */
 typedef ioportid_t gpio_port_t;
 
-/**
- * Setup one or more pins of the given GPIO port as outputs.
- * @param[in] port
- * @param[in] gpios
- */
-extern void gpio_setup_output(ioportid_t port, uint16_t gpios);
-
-/**
- * Setup one or more pins of the given GPIO port as inputs.
- * @param[in] port
- * @param[in] gpios
- */
-extern void gpio_setup_input(ioportid_t port, uint16_t gpios);
 
 /**
  * Setup one or more pins of the given GPIO port as inputs with pull up resistor enabled.
  * @param[in] port
  * @param[in] gpios
  */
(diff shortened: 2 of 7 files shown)


fvantienen:
Nice, maybe we can later witch to line's instead of PIN/PORT as well. That makes for a lot less defines/configures.

fvantienen (dismissed):

sw/airborne/mcu_periph/gpio.h
```diff
@@ -25,11 +25,11 @@
...
- * - gpio_setup_output(port, gpios)
- * - gpio_setup_input(port, gpios)
+ * - pprz_gpio_setup_output(port, gpios)
+ * - pprz_gpio_setup_input(port, gpios)
  * - gpio_get(port, gpios)
- * - gpio_set(port, gpios)
- * - gpio_clear(port, gpios)
+ * - pprz_gpio_set(port, gpios)
+ * - pprz_gpio_clear(port, gpios)
  * - gpio_toggle(port, gpios)
```
gautierhattenberger: Shouldn't we have all functions starting with `pprz_` here ? Coherence will help for development.

sw/airborne/arch/stm32/mcu_periph/gpio_arch.c
```diff
@@ -92,14 +100,14 @@ void gpio_setup_input(uint32_t port, uint16_t gpios)
 void gpio_setup_input_pullup(uint32_t port, uint16_t gpios)
 {
   gpio_enable_clock(port);
-  gpio_set(port, gpios);
+  pprz_gpio_set(port, gpios);
```
gautierhattenberger: Since this is internal to the gpio driver, it could directly use `gpio_set/clear`.

sw/airborne/mcu_periph/gpio.h
Fabien-B: I agree, but not all functions are implemented for all architectures, so I renamed only the common functions for now.

sw/airborne/arch/stm32/mcu_periph/gpio_arch.c
Fabien-B: That's in fact the reason I renamed the functions: since `gpio_set` and `gpio_clear` already exists in libopencm3, we cannot declare them in `gpio.h` (technically we *can*, but with a tons of compiler warnings...).

sw/airborne/mcu_periph/gpio.h
gautierhattenberger: The functions that are not in implemented for all architectures can stay as they are now, but at least the one in `gpio.h` should be similar.

sw/airborne/arch/stm32/mcu_periph/gpio_arch.c
gautierhattenberger: yes, but here you really want to use the functions of libopencm3.

gautierhattenberger:
As a side note, with your plan to drop the support of stm32 baremetal, this will be no longer necessary to rename these functions....

Fabien-B:
So let's put this PR to sleep meanwhile.

PR #3586, opened 2026-02-04
Remove stm32 arch

Fabien-B:
- Remove `sw/airborne/arch/stm32`
- remove `libopencm3`
- remove `luftboot` that depends on libopencm3

Is that ok to remove luftboot ? Does anyone still use it ?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,3 @@
-[submodule "sw/ext/libopencm3"]
-	path = sw/ext/libopencm3
-	url = https://github.com/libopencm3/libopencm3.git
-[submodule "sw/ext/luftboot"]
-	path = sw/ext/luftboot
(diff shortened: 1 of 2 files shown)


fvantienen (approved):

gautierhattenberger:
Do we have an alternate solution for a bootloader if we remove luftboot ? @fvantienen how does it work on the pixhawk boards supported by paparazzi ? Is it the original bootloader from PX4 ?

There are still several references to luftboot in the makefiles where `HAS_LUFTBOOT` have been abused for years to tell if there is a bootloader or not. I guess we can merge this and see how it goes, but a little bit of cleaning / renaming would be good.

fvantienen:
Yes we use the PX4 bootloader for the pixhawks.

PR #3587, opened 2026-02-04
Remove old GCS and supervision.

Fabien-B:
I suggest that we also remove the old GCS and the old paparazzicenter.
The new ones have been used for quite some time now.


fvantienen (approved):

sw/tools/gcs_launch.py
gautierhattenberger, line 1: It is possible that people with a custom control panel still use this script ? Maybe it should be kept for now with a deprecated warning.

conf/tools/pprzgcs.xml
```diff
@@ -1,2 +1 @@
-<program command="sw/tools/gcs_launch.py" name="PprzGCS" />
-
+<program command="$pprzgcs" name="PprzGCS" />
```
gautierhattenberger: add `icon="gcs.svg" favorite="true"` ?

sw/tools/gcs_launch.py
Fabien-B: Only custom tools should be in the control panel, so it should not be the case.
However I restored the file, to just print an error message.
The "PprzGCS" tool was introduced after the tools have been moved out of the control panel, so it should not happen.

PR #3595, opened 2026-02-13
Add missing XCB for QT5 to install script.

dewagter:
Fixes:

```
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

./install.sh: line 58:  3893 Aborted                 (core dumped) python3 ./sw/tools/install.py
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,9 @@ then
 fi
 
 
+# Solves: Could not load the Qt platform plugin "xcb"
+sudo apt install libxcb-xinerama0 libxcb-cursor0 libxkbcommon-x11-0 libglu1-mesa
+
 if [ "$USE_VENV" = true ]
 then
   sudo apt install -y python3 python3-venv


dewagter:
Also needed in 24.04 LTS

dewagter:
> I don't use this script, but if it helps

Do you always copy-paste install instructions from the wiki? ```git clone && ./install``` that adapts to your OS and only installs what you need is not a good idea?

@Fabien-B : Qt5 (as installed by ```python setup.py```) is giving quite some trouble and prevents the install window from showing up, while apparently you can run Paparazzi without using: ```./install.py -n```, so can you tell why this was needed?

Fabien-B:
Using a venv is not strictly necessary, but it helps managing the python requirements.
In the paparazzi center, the "documentation" page use `PyQtWebEngine` that depends itself on `PyQt5`.

PR #3598, opened 2026-02-13
Add UnifiedMocapRouter

dewagter:
The default option for indoor motion tracking interfacing: UnifiedMoCapRouter


dewagter:
Add a ```make clean``` for the ```ext```.

gautierhattenberger:
We will test it next week at Enac

dewagter:
Tested in Ubuntu 18, 20, 22, 24.

For 18.04 you do need a ```cmake``` update:

```
wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add -
sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
sudo apt-get update && sudo apt-get install cmake
```

Fabien-B:
Could you add a file for it in `conf/tools` ?

gautierhattenberger:
It works for us as well, however it is not clear what are the dependencies. It needs `cmake` and some boost libraries, however I ended up installing `libboost-all-dev` as mentioned in the README, but isn't it a bit to much ?
We don't have these dependencies in the current `paparazzi-dev` packages, so compilation of paparazzi on a fresh system is likely to fail. What is the best option for this ? Skip build with a warning if cmak / boost are missing ?

dewagter:
> Could you add a file for it in `conf/tools` ?

Done.

dewagter:
> It works for us as well, however it is not clear what are the dependencies. It needs `cmake` and some boost libraries, however I ended up installing `libboost-all-dev` as mentioned in the README, but isn't it a bit to much ? We don't have these dependencies in the current `paparazzi-dev` packages, so compilation of paparazzi on a fresh system is likely to fail. What is the best option for this ? Skip build with a warning if cmak / boost are missing ?

Indeed. I didn't notice yet because ```gazebo``` installs all requirements.

You only need ```libboost-program-options-dev``` and ```libboost-filesystem-dev```. I also added a safeguard.

dewagter:
- [x] safeguard that explains missing items
 - [x] minimal install packages added to ```install.py``` waiting for the next ```paparazzi-dev``` update
 - [x] tools added
 - [x] tested in every ubuntu from 18 to 24 
 - [x] tested in delft and enac

gautierhattenberger:
It looks fine, but needs to fix conflict with master

dewagter:
> It looks fine, but needs to fix conflict with master

rebased.

PR #3600, opened 2026-02-15
Upgrade to OpenCV 4.10.0 for Bebop - Fix compiling on Ubuntu 22 and 24

dewagter:


dewagter:
- [x] tested in Ubuntu 20.04
 - [x] tested in Ubuntu 22.04
 - [x] tested in Ubuntu 24.04

PR #3613, opened 2026-03-06
[guidance] attempt to clean guidance indi code

gautierhattenberger:
- remove super dangerous RC_DEBUG flag
- use new filter *_types
- fix filters documentation (time constant -> omega)
- outer loop control separated as a WEAK function


gautierhattenberger:
@dewagter do you have a feedback for this one ?

dewagter (commented):
I like the idea a lot, but it will probably require testing several airframes.

gautierhattenberger:
We don't have an hybrid with `guidance_indi_hybrid` ready to fly at the moment, only in simulation. If you have the opportunity to test it, that would be nice.

EwoudSmeur (approved):

PR #3629, opened 2026-04-02
[gps] Fix gps uavcan

Fabien-B:


sw/airborne/modules/gps/gps_uavcan.c
```diff
@@ -130,9 +130,9 @@ static void gps_uavcan_cb(struct uavcan_iface_t *iface __attribute__((unused)),
   SetBit(state.valid_fields, GPS_VALID_VEL_NED_BIT);
 
   // norm of 2d ground speed in cm/s
-  state.gspeed = VECT2_NORM2(state.ned_vel);
+  state.gspeed = sqrtf(VECT2_NORM2(state.ned_vel));
```
gautierhattenberger, line 133: something like https://github.com/paparazzi/paparazzi/blob/master/sw/airborne/math/pprz_algebra_float.h#L132 ?

sw/airborne/modules/gps/gps_uavcan.c
```diff
@@ -49,7 +49,7 @@ static void gps_uavcan_cb(struct uavcan_iface_t *iface __attribute__((unused)),
     return;   // decode error
   }
 
-  struct GpsState state;
+  struct GpsState state = {0};
```
gautierhattenberger, line 52: This is not the only place where we should initialise this. Can you make a `gps_state_init(...)` function in `gps.h` and call it in other gps modules ?

gautierhattenberger (approved):

PR #3637, opened 2026-04-12
update to pcre2

Fabien-B:
Update various paparazzi tools to PCRE2, as pcre[13] is not available anymore in Ubuntu26.04.
PCRE2 is available in Jammy and Noble, so ivy-c (3.16) and ivy-ocaml (1.3.4) have been updated to use pcre2.
This commit will be needed after updating those packages.


gautierhattenberger:
I remove some old tools, a few of them where also using libpcre.

gautierhattenberger (approved):
CI failing due to gazebo, not related to this.

PR #3597, opened 2026-02-13
Gz Harmonic

dewagter:
- Gazebo Classic (```gazebo9, gazebo11```) is supported only until Ubuntu22
 - The new ```gz_harmonic``` should be used for ubuntu 22 and 24

PR #3640, opened 2026-04-15
[docker] add correct gazebo version

gautierhattenberger:


dewagter:
@gautierhattenberger does **gz** work with paparazzi? The whole point of this docker image was because gazebo classic is no longer supported and gz-harmonic is not yet supported.

gautierhattenberger:
I'm not really sure since I don't use gazebo...
We wanted to have the docker image set to ubuntu 24.04 and I copied the the updated install script that you have done. If you think this is the wrong way, I can still revert that, or even stick to ubuntu 22.04.

gautierhattenberger:
@dewagter It seems that `gz-harmonic` is really not working. What to you install on ubuntu 24.04 to use Gazebo ? If it is no longer working, can I remove the airframe using it from the compilation tests until this is solved ?

dewagter:
Support for the new gazebo started (See #3597).

In Ubuntu 2024, no version work with paparazzi yet. That is why we had the 22.04 based docker with the latest gazebo classic.

The install script adapts and only shows compatible versions.

PR #3648, opened 2026-04-21
remove cflags sort

Fabien-B:
Some flags reordering lead to a build fail on Ubuntu 26.04.


gautierhattenberger (approved):
maybe sorting source files could be removed as well ?

Issue #3647, opened 2026-04-21
Bluegiga USB dongle fail to build


Fabien-B:
The `bluegiga_usb_dongle` fail to build with gcc 15.2.

Is anyone using it ?
It should be easy to fix, but If no-one is using it, the best may be to remove it.

```
make -C sw/tools/bluegiga_usb_dongle
main.c: In function ‘main’:
main.c:1103:37: error: passing argument 3 of ‘pthread_create’ from incompatible pointer type [-Wincompatible-pointer-types]
 1103 |   pthread_create(&threads[0], NULL, send_msg, NULL);
      |                                     ^~~~~~~~
      |                                     |
      |                                     void * (*)(void)
In file included from main.c:153:
/usr/include/pthread.h:204:36: note: expected ‘void * (*)(void *)’ but argument is of type ‘void * (*)(void)’
  204 |                            void *(*__start_routine) (void *),
      |                            ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
main.c:845:7: note: ‘send_msg’ declared here
  845 | void *send_msg()
      |       ^~~~~~~~
main.c:1104:37: error: passing argument 3 of ‘pthread_create’ from incompatible pointer type [-Wincompatible-pointer-types]
 1104 |   pthread_create(&threads[1], NULL, recv_paparazzi_comms, NULL);
      |                                     ^~~~~~~~~~~~~~~~~~~~
      |                                     |
      |                                     void * (*)(void)
/usr/include/pthread.h:204:36: note: expected ‘void * (*)(void *)’ but argument is of type ‘void * (*)(void)’
  204 |                            void *(*__start_routine) (void *),
      |                            ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
main.c:896:7: note: ‘recv_paparazzi_comms’ declared here
  896 | void *recv_paparazzi_comms()
```

spacewolfXfr:
If removed, this may also be reflected in [Pprzlink](https://github.com/paparazzi/pprzlink), as there is a "BLUEGIGA" message (id 106) defined in the "Telemetry" class in the [messages.xml](https://github.com/paparazzi/pprzlink/blob/master/message_definitions/v1.0/messages.xml) file.

OpenUAS:
We have quite some, regardless, plz remove no one is using it anymore AFAIK

PR #3649, opened 2026-04-21
Remove Bluegiga

Fabien-B:
See #3647.
If nobody use it, I propose to remove it.
Otherwise, please fix the compilation issues.


dewagter (approved):
All new projects are on WiFi or UWB.

PR #3651, opened 2026-04-24
[indi] used proper quaternion 2nd order filtering

gautierhattenberger:
The filtering of euler angles was not correct, even with the correction for yaw from #3624.

The model applied here is a pseudo continuous 2nd order filter based on the reference model used for attitude control (but without the saturations). It assumes a sampling freq larger than the cutoff frequency, which is usually true here.

The `calcG` functions receive a filtered quaternion and are converting it to the Euler format that suits them. It still needs to match the expected "increments" that will be applied after running WLS (or matrix inverse).

### DIFF
--- a/sw/airborne/filters/quaternion_filter.h
+++ b/sw/airborne/filters/quaternion_filter.h
@@ -1,0 +1,191 @@
+/*
+ * Copyright (C) 2026 Gautier Hattenberger <gautier.hattenberger@enac.fr>
+ *
+ * This file is part of paparazzi
+ *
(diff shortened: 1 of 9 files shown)


Fabien-B (approved):

gautierhattenberger:
I want to do some extra test in real flight before merging, but @dewagter and @EwoudSmeur, do you have feedback for this ? Do you see any potential issue ?

gautierhattenberger:
Tested and working

PR #3653, opened 2026-05-05
Integration of Normalized Parametric Guiding Vector Fields

spacewolfXfr:
This is an implementation of the step adaptation strategy for [Parametric Guiding Vector Fields](https://ieeexplore.ieee.org/abstract/document/9312173), yielding the so-called [Normalized Parametric Guiding Vector Field](https://hal.science/hal-04550976).

The method is based on locally computing the step size at each round, in order to have the guiding point speed match the aircraft's. This formulation makes the parameters easier to tune as they can be related to physical values or be set to "neutral values", i.e. 1.

The step adaptation computation is based on a second order Taylor approximation of the parametric curve, which yields a polynomial equation solved using [Halley's method](https://en.wikipedia.org/wiki/Halley%27s_method).


sw/airborne/modules/guidance/gvf_parametric/gvf_adapted_step.c
gautierhattenberger, line 1: Can you clean a bit the code (unused function, commented debug lines, ...) ? Thanks

gautierhattenberger (changes requested):

gautierhattenberger (approved):

PR #3655, opened 2026-05-05
Draw shapes on the GCS directly from the aircraft

spacewolfXfr:
A new telemetry message is introduced: DRAW (Id 212, cf the matching [`pprzlink` pull requrest](https://github.com/paparazzi/pprzlink/pull/213).

This message is made to match the SHAPE ground message (Id 38), with some limitations on colors: only a finite subset of line and fill colors are available. It is done to limit the message size ; the fields `linecolor`, `fillcolor` and `opacity` have been merged into a single `uint8` with subparts accessible using bitfields.


sw/ext/pprzlink
gautierhattenberger, line 1: The reference to pprzlink should be updated after merging https://github.com/paparazzi/pprzlink/pull/213

gautierhattenberger (changes requested):

gautierhattenberger (approved):

PR #3656, opened 2026-05-21
BUGFIX: Incorrect frequency display (conversion error)

spacewolfXfr:
Also change sender discovery; it is now based on detecting the ALIVE message (rather than any message).


sw/ext/chibios
gautierhattenberger, line 1: You should not change this submodule

gautierhattenberger (changes requested):

gautierhattenberger (approved):

PR #3659, opened 2026-06-01
Paparazzi Center: Upgrade tool command parsing to allow quoted arguments 

spacewolfXfr:
Simple upgrade, using the [`shlex`](https://docs.python.org/3/library/shlex.html) Python module for parsing commands instead or splitting on spaces.

For testing, added tools allowing raw command execution (either standard CMD or using Python from pprzEnv)


spacewolfXfr:
Example of possible command:

```sh
~/paparazzi/pprzEnv/bin/python -c "import time;[print('Time: {:.3f} ; ToW (ms) : {} ({})'.format(time.time(),int((time.time()%604800)*1000),time.sleep(0.5)),flush=True) for _ in range(1 << 20)]"
```

gautierhattenberger (approved):

Issue #3664, opened 2026-06-17
Intermittent crash (code 11) in pprzsim-launch when simulating multiple aircraft


OpenUAS:
Description:
The Paparazzi simulator unexpectedly crashes or stops working after running normally for a period of time. This seems to occur when running multiple simulated aircraft simultaneously through a session in the Python-based control center. It is currently unclear exactly what triggers the crash, but it results in a specific exit code.

Environment / Setup

Launcher: Python-based control center
Session: Sim All New QT
Commands Executed: * 
```
pprzsim-launch -t sim -a Twisty
pprzsim-launch -t sim -a Easystar_3
```

Steps to Reproduce

- Open the Python-based control center.
-  Launch a Sim session to start the multiple applications and parameters.
-  Ensure both the Twisty and Easystar_3 aircraft are launched via pprzsim-launch as shown above.
-  Let the simulation run....
-  Wait. The crash happens intermittently "after a while."

Expected Behavior
The simulator should continue running indefinitely without freezing or terminating, maintaining the simulation for both aircraft.

Actual Behavior
The simulation suddenly stops working. Either pprzsim-launch or a subprocess it calls crashes completely.

Error Messages / Logs
When the crash occurs, the following message is emitted:

    Simulator terminated with code 11

Note: Exit code 11 typically indicates a Segmentation Fault (SIGSEGV), suggesting a memory access violation in the simulator or an underlying dependency.

The crash is caused by a bug in nps_ivy.c  see for a fix #3666

PR #3665, opened 2026-06-17
Fixes Sim Crash of issue #3664

OpenUAS:
# Why the simulator crashes with "code 11", and the fix

## Symptom

Running the *Sim All New QT* session with two sims (Twisty + Easystar_3) intermittently produces **"Simulator terminated with code 11"** in the Paparazzi Center console, "after a while". That message is emitted at console_widget.py when the sim's `QProcess` finishes with a non-zero exit; **code 11 = SIGSEGV** in the native `simsitl` binary (`pprzsim-launch` `os.execv`'s into it).

## Root cause — a data race inside the Ivy bus library

The `sim` target is built on the NPS infrastructure. NPS runs Ivy (the telemetry bus) across **two threads**:

- `th_display_ivy` → nps_main_common.c loops calling `nps_ivy_display()` → `IvySendMsg()`.
- `th_ivy_main` → nps_ivy.c runs `IvyMainLoop()`, which **adds, frees and rewrites** the global client/binding lists every time a peer connects, disconnects, or registers a regex.

`simsitl` links `libivy.so.3`, which is **single-threaded by design**: `IvySendMsg()` and `IvyMainLoop()` both walk and mutate the *same* global lists in the library's BSS **with no internal locking**. So while the display thread is iterating the client list to send telemetry, the loop thread can free a node out from under it → use-after-free → **SIGSEGV**. More agents on the bus (two sims + GCS + server, peers coming and going) = more list churn = higher probability = the "after a while" behavior.

The pre-existing `ivy_mutex` (from PR #2835) only wrapped the app's *own* `IvySendMsg` calls — `IvyMainLoop()` never took it, so the two threads were never actually mutually excluded.

### Deterministic proof (valgrind DRD)

DRD on the unfixed binary, with a peer connected and clients churning, reported **137 race contexts / 210,772 occurrences**, 129 of them the exact send-vs-loop pair on a `libivy.so.3` BSS address:

```
Conflicting load by thread 3 at 0x04921860      ← display thread
   at IvySendMsg (libivy.so.3.16)
   by nps_ivy_display (nps_ivy.c:235)
Allocation context: BSS section of libivy.so.3.16
Other segment (thread 4):                        ← ivy loop thread
   ... IvyBindingCompile ... IvyMainLoop (libivy.so.3.16)
   by ivy_main_loop (nps_ivy.c:53)
```

## The fix (only nps_ivy.c, +48/−6 lines)

`libivy`'s select-based `IvyMainLoop()` exposes **before/after-select hooks**. I use them to hold `ivy_mutex` during *all* event processing and release it only while the loop is idle in `select()`. Because every app-side Ivy call already takes `ivy_mutex`, sends and the loop's list bookkeeping can no longer overlap:

1. **The crash fix** — install `IvySetBeforeSelectHook`(unlock) / `IvySetAfterSelectHook`(lock) and enter the loop with the lock held.
2. **A second, separate bug DRD surfaced** — nps_ivy.c copied the shared `fdm` struct holding only `ivy_mutex`, not `fdm_mutex` (every other reader uses `fdm_mutex`). Memory-safe (numeric data → at worst garbage/NaN telemetry, not a crash), but a real race. Wrapped the `memcpy` in `fdm_mutex`.
3. **A benign one-byte flag race** — the `nps_ivy_send_world_env` signal flag was written under `fdm_mutex` but read/cleared under `ivy_mutex`. Made `fdm_mutex` its sole owner via a test-and-clear at the single consumer.

Lock ordering stays one-directional (`ivy_mutex → fdm_mutex`; the writer never takes `ivy_mutex`), so there is no deadlock risk.

## Validation

| Build | DRD race contexts | libivy send-vs-loop | FDM-state | Peers churned |
|---|:-:|:-:|:-:|:-:|
| Unfixed (HEAD) | 137 | 129 | 37 | 0 |
| + Ivy hooks | 4 | **0** | 4 | 28 |
| + fdm_mutex | 1 | 0 | 1 (flag) | 29 |
| **Complete fix** | **0** | **0** | **0** | 32 |

The complete fix yields **"0 errors from 0 contexts"** under DRD while 32 clients connect/disconnect. Both sims rebuild clean, the hooks resolve at runtime (`objdump` shows 2 each), and telemetry keeps flowing (no deadlock).

**Summary of what was delivered:**

- **Root cause identified and proven**: The crash is a use-after-free data race in `libivy.so.3` (single-threaded by design), driven concurrently by NPS's display thread (`IvySendMsg`) and ivy thread (`IvyMainLoop`). Proven deterministically with valgrind DRD — 129 send-vs-loop race contexts on the unfixed build.

- **Fix implemented** in nps_ivy.c (+48/−6, no other files): serializes the loop against sends via libivy's before/after-select hooks, plus two related synchronization bugs DRD surfaced in the same file.

- **Validated**: complete fix → DRD reports **0 errors from 0 contexts** with 32 clients churning; both sims build clean; telemetry flows with no deadlock.

### DIFF
--- a/sw/simulator/nps/nps_ivy.c
+++ b/sw/simulator/nps/nps_ivy.c
@@ -31,7 +31,11 @@ pthread_t th_ivy_main; // runs main Ivy loop
 static MsgRcvPtr ivyPtr = NULL;
 static int seq = 1;
 static int ap_launch_index;
-static pthread_mutex_t ivy_mutex; // mutex for ivy send
+/* Serializes ALL access to the (non thread-safe) Ivy bus state.
+ * Held by the IvyMainLoop thread while it processes events (via the
+ * before/after-select hooks installed in ivy_main_loop()) and by the
+ * display thread around every IvySendMsg/IvyBindMsg/IvyUnbindMsg call. */
+static pthread_mutex_t ivy_mutex = PTHREAD_MUTEX_INITIALIZER;
 
 /* Gaia Ivy functions */
 static void on_WORLD_ENV(IvyClientPtr app __attribute__((unused)),
@@ -47,10 +51,26 @@ void* ivy_main_loop(void* data __attribute__((unused)));
 
 int find_launch_index(void);
 
+static void ivy_loop_before_select(void *data __attribute__((unused)))
+{
+  pthread_mutex_unlock(&ivy_mutex);
+}
+
+static void ivy_loop_after_select(void *data __attribute__((unused)))
+{
+  pthread_mutex_lock(&ivy_mutex);
+}
 
 void* ivy_main_loop(void* data __attribute__((unused)))
 {
+  IvySetBeforeSelectHook(ivy_loop_before_select, NULL);
+  IvySetAfterSelectHook(ivy_loop_after_select, NULL);
+
+  /* The loop expects the lock to be held on entry: its first action each
+   * iteration is the before-select hook, which releases ivy_mutex. */
+  pthread_mutex_lock(&ivy_mutex);
   IvyMainLoop();
+  pthread_mutex_unlock(&ivy_mutex);
 
   return NULL;
 }
@@ -139,9 +159,11 @@ void nps_ivy_send_WORLD_ENV_REQ(void)
   // Bind to the reply
   ivyPtr = IvyBindMsg(on_WORLD_ENV, NULL, "^%d_%d (\\S*) WORLD_ENV (\\S*) (\\S*) (\\S*) (\\S*) (\\S*) (\\S*)", pid, seq);
 
-  // Send actual request
+  // Send actual request (copy the shared fdm under fdm_mutex, like nps_ivy_display)
   struct NpsFdm fdm_ivy;
+  pthread_mutex_lock(&fdm_mutex);
   memcpy(&fdm_ivy, &fdm, sizeof(struct NpsFdm));
+  pthread_mutex_unlock(&fdm_mutex);
 
   IvySendMsg("nps %d_%d WORLD_ENV_REQ %f %f %f %f %f %f",
       pid, seq,
@@ -152,8 +174,6 @@ void nps_ivy_send_WORLD_ENV_REQ(void)
       (fdm_ivy.ltpprz_pos.y),
       (fdm_ivy.ltpprz_pos.z));
   seq++;
-
-  nps_ivy_send_world_env = false;
 
   pthread_mutex_unlock(&ivy_mutex);
 }
@@ -289,7 +309,15 @@ void nps_ivy_display(struct NpsFdm* fdm_data, struct NpsSensors* sensors_data)
 
   pthread_mutex_unlock(&ivy_mutex);
 
-  if (nps_ivy_send_world_env) {
+  /* Test-and-clear the world-env request flag under fdm_mutex - the same lock
+   * the sim-loop thread holds when it sets the flag in nps_atmosphere_update()
+   * - so the flag is never accessed concurrently without synchronization. */
+  pthread_mutex_lock(&fdm_mutex);
+  bool do_world_env_req = nps_ivy_send_world_env;
+  nps_ivy_send_world_env = false;
+  pthread_mutex_unlock(&fdm_mutex);
+
+  if (do_world_env_req) {
     nps_ivy_send_WORLD_ENV_REQ();
   }
 }


OpenUAS:
Compared this solution to another "solution" Well the whole story:

# The Hunt for the Eleven-Legged Ghost

*Being a truthful account, more or less, of a crash in the simulator, and of two
fixes that came to settle the matter like gentlemen — though only one of them
was carrying a loaded pistol.*

---

Now I have seen a good many troubles in my time, and most of them never
happened; but the trouble with the simulator was the genuine article. You would
start her up — two little airplanes, Twisty and the Easystar, honest as the day
is long — and she would purr along sweet as a parlor cat. Then, after a while,
with no more warning than a mule gives, she would keel over dead and leave
behind her a tombstone that read, in its entire and unhelpful eloquence:

> **Simulator terminated with code 11.**

Eleven. That is all. Not a note, not a forwarding address, not so much as a
how-do-you-do. A man could grow old reading that number and learn nothing from
it but the cultivation of patience and profanity.

## The Scene of the Crime

So I rolled up my sleeves and went poking about in the cellar of the thing,
where the telemetry bus lives — a sociable contraption called **Ivy**. And here
is where I found the rot. There were **two threads** in that house, and both of
them had been given the run of the same cupboard, and *neither* had been told
the other existed.

The first thread, a tidy fellow, spent all day **sending** messages, walking
politely down the line of guests reading off their names. The second thread, the
**loop**, was the doorman — and a busy one, what with company arriving and
leaving every minute. Trouble was, every time a guest left, the doorman would
**burn the man's chair, the man's name-card, and the floorboards he stood on** —
while the first thread was *still reading the name off that very card.*

You can imagine the result. One of them reaches for a thing; the other has
already set it on fire and danced on the ashes. In the polite society of
computers they call this a *use-after-free*. On the river we called it
*drowning*, and it came to the same end.

There *was* a lock on the cupboard door — somebody had nailed it up in a fit of
good intentions a while back. But only the *sender* ever bothered to turn the
key. The doorman walked straight through the wall. A lock that one party honors
and the other ignores is not a lock at all; it is a *suggestion*, and the loop
thread was not the suggestible kind.

## Calling in the Bloodhound

Now a sensible man does not argue with a ghost; he hires a better tracker than
himself. So I sent for **DRD**, a valgrind of the genus bloodhound, who can
smell a race three counties off and will not be hurried. I turned loose
two-and-thirty guests to come and go like a politician's promises, and I let the
dog work.

He came back lathered and certain. **One hundred and thirty-seven** separate
quarrels. **Two hundred and ten thousand** snarls, near enough. And he laid the
guilty pair right at my boots: the sender, mid-sentence — and the loop, match in
hand. There was no arguing with it. The ghost had a name, an address, and a
criminal record as long as a steamboat.

## The Cure

The remedy, once you see it, is almost an insult in its plainness: *make the
doorman carry the same key.* The loop drops the lock only while it is dozing on
the porch waiting for company, and clutches it tight the whole time it is
burning furniture. Now the two of them can never be in the cupboard at once,
and the fire and the reading happen in a civilized rotation, like church and
saloon on a Sunday.

While I had the floor up I found **two more small leaks** the dog had sniffed
out — a fellow copying the airplane's whereabouts without asking the right
permission, and a one-bit weathervane flag that two threads kept flipping behind
each other's backs. Neither would *kill* you outright; they would only feed you
nonsense and call it weather. I patched them both, since I had the hammer out.

The dog ran the house again. **Zero quarrels. Zero snarls. Not one.** Two-and-
thirty guests trampling in and out, and the place quiet as a graveyard at noon —
the good kind of quiet, where everybody is alive.

## The Duel

Word got around, as it does, that there was a *second* fix in town — a stranger
named `nps_ivy_fixed.c`, who walked with a swagger and let it be known he had
Settled The Matter. Naturally we arranged a meeting. It is only neighborly.

The stranger's whole strategy, it developed, was to ask the departing guests to
hand back their name-cards a little more *quietly*. A fine and gentlemanly
notion. It addressed everything about the disaster except **the doorman, who was
still cheerfully burning the house down.** He had never once looked at the lock.
His loop was the same loop as the corpse we started with — I could have matched
it line for line in the dark.

We put him in front of the bloodhound all the same, fair and square, same
guests, same hour. The dog counted **one hundred and thirteen** quarrels and
better than **three hundred thousand** snarls — which, you will note, is *more*
uproar than the ghost made before anybody tried to help. The stranger had ridden
all that way to fix a fire by adjusting the curtains, and managed to make the
smoke worse.

## Moral

If a thing is to be guarded by a lock, then **every soul in the house must carry
the key** — the busy and the idle, the saint and the doorman alike. A precaution
that only the polite observe is no precaution; it is merely a record of who was
going to be surprised.

The ghost is buried. The house is quiet. And the gravestone that read *code 11*
now reads nothing at all — which, in this line of work, is the finest sentence a
man can hope to write.

*— faithfully set down, with only the usual improvements.*

OpenUAS:
Original change before this one still did not help.

OpenUAS:
Old fix still did not fully resolve the issue.

OpenUAS:
Fix #3657 is good enough for one part of the crash. Will make another PR for the remainder of stalled telemety and hardened  fdm_mutex

PR #3667, opened 2026-06-26
All previous OCAML tools now a native QT6 app

OpenUAS:
Here PR for drop-in replacment for old OCAML tools: **plotter, play, logplotter, gaia, messages** as native C++ QT6 application having all the functionality of previous ones. A new GCS is not in this PR, that will be separate, build system is already ready for it though.
Even the executable filenames are the same and in the same, even without control_panel.xml changes it still should work.

Test it out an report issues plz.

The new code uses very few libraries and a:

`sudo apt update && sudo apt install -y build-essential cmake git extra-cmake-modules qt6-base-dev qt6-svg-dev qt6-charts-devlibxml2-dev libzip-dev ` 

should be enough to get the source libs to compile. should be compatible with QT6.2.4 and upto QT6.10.2  thus from Ubuntu 22.04 upto 26.04 but very likely more version are OK. note that older system the lib may be  called differently, e.g.  libqt6svg6-dev , but that could be handled by PPRZ docs and pprz dev libs.

Then the normal procedure of: `cd ~/paparazzi && git submodule sync && git submodule update && make clean && make`


Fabien-B:
Nice.
There are still errors, I did not managed to build:
```
/home/fabien/paparazzi/sw/logalizer/play.cpp: In member function ‘bool PlayCore::loadLog(const QString&)’:
/home/fabien/paparazzi/sw/logalizer/play.cpp:408:23: error: ‘ParseResult’ is not a member of ‘QDomDocument’
  408 |         QDomDocument::ParseResult result = doc.setContent(content);
      |                       ^~~~~~~~~~~
/home/fabien/paparazzi/sw/logalizer/play.cpp:409:14: error: ‘result’ was not declared in this scope
  409 |         if (!result) {
      |              ^~~~~~
/home/fabien/paparazzi/sw/logalizer/play.cpp: In static member function ‘static QDomElement PlayCore::openProtocol(const QString&, QDomDocument&)’:
/home/fabien/paparazzi/sw/logalizer/play.cpp:806:29: error: ‘ParseResult’ in ‘class QDomDocument’ does not name a type
  806 |         const QDomDocument::ParseResult res = docOut.setContent(data);
      |                             ^~~~~~~~~~~
/home/fabien/paparazzi/sw/logalizer/play.cpp:807:14: error: ‘res’ was not declared in this scope
  807 |         if (!res)
```

Its the same error on both 22.04 and 24.04.

According to the doc, `ParseResult` was introduced in Qt 6.5.

sw/logalizer/play.cpp
```diff
@@ -0,0 +1,2064 @@
...
+        });
+
+        // Build the per-stream frame index now that m_log is time-sorted; this powers
+        // O(log n) snapshot lookups for responsive live scrubbing (see broadcastSnapshot).
+        buildStreamIndex();
+
+        // Now parse the standard XML properties (Conf and Protocols). Many Paparazzi .log files
+        // miss a single root element wrapper.
+        QDomDocument doc;
+        QDomDocument::ParseResult result = doc.setContent(content);
```
Fabien-B: `QDomDocument::ParseResult` does no exists in Qt 6.2.4

OpenUAS:
OK, plz retest with QT6.2.4, on higher QT6 versions you will get ofcourse depreciated warnings, but that is to be expected. It still works and compiles then ofcourse if  QT6 >= 6.5

gautierhattenberger:
@OpenUAS You can check #3679, it should add the packages you need to the CI docker image to run the compilation tests.

I have also tested locally the build. I got some issues because the `ivy-qt` and `pprzlink-qt` submodules are not initialized automatically from the top level `make`.

I also tried some of the tools, but they are not all working as expected:
- the drag&drop between message and rt-plotter
- the log plotter as some weird behavior when loading several curves
- the spin boxes in gaia have dark background that make them unreadable (are you using a dark theme ?)
I have not tested everything yet. I'll report here if I find some other bugs.

OpenUAS:
@gautierhattenberger Thanks for testing, a question:
Would love to have some extra info of: 

- Which OS?
- Full Wayland or X11?
- Which Theme

Or other info that you like to share that would be helpful to try and reproduce.

1) The drag&drop between message and rt-plotter
Ok, for me working, but then will test on other OS and most important X11? and fix where needed. The drag and drop was quite involved since the Wayland hassle where not all is allowed per default and some QT6 limits, a workaround was created.

2) The log plotter as some weird behavior when loading several curves.
A screenshot would be nice if possible

>the spin boxes in gaia have dark background that make them unreadable (are you using a dark theme ?)
Yes, dark theme ,and adde aspecific tweak, since default dark theme was difficult to read. Will disable the patch first and live with Them default colors, which are not super) and later re-add tweak for both Light and Dark theme. Not a PPRZ issue perse but not a good Dark theme on QT6 on Ubuntu 26.04 under Wayland.

>I have not tested everything yet. I'll report here if I find some other bugs.
TIA

gautierhattenberger:
I'm using Kubuntu 24.04 (Ubuntu with KDE desktop, normal color, not dark theme), with x11.

See below an example with environment simulator

image

I don't have the problem for log plotter with the multiple curves, we'll see if it comes back later. However, the right click on the plot is not like before (vertical auto scaling), but it does reset the full view in a strange way (see below after a right click). The input boxes are also dark.
image

And finally, it is not possible to support ubuntu 20.04 since some of the qt6 packages that you need are not available. I know this version is not supported anymore, but are still using it for some of you GCS. So I'd like to have some protections in the makefiles where the new tools are not compiled if not available and the old are not (yet) removed and compiled.

gautierhattenberger:
When opening a new window (ctrl-N) of the log plotter and then closing the tool from the paparazzi center, only the first original window is closed.
It is not the case with the real-time plotter were all windows are correctly closed (although I still can't use it because drag and drop doesn't work...).

OpenUAS:
@gautierhattenberger Awesome, very useful feedback, will address all issues and refresh PR.

Issue #3663, opened 2026-06-12
Submodule issues of mavlink and dronecan not building on clean OS


OpenUAS:
### Intro
When installing Paparazzi via official documentation the Mavlink and Dronecan submodules are not building. Not a Paparazzi fault per se. Not a personal issue either, ofcoure could install, patch  and upfix the needed packages oneself. But somehow this must be fixed cleanly for regular users.

### Issue encountered:
* mavlink not installed, missing packages "imp" or "future"
* dronecan not installed, missing package "em"

Since PPRZ mavlink tracks upstream mavlink/mavlink (with nested ardupilot/pymavlink), and upstream already removed both future and the imp usage (the imp-removed-in-3.12 problem that bites Ubuntu 24.04+), 

- [ ] Please advance the mavlink git submodule to a later version.
- [ ] Dronecan DSDLC needs an upstream fix in https://github.com/dronecan/dronecan_dsdlc

😠 _Yeah fine grrrr.. now I cannot testfly today, I need a Dronecan fix today!"_

OK we hear you... cut paste this line this in you terminal and [enter]:

 `sudo apt install python3-empy`
 
replace the the file

 ~/paparazzi/sw/ext/dronecan/dronecan_dsdlc/dronecan_dsdlc.py
 
with this file https://github.com/user-attachments/files/30129779/dronecan_dsdlc.py attached to this issue, then after a:

`make clean && make`

...you are good to go!

OpenUAS:
See https://github.com/dronecan/dronecan_dsdlc/pull/42

PR #3671, opened 2026-06-30
Swarm supervision

Fabien-B:
Make multiple Aicraft selection possible.
I allows to select common configuration for all selected ACs, or to clean/build/Flash in batch.
image


OpenUAS:
A good idea. Just do it if it does not work convenient, we still can change it. Only the missing screen real estate on smaller screens is quite problematic with this layout. Well ... I fix that part then in the next GUI iteration.

Fabien-B:
At first I kept the previous header, but I generalized the side panel to do both the swarm and single AC, so I decided to remove the header.
I will look into adding back an header to be shown when only one AC is selected.

OpenUAS:
@Fabien-B difficult to test since I cannot build due to 3rd party blocker for PPRZ see https://github.com/dronecan/dronecan_dsdlc/pull/42    I made a fix for that issue and also a PR it but is it is not in their branch yet.  Will patch locally and retry.  For a fix see also https://github.com/paparazzi/paparazzi/issues/3663

Oh, and a fix fro mavlink 3rd party I already made but also there depends on 3rd party. 

Maybe our sw/ext better to point at a fork? This why we could patch it self and after submodule update have the working version already, no need to wait fro a 3rd party.

Anyhow will make temp fix and retest...

OpenUAS:
Yes works well, although GUI kind of confusing, but that can be fixed later... For example i I  compile multiple the build states ony one name, see arrows. But all OK. Would be nice to have a way to fold or hide the list of AC's though.

screenshot_pprz_center_01

OpenUAS (approved):
All OK for first iteration. GUI could be improved though. E.g. List folded out of view per default would be best IMHO. Suggesting changes but still apprive, else I feel it will never land up in master, anyhow thanks, really useful for some.

gautierhattenberger:
There are some glitches that would have required more testing before merging, we should have marked this as a draft. Now that it is merged, is it possible to add the option to display in single mode (@Fabien-B you show me something like that at some point) ?

Issue #3674, opened 2026-07-16
Make clean and make balks and stops build


OpenUAS:
When performing a  `make clean && make ` This appears:

```
 "[ -L conf/conf.xml ] || [ -f conf/conf.xml ] || cd conf && ln -s conf_example.xml conf.xml
ln: Already exists
make: *** [Makefile:136: conf/conf.xml] Error "
```
Yeah, since I use a lot of symlinks to all kind of configs, my conf.xml symlink indeed exists... as already using this method in PPRZ 12 years... works well if you switch branches and config a lot during the day. Anyhow, the fake :brain: give some quick help:
**What is incorrect** In the recipe

```makefile
[ -L $@ ] || [ -f $@ ] || cd conf && ln -s conf_example.xml conf.xml
```

shell precedence groups this as `( ... || ... || cd conf ) && ln -s ...`. So whenever conf.xml already existed, the guard succeeded, `cd conf` was skipped, and `ln -s` ran anyway from the top-level directory — failing with "Already exists" (or worse, creating a stray conf.xml in the repo root).

**The fix:**

```makefile
conf/conf.xml: conf/conf_example.xml
	[ -L $@ ] || [ -f $@ ] || ln -s $(<F) $@
```

- No `cd` needed — `ln -s conf_example.xml conf/conf.xml` creates the symlink in conf/ with the correct relative target.
- Uses automatic variables `$@` and `$(<F)`, matching the style of the generic `conf/%.xml` rule, so it stays correct if paths ever change.
- Verified idempotent: it preserves your existing conf.xml (your OPENUAS swarm conf symlink was untouched), and correctly creates the symlink from scratch when absent.

gautierhattenberger:
closed with #3677

PR #3677, opened 2026-07-17
Fix make with existing Symlink

OpenUAS:
Please **read all of the #3674** it ixes bug #3674 ** [Makefile:136: conf/conf.xml] Error


gautierhattenberger (approved):
My bad, I didn't know the `@F` variable, but this was what I wanted to do, indeed.

Issue #3593, opened 2026-02-12
Persistent settings


Fabien-B:
> Note that persistent settings not available for Chibios, need support fixed 

 _Originally posted by @OpenUAS in [#3584](https://github.com/paparazzi/paparazzi/issues/3584#issuecomment-3882927619)_

OpenUAS:
Fixed with PR #3678

Issue #3466, opened 2025-06-06
[Tracking] Track fixes for minor anoiances, drivers, bugs etc.


OpenUAS:
Over the course of years we have a branch with small annoyances fixed, bugs fixed, drivers added, drivers improved, example airframes added, etc. etc. They tend never to land up in master. That needed to change.

Therefore just to make sure all this lands up to benefit more this issue is to keep track of the task and Pull request numbers involved. 

This issue is a living document where PR are added as well as tasks in random order still.

- [ ] Ongoing minor issues
- [x] Get honored the FMU2.4 fixes (Pixracer) 
- [x] add_sonar_i2c driver  PR #3540
- [x] fix_better_pwm driver_reading
- [x] Get honored he serial servo PR #3401
- [x] Add board Mateksys H743 wing v3 see PR #3552
- [x] Get PR honored for fix_eagletree_airspeed_improved see PR #3534
- [x] Get PR honored for fix_better_sonar_adc_readings see PR #3542
- [x] add and refactor Rangefinder PWM driver improved (LOCAL OK)
- [x] add_sf20_rangefinder and combine with  sf11/sf30 and improved (LOCAL OK)
- [x] add VB50A and VB22A rangefinder sensors (LOCAL OK)
- [ ] add_sensor_micoair_mr_06_ranger
- [ ] improve_teraranger_driver and add other models (LOCAL OK)
- [ ] add_esc_kiss_telemetry (LOCAL OK)
- [ ] fix_lisa_(mx) board options safety LED's  (LOCAL OK)
- [ ] fix_lisa_(mx) board_full_options
- [ ] add_parrot_minidrones_new (LOCAL OK)
- [ ] add_sensor_baro_dps310_i2c
- [ ] Finish and test Goermicro SPA06-003 driver
- [ ] add_sensor_micoair_all_mfr_xx_optical_flow
- [ ] fixed TCAS  for rotorcraft with FW integrate
- [ ] add_alpha_esc redo driver
- [ ] redux_integrate_parrot_isp
- [ ] testfly_parrot_disco_phase2_remaster_v4
- [ ] add board FlyingRC H7
- [ ] add board FlyingRC small F4
- [ ] add_chibios_for_lisa_s_fc
- [ ] add_sensor_bmm_150_magneto
- [ ] add_sensor_bmm_350_magneto
- [ ] add_sensor_hereflow
- [ ] add_chibios_boardfile_crazybee_f4_pro                            
[...]

PR #3678, opened 2026-07-17
Persistent setting re-created for Chibios

OpenUAS:
To fix issue #3593 Persistent setting that where gone where baremetal STM32 was removed; now re-created for Chibios. The whole story of my fake :brain:  below for who is interested.

### DIFF
--- a/sw/airborne/arch/chibios/modules/core/settings_arch.c
+++ b/sw/airborne/arch/chibios/modules/core/settings_arch.c
@@ -1,34 +1,1061 @@
 /*
- * Copyright (C) 2015 The Paparazzi Team
+ * Copyright (C) 2026 OpenUAS
  *
- * This file is part of Paparazzi.
+ * This file is part of paparazzi.
  *
- * Paparazzi is free software; you can redistribute it and/or modify
+ * paparazzi is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 2, or (at your option)
  * any later version.
  *
- * Paparazzi is distributed in the hope that it will be useful,
+ * paparazzi is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with Paparazzi; see the file COPYING.  If not, write to
- * the Free Software Foundation, 59 Temple Place - Suite 330,
- * Boston, MA 02111-1307, USA.
+ * along with paparazzi; see the file COPYING.  If not, see
+ * <http://www.gnu.org/licenses/>.
  *
  */
 
 /**
  * @file arch/chibios/modules/core/settings_arch.c
- * Persistent settings low level flash routines stm32.
+ * Flight safe persistent settings flash routines for ChibiOS,
+ * supporting the STM32F1, STM32F4, STM32F7 and STM32H7 families.
  *
- * FIXME dummy file
+ * The settings live in the last flash sector as an append only
+ * journal of self contained records:
  *
+ *   sector: | record | record | ... | erased space ...             |
+ *   record: | magic, data size | data (padded) | crc32 (padded)   |
+ *
+ * with all record fields aligned to the family programming unit (at
+ * least 4 bytes). Reading returns the newest record with a valid
+ * checksum; storing appends a new record into erased space. The flash
+ * registers are accessed directly (CMSIS definitions provided through
+ * the ChibiOS HAL headers), so no HAL/EFL driver configuration is
+ * required.
+ *
+ * Why it is safe to store settings in flight:
+ *
+ * - No sector erase is ever performed in flight. Appending makes an
+ *   erase (which stalls the CPU on flash fetches for up to ~2 seconds)
+ *   unnecessary until the sector is full: then persistent_write()
+ *   returns PFLASH_ERR_FULL in flight, and reclaims the sector with an
+ *   erase only when the autopilot reports motors off and not in flight
+ *   (persistent_clear() is refused in flight the same way).
+ *
+ * - Programming a single unit stalls flash fetches only for tens of
+ *   microseconds, and even that is hidden: the programming routines
+ *   execute from RAM (ChibiOS .ramtext) with interrupts masked
+ *   (chSysLock) for exactly that window, so neither the CPU nor a
+ *   preempting ISR can fetch from flash while the flash bus is busy.
+ *   Between units, interrupts are served normally.
+ *
+ * - All operations are serialized by a mutex (thread safe), and every
+ *   busy poll is bounded, so a faulty flash controller degrades to an
+ *   error return, never a hang.
+ *
+ * - A power cut during a store can at worst produce a torn record,
+ *   which the checksum rejects on the next boot: the previous valid
+ *   record (or the airframe defaults) are used instead. A sector
+ *   holding torn data beyond the valid record chain is never appended
+ *   to again (a partially programmed flash word must not be programmed
+ *   twice); it is reclaimed by the next on ground store or clear.
+ *   Erases and programs are verified by reading back.
  */
 
 #include "modules/core/settings.h"
+#include "autopilot.h"
+
+#include <ch.h>
+#include <hal.h>
+#include <string.h>
+
+#if defined(STM32F1XX) || defined(STM32F4XX) || defined(STM32F7XX) || defined(STM32H7XX)
+
+/** Base address of the internal flash on all supported STM32 families. */
+#define PFLASH_BASE 0x08000000UL
+
+/* ST flash unlock key sequence, identical on all supported families
+ * (defined here only if the CMSIS device header does not provide it) */
+#ifndef FLASH_UNLOCK_KEY1
+#define FLASH_UNLOCK_KEY1 0x45670123UL
+#endif
+#ifndef FLASH_UNLOCK_KEY2
+#define FLASH_UNLOCK_KEY2 0xCDEF89ABUL
+#endif
+
+/**
+ * Return codes of the persistent_write/read/clear API.
+ * All errors are negative, so callers may simply test for != 0.
+ */
+enum pflash_status {
+  PFLASH_OK           = 0,   ///< success
+  PFLASH_ERR_SECTOR   = -1,  ///< no usable settings sector (geometry or overlap)
+  PFLASH_ERR_SIZE     = -2,  ///< requested size is zero or does not fit a sector
+  PFLASH_ERR_NODATA   = -3,  ///< no valid stored settings (or stored size mismatch)
+  PFLASH_ERR_VERIFY   = -4,  ///< erase/program failed or read back mismatch
+  PFLASH_ERR_FULL     = -5,  ///< journal full, erase deferred until on the ground
+  PFLASH_ERR_INFLIGHT = -6,  ///< sector erase refused while in flight
+};
+
+/**
+ * Upper bound for busy polling an erase, far above the worst case
+ * erase time. It only turns a (theoretical) stuck flash controller
+ * into an error return instead of an infinite loop in the autopilot.
+ */
+#define PFLASH_WAIT_MAX_POLLS 400000000UL
+
+/**
+ * Upper bound for busy polling one unit program operation, which runs
+ * with interrupts masked: generous for the worst case programming time
+ * (tens of microseconds), while still bounding the interrupt blackout
+ * to milliseconds should the flash controller ever get stuck.
+ */
+#define PFLASH_PROG_MAX_POLLS 2000000UL
+
+/**
+ * Place a function in RAM (ChibiOS .ramtext, copied to RAM together
+ * with .data by crt0) so it executes without a single flash fetch, and
+ * forbid inlining it back into a flash resident caller.
+ */
+#define PFLASH_RAMFUNC __attribute__((noinline, section(".ramtext")))
+
+/** Serializes all persistent settings operations (thread safety). */
+static MUTEX_DECL(pflash_mutex);
+
+/** Location and geometry of the settings sector, filled by pflash_detect(). */
+struct FlashInfo {
+  uint32_t addr;      ///< address of the settings sector
+  uint32_t page_size; ///< size of the settings sector
+  uint32_t snb;       ///< sector number encoding for the SNB register field (unused on F1)
+};
+
+/*
+ * Family specific primitives:
+ *
+ * pflash_detect       locate the last flash sector
+ * pflash_erase        erase (only) the settings sector, ground only
+ * pflash_ram_program  program one unit, executes from RAM (.ramtext)
+ * pflash_program_unit flight safe wrapper around pflash_ram_program:
+ *                     interrupts masked only for the microseconds the
+ *                     flash bus is actually busy
+ * pflash_cache_flush  make cached flash reads coherent again
+ */
+
+#if defined(STM32F1XX)
+
+/** F1 flash programming unit: one 16bit half word. */
+#define PFLASH_PROG_SIZE 2
+
+/** All F1 FPEC error flags: programming error, write protection error. */
+#define PFLASH_F1_SR_ERRORS (FLASH_SR_PGERR | FLASH_SR_WRPRTERR)
+
+/**
+ * Detect the flash geometry and locate the last page.
+ *
+ * Low/medium density parts (<= 128k) have 1k pages, high density and
+ * connectivity line parts have 2k pages. XL density parts (> 512k)
+ * are not supported, their upper flash bank has a second FPEC with
+ * its own registers.
+ *
+ * @return 0 on success
+ */
+static int32_t pflash_detect(struct FlashInfo *flash)
+{
+  /* flash size in kBytes, from the factory programmed size register */
+  uint32_t size_kb = *(volatile const uint16_t *)FLASHSIZE_BASE;
+  if ((size_kb < 16) || (size_kb > 512)) { return -1; }
+
+  flash->page_size = (size_kb <= 128) ? 0x400 : 0x800;
+  flash->addr = PFLASH_BASE + size_kb * 1024 - flash->page_size;
+  flash->snb = 0; /* not used, F1 erases by page address */
+
+  return 0;
+}
+
+/**
+ * Busy-wait until the flash controller is idle.
+ * @return 0 when idle, -1 if the poll limit was exceeded
+ */
+static int32_t pflash_wait(void)
+{
+  for (uint32_t i = 0; i < PFLASH_WAIT_MAX_POLLS; i++) {
+    if (!(FLASH->SR & FLASH_SR_BSY)) { return 0; }
+  }
+  return -1;
+}
+
+/** Unlock the flash control register (no-op when already unlocked). */
+static void pflash_unlock(void)
+{
+  if (FLASH->CR & FLASH_CR_LOCK) {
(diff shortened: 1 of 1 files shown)


OpenUAS:
# Persistent Settings for ChibiOS

## What Was Built

Persistent Settings allow selected autopilot parameters to survive a restart. Tune a value, store it, power-cycle the aircraft, and the new value comes back automatically.

The completed design turned the old ChibiOS placeholder into a real internal-flash storage system for:

- STM32F1
- STM32F4
- STM32F7
- STM32H7

It required no external EEPROM, SD card, or board-specific storage configuration.

## Features

### Append-Only Storage

Saving settings no longer erased the flash sector every time.

Each save added a small record to the next free location:

```text
| saved settings | newer settings | newest settings | free space |
```

At startup, the autopilot scanned these records and loaded the newest valid one.

This provided:

- Fast saves
- Far fewer flash erases
- Much lower flash wear
- Recovery from interrupted writes
- Many saves before cleanup became necessary

### Data Integrity

Every record contained:

- A recognizable format marker
- The data size
- The settings data
- A CRC-32 checksum

Before loading anything, the driver checked the complete record. Corrupt, incomplete, or incompatible data was rejected, and Paparazzi safely kept its airframe defaults.

Every flash operation was also read back and verified. The code never claimed that settings were stored unless the bytes actually matched.

### Thread Safety

A ChibiOS mutex protected reading, writing, and clearing.

If two threads attempted to access Persistent Settings simultaneously, their operations were serialized instead of racing over the flash controller or journal.

### Clear Error Reporting

Named errors distinguished between conditions such as:

- Unsupported or unsafe flash geometry
- Invalid settings size
- No valid stored settings
- Verification failure
- Journal full during flight
- Erase refused during flight

Existing callers still worked because success remained `0` and every failure remained negative.

---

## Safe Storage During Flight

This was the most important part of the design.

A normal STM32 flash-sector erase can stall instruction fetching for hundreds of milliseconds or even seconds. That is unacceptable during flight.

The implementation addressed this using several layers.

### No Erasing in Flight

Normal saves only appended a record into erased space.

If the sector became full during flight, the function returned a safe “journal full” error. It did **not** erase the sector.

Sector cleanup was permitted only when:

- The autopilot was not in flight, and
- The motors were off

`persistent_clear()` used the same protection.

### Flash Programming Ran from RAM

Even programming one small flash word can briefly block the flash bus.

The low-level programming routine was therefore placed in ChibiOS’s `.ramtext` section. ChibiOS copied it to RAM during startup, allowing it to execute without fetching instructions from flash.

This placement was confirmed in the final executable and under link-time optimization. For example, the F4 test placed the routine at RAM address `0x20002bb0`.

### Very Short Interrupt Protection

Interrupts were masked only while one flash programming unit was physically being written.

That window lasted microseconds, not milliseconds or seconds. Interrupts ran normally between units.

The RAM routine contained:

- No calls back into flash code
- No flash-resident constants
- No unbounded waits
- Only the minimum register operations required

The generated machine instructions were inspected to verify those properties.

---

## Robustness Across STM32 Families

The common journal, checksum, validation, and safety logic was shared. Only the hardware-specific flash operations differed.

### STM32F1

- Factory flash-size detection
- Correct 1 KB or 2 KB page geometry
- 16-bit half-word programming
- FPEC page erase
- Explicit rejection of unsupported flash arrangements

### STM32F4

- Correct 16 KB, 64 KB, and 128 KB sector geometry
- Correct single-bank and dual-bank handling
- 32-bit programming
- ART cache invalidation after flash changes
- Defensive masking of sector-number register fields

### STM32F7

- Correct handling of its larger sector variants
- Correct optional dual-bank geometry
- 32-bit programming
- Data-cache invalidation before verification
- RAM placement checked with optimization enabled

### STM32H7

- Correct bank-two flash registers
- 128 KB sector geometry for supported 2 MB devices
- Required 256-bit, 32-byte flash-word programming
- Cache maintenance
- Protection against programming the same flash word twice

All controller busy loops had timeouts. A defective or stuck flash controller produced an error instead of freezing the flight computer indefinitely.

---

## The Investigation

The first clue looked harmless: one missing function.

`persistent_clear()` did not exist in the ChibiOS implementation, causing the linker to stop. But opening the file revealed something more serious: it was only a dummy. Writes always failed, while reads claimed success without actually loading settings.

The old bare-metal implementation supplied the map, but not the whole answer. ChibiOS, multiple STM32 families, caches, dual banks, and in-flight timing all changed the rules.

Then came the first hidden danger: **F4 dual-bank geometry**.

An early interpretation treated F4 dual-bank sectors like F7 sectors and halved their size. On a 2 MB STM32F427, including the Pixracer target, that calculated the wrong address and sector number. The bad number could even spill into neighboring flash-control bits. Reference-manual verification exposed it, and the geometry was corrected.

The next danger was the obvious one hiding in plain sight: **erase time**.

A design that correctly stored settings could still stop the flight controller for seconds. That led to the append-only journal, strict in-flight erase prohibition, and RAM-executed programming.

Finally, the power-failure investigation uncovered the quietest bug.

After an interrupted save, the journal could contain a torn record. An early recovery rule could append a new record after that damaged data, but the next boot scan would stop at the torn record and never reach the new one. On H7, attempting another approach could reprogram an already touched flash word and trigger a hard flash/ECC error.

The final rule became simple and strict:

> Append only at the end of the verified record chain. If anything suspicious exists beyond it, do not touch it in flight. Reclaim the sector after landing.

That rule prevented unreachable records, repeated programming, and unsafe cleanup.

---

## What Was Checked

- F4 full build and link using Twisty/Pixracer
- F7 full build and link using Martyn/px4fmu 5.0
- H7 compilation using T1_Ranger_VTOL/tawaki
- F1 compilation using a temporary Lisa/M ChibiOS target
- Zero warnings from the implementation on all four families
- RAM-section placement
- RAM routine disassembly
- LTO compatibility
- Cache-coherency behavior
- Flash geometry and bank selection
- Firmware/settings-sector overlap protection
- Power-loss behavior
- Torn-record recovery
- Journal-full behavior
- In-flight erase refusal
- Mutex protection
- Timeout coverage
- Cleanup of all temporary verification files

## What Users Would Experience

With the implementation restored, using Persistent Settings remains simple:

```xml
<define name="USE_PERSISTENT_SETTINGS" value="TRUE"/>
```

From there, Paparazzi handles storage and recovery automatically:

- Save tuning changes without immediately wearing the flash.
- Restart and recover the newest verified settings.
- Keep flying safely while a save is appended.
- Receive an error instead of an in-flight erase when storage is full.
- Fall back safely to airframe defaults if stored data is damaged.

Have fun...

Fabien-B:
Hi @OpenUAS,
It seems that your "fake 🧠" as re-invented the wheel. Could you use the "Embedded Flash Driver" (EFL) from ChibiOS, or even better, the "Managed Flash Storage Driver" (MFS) on top of EFL ?
It will allows much shorter, readable and maintainable code.

Also, how "persistent" are those settings ?
Do they persists across flashing a new firmware ? If so, the list and ids of settings can change, how is it handled ?

My point of view is that those settings should be persistent across reboots, but they should be erased when flashing a new firmware.

https://chibiforge.org/doc/21.11/hal/group___h_a_l___m_f_s.html
https://chibiforge.org/doc/21.11/hal/group___h_a_l___e_f_l.html
https://github.com/ChibiOS/ChibiOS/blob/master/testhal/STM32/multi/EFL-MFS

OpenUAS:
Hi thanks for looking into it. Thanks for pointing at the ChiBios Functionality available.

It is so elaborate that the big difference is that the changes t are set-able LIVE in air while flying without hiccup.  It is so complex since erase should be avoided.

@Fabien-B  you asked: "those settings should be persistent across reboots," Yes, they are.

Indeed I agree if the question is, if re-flashed, do the same values still exist, Often not needed  since we flash for a reason, most of the time changing the AP functionality, but one could make an auto update settings path. For now: settings gone after a re-flash.

But if one wants setting to survive after re-flash: Yeah, that is possible. FYI just made a change, not refreshed the PR yet for be able to flash if there is an Ardupilot bootloader and use original Ardu storage bank 1 as storage and bank 2 as backup). This survives re-flashing flashing.(but parameters still can be not synced)

Will work on improving persistent storage more,  But first I try to find time to finish more urgent things to fix see: [#3466](https://github.com/paparazzi/paparazzi/issues/3466) and that is way incomplete even ...

OpenUAS:
@Fabien-B  A quick compare, nice that you pointed out  thise MFS. Before we had none persistent storage, now we have a solution, maybe not ideal as may things. I take a look in a year if the state of Chibios has improved in respect to H7, or who knows create a fix for Chibios. For now being pragmatic, for time being mybe use just this one C file based on previous code.

The whole story, good before bedtime if you are unable to get to sleep ;)

The comparison across all four families currently supported by settings_arch.c: F1, F4, F7, and H7. 
The deciding test is not whether generic MFS exists, but whether this vendored ChibiOS has a working EFL low-level driver and test coverage for each family, plus compatible program/erase behavior for in-flight use.

**The conclusion** is cross-family: 
MFS would simplify the visible Paparazzi wrapper, BUT NOT the system as a whole in this vendored ChibiOS. 
F1/F4 have EFL, only selected F7 parts do, H7 has none, and unmodified MFS may erase during writes or mount repair without an autopilot flight-state hook;  **retaining the current backend is the safer and actually simpler production choice today**.

**Conclusion**

Using ChibiOS EFL and MFS would make Paparazzi’s visible `persistent_read/write/clear()` wrapper shorter, but it would **not simplify the complete implementation enough to justify migration**.

The current settings_arch.c should be retained for now because:

- STM32H7 has no EFL driver in the vendored ChibiOS.
- Some Paparazzi STM32F7 MCUs are unsupported by the vendored F7 EFL driver.
- MFS may automatically erase sectors during a write or even during startup repair.
- MFS has no awareness of whether the aircraft is flying.
- The current backend provides stronger and more specific flight-safety guarantees.
- The current compiled backend is only about 1.9 KiB despite having 1,153 source lines.

The vendored ChibiOS is `ver21.11.1@15221-404-g48e2fd03c`, so conclusions are based on the actual local implementation rather than only the public 21.11 documentation.

**Layer Comparison**

| Option | What it replaces | What remains custom | Assessment |
|---|---|---|---|
| Current implementation | Nothing | Flash primitives, journal, mirror, safety | Best current option |
| EFL only | Direct register access | Journal, CRC, mirroring, flight policy | Limited simplification |
| EFL + MFS | Flash access and journal | Flight-state wrapper, board layout, H7 fallback | Cleaner wrapper, more total complexity |
| Modified MFS + custom H7 EFL | Almost everything | Significant ChibiOS maintenance | Possible long term, not simpler today |

## EFL Only

EFL is a generic embedded-flash abstraction. It provides:

- Flash descriptors and sector geometry.
- `flashRead()`
- `flashProgram()`
- `flashStartEraseSector()`
- `flashWaitErase()`
- `flashVerifyErase()`
- Optional mutual exclusion.

Using it could remove much of the family-specific register code from settings_arch.c.

It does **not** provide:

- Persistent-record formatting.
- CRC-protected journals.
- Power-loss recovery.
- Primary/backup mirroring.
- Selection of the newest committed settings.
- Airframe-default fallback.
- “Never erase in flight” policy.

EFL alone would therefore remove perhaps the lower-level third of the implementation, while the complete journal and safety layer would remain.

It also requires enabling `HAL_USE_EFL=TRUE`; Paparazzi currently defines it as `FALSE` in halconf.h.

## MFS Plus EFL

MFS is much closer to the required feature set. It provides:

- Two-bank managed storage.
- Indexed variable-sized records.
- CRC16 integrity checks.
- Append-style record updates.
- Wear levelling through bank swapping.
- Garbage collection.
- Write verification.
- Mount-time repair after interrupted operations.
- Atomic transactions across multiple record IDs.

A Paparazzi wrapper could be conceptually small:

```c
persistent_read()  -> mfsReadRecord(..., 1, ...)
persistent_write() -> mfsWriteRecord(..., 1, ...)
persistent_clear() -> mfsEraseRecord(..., 1)
```

For the Matek board, the intended configuration would be roughly:

```c
MFSConfig config = {
  .flashp = (BaseFlash *)&EFLD1,
  .erased = 0xFFFFFFFF,
  .bank_size = 16384,
  .bank0_start = 1,
  .bank0_sectors = 1,
  .bank1_start = 2,
  .bank1_sectors = 1
};
```

This maps naturally onto sectors 1 and 2.

However, MFS treats those sectors as two banks of one managed store, not two permanent identical clones. During garbage collection it:

1. Copies current records to the alternate bank.
2. Writes the destination bank header.
3. Erases the old bank.
4. Switches the active bank.

This is power-loss safe and provides redundancy while transitioning, but it does not keep both sectors as identical long-lived copies.

## Critical Flight-Safety Difference

The current backend only erases when:

```c
!autopilot_get_motors_on() && !autopilot_in_flight()
```

If a journal is full in flight, it returns `PFLASH_ERR_FULL`. It never starts an erase.

Unmodified MFS can call garbage collection automatically from `mfsWriteRecord()` when contiguous space is insufficient. Garbage collection erases a bank.

MFS can also erase during `mfsStart()` or `mfs_mount()` while repairing:

- Two valid banks.
- Garbage in one bank.
- Both banks unreadable.
- Torn records found during strong checking.

Therefore, merely checking `autopilot_in_flight()` before calling `mfsWriteRecord()` is insufficient. Space could become insufficient inside that call, causing an erase anyway.

To preserve the current contract, MFS would need one of:

- A dry-run API guaranteeing no garbage collection is needed.
- A “no GC” write mode.
- A callback authorizing erase.
- A modification to MFS itself that returns a deferred-GC result while flying.

None of those policies exists in the vendored public API.

Mount-time repair is less dangerous because it occurs during boot, before flight, but it still means MFS controls when sectors are erased rather than Paparazzi.

## MCU Support

| Family | EFL available locally | MFS practical | Important limitation |
|---|---:|---:|---|
| STM32F1 | Yes | Possible | No bundled EFL-MFS test target |
| STM32F4 | Yes | Possible | No bundled F405 EFL-MFS test target |
| STM32F7 | Partial | Some boards | Only F765/767/769/777/779 supported |
| STM32H7 | **No** | **No** | No H7 EFL low-level driver |

### STM32F1

The local EFL driver supports F1 page geometry and half-word programming. MFS could work, but the bundled EFL-MFS test suite contains no current F1 make target.

The existing custom backend already handles:

- 1 KiB and 2 KiB page geometry.
- Half-word programming.
- Bounded busy waits.
- RAM-resident programming.

Migration would need hardware testing on each active F1 density variant.

### STM32F4

The EFL driver supports the STM32F405 through the `STM32F40_41xxx` path and contains the correct sector table:

- Sectors 0–3: 16 KiB
- Sector 4: 64 KiB
- Remaining sectors: 128 KiB

MFS should technically operate on the F405 sectors 1 and 2. However, the bundled EFL-MFS test matrix has no F405/F4 target, despite stale Eclipse configuration names mentioning older F4 tests.

### STM32F7

The local F7 EFL driver only supports:

- F765
- F767
- F769
- F777
- F779

It does not support several Paparazzi boards:

- BetaFPV AIO F7: STM32F722
- T-Motor AIO F7: STM32F722
- Holybro Kakute F7: STM32F745

These would fail at compile time with:

```text
This EFL driver does not support the selected device
```

The bundled MFS test does cover an STM32F767, so supported high-end F7 variants have reasonable upstream test coverage.

### STM32H7

There is no `STM32H7xx/hal_efl_lld.c` or `.h` in this vendored ChibiOS.

Therefore MFS cannot use internal H7 flash through EFL. Current H7 boards include:

- Matek H743 Slim
- Matek H743 Wing
- Cube Orange
- Pixhawk 6X/H753
- Tawaki H743/H753

The custom backend currently provides the missing H7 implementation itself:

- 32-byte flash-word programming.
- Bank-2 operation.
- Cache invalidation.
- Direct `CR2`, `SR2`, and `KEYR2` handling.
- RAM-resident programming.
- No erase in flight.

Replacing the backend with MFS would require first writing and maintaining a new ChibiOS H7 EFL driver.

## Program-Time Safety

The custom backend deliberately places its flash programming primitive in `.ramtext` and masks interrupts only around each hardware programming unit.

That prevents an interrupt handler or flash-resident caller from fetching instructions while the flash bus is busy.

The vendored EFL implementations perform their complete programming loops inside normal flash-resident functions. On F1/F4/F7, the processor can stall while flash is busy. This can be acceptable for generic configuration storage, but it is a weaker real-time guarantee than the current implementation.

The current backend’s behavior is better suited to an active flight controller:

- Very short bounded interrupt blackout.
- Explicitly RAM-resident programming routine.
- Interrupts run normally between programming units.
- Long sector erase prohibited in flight.

## Integrity and Recovery

| Feature | Current backend | MFS |
|---|---|---|
| Record checksum | CRC32 | CRC16 |
| Write verification | Yes | Optional, enabled by default |
| Torn-write handling | Reject record, retain earlier one | Sealed headers and mount repair |
| Permanent clone | Yes for Matek configuration | No, active/standby banks |
| Wear levelling | Append journal, erase when full | Better bank-swap wear levelling |
| Multiple record IDs | No | Yes |
| Atomic multi-record transaction | No | Yes |
| Airframe-default fallback | Directly integrated | Requires adapter |
| Erase authorization | Paparazzi flight state | None |

For a single 12-byte settings blob, MFS’s multiple IDs and transactions provide little practical benefit.

## Complexity

Current implementation:

```text
Source: approximately 1,153 lines
Compiled Adam object:
  text: 1884 bytes
  data:   16 bytes
```

MFS itself is approximately 1,547 lines, and F4 EFL is approximately 965 lines. Even though those are shared library components, adopting them would add:

- MFS driver state.
- Record descriptor array.
- Transaction state unless compiled out.
- Shared copy buffer.
- EFL driver state.
- A Paparazzi adapter.
- H7 custom fallback.
- Flight-safe GC modifications.

MFS could be configured smaller:

```c
#define MFS_CFG_MAX_RECORDS 1
#define MFS_CFG_TRANSACTION_MAX 0
#define MFS_CFG_MAX_RECORD_SIZE ...
```

But the resulting system would still have two implementations:

- EFL/MFS for supported F1/F4/F7 variants.
- Existing custom backend for H7 and unsupported F7 variants.

That is more maintenance, not less.

## Recommendation

Keep the current settings_arch.c implementation.

It is specialized for exactly what Paparazzi needs:

- One small atomic settings blob.
- CRC32 integrity.
- Board-selected mirrored sectors.
- Default last-sector operation elsewhere.
- F1/F4/F7/H7 support.
- No erase in flight.
- RAM-resident programming.
- Small compiled footprint.
- No additional HAL configuration.

A future MFS migration would become reasonable only after all of these are true:

1. ChibiOS gains an H7 EFL driver.
2. F722 and F745 are supported by F7 EFL.
3. MFS gains a no-GC write mode or erase-authorization callback.
4. F1, F405, F722/F745/F767, and H743 test targets exist.
5. Real-time behavior during program and erase is validated.
6. The settings format migration and ArduPilot-sector ownership are defined.

Until then, EFL could modestly reduce register-level code on selected families, but mixing EFL and direct implementations would make the backend less uniform. MFS would hide a large amount of correct generic storage logic, but it would also weaken the flight-specific safety contract and remove support for part of the board fleet.

Fabien-B:
@OpenUAS 
I am not very fond of (not) reading heavy loads of AI generated text, and if I'm not against AI generated code, I prefer when it's been reviewed and tested. That will not be easy to test this code on all existing STM32 ever used in an autopilot, and let's be honest: we Paparazzi users are not a big community, so we cannot rely on that to catch all kind of bugs.

This is why it would be best to use the ChibiOS implementation for that: better review, potentially wider community testing.

I understand that EFL seems to not be supported on every STM32 out there, but maybe the "clean" solution is to contribute back to ChibiOS by implementing it on the missing targets.
That would help everyone, and the ChibiOS maintainers could give good advice for this implementation.

The official ChibiOS is now here: https://github.com/chibios-upstream/chibios/

https://discord.gg/YCqBzhYYv

PR #3679, opened 2026-07-18
[ci] add necessary packages to compile qt6 tools

gautierhattenberger:
This is temporary solution for testing and developing tools from #3667 before integrating to paparazzi-dev package.
For now, it will be integrated to the docker image used by the CI server.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -20,4 +20,5 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y \
     python3-lxml \
     python3-empy \
     python3-pexpect \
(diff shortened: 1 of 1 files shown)


docker/ci/Dockerfile
```diff
@@ -20,4 +20,5 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y \
     python3-lxml \
     python3-empy \
     python3-pexpect \
+    build-essential cmake extra-cmake-modules qt6-base-dev qt6-svg-dev qt6-charts-devlibxml2-dev libzip-dev \
```
OpenUAS, line 23: Yes, that should do it AFAIK

OpenUAS (approved):
Yes, AFAIK that should be enough. Will somehow do a renewed PR of #3679  but that needs a review approval before the CI tries to rebuild it. I do not know which OS the Docker CI runs, so hope the package names are still the same.

PR #3680, opened 2026-07-20
[wgs84] compute the geoid/ellipsoid with the same model

gautierhattenberger:
Use the same model by default (WGS84) in the flight plan generator (instead of EGM96) and the airborne code.
Small differences can propagate and lead to different NED altitude for the waypoints, especially when using `WAYPOINTS_LLA_ALT` option with EKF2.
A 'configure' allows to deactivate this option, which is acceptable for a local flight.


gautierhattenberger:
Small addition: add an option to force using the reference from the flight plan with EKF2, and not one computed after receiving a good valid GPS position. It allows to have a fleet of drones to fly with the same reference and and comparable local positions.

Fabien-B (approved):

PR #3682, opened 2026-07-21
Fix pmw3901 driver

Fabien-B:
- Fix PMW3901 driver.
- Fix sys_time_usleep for durations less than 1ms,
- Do not restart SPI driver if the chip select pin as not changed: when CPOL=1 (CLK idle HIGH), restarting the driver cause a short CLK toggle, interpreted as the first clock cycle by the sensor.


OpenUAS (approved):
I do have this sensor, but no opportunity to test in a short timeframe. I assume you use it yourself and came across the issue and fixed it. Let's honor the PR and if it does not work for others an we get a huge mass of people complaining 😉  we revert.

PR #3691, opened 2026-07-24
[guidance] reference: integer route-vector normalization, no atan2f

agentperfopt:
## Summary

`gh_compute_route_ref()`, the function PR #717's callers were built around, found the route direction with `atan2f` to get an angle, then read `|cos|`/`|sin|` of that angle back off a trig table. Those are just the route vector's own normalized components, so this replaces the round trip with a direct integer normalization: no `atan2f`, no float, no trig table.

## Why this doesn't build

The fixed-point "route reference angle" model this function belonged to doesn't exist anymore — `57ff4b982` ("Ref model float", #2842) replaced it with a float-vector reference model well after #717 merged, and the current `gh_saturate_speed()`/`gh_saturate_accel()` do a plain vector-magnitude clamp with no route direction left to compute. This is a not-built reference reconstruction under `guidance/legacy_gh_ref/`: the pre-rewrite function copied verbatim from paparazzi's own history (commit `603b40a51`, the last commit before the rewrite), next to the optimized version, both built against the real `pprz_algebra_int.h`/`pprz_trig_int.h` still in this tree. Posting it because the technique holds up and seemed worth documenting even without a live target.

## Mechanism

`atan2f(-y, -x)` gets an angle, then the trig table reads `|cos|`/`|sin|` of it back — but `cos`/`sin` of that angle are just `x/mag` and `y/mag` for the route vector's own magnitude. Computing the magnitude with an integer square root (the larger component gets shifted into `[2^14, 2^15)` first, so `x² + y²` always fits in a `uint32_t`, then four Newton's-method iterations) and dividing gets there directly.

## Results

Real `pprz` integer math, real `atan2f`, real trig table, x86 wall-clock (hardware FPU, so a conservative floor):

| workload | current | this change | speedup |
|---|---|---|---|
| tracking a moving position setpoint | 47.44 ns | 35.22 ns | 1.35x |
| tracking a speed setpoint (2 atan2f/cycle) | 66.07 ns | 44.75 ns | 1.48x |
| hover-hold (route vector converged to 0) | 5.63 ns | 4.62 ns | ties |

On a Cortex-M4 with no FPU, the real target for a lot of this fleet, the float/`atan2f` library call count in the reference update goes from 5 to 0 — hundreds of cycles each, not counted in the x86 numbers above.

## Verification

Checked the integer path against the true (double-precision) normalized components over 2,000,000 randomized route vectors spanning the full `int32_t` range: max error 1.9 LSB out of a 14-bit scale, 0 overflow cases in the intermediate `x² + y²` sum the range-normalization step exists to prevent. Not bit-identical to the `atan2f` path, it's a genuinely different computation, but the closer of the two to the true value.

An automated perf-optimization pass profiling merged PRs found this; I rebuilt the historical function from paparazzi's own git history to check it against, since the live target is gone, and ran the verification above myself.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/guidance/legacy_gh_ref/README.md
+++ b/sw/airborne/firmwares/rotorcraft/guidance/legacy_gh_ref/README.md
@@ -1,0 +1,21 @@
+# legacy_gh_ref (historical reference, not built)
+
+This directory is **not referenced by any build file and is not compiled**. `gh_compute_route_ref()`,
+the function [PR #717](https://github.com/paparazzi/paparazzi/pull/717) modified callers of, no longer
+exists in this tree: the fixed-point "route reference angle" horizontal-guidance model it belonged to
+was replaced by a float-vector reference model in `57ff4b982` ("Ref model float", #2842).
+`gh_saturate_speed()`/`gh_saturate_accel()`, its current successors in `guidance_h_ref.c`, do a plain
+vector-magnitude clamp and have no route-direction computation left to optimize.
+
+`gh_compute_route_ref_reference.h`'s `gh_compute_route_ref_reference()` is copied verbatim from the
+real pre-rewrite source (paparazzi commit `603b40a51`, immediately before `57ff4b982`), using the real
+`pprz_algebra_int.h`/`pprz_trig_int.h` (both still present, unchanged, in this tree) rather than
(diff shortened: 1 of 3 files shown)


gautierhattenberger:
the function doesn't exist anymore...

PR #3698, opened 2026-08-02
Upload Progresbar Fixed

OpenUAS:
# Cleaner Firmware Upload Progress

## Why this change?

Firmware upload tools redraw `Erase`, `Program`, and `Verify` progress with carriage returns. Paparazzi Center treated each redraw as a new message, filling the console with repetitive progress lines.

### Before: every update creates another line

paparazzi-flash-classic

### After: one line follows the upload from start to finish

paparazzi-flash-latest

## What improves?

- **One clear status line:** each upload stage replaces the previous update in place.
- **Useful stall feedback:** after 3 seconds without progress, the leading `=` blinks without shifting the bar.
- **Clean completion:** the progress row becomes `Done with flashing <aircraft>` after a successful upload.
- **Future-ready progress data:** a Qt signal exposes encoded Erase, Program, and Verify progress for a later graphical progress bar.

## Low-risk scope

The special handling applies only to programs named `Flash ...`. Build output and all other console messages retain their existing behavior.

Changes are contained in `sw/supervision/python/console_widget.py`; no firmware or airframe behavior is affected.

## Validation

- Python compilation passes.
- `progress_improved.patch` applies cleanly against the pre-change source.
- The patch changes one source file only.

And to show how it went even out of hand in the past:
or_even_worse_upload

### DIFF
--- a/sw/supervision/python/console_widget.py
+++ b/sw/supervision/python/console_widget.py
@@ -1,8 +1,9 @@
-# Copyright (C) 2008-2022 The Paparazzi Team
+# Copyright (C) 2008-2026 The Paparazzi Team
 # released under GNU GPLv2 or later. See COPYING file.
 from generated.ui_console import Ui_Console
 from PyQt5.QtWidgets import *
-from PyQt5.QtCore import QProcess, QByteArray, Qt
+from PyQt5.QtCore import QProcess, QByteArray, Qt, QTimer, pyqtSignal
+# QTextCursor likly no longer needed
 from PyQt5.QtGui import QTextCursor
 import utils
 from program_widget import ProgramWidget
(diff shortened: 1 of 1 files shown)


gautierhattenberger (commented):
That looks a nice addition, but I can't test it before end of August. Maybe @Fabien-B, if he has a board with him to test upload.

OpenUAS:
@gautierhattenberger , thanks, no hurry. Yes, could not look at it anymore while uploading 100 times a day during board development, needed an intermezzo break regular tasks..

Fabien-B:
I appreciate the idea but...


image
image

We probably need a true VT100 interpreter, but that's not easy.
And that leave the question of how to handle the output of multiple programs running in parallel...

Fabien-B:
@OpenUAS please see #3701, and test on your side.

PR #3700, opened 2026-08-06
[perf] simplify yaw calculation

claudeyj:
`quat_from_earth_cmd_f()` currently finds the yaw angle with `atan2()`, then immediately uses that angle to calculate `cosf(yaw / 2)` and `sinf(yaw / 2)`.

The two vectors already give us the values needed for this calculation. Let `a` and `b` be the vectors being aligned:

```text
d = a · b
c = |a × b|
r = sqrt(d² + c²)
```

For the angle between them:

```text
cos(yaw) = d / r
|sin(yaw)| = c / r
```

The quaternion needs half of this angle. Using the standard half-angle formulas:

```text
cos(yaw/2) = sqrt((1 + cos(yaw)) / 2)
|sin(yaw/2)| = sqrt((1 - cos(yaw)) / 2)
```

The sign of `sin(yaw/2)` still comes from `(a × b) · thrust_vect`, so the rotation direction stays the same.

This avoids calculating an angle only to convert it back into sine and cosine. It uses basic arithmetic and `sqrtf()` instead, which should be cheaper on flight controllers where these trigonometric functions are expensive. The value passed to the square roots is clamped to handle small floating-point errors. The resulting quaternion and the roll/pitch calculation are otherwise unchanged.

### DIFF
--- a/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.c
+++ b/sw/airborne/firmwares/rotorcraft/stabilization/stabilization_attitude_quat_transformations.c
@@ -83,8 +83,6 @@ void quat_from_earth_cmd_f(struct FloatQuat *quat, struct FloatVect2 *cmd, float
   struct FloatVect3 thrust_vect;
   VECT3_ASSIGN(thrust_vect, R_rp.m[6], R_rp.m[7], R_rp.m[8]);
 
-  /// @todo optimize yaw angle calculation
-
   /*
    * Instead of using the psi setpoint angle to rotate around the body z-axis,
    * calculate the real angle needed to align the projection of the body x-axis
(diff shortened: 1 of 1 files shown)


gautierhattenberger:
Closing for the same reason than #3689. At least, the other one was doing better than testing `n == 0.f`...

claudeyj:
Thanks, that’s fair. The main concern is the lack of board testing. I’ll revisit this only after testing it on hardware and simplifying the code. Btw, is there any chance that I could contribute to pprz without having a board on my hand?

gautierhattenberger:
@claudeyj it demands the kind of contribution you are looking for. If it is about improving the performances of the embedded code, it is hardly possible without testing it on a real board.
Some functionality improvements can be tested with simulations. You can have a look at the open issues.

claudeyj:
Got it, thank you very much.

PR #3706, opened 2026-08-31
Slight fix of inconsistency between `motors_on` and `kill_throttle` for fixedwing autopilot

spacewolfXfr:
`autopilot_set_kill_throttle` now directly set the kill variable before calling `autopilot_set_motors_on`
Also added `NavResurrect` for fixedwing (set kill to false)

### DIFF
--- a/sw/airborne/autopilot.c
+++ b/sw/airborne/autopilot.c
@@ -301,11 +301,8 @@ bool autopilot_get_motors_on(void)
  */
 void autopilot_set_kill_throttle(bool kill)
 {
-  if (kill) {
-    autopilot_set_motors_on(false);
-  } else {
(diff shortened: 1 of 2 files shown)


OpenUAS:
Thanks for trying to improve PPRZ. I'm not fully convinced about the "NavResurrect" since it is not tested for AP MODE. I would like to see more about possible impact of change. More so since it could be quite a disaster if not validated. Flying PPRZ from 2008 I felt no restriction with older implementation, but maybe I missed a scenario.
Mind You I DO like if it is migration to toward common rotorcraft and fw, as stated. This is a tricky one.

Could do myself but .. no time excuse ;)  well not before IMAV2026 it is.

